PR #17, opened 2016-01-25
Fixes em cpu freq

deggeman:
Hi Patrick,

two small patches for HWmon energy meter and the fix for rfc test to be able to deal with single freq domain systems with ondemand cpufreq governor.


deggeman:
OK, I haven't tested TC2 platform. Let's skip this patch in this case. Do I have to push another request?

derkling:
Yes, remove the patch from the branch and push again the branch to your repo _with the same name_.
That should be enough to refresh the current pull request.

PR #20, opened 2016-01-25
Fix #4

### LINKED ISSUE #4: [Conf] Simplify configuration file syntax
We currently use JSON as format for all the configuration file which is quite fragile when it comes to add/remove entries because of the dependency on proper placement of commas.

We should switch to a more robust syntax, possibilities are:
1. extend libs/utils/conf.py::JsonConf() to parse a comma-less file and properly add commas when needed just before parsing with the standard JSON loader
2. evaluate the usage of YAML as input format

JaviMerino:
A fix for #4 and a small drive-by fix at the regexp while I'm at it.

### DIFF
--- a/libs/utils/conf.py
+++ b/libs/utils/conf.py
@@ -52,6 +52,12 @@ class JsonConf(object):
                 content = content[:match.start()] + content[match.end():]
                 match = JSON_COMMENTS_RE.search(content)
 
+            # Allow trailing commas in dicts an lists in JSON
+            # Note that this simple implementation will mangle things like:
+            # {"config": ", }"}
+            content = re.sub(r',[ \t\r\n]+}', '}', content)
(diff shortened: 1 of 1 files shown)


derkling:
I think we could still have issues with fragments like for example:
```json
    "tftp"  : {
        "folder"    : "/var/lib/tftpboot",
        "kernel"    : "juno1/kernel.img",
        "dtb"       : "juno1/dtb.img"
    },
```

In that case, if we remove the "dtb" attribute we should remove the previous comma as well.
It seems to be that the proposed patch does not fixes all these cases, thus we will end with a partial solution which could be misleading in the sense that if we can remove some commas the user could think that all the commas can be removed.

What do you think?

derkling:
Tested and it works perfectly! Thanks.

JaviMerino:
:sun_with_face:

PR #36, opened 2016-02-03
Vagrantfile: initial version

JaviMerino:
Add a simple vagrant file that sets up a VM to use with lisa.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,26 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "ubuntu/trusty64"
+
+  # Compiling pandas requires 1Gb of memory
+  config.vm.provider "virtualbox" do |v|
+    v.memory = 1024
+  end
+
+  # Forward ipython notebook's port to the host
+  config.vm.network "forwarded_port", guest: 8888, host: 8888
+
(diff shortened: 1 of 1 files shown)


JaviMerino:
It's by no means complete.  I haven't even tried to run the tests with it.  It's more of a proof of concept to think how we can deploy this without it being too much of a pain.  Not sure if vagrant is overkill.

derkling:
I like the idea, this provides also a more "safe" environment to run a notebook server and experiment without to worry about breaking your system.

On the overhead side, I think that if it's functional we should have that option merged just an one of the multiple possibilities for the user to use the framework. Perhaps we should add just a paragraphs in the installation instructions which call out for that possibility and points to the relevant documentation or directly reports which commands are required to build that VM.

Another possibility we could think about is to provide a "docker" based solution, which can be more convenient in case you do not really need a full VM.

JaviMerino:
vagrant lets you use docker as a provider instead of the default (which is virtualbox).  It should be possible to have a file that supports both.  It defaults to virtualbox to allow for people with macs and windows computers to use it.

As to whether to put it in a subfolder, vagrant assumes that the location of the file is the root of the project.  You can probably change that in the `Vagrantfile`.

JaviMerino:
New version installing `sshpass` as well.

derkling:
Tested and works fine! Let's merge this now and we can add new features in the future.

PR #38, opened 2016-02-04
README.md: added some tools dependencies

mike2390:


mike2390:
Going to send a new request with just the patches of interest.

PR #43, opened 2016-02-09
Add ping_time parameter to target.config and reboot

mike2390:
When rebooting, the reboot function repeatedly pings the target.
However, rebooting might require some seconds and in order to avoid
pinging the machine before it actually reboots, we can set the
ping_time option in target.config to tell the function how long to
wait before pinging.

Also, a comment has been added in target.config to specify what the
file represents and to highlight that commented parameters are
optional.

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -615,6 +615,12 @@ class TestEnv(ShareState):
                 logging.debug('%14s - %s', 'HostResolver', cmd)
                 os.popen(cmd)
 
+        return self.parse_arp_cache(host)
+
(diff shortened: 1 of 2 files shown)


mike2390:
I've done the changes to `target.config` and added the `reboot_time` parameter to the file.

Also, I've pushed the changes to the `reboot` function to make it more robust when trying to access the target by using the MAC address.

This should now be ready for merge. Please review it and if anything has to be changed, let me know.

derkling:
Another step towards a more robust reboot support ;-)

jlelli:
Nice! :+1:

PR #48, opened 2016-02-11
rfc_stune: Add the cgroup module back

JaviMerino:
4e2673a38945 ("libs/utils/env: move modules definition into board
specific section") trimmed down the list of modules defined for each
testsuite.  For STune, it trimmed it too much, removing the cgroups
module which is needed by the test.  Add it back.


derkling:
I noticed the same error me to, but not pushed a patch.
Thanks for pushing that update.

PR #51, opened 2016-02-15
test/eas/rfc: fix calculation of the first period of the task

JaviMerino:
We avoid the first period of the rtapp task in the boosted utilization
signal test because of the initial high load.  The period we read from
the parameters is expressed in milliseconds but the index of the
dataframes returned by trappy is in seconds.  Convert rtapp_period to
seconds before adding it to task_start to correctly add the two
variables in the same unit (seconds).


derkling:
Did not notice that simple error when I've refactored the code, neat!

PR #58, opened 2016-03-01
workload: fix background run of a workload

mdigiorgio:
Devlib provides a specific method for running a command in background which is
not what it was used in LISA. Before, LISA was calling execute() with a
background parameter set to True; now it just calls the background() function
which does the job.

This change only works after [this PR](ARM-software/devlib#25) is merged.


mdigiorgio:
[background fix](ARM-software/devlib#25) has been merged in mainline devlib. This PR should now be ready for merge too.

derkling:
The previous link seems broken, but since the PR (ARM-software/devlib#25) for devlib has been merged, let's fix this issue on LISA side as well. Thanks ;-)

PR #59, opened 2016-03-01
Schedtune boost margin improvements

JaviMerino:
Update to trappy v5.2.0 and use it's improvements to simplify the schedtune boost margin test.


derkling:
I love "bartizzation" of tests! ;-)

PR #63, opened 2016-03-02
workload: add possibility to set pauses before and after executing a workload

mdigiorgio:
By means of a start and an end pause parameter it is now possible to set a
period of time to wait after starting collecting traces and before the
execution of workload (for example to allow something to settle down). In
addition, you can also set a period of time to wait after the workload has
executed and before stopping the trace collection.

The two parameters are optional and their default value is zero.


derkling:
Thanks!

PR #65, opened 2016-03-04
shell: advertise lisa-test

JaviMerino:
The lisa-shell has a lisa-test command.  It's in its early stages but
it's useful in its current form so start advertising it.


derkling:
Right, well done!
We still miss some useful commands to list available tests and to easily access to their configuration file. Let's keep these on a FR #66.

PR #73, opened 2016-03-22
Usage of device ID in target configuration for Android devices

mdigiorgio:
When multiple Android devices are connected to the host machine, the user must specify the ID of the device he wants to connect to. This series shows an example of usage in the `target.config` file and fixes a bug when passing the device ID to the test environment.


jlelli:
Tested. Looks good to me!

derkling:
Thanks for fixing this! :+1:

PR #80, opened 2016-03-31
eas test rfc.config comments for ftrace

msrasmussen:
Add a few helpful comments to eas test rfc.config describing how to enable ftrace.


derkling:
Thanks, this contributes to the improvement of the so bad documentation we unfortunately still have.

PR #81, opened 2016-03-31
libs/utils/trace: add support for normalize_time

### LINKED ISSUE #57: [Trace] Feature request to pass  'normalize_time = False' into  class Trace::__init__()
To be able to sync the time (x axis) between kernelshark trace.dat and any TRAPpy based plot it would be nice to be able to see the same time on both of them. Lisa's Trace class should expose a parameter 'normalize_time' w/ a default value of True.

JaviMerino:
Pass normalize_time down to trappy.  Sometimes it's useful to have the
same time between kernelshark and trappy.  Trappy supports this with
normalize_time but the Trace class doesn't allow you to specify it.  Let
Trace accept this parameter and pass it down to trappy.

close #57

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -33,6 +33,7 @@ class Trace(object):
 
     def __init__(self, platform, datadir, events,
                  tasks=None, window=(0,None),
+                 normalize_time=True,
                  trace_format='FTrace'):
 
         # The platform used to run the experiments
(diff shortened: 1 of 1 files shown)


JaviMerino:
Rebased on top of the new master and added a couple of commits addressing my review of #82

derkling:
Fully agree with all the proposed code simplifications, going to merge! ;-)
Worth also to expose all the TRAPpy Trace parameters. Going to merge...

PR #83, opened 2016-04-01
Update submodules

JaviMerino:
Update devlib and trappy to the latest and greatest


derkling:
Let's boost LISA by using the most shining version of the main dependency libs ;-)

PR #86, opened 2016-04-12
ipynb: fix resetting frame statistics

mdigiorgio:
Frame statistics shall be reset before collecting them for the current
workload. This was done in the wrong way before, because the reset
option needs a double dash in front to work properly.


ipynb/android/workloads/Android_Recents_Fling.ipynb
```diff
@@ -292,7 +292,7 @@
     "    # Allow the activity to start\n",
     "    sleep(5)\n",
     "    # Reset framestats collection\n",
-    "    target.execute('dumpsys gfxinfo com.android.systemui reset')\n",
```
derkling: This means that before we was also limiting the stats just to the "com.android.systemui" app?
If that's the case, well done! It's better to remove this constraint and to consider system-wide stats ;-)

mdigiorgio: I'm not sure about that, but certainly the stats were not reset,

derkling:
Let's merge it as it it... however, it would be worth to better understand what is the meaning of an app being specified. Maybe it could be useful to have a more generic interface where we can focus the stats to just a single application.

PR #87, opened 2016-04-13
libs/utils/trace_analysis: add cluster frequency residency plots

mdigiorgio:
Add a plot method to show the frequency residency in each cluster. By frequency
residency we mean the amount of time spent at that frequency in a given cluster.

### DIFF
--- a/libs/utils/trace_analysis.py
+++ b/libs/utils/trace_analysis.py
@@ -788,48 +788,6 @@ class TraceAnalysis(object):
         figname = '{}/{}schedtune_conf.png'.format(self.plotsdir, self.prefix)
         pl.savefig(figname, bbox_inches='tight')
 
-    def _computeNonIdleTime(self, t_prev, t, tasks_time_intervals):
-        """
-        Compute the non-idle time spent by tasks within the time interval
-        [t_prev. t].
-
-        :param t_prev: Time interval init boundary
-        :type t_prev: double
-
-        :param t: Time interval end boundary
-        :type t: double
-
-        :param tasks_time_intervals: DataFrame containing start and end times
-            of tasks
-        :type tasks_time_intervals: :mod:`pandas.DataFrame`
-        """
-        nonidle_time = 0.0
-        for _, row in tasks_time_intervals.iterrows():
-            start = row['start']
-            end = row['end']
-            if start > t:
-                # Task not running at f_prev
-                continue
-            elif start >= t_prev and start <= t:
-                # Task starting at instant of time where frequency is f_prev
-                if end <= t:
-                    # Task runs at f_prev for the whole interval
-                    nonidle_time += end - start
-                else:
-                    # Task continues running at different frequency
-                    nonidle_time += t - start
-            else:
-                # Task started with previous frequency
-                if end > t_prev:
-                    # Task is still running now, at f_prev
-                    if end <= t:
-                        nonidle_time += end - t_prev
-                    else:
-                        # Task will continue running at different frequency
-                        nonidle_time += t - t_prev
-
-        return nonidle_time
-
     def getCPUFrequencyResidency(self, cpu):
         """
         Get a DataFrame with per-CPU frequency residency, i.e. amount of
@@ -839,70 +797,7 @@ class TraceAnalysis(object):
         :param cpu: CPU ID
         :type cpu: int
         """
-
-        if not self.trace.hasEvents('cpu_frequency'):
-            logging.warn('Events [cpu_frequency] not found, '\
-                         'plot DISABLED!')
-            return None
-        if not self.trace.hasEvents('cpu_idle'):
-            logging.warn('Events [cpu_idle] not found, '\
-                         'plot DISABLED!')
-            return None
-
-        freq_dfr = self.trace.df('cpu_frequency')
-        cpu_freqs = freq_dfr[freq_dfr.cpu == cpu]
-        idle_dfr = self.trace.df('cpu_idle')
-        cpu_idle = idle_dfr[idle_dfr.cpu_id == cpu]
-
-        ### Compute TOTAL time spent at each frequency ###
-        time_intervals = cpu_freqs.index[1:] - cpu_freqs.index[:-1]
-        total_time = pd.DataFrame({
-            'TOTAL Time' : time_intervals,
-            'Frequency [MHz]' : [f/1000 for f in cpu_freqs.iloc[:-1].frequency]
-        })
-        total_time = total_time.groupby(['Frequency [MHz]']).sum()
-
-        ### Compute ACTIVE time spent at each frequency ###
-        # time instants where CPU is idle
-        idle_times = cpu_idle[cpu_idle['state'] != IDLE_STATE]
-        # time instants where CPU is not idle
-        non_idle_times = cpu_idle[cpu_idle['state'] == IDLE_STATE]
-
-        if idle_times.index[0] < non_idle_times.index[0]:
-            idle_times = idle_times.iloc[1:]
-
-        time_intervals = [t2 - t1 for t1, t2 in zip(non_idle_times.index,
-                                                    idle_times.index)]
-
-        times_len = min(len(idle_times), len(non_idle_times))
-        tasks_time_intervals = pd.DataFrame({
-            "start": non_idle_times.index[:times_len],
-            "end": idle_times.index[:times_len]
-        })
-
-        # Compute the non-IDLE time spent at a certain frequency
-        time_intervals = []
-        frequencies = []
-        row_iterator = cpu_freqs.iterrows()
-        # Take first item from row_iterator
-        t_prev, f_prev = row_iterator.next()
-        for t, f in row_iterator:
-            nonidle_time = self._computeNonIdleTime(t_prev,
-                                                    t,
-                                                    tasks_time_intervals)
-            freq = f_prev['frequency']
-            time_intervals.append(nonidle_time)
-            frequencies.append(freq)
-            t_prev = t
-            f_prev = f
-
-        active_time = pd.DataFrame({
-            'ACTIVE Time' : time_intervals,
-            'Frequency [MHz]' : [f/1000 for f in frequencies]
-        })
-        # Sum the time intervals with same frequncy
-        active_time = active_time.groupby(['Frequency [MHz]']).sum()
-        return total_time, active_time
+        return self.getClusterFrequencyResidency([cpu])
 
     def plotCPUFrequencyResidency(self, cpus=None):
         """
@@ -951,6 +846,84 @@ class TraceAnalysis(object):
                           loc='lower right')
             plt.subplots_adjust(hspace=0.5)
 
+    def _getClusterActiveSignal(self, cluster):
+        """
+        Build a square wave representing the active (i.e. non-idle) cluster
+        time, i.e.:
+            cluster_active[t] = 1 if there is a task running in the cluster at
+                                time t
+            cluster_active[t] = 0 otherwise
+
+        :param cluster: list of CPU IDs belonging to a cluster
+        :type cluster: list(int)
+        """
+        freq_dfr = self.trace.df('cpu_frequency')
+        cluster_freqs = freq_dfr[freq_dfr.cpu.isin(cluster)]
+        idle_dfr = self.trace.df('cpu_idle')
+
+        cpu_activation = {}
+        for cpu in cluster:
+            cpu_freqs = freq_dfr[freq_dfr.cpu == cpu]
+            cpu_idle = idle_dfr[idle_dfr.cpu_id == cpu]
+
+            # time instants where CPU goes to an idle state
+            idle_times = cpu_idle[cpu_idle.state != IDLE_STATE]
+            # time instants where CPU leaves an idle state
+            non_idle_times = cpu_idle[cpu_idle.state == IDLE_STATE]
+
+            if idle_times.index[0] < non_idle_times.index[0]:
+                idle_times = idle_times.iloc[1:]
+
+            activation_indexes = sorted(
+                non_idle_times.index.values.tolist() + \
+                idle_times.index.values.tolist()
+            )
+            if activation_indexes[0] != 0.0:
+                activation_indexes.insert(0, 0.0)
+                activations = [i%2 for i in range(len(activation_indexes))]
+            else:
+                activations = [(i+1)%2 for i in range(len(activation_indexes))]
+            cpu_activation[cpu] = pd.Series(activations,
+                                            index=activation_indexes)
+
+        activations = pd.DataFrame(cpu_activation)
+        activations.loc[0.0].fillna(0, inplace=True)
+        activations.fillna(method='ffill', inplace=True)
+
+        cluster_active = activations[activations.columns[0]].astype(int)
+        for c in activations.columns[1:]:
+            cluster_active |= activations[c].astype(int)
+
+        return cluster_active
+
+    def _getCPUActiveSignal(self, cpu):
+        """
+        Build a square wave representing the active (i.e. non-idle) CPU time,
+        i.e.:
+            cpu_active[t] = 1 if there is a task running in the CPU at time t
+            cpu_active[t] = 0 otherwise
+
+        :param cpu: CPU ID
+        :type cpu: int
+        """
+        return self._getClusterActiveSignal([cpu])
+
+    def _integrate_square_wave(self, sq_wave):
+        """
+        Compute the integral of a square wave time series.
+
+        :param sq_wave: square wave assuming only 1.0 and 0.0 values
+        :type sq_wave: :mod:`pandas.Series`
+        """
+        sq_wave.iloc[-1] = 0.0
+        # Compact signal to obtain only 1-0-1-0 sequences
+        comp_sig = sq_wave.loc[sq_wave.shift() != sq_wave]
+        # First value for computing the difference must be a 1
+        if comp_sig.iloc[0] == 0.0:
+            return sum(comp_sig.iloc[2::2].index - comp_sig.iloc[1:-1:2].index)
+        else:
+            return sum(comp_sig.iloc[1::2].index - comp_sig.iloc[:-1:2].index)
+
     def getClusterFrequencyResidency(self, cluster):
         """
         Get a DataFrame with per cluster frequency residency, i.e. amount of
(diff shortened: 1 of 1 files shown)


mdigiorgio:
@derkling I'd like to add this to the trace analysis tutorial, but this requires to run again the workload on the target device to collect `cpu_idle` events as well. I would like to run the same workload you ran, could you tell me where can I find it?

libs/utils/trace_analysis.py
```diff
@@ -785,3 +787,117 @@ def plotSchedTuneConf(self):
...
+                         'plot DISABLED!')
+            return None
+        if not self.trace.hasEvents('cpu_idle'):
+            logging.warn('Events [cpu_idle] not found, '\
+                         'plot DISABLED!')
+            return None
+
+        freq_dfr = self.trace.df('cpu_frequency')
+        cpu_freqs = freq_dfr[freq_dfr.apply(lambda x: x.cpu in cluster,
+                                            axis=1)]
```
JaviMerino: Even better:
``` python
cpu_freqs = freq_dfr[freq_dfr["cpu"].isin[cluster]]
```

libs/utils/trace_analysis.py
```diff
@@ -785,3 +787,117 @@ def plotSchedTuneConf(self):
...
+        # time instants where CPUs go to an idle state
+        idle_times = cpu_idle[cpu_idle['state'] != IDLE_STATE]
+        # time instants where CPUs leave an idle state
+        non_idle_times = cpu_idle[cpu_idle['state'] == IDLE_STATE]
+
+        if idle_times.index[0] < non_idle_times.index[0]:
+            idle_times = idle_times.iloc[1:]
+
+        time_intervals = [t2 - t1 for t1, t2 in zip(non_idle_times.index,
+                                                    idle_times.index)]
```
JaviMerino: You can't substract like this because you may substract a time when one cpu goes out of idle from when another cpu went idle.  For example, in the following trace:
```
0.016381 	<idle> 	3 	0 	3 	4294967295
0.016424 	<idle> 	3 	0 	3 	0
0.017242 	<idle> 	2 	0 	2 	1
0.017372 	<idle> 	3 	0 	3 	4294967295
0.017544 	<idle> 	3 	0 	3 	0
0.017618 	<idle> 	3 	0 	3 	4294967295
0.017754 	<idle> 	3 	0 	3 	0
0.017870 	<idle> 	2 	0 	2 	4294967295
0.018141 	<idle> 	3 	0 	3 	4294967295
```
give you the following `time_intervals`:
```
[4.3000000005122274e-05,
 -0.00013000000035390258,
 -7.3999999585794285e-05,
 -0.0001159999992523808]
```

We have to split the `non_indle_times` and `idle_times` into per cpu arrays before doing the substraction. The previous four lines should be something like:
``` python
        idle_non_idle = [(v.index, v2.index) for (_, v), (_, v2) in
                         zip(idle_times.groupby("cpu_id"), non_idle_times.groupby("cpu_id"))]

        time_intervals = []
        for (idle, non_idle) in idle_non_idle:
            if idle[0] < non_idle[0]:
                idle = idle[1:]

            time_intervals.extend([t2 - t1 for t1, t2 in zip(non_idle, idle)])
```

mdigiorgio: Good point! I think we should implement a `getCPUResidency` because in another situations we want per-CPU info and just get the data from there. What do you think?

JaviMerino: if you think we are going to need info per-cpu then yes, create `getCPUResidency`.  The code I suggested in the previous comment was the minimal change I could think of to make it work for me, but there are probably better ways of doing it.

mdigiorgio:
I've just pushed what should be the last version of the frequency residency plot code.

@derkling, @jlelli and @JaviMerino could you please have a look at it and try see if the results make sense?

libs/utils/trace_analysis.py
```diff
@@ -950,3 +950,158 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+            cpu_activation[cpu] = pd.Series(activations,
+                                            index=activation_indexes)
+
+        activations = pd.DataFrame(cpu_activation)
+        activations.loc[0.0].fillna(0, inplace=True)
+        activations.fillna(method="ffill", inplace=True)
+
+        cluster_active = activations[activations.columns[0]].astype(int)
+        for c in range(1, len(activations.columns)):
+            cluster_active |= activations[activations.columns[c]].astype(int)
```
JaviMerino: You could use `reduce` instead of those 3 lines:
``` python
cluster_active = reduce(operator.or_, [cpu_activations.astype(int) for _, cpu_activations in activations.iteritems()])
```

You'd have to `import operator` at the beginning of the file.  Not sure whether the code is clearer.

libs/utils/trace_analysis.py
```diff
@@ -1009,52 +888,107 @@ def getClusterFrequencyResidency(self, cluster):
...
+        if not self.trace.hasEvents('cpu_idle'):
+            logging.warn('Events [cpu_idle] not found, '\
+                         'plot DISABLED!')
+            return None
+
+        freq_dfr = self.trace.df('cpu_frequency')
+        cluster_freqs = freq_dfr[freq_dfr.cpu.isin(cluster)]
+        idle_dfr = self.trace.df('cpu_idle')
+
+        ### Computer TOTAL Time ###
```
JaviMerino: Should be "Compute"

libs/utils/trace_analysis.py
mdigiorgio: Yeah, probably makes clear we are applying the or to different signals

JaviMerino:
Another minor nit: in the last commit message: "This ptimises" :arrow_right: "This optimises"

libs/utils/trace_analysis.py
```diff
@@ -29,6 +29,8 @@
 # Configure logging
 import logging
 
+IDLE_STATE = 4294967295
```
mdigiorgio: To be renamed as `NON_IDLE_STATE`

derkling:
For this first run of review I have mainly updates related to data presentation.
The main updates I'm proposing for the first two plots are:

1. Use a figure title to embed a textual legend
2. Have all plots on a single column to make it easy compare different CPUs Clusters
3. Other minor eye candies to make the plots more similar to other ones we already produce in the trace_analysis module.

All the updates are collected into this patch:
```diff
diff --git a/libs/utils/trace_analysis.py b/libs/utils/trace_analysis.py
index 3728689..a6047cf 100644
--- a/libs/utils/trace_analysis.py
+++ b/libs/utils/trace_analysis.py
@@ -818,44 +818,48 @@ class TraceAnalysis(object):
         :type cpus: list(str)
         """
 
+        if not self.trace.hasEvents('cpu_frequency'):
+            logging.warn('Events [cpu_frequency] not found, '\
+                         'plot DISABLED!')
+            return None
+
         if not cpus:
-            cpus = sorted(self.platform['clusters']['little'] + self.platform['clusters']['big'])
+            # Generate plots only for available CPUs
+            cpufreq_data = self.trace.df('cpu_frequency')
+            cpus = range(cpufreq_data.cpu.max()+1)
         else:
             cpus.sort()
 
-        n_cpus = len(cpus)
-        if n_cpus == 1:
-            gs = gridspec.GridSpec(1, 1)
-        else:
-            gs = gridspec.GridSpec(n_cpus/2 + n_cpus%2, 2)
-
+        gs = gridspec.GridSpec(len(cpus), 1)
         fig = plt.figure()
+        plt.suptitle("OPP Residency Time by CPU (Green: Total, Red: Idle)",
+                     fontsize=16, horizontalalignment='center');
 
-        axis = None
-        for idx, cpu in enumerate(cpus):
+        # Plot CPUs residency, assuming big cores have higher IDs
+        for idx, cpu in enumerate(sorted(cpus, reverse=True)):
             total_time, active_time = self.getCPUFrequencyResidency(cpu)
 
-            axis = fig.add_subplot(gs[idx])
-            total_time.plot.barh(ax=axis,
+            xmax = 1.05 * total_time.max().values[0]
+            yrange = 0.3 * len(total_time) * len(cpus)
+
+            axes = fig.add_subplot(gs[idx])
+            total_time.plot.barh(ax=axes,
                                  color='g',
                                  legend=False,
-                                 figsize=(16,8),
-                                 fontsize=16);
-            active_time.plot.barh(ax=axis,
+                                 figsize=(16,yrange));
+            active_time.plot.barh(ax=axes,
                                   color='r',
                                   legend=False,
-                                  figsize=(16,8),
-                                  fontsize=16);
-            axis.set_title('CPU{}'.format(cpu))
-            axis.set_xlabel('Time [s]')
-            axis.set_xlim(0, self.x_max)
-
-        if axis:
-            handles, labels = axis.get_legend_handles_labels()
-            plt.figlegend(handles,
-                          labels,
-                          loc='lower right')
-            plt.subplots_adjust(hspace=0.5)
+                                  figsize=(16,yrange));
+
+            axes.set_title('CPU{}'.format(cpu))
+            axes.set_xlabel('Time [s]')
+            axes.set_xlim(0, xmax)
+
+            axes.grid(True);
+            if (idx+1 < len(cpus)):
+                axes.set_xticklabels([])
+                axes.set_xlabel('')
 
     def _getClusterActiveSignal(self, cluster):
         """
@@ -1002,7 +1006,7 @@ class TraceAnalysis(object):
         active_time.index.name = 'Frequency [MHz]'
         return total_time, active_time
 
-    def plotClusterFrequencyResidency(self, clusters=['big', 'little']):
+    def plotClusterFrequencyResidency(self, clusters=['big', 'LITTLE']):
         """
         Plot the frequency residency in a given cluster, i.e. the amount of
         time cluster `cluster` spent at frequency `f_i`. By default, both 'big'
@@ -1012,14 +1016,17 @@ class TraceAnalysis(object):
         :type clusters: list(str)
         """
 
-        if len(clusters) == 1:
-            gs = gridspec.GridSpec(1, 1)
-        else:
-            gs = gridspec.GridSpec(1, 2)
+        if not self.trace.hasEvents('cpu_frequency'):
+            logging.warn('Events [cpu_frequency] not found, '\
+                         'plot DISABLED!')
+            return None
+
+        gs = gridspec.GridSpec(len(clusters), 1)
         fig = plt.figure()
-        fig.subplots_adjust(hspace=.5);
+        plt.suptitle("OPP Residency Time by Cluster (Green: Total, Red: Idle)",
+                     fontsize=16, horizontalalignment='center');
 
-        axis = None
+        axes = None
         for idx, c in enumerate(clusters):
             total_time, active_time = self.getClusterFrequencyResidency(
                 self.platform['clusters'][c.lower()])
@@ -1028,25 +1035,26 @@ class TraceAnalysis(object):
                 plt.close(fig)
                 return
 
-            axis = fig.add_subplot(gs[idx])
-            total_time.plot.barh(ax = axis,
+            xmax = 1.05 * total_time.max().values[0]
+            yrange = 0.4 * len(total_time) * len(clusters)
+
+            axes = fig.add_subplot(gs[idx])
+            total_time.plot.barh(ax = axes,
                                  color='g',
                                  legend=False,
-                                 figsize=(16,8),
-                                 fontsize=16);
+                                 figsize=(16,yrange));
 
-            active_time.plot.barh(ax = axis,
+            active_time.plot.barh(ax = axes,
                                   color='r',
                                   legend=False,
-                                  figsize=(16,8),
-                                  fontsize=16);
-
-            axis.set_title('{} Cluster'.format(c))
-            axis.set_xlabel('Time [s]')
-            axis.set_xlim(0, self.x_max)
-
-        if axis:
-            handles, labels = axis.get_legend_handles_labels()
-            plt.figlegend(handles,
-                          labels,
-                          loc='lower right')
+                                  figsize=(16,yrange));
+
+            axes.set_title('{} Cluster'.format(c))
+            axes.set_xlabel('Time [s]')
+            axes.set_xlim(0, xmax)
+
+            axes.grid(True);
+            if (idx+1 < len(clusters)):
+                axes.set_xticklabels([])
+                axes.set_xlabel('')
+
```

libs/utils/trace_analysis.py
```diff
@@ -1009,52 +888,107 @@ def getClusterFrequencyResidency(self, cluster):
...
+            )
+            active_t = cluster_active * freq_active
+            # Sum up the time intervals by computing the integral of the square
+            # wave
+            nonidle_time.append(self._integrate_square_wave(active_t))
+
+        active_time = pd.DataFrame({"ACTIVE Time" : nonidle_time},
+                                   index=[f/1000 for f in available_freqs])
+        active_time.index.name = 'Frequency [MHz]'
         return total_time, active_time
```
derkling: Do you think it is possible to use a memoize decorator to save both active and total time once computed by this function: https://wiki.python.org/moin/PythonDecoratorLibrary#Memoize

This would allow to call the same function multiple times without redoing the computations.

Does it make sense?
I was thinking that for example this can be useful when we want to figure out the x_max across multiple clusters.
In this case we likely need to call the function one time for each cluster to compute the x_max and than again to plot the values.

libs/utils/trace_analysis.py
```diff
@@ -1009,52 +888,107 @@ def getClusterFrequencyResidency(self, cluster):
...
-            t_prev = t
-            f_prev = f
-
-        active_time = pd.DataFrame({
-            'ACTIVE Time' : time_intervals,
-            'Frequency [MHz]' : [f/1000 for f in frequencies]
+        :param cluster: list of CPU IDs belonging to the cluster
+        :type cluster: list(int)
+        """
+
```
derkling: What about allowing cluster to be just a string if we want just one cluster?

In this case, this should make the trick for the following code:
```python
if not isinstance(cluster, list):
    cluster = [cluster]
```

mdigiorgio: Yes, we could allow both the list of strings and a single string...

derkling: well... here cluster is a list of CPU... but still can we have "clusters" of just a CPU?

libs/utils/trace_analysis.py
```diff
@@ -1009,52 +888,107 @@ def getClusterFrequencyResidency(self, cluster):
...
+            logging.warn('Events [cpu_idle] not found, '\
+                         'plot DISABLED!')
+            return None
+
+        freq_dfr = self.trace.df('cpu_frequency')
+        cluster_freqs = freq_dfr[freq_dfr.cpu.isin(cluster)]
+        idle_dfr = self.trace.df('cpu_idle')
+
+        ### Computer TOTAL Time ###
+        time_intervals = cluster_freqs.index[1:] - cluster_freqs.index[:-1]
```
derkling: It seems to me that here you are computing time intervals by subtracting the index of every two successive events.
Is that right?

In that case you end up with intervals for frequency events from the same CPU, which imply that you are assuming all the CPUs are sync wrt frequency switching.

In that case would not be simpler to just consider the events of the first CPU of the cluster?

libs/utils/trace_analysis.py
JaviMerino: In trappy we have `listify()` that does exactly what @derkling describes in the previous commit.  We use it a lot for this type of arguments, that can either be a single element and a list.  I don't think it's worth using the trappy implementation here, but maybe it's worth copying the definition.

derkling:
Just realized that the second patch refactors the first one... for next posting we should cleanup the set and send two fresh&clean incremental patches ;-)

libs/utils/trace_analysis.py
mdigiorgio: sorry I meant list of int (CPU IDs) and a string being either `"big"` or `"little"`

libs/utils/trace_analysis.py
```diff
@@ -951,46 +846,30 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
             plt.subplots_adjust(hspace=0.5)
 
-    def getClusterFrequencyResidency(self, cluster):
+    def _getClusterActiveSignal(self, cluster):
         """
-        Get a DataFrame with per cluster frequency residency, i.e. amount of
-        time spent at a given frequency in each cluster.
+        Build a square wave representing the active (i.e. non-idle) cluster
+        time, i.e.:
+            cluster_active[t] = 1 if there is a task running in the cluster at
```
derkling: That comment is misleading. We are looking at cpu_idle events, thus maybe better to say that:
```
   cluster_active[t] == 1 if the CPU is reported to be exited from an idle state by a cpu_idle event
```

libs/utils/trace_analysis.py
```diff
@@ -951,46 +846,30 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
 
             # time instants where CPU goes to an idle state
-            idle_times = cpu_idle[cpu_idle['state'] != IDLE_STATE]
+            idle_times = cpu_idle[cpu_idle.state != IDLE_STATE]
             # time instants where CPU leaves an idle state
-            non_idle_times = cpu_idle[cpu_idle['state'] == IDLE_STATE]
+            non_idle_times = cpu_idle[cpu_idle.state == IDLE_STATE]
 
             if idle_times.index[0] < non_idle_times.index[0]:
                 idle_times = idle_times.iloc[1:]
```
derkling: Is seems to me that here you are trying to force "the first IDLE_EXIT to follow the first IDLE_ENTER", is that correct?
Why do we need that?
... and here I'm lost because the code in my branch is different that what I see on that patch... :-/

libs/utils/trace_analysis.py
mdigiorgio: Should be possible yes ;).

In the patch on the previous comment, the title should be "...Red: Active" rather than "...Red: Idle".

mdigiorgio: Also, why do you need to sort the cpus again in the for loop? And why do you need to assume big CPUs have higher IDs?

libs/utils/trace_analysis.py
mdigiorgio: Correct, this works under the assumption that CPUs belonging to a cluster cannot run at different frequencies.

Therefore, the answer to your second question is yes we can just consider one CPU in the cluster. I'll write a comment explaining this

libs/utils/trace_analysis.py
mdigiorgio: This was needed before the optimisation introduced in the last commit, because sometimes the two dataframes had different  length. I think this is not needed any more, because by building square wave signal that tells us when the CPU is active we do not care about computing time differences.

I'll fix this and the following lines as well to make more clear what we want to get.

derkling:
This is now tracked by #132

PR #88, opened 2016-04-14
ipynb: add ChromeOS-related notebook section

mdigiorgio:


ipynb/chromeos/ChromeOS_Tests.ipynb
```diff
@@ -0,0 +1,827 @@
...
+    "**NOTE**: if you want to receive a token from the benchmark before starting trace collection and power measurement, you need to modify the benchmark such that a specific UDP packet is sent to the host machine. As a reference, consider the following python script and convert it to the language in which the benchmark is written accordingly:\n",
+    "\n",
+    "```python\n",
+    "import socket\n",
+    "sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)\n",
+    "sock.sendto(\"POWER\", (<TARGET_IP>, 1234))\n",
+    "sock.close()\n",
+    "```\n",
+    "\n",
+    "You should put this right before the instruction that starts the execution of the actual workload in the benchmark."
```
derkling: Can we add here, as an example, the description of how to apply this patch for WebGL acquarium?
I mean, describe where the file with the benchmark is located (relative to a ChromeOS installation) and in which function it has to be placed?
This can be useful as a reference for ourself too.

ipynb/chromeos/ChromeOS_Tests.ipynb
```diff
@@ -0,0 +1,827 @@
...
+    "    \"password\"    : 'test0000',\n",
+    "    \n",
+    "    # RTApp calibration values (comment to let LISA do a calibration run)\n",
+    "    \"rtapp-calib\" :  {\n",
+    "        \"0\": 179, \"1\": 177, \"2\": 71, \"3\": 71\n",
+    "    },\n",
+    "    \n",
+    "    # Tools required by the experiments\n",
+    "    #\"tools\"   : [ 'rt-app', 'trace-cmd' ],    \n",
+    "    \"exclude_modules\" : ['hwmon'],\n",
```
derkling: This should not be required... I do not think we load HWMON by default on OAK... but better to check

ipynb/chromeos/ChromeOS_Tests.ipynb
```diff
@@ -0,0 +1,827 @@
...
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Get host IP\n",
+    "s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)\n",
+    "s.connect((\"gmail.com\",80))\n",
```
derkling: Interesting trick! :-)
But we do really need to setup a connection to get the local IP?
This breaks the notebook if we run it without internet access... maybe there is a Python API to get the local IP address?

ipynb/chromeos/ChromeOS_Tests.ipynb
```diff
@@ -0,0 +1,827 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "········\n"
+     ]
+    }
+   ],
+   "source": [
+    "# ask user for host password\n",
+    "password = getpass.getpass()"
```
derkling: Does this requires to input the password each time?
Otherwise we can try to use the "standard" GNOME keyring to store and recover the password in an automated way each time we execute the cell.
Here is an example of code I use for my dotfiles:
   https://github.com/derkling/dotfiles/blob/master/bin/gnome_keyring.py
We just need the set_credentials and get_credentials functions.

ipynb/chromeos/ChromeOS_Tests.ipynb
```diff
@@ -0,0 +1,827 @@
...
+      "    \"avg_interframe_time_1000_fishes{perf}\": 0.0273612376448, \n",
+      "    \"avg_interframe_time_0050_fishes{perf}\": 0.0203918644533, \n",
+      "    \"avg_fps_0050_fishes{perf}\": 49.0385253742\n",
+      "}\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Print Acquarium results\n",
+    "print json.dumps(acquarium_res['results'], indent=4)"
```
derkling: What about something like:
```python
df = pd.DataFrame.from_dict(results, orient="index")
df.columns = ['Values']
df
```

ipynb/chromeos/ChromeOS_Tests.ipynb
```diff
@@ -0,0 +1,827 @@
...
+    "x = big_pwr.index.get_values()\n",
+    "y = big_pwr.Power.get_values()\n",
+    "bnrg = integrate.simps(y, x=x)\n",
+    "print \"big Cluster Measured Energy = {} mJ\".format(bnrg)\n",
+    "\n",
+    "x = little_pwr.index.get_values()\n",
+    "y = little_pwr.Power.get_values()\n",
+    "lnrg = integrate.simps(y, x=x)\n",
+    "print \"LITTLE Cluster Measured Energy = {} mJ\".format(lnrg)\n",
+    "print \"Total Measured Energy = {} mJ\".format(bnrg + lnrg)"
```
derkling: As before, what about:
```python
results = {
    'big Cluster' : bnrg,
    'LITTLE Cluster' : lnrg,
    'Total' : bnrg + lnrg,
}

df = pd.DataFrame.from_dict(results, orient="index")
df.columns = ['Energy [mJ]']
df
```

README.md
```diff
@@ -46,6 +46,7 @@ Debian system.
 ##### Install (upgrade) required Python libraries
 
 	$ sudo pip install --upgrade trappy bart-py devlib
+	$ sudo pip install netifaces
```
derkling: Can't this be installed just via apt-get?
This allows to use pip only for main dependency libs...

mdigiorgio: Yes, it can be installed via apt-get, but since it is a python package isn't it more reliable to install it via pip?

derkling: well, we should define a reasonable strategy... so far I've used apt-get for python modules which are not direct LISA submodules. That's why there is a pip install command for just trappy&friends...

mdigiorgio: OK, let's add it next to `python-nose` then

derkling:
This seems to be a broken PR... the notebook is either empty or broken. It does not load.

PR #95, opened 2016-04-27
Fix LISA-shell paths

msrasmussen:
lisa-help only works when called from the LISA root folder. Fix this by
storing LISA_HOME correctly and add a useful error message if someone
tries to initialize the LISA-shell outside the LISA root folder.

Signed-off-by: Morten Rasmussen <morten.rasmussen@arm.com>


derkling:
This seems to be fixed in mainline now.

PR #96, opened 2016-04-27
lisa: Add support for new Hwmon node names with 'mainline' driver

credp:
The names of the nodes for energy have changed - they are no longer
of the form 'a53_energy' for the CPU clusters, but are now called
'BOARD_LITTLE_ENERGY'. Likewise, all the other sensor names are different
but we don't monitor those in lisa.

In order to support those, add a new device named juno2. Once both types
are no longer in circulation, the names can be changed.

### DIFF
--- a/libs/utils/energy.py
+++ b/libs/utils/energy.py
@@ -40,6 +40,13 @@ DEFAULT_ENERGY_METER = {
             'kinds' : [ 'energy' ]
         }
     },
+    'juno2' : {
+        'instrument' : 'hwmon',
+        'conf' : {
+            'sites' : [ 'BOARDLITTLE', 'BOARDBIG' ],
+            'kinds' : [ 'energy' ]
+        }
+    },
 
     # Hikey: by default use AEP
     'hikey' : {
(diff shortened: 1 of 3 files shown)


libs/utils/energy.py
```diff
@@ -167,6 +174,10 @@ def report(self, out_dir, out_file='energy.json'):
                 clusters_nrg['LITTLE'] = '{:.6f}'.format(nrg_total)
             elif self._target.big_core.upper() in ch.upper():
                 clusters_nrg['big'] = '{:.6f}'.format(nrg_total)
+            elif 'LITTLE' in ch.upper():
```
derkling: What about using an OR on the previous check at line 166/173?

libs/utils/env.py
```diff
@@ -343,6 +343,9 @@ def _init_target(self, force = False):
         elif self.conf['board'].upper() == 'JUNO':
             platform = devlib.platform.arm.Juno()
             self.__modules = ['bl', 'hwmon', 'cpufreq']
+        elif self.conf['board'].upper() == 'JUNO2':
```
derkling: What about using an OR on the previous check at line 343?

libs/utils/energy.py
credp: Sure, if you prefer it like that. Let me change it :)

credp:
I have reworked this a bit, I'll post a new branch and close this one.

PR #97, opened 2016-04-27
lisa: Add support for new Hwmon node names with 'mainline' driver

credp:
The names of the nodes for energy have changed - they are no longer
of the form 'a53_energy' for the CPU clusters, but are now called
'BOARD_LITTLE_ENERGY'. Likewise, all the other sensor names are different
but we don't monitor those in lisa.

In order to support those, add a new device named juno2. Once both types
are no longer in circulation, the names can be changed.


derkling:
Big improvements, thanks for the effort and the contribution! ;-)

PR #103, opened 2016-05-03
submodules: update trappy to v5.4.0

JaviMerino:


derkling:
Is it legal to reject or even just delay such a beloved PR?!?

However, I would like to get this opportunity to improve the lisa-update script. Right now we need to pull LISA::master and than call lisa-update... which seems to not make a lot of sense.
It would be probably better to extend the lisa-update command to pull master before to update submodules. What do you think?

If you agree we can try to merge before the update on lisa-update and than use this new HEAD to verify that the trappy bump is properly pulled with just a single lisa-update command...

derkling:
Just to be clear... the repose to the first question was a bold NO! ;-)

PR #104, opened 2016-05-04
lisa-update all

JaviMerino:
Teach lisa-update to update itself.  As requested in #103

### DIFF
--- a/src/shell/lisa_shell
+++ b/src/shell/lisa_shell
@@ -73,15 +73,15 @@ echo -ne "$LISASHELL_DEFAULT"
 
 function _lisa-update-usage {
     echo "Usage: lisa-update (CMD)"
-	echo "  CMD: what to update (default: submodules)"
+	echo "  CMD: what to update (default: all)"
+	echo "       all - update lisa and all the external dependencies"
 	echo "       submodules - update external dependencies provided by submodules"
 	echo
     echo "Update submodules (if DEVMODE enabled)"
 }
 
-function _lisa-update-all {
+function _lisa-update-submodules {
 echo -ne "${LISASHELL_BLUE}"
-echo -e "*** Update LISA installation"
 if [ "x$DEVMODE" == "x1" ]; then
 	# Force update existing modules
 	echo
@@ -93,10 +93,55 @@ fi
 echo -ne "$LISASHELL_DEFAULT"
 }
 
+function _lisa-update-all {
+echo -ne "${LISASHELL_BLUE}"
+echo -e "*** Update LISA installation"
+
+git update-index -q --refresh
+ret=$?
+if [ $ret -ne 0 ]; then
+    echo "LISA internal error: git update-index failed"
+    echo "Please report it: https://github.com/ARM-software/lisa/issues"
+    return $ret
+fi
+
+git diff-index --quiet HEAD
+ret=$?
+if [ $ret -ne 0 ]; then
+    echo "There are outstanding uncommitted changes."
+    echo "Please, commit your changes or stash them before you can update lisa"
+    return $ret
+fi
+
+curr_commit=$(git rev-parse HEAD)
+remote_name=$(git remote -v | grep ARM-software/lisa.git | grep -m 1 fetch | awk '{print $1}')
+git merge-base --is-ancestor $curr_commit $remote_name/master
+ret=$?
+if [ $ret -ne 0 ]; then
+    echo "You have committed changes that are not part of $remote_name/master"
+    echo "Please move to the master branch before running lisa-update"
+    return $ret
+fi
+
+git pull --ff-only $remote_name master
+ret=$?
+if [ $ret -ne 0 ]; then
+    # git pull should have printed some error.  Abort and propagate the error code.
+    return $ret
+fi
+
+_lisa-update-submodules
+
+echo -ne "$LISASHELL_DEFAULT"
+}
+
 function lisa-update {
 CMD=${1:-all}
 echo
 case "x${CMD^^}" in
+'xSUBMODULES')
+	_lisa-update-submodules
+	;;
 'xALL')
 	_lisa-update-all
 	;;


src/shell/lisa_shell
```diff
@@ -93,13 +93,27 @@ fi
 echo -ne "$LISASHELL_DEFAULT"
 }
 
+function _lisa-update-all {
+echo -ne "${LISASHELL_BLUE}"
+echo -e "*** Update LISA installation"
+
+git pull
```
derkling: Is this not going to produce a merge commit if by chance we issue a "lisa-update all" while not being on origin/master?

Maybe we can either:
1) check before to be on origin/master (a local master branch still is not valid, since it can contain local user patches) and abort with a warning explaining that this operation is possible only while being on origin/master
2) replace this pull with:
``` shell
git fetch origin master
```
Which should safely update the origin/master head and call a sub-modules update.
Unfortunately the submodules update will not see the new origin/master... but this can be solved with a trick:
BEFORE the submodules update
- stash current modifications with a certain name (e.g. __lisa_update)
- get a reference to the current branch (CURR_BRANCH)
- checkout origin/master as a detached branch
- UPDATE submodules at this time
- go back and checkout CURR_BRANCH
- stash pop "__lisa_update"

Do you think this can work and be useful?

Otherwise I would just go for solution 1 and warn the user that she needs to be in origin/master to do an update.
It's on the user responsibilities to save the current work...

src/shell/lisa_shell
```diff
@@ -93,13 +93,27 @@ fi
...
+
+_lisa-update-submodules
+
+echo -ne "$LISASHELL_DEFAULT"
+}
+
 function lisa-update {
 CMD=${1:-submodules}
```
derkling: I would propose to set "all" as default, since this command does nothing unless we update origin/master, right?

JaviMerino: it does if you've manually updated `origin/master`.  In any case, if you prefer "all" to be the default I'll change it.

src/shell/lisa_shell
JaviMerino: [ github ate my reply so resending :cry: ]

I prefer 1). I would do:

1. Abort if there are any uncommitted changes
2. Check that the top-most commit is part of `origin/master`.  If it isn't that means that the user has commits that aren't part of upstream, so we abort.
3. `git pull origin master`.  I went for just `git pull` because I didn't want to make an assumption about the names of the remote but I guess you're right that it's very likely that `origin` points to this repository.

What do you think?

derkling: I think that solution it's sound and reasonably safe at the same time ;-)
Most of the users will probably always be on master... while, it they get a bice and bold
   WARNING: you have uncommitted changes
or
  WARNING: you are not on master branch
they will know what is required before proceeding with the update.

JaviMerino: Ok, I've updated the pull request and incorporated your suggestions :smile:

src/shell/lisa_shell
```diff
@@ -93,10 +93,49 @@ fi
...
 }
 
+function _lisa-update-all {
+echo -ne "${LISASHELL_BLUE}"
+echo -e "*** Update LISA installation"
+
+git update-index -q --refresh
+ret=$?
+if [ $ret -ne 0 ]; then
+    echo "lisa internal error: git update-index failed.  Please report it: https://github.com/ARM-software/lisa/issues"
```
derkling: Can we split this in two lines, like you did for the following messages, i.e.:

```bash
echo "LISA internal error: git update-index failed"
echo "Please report it: https://github.com/ARM-software/lisa/issues"
```

Notice the usage of LISA (capitalized)... I would prefer to use that for whenever we refer to the acrimonious.

src/shell/lisa_shell
```diff
@@ -93,10 +93,49 @@ fi
...
+if ! git diff-index --quiet HEAD; then
+    echo "There are outstanding uncommitted changes."
+    echo "Please, commit your changes or stash them before you can update lisa"
+    return 1
+fi
+
+curr_commit=$(git rev-parse HEAD)
+if ! git merge-base --is-ancestor $curr_commit origin/master; then
+    echo "You have committed changes that are not part of origin/master"
+    echo "Please move to the master branch"
```
derkling: "Please move to the master branch..." I would add also  "before running the lisa-update command"

src/shell/lisa_shell
```diff
@@ -93,10 +93,49 @@ fi
...
+    return 1
+fi
+
+git pull --ff-only origin master
```
derkling: So at the end we assume that "origin" is named origin?
Was that something you was against?
If we can make it more generic for me it's ok... perhaps because of the previous comments now we know that we are on master, isn't it?

src/shell/lisa_shell
```diff
@@ -93,10 +93,49 @@ fi
...
+echo -e "*** Update LISA installation"
+
+git update-index -q --refresh
+ret=$?
+if [ $ret -ne 0 ]; then
+    echo "lisa internal error: git update-index failed.  Please report it: https://github.com/ARM-software/lisa/issues"
+    return $ret
+fi
+
+if ! git diff-index --quiet HEAD; then
```
derkling: Just for consistency, what about using:
```bash
git diff-index --quiet HEAD
ret=$?
[ $ret -ne 0 ]; then
   .....
```

derkling:
Apart from some minor corrections I think we have now a much more robust update procedure.
Thanks for the effort :+1:

src/shell/lisa_shell
JaviMerino: and do
``` bash
    return $ret
```
instead of making up a return code.  Yes, that's better.

src/shell/lisa_shell
JaviMerino: I've changed it to:
``` bash
    echo "Please move to the master branch before running lisa-update"
```

src/shell/lisa_shell
JaviMerino: You're right, I don't like the hardcoding of origin.  I guess it could be changed to:
``` bash
origin_name=$(git remote -v | grep ARM-software/lisa.git | grep -m 1 fetch | awk '{print $1}')
```

Do you prefer that?

JaviMerino:
I've updated the pull request again and incorporated your suggestions.  I've done the generic remote name, I hope that the logic doesn't add too much complexity.

src/shell/lisa_shell
```diff
@@ -93,10 +93,55 @@ fi
...
+git diff-index --quiet HEAD
+ret=$?
+if [ $ret -ne 0 ]; then
+    echo "There are outstanding uncommitted changes."
+    echo "Please, commit your changes or stash them before you can update lisa"
+    return $ret
+fi
+
+curr_commit=$(git rev-parse HEAD)
+remote_name=$(git remote -v | grep ARM-software/lisa.git | grep -m 1 fetch | awk '{print $1}')
```
derkling: :heart_eyes:

derkling:
Thanks for all the passionate effort... let's merge it! ;-)

PR #107, opened 2016-05-10
lisa_shell: bind ipython server to eth0 when run from a vagrant environment

JaviMerino:
When running inside vagrant with the default VirtualBox configuration, the host accesses the vm using a virtual ethernet link.  `lisa-ipython start` uses loopback as the default interface, which makes the ipython server bind only to loopback and refuse connections from the host.

Make `lisa-ipython start` bind to eth0 by default, which is the more sensible default in that environment.

Reported by @raw-bin


mdigiorgio:
Thanks for this :+1:

derkling:
love it!

PR #108, opened 2016-05-10
lisa_shell: document the PORT parameter of lisa-ipython

JaviMerino:
0bb8438f0157 ("lisa_shell: add support for multiple instances") introduced a third parameter to lisa-ipython.  After the network interface you can set the port.  Document it in lisa-ipython's usage.


derkling:
Thanks for spotting this... +1

PR #102, opened 2016-04-30
README: add instructions on VM based setup

derkling:
The easiest way to install and run LISA tools is probably by using a VM.
This patch adds a minimum set of instructions on how to build and run
a VitualBox image using the Vagrant tool to automate the build and
start of the VM.

PR #113, opened 2016-05-11
Vagrantfile: update virtual machine creation

### LINKED ISSUE #117: vagrant: add support for Android SDK
1. Install the android SDK (android-sdk_r24.4.1-linux.tgz cmdline tool - upgrade after the installation for the platform-tools) in the *same share area* where vagrant is sharing the host fs folder.
2. define ANDROID_HOME (I added the platform-tools to the PATH also) in the bash.bashrc of the VM instance created

For the first point I would suggest to install the Android SDK on the same lisa source tree, which is mounted under /vagrant. We have a lisa/tools folder, we can unzip the SDK in there so that it is available for all the VM and/or container we want to use starting from the LISA source tree.
The Vagrant provisioning script thus has to check before if the SDK is already available, otherwise download and install it under  /vagrant/tools/android-sdk. This is the path used for the ANDROID_HOME.

mdigiorgio:
This series addresses the following issues:

* Avoiding warnings with locales
* Installing Android SDK under `lisa/tools` if not yet present and setting `ANDROID_HOME` properly

Solves one of the warnings reported in #102.

close #117 

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,7 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+
 Vagrant.configure(2) do |config|
   config.vm.box = "ubuntu/trusty64"
 
@@ -14,17 +15,40 @@ Vagrant.configure(2) do |config|
 
   config.vm.provision "shell", inline: <<-SHELL
     sudo apt-get update
-    sudo apt-get install -y autoconf automake build-essential git libfreetype6-dev libpng12-dev libtool nmap pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd tree
+    sudo apt-get install -y autoconf automake build-essential expect git \
+        libfreetype6-dev libpng12-dev libtool nmap openjdk-7-jdk \
+        openjdk-7-jre pkg-config python-all-dev python-matplotlib \
+        python-nose python-numpy python-pip python-zmq sshpass trace-cmd \
+        tree wget
     sudo pip install ipython[notebook] pandas
     sudo apt-get remove -y w3m
 
     ln -s /vagrant /home/vagrant/lisa
+
+    cd /home/vagrant/lisa
+    ANDROID_SDK_URL="https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz"
+    if [ ! -e ./tools/android-sdk-linux ]; then
+        echo "Downloading Android SDK [$ANDROID_SDK_URL]..."
+        wget -qO- $ANDROID_SDK_URL | tar xz -C tools
+        expect -c '
+            set timeout -1;
+            spawn ./tools/android-sdk-linux/tools/android update sdk --no-ui \
+                -t tool,platform-tool,platform;
+            expect {
+                "Do you accept the license" { exp_send "y\r" ; exp_continue }
+                eof
+            }
+        '
+    fi
+
     chown vagrant.vagrant /home/vagrant/lisa
     echo cd /home/vagrant/lisa >> /home/vagrant/.bashrc
     for LC in $(locale | cut -d= -f1);
     do
         echo unset $LC  >> /home/vagrant/.bashrc
     done
+    echo "export ANDROID_HOME=/vagrant/tools/android-sdk-linux" >> /home/vagrant/.bashrc
+    echo "export PATH=$ANDROID_HOME/platform-tools:$ANDROID_HOME/tools:$PATH" >> /home/vagrant/.bashrc
     echo source init_env >> /home/vagrant/.bashrc
   SHELL
 end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -50,5 +50,17 @@ Vagrant.configure(2) do |config|
     echo "export ANDROID_HOME=/vagrant/tools/android-sdk-linux" >> /home/vagrant/.bashrc
     echo "export PATH=$ANDROID_HOME/platform-tools:$ANDROID_HOME/tools:$PATH" >> /home/vagrant/.bashrc
     echo source init_env >> /home/vagrant/.bashrc
+
+    echo "Virtual Machine Installation completed successfully!                "
+    echo "                                                                    "
+    echo "You can now access and use the virtual machine by running:          "
+    echo "                                                                    "
+    echo "    $ vagrant ssh                                                   "
+    echo "                                                                    "
+    echo "NOTE: if you exit, the virtual machine is still running. To shut it "
+    echo "      down, please run:                                             "
+    echo "                                                                    "
+    echo "    $ vagrant suspend                                               "
+    echo "                                                                    "
   SHELL
 end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -21,6 +21,10 @@ Vagrant.configure(2) do |config|
     ln -s /vagrant /home/vagrant/lisa
     chown vagrant.vagrant /home/vagrant/lisa
     echo cd /home/vagrant/lisa >> /home/vagrant/.bashrc
+    for LC in $(locale | cut -d= -f1);
+    do
+        echo unset $LC  >> /home/vagrant/.bashrc
+    done
     echo source init_env >> /home/vagrant/.bashrc
   SHELL
 end
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 results*
 *.json
 *.log
+tools/android-sdk-linux


Vagrantfile
```diff
@@ -21,6 +21,10 @@ Vagrant.configure(2) do |config|
     ln -s /vagrant /home/vagrant/lisa
     chown vagrant.vagrant /home/vagrant/lisa
     echo cd /home/vagrant/lisa >> /home/vagrant/.bashrc
+	for LC in $(locale | cut -d= -f1);
```
derkling: Can you repost using spaces instead of tabs... just to be consistent within the file... ;-)

Vagrantfile
```diff
@@ -21,6 +21,10 @@ Vagrant.configure(2) do |config|
     ln -s /vagrant /home/vagrant/lisa
     chown vagrant.vagrant /home/vagrant/lisa
     echo cd /home/vagrant/lisa >> /home/vagrant/.bashrc
+    for LC in $(locale | cut -d= -f1);
+    do
+        echo export $LC=""  >> /home/vagrant/.bashrc
```
JaviMerino: Being pedantic, it's better to unset the variable:
``` bash
echo unset $LC >> /home/vagrant/.bashrc
```

derkling:
Can we add to this PR another set of patches to address Basil's comments?
I've collected the requirements in #117

mdigiorgio:
:+1: going to do it

mdigiorgio:
@derkling I pushed a new commit with what you've asked. When you have time could you please have a look at it?

For the Android APIs either we explicitly set which one to install in the filter option (we can only use numbers that are going to change) or we install everything (as it is currently).

Vagrantfile
```diff
@@ -14,17 +14,34 @@ Vagrant.configure(2) do |config|
 
   config.vm.provision "shell", inline: <<-SHELL
     sudo apt-get update
-    sudo apt-get install -y autoconf automake build-essential git libfreetype6-dev libpng12-dev libtool nmap pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd
+    sudo apt-get install -y autoconf automake build-essential git libfreetype6-dev libpng12-dev libtool nmap pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd openjdk-7-jdk openjdk-7-jre expect
```
JaviMerino: Please keep the alphabetical order.  I know, I can be very pedantic :smirk:

Vagrantfile
```diff
@@ -14,17 +14,34 @@ Vagrant.configure(2) do |config|
...
     sudo pip install ipython[notebook] pandas
     sudo apt-get remove -y w3m
 
     ln -s /vagrant /home/vagrant/lisa
+
+    cd /home/vagrant/lisa
+    if [ ! -d ./tools/android-sdk-linux ]; then
+        wget -q https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz
```
JaviMerino: Why the `-q`?  Showing some progress so that the user is not wondering why is this stalled is a good thing.  The last thing you can read before this is that w3m is being removed.  I would leave `wget`'s output, but if that's not possible or it's too noisy, then at least put an `echo Downloading Android SDK...` before executing `wget`.

Vagrantfile
```diff
@@ -14,17 +14,34 @@ Vagrant.configure(2) do |config|
...
+        tar xf android-sdk_r24.4.1-linux.tgz -C tools
+		rm android-sdk_r24.4.1-linux.tgz
```
JaviMerino: Remove the extra tab

Vagrantfile
```diff
@@ -14,17 +14,34 @@ Vagrant.configure(2) do |config|
...
+    sudo apt-get install -y autoconf automake build-essential git libfreetype6-dev libpng12-dev libtool nmap pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd openjdk-7-jdk openjdk-7-jre expect
     sudo pip install ipython[notebook] pandas
     sudo apt-get remove -y w3m
 
     ln -s /vagrant /home/vagrant/lisa
+
+    cd /home/vagrant/lisa
+    if [ ! -d ./tools/android-sdk-linux ]; then
+        wget -q https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz
+        tar xf android-sdk_r24.4.1-linux.tgz -C tools
```
JaviMerino: Add `tools/android-sdk-linux` to `.gitignore`

Vagrantfile
```diff
@@ -14,17 +14,34 @@ Vagrant.configure(2) do |config|
...
     sudo apt-get update
-    sudo apt-get install -y autoconf automake build-essential git libfreetype6-dev libpng12-dev libtool nmap pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd
+    sudo apt-get install -y autoconf automake build-essential git libfreetype6-dev libpng12-dev libtool nmap pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd openjdk-7-jdk openjdk-7-jre expect
     sudo pip install ipython[notebook] pandas
     sudo apt-get remove -y w3m
 
     ln -s /vagrant /home/vagrant/lisa
+
+    cd /home/vagrant/lisa
+    if [ ! -d ./tools/android-sdk-linux ]; then
```
JaviMerino: What about `-e` instead of `-d`?  That way the user can symlink the android-sdk-linux installation that they have installed in the system if that's the case.

Vagrantfile
```diff
@@ -14,17 +14,35 @@ Vagrant.configure(2) do |config|
 
   config.vm.provision "shell", inline: <<-SHELL
     sudo apt-get update
-    sudo apt-get install -y autoconf automake build-essential git libfreetype6-dev libpng12-dev libtool nmap pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd
+    sudo apt-get install -y autoconf automake build-essential expect git libfreetype6-dev libpng12-dev libtool nmap openjdk-7-jdk openjdk-7-jre pkg-config python-all-dev python-matplotlib python-nose python-numpy python-pip python-zmq sshpass trace-cmd
```
derkling: Can we split these lines? The usual "\" syntax should work... and we can keep the 80 cols...

Vagrantfile
```diff
@@ -14,17 +14,35 @@ Vagrant.configure(2) do |config|
...
     sudo pip install ipython[notebook] pandas
     sudo apt-get remove -y w3m
 
     ln -s /vagrant /home/vagrant/lisa
+
+    cd /home/vagrant/lisa
+    if [ ! -e ./tools/android-sdk-linux ]; then
+        echo Downloading Android SDK...
```
derkling: Better quoting the text?

Vagrantfile
```diff
@@ -14,17 +14,35 @@ Vagrant.configure(2) do |config|
...
+        wget -q https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz
```
derkling: Let's use a var for this path and report the same in the previous echo:
```bash
ANDROID_SDJK_URL=https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz
...
echo "Downloading Android SDK [$ANDROID_SDJK_URL]..."
```

derkling: Should we add wget to the list of apt-get installed packages... just to be sure...

Vagrantfile
```diff
@@ -14,17 +14,35 @@ Vagrant.configure(2) do |config|
...
+        tar xf android-sdk_r24.4.1-linux.tgz -C tools
+        rm android-sdk_r24.4.1-linux.tgz
```
derkling: Maybe the previous tree lines can be replaced with something like:
```
wget -qO- $ANDROID_SDK_URL | tar xz -C tools
```

Vagrantfile
```diff
@@ -14,17 +15,37 @@ Vagrant.configure(2) do |config|
...
+        openjdk-7-jre pkg-config python-all-dev python-matplotlib \
+        python-nose python-numpy python-pip python-zmq sshpass trace-cmd wget
     sudo pip install ipython[notebook] pandas
     sudo apt-get remove -y w3m
 
     ln -s /vagrant /home/vagrant/lisa
+
+    cd /home/vagrant/lisa
+    if [ ! -e ./tools/android-sdk-linux ]; then
+		ANDROID_SDK_URL="https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz"
```
derkling: Either fix indent or move it before the if

mdigiorgio: Updated version online!

Vagrantfile
```diff
@@ -14,17 +15,39 @@ Vagrant.configure(2) do |config|
...
+        '
+    fi
+
     chown vagrant.vagrant /home/vagrant/lisa
     echo cd /home/vagrant/lisa >> /home/vagrant/.bashrc
     for LC in $(locale | cut -d= -f1);
     do
         echo unset $LC  >> /home/vagrant/.bashrc
     done
+    echo "export ANDROID_HOME=/vagrant/tools/android-sdk-linux" >> /home/vagrant/.bashrc
```
derkling: We should add also:
```bash
echo "export PATH=$ANDROID_HOME/platform-tools:$ANDROID_HOME/tools:$PATH" >> /home/vagrant/.bashrc
```

Vagrantfile
```diff
@@ -14,17 +15,39 @@ Vagrant.configure(2) do |config|
...
     echo source init_env >> /home/vagrant/.bashrc
   SHELL
```
derkling: Before returning from the provisioning script, let's notice the user that the VM is ready and she can access and use it by running the
```bash
vagrant ssh
```

And notice that if you exit the SSH the machine is still running and this command
```bash
vagrant suspend
```
to actually shout down the machine when not required.

PR #116, opened 2016-05-17
ipynb: energy_model: add example for energy model analysis

Leo-Yan:
Add ipynb script to analyse energy model and display the power
efficiency and capacity between two clusters.

This will be helpful to check if energy model is reasonable or not,
and it's straightforward to get to know if there have overlap between
two cluster's energy model; If have overlap, which it's likely to
introduce wrong behavior for EAS energy calculation.

### DIFF
--- a/ipynb/utils/energy_model_example.ipynb
+++ b/ipynb/utils/energy_model_example.ipynb
@@ -1,0 +1,134 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "\n",
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "from __future__ import division\n",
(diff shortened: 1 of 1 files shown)


derkling:
It would be great if such a notebook can be converted into a proper EM analysis module, under libs/utils, which uses data collected from platform.json to plot the OPP curves as well as to highlight and report suboptimal points... I would suggest to open a new issue to track this as a new feature request.

IMO, this notebook for the time being is not generic enough to be integrated.
However, it can still be useful to provide an example... in any case I would like to have a dedicated section, e.g. ipynb/energy_model where we can collect examples and more advances notebooks.
For example we have already in mind to add example notebooks to show how to build an energy model.

derkling:
We now have an "official" set of Notebooks to generate energy models which also produce a predefined and "standard" set of EM representation. These notebooks have been merged in #172 and they are available here:
   https://github.com/ARM-software/lisa/tree/master/ipynb/energy

It would be nice if we can integrate your plots to the ones already available in these official notebooks.
Thus, let's close this PR for the time being.

PR #120, opened 2016-05-31
[Android] add wrapper for workload-automation

mdigiorgio:
Workload-automation (WA) already provides many workloads for Android targets.
This is a wrapper for LISA to use WA.

In order to use one of the WA workloads (`wa list workloads` tells you what is
available) you need to implement a <workload__class_name> class in a
<workload_name>.py file stored under libs/utils/android/workloads.

<workload_class_name> inherits Wlauto and has to implement the methods from
Workload and (eventually) Wlauto.

This way, the workload is visible by the Workload class that will list it as
available and will eventually check that the related package is installed on the
Android target.

Signed-off-by: Michele Di Giorgio michele.digiorgio@arm.com

### DIFF
--- a/libs/utils/android/wlauto.py
+++ b/libs/utils/android/wlauto.py
@@ -1,0 +1,186 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2015, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import subprocess as sp
+import tempfile
+import yaml
+import glob
+
+from . import Workload
+from distutils.spawn import find_executable
+from shutil import copyfile, rmtree
+
+import logging
+
+WA_RUN_CMD = 'wa run -f {} -d {}'
+WA_REPO = 'https://github.com/ARM-software/workload-automation.git'
+
+# Workloads directory path
+workloads_dir = os.path.dirname(os.path.abspath(__file__))
+workloads_dir = os.path.join(workloads_dir, 'workloads')
+agendas_dir = os.path.join(workloads_dir, 'agendas')
+
+class Wlauto(Workload):
+    """
+    Wrapper class for workload-automation
+
+    Workload-automation (WA) already provides many workloads for Android
+    targets.
+
+    In order to use one of the WA workloads (`wa list workloads` tells you what
+    is available) you need to implement a <workload__class_name> class in a
+    <workload_name>.py file stored under libs/utils/android/workloads.
+
+    <workload_class_name> inherits Wlauto and has to implement the methods from
+    Workload and (eventually) Wlauto.
+
+    This way, the workload is visible by the Workload class that will list it
+    as available and will eventually check that the related package is
+    installed on the Android target.
+
+    :param test_env: Test Environment object
+    :type test_env: TestEnv object
+    """
+
+    te = None
+
+    def __init__(self, test_env):
+        if not find_executable('wa'):
+            raise RuntimeError(
+                'wa not installed. Please, install it from %s', WA_REPO
+            )
+
+        Wlauto.te = test_env
+        super(Wlauto, self).__init__(test_env)
+        self._log = logging.getLogger('Wlauto')
+        self._log.debug('WA wrapper created')
+
+
+    @staticmethod
+    def _wload_is_available(workload_name):
+        """
+        Check if the specified workload is available in workload-automation.
+
+        :param workload_name: Name of the workload
+        :type workload_name: str
+        """
+        _log = logging.getLogger('Wlauto')
+        _log.debug('Checking for %s', workload_name)
+
+        if sp.check_output(['wa', 'list', '-n', workload_name, 'workloads']):
+            return True
+
+        _log.error('%s not available in WA', workload_name)
+        return False
+
+    @staticmethod
+    def wa_run(exp_dir, agenda, workload_name, collect=''):
+        """
+        Run workload automation using the specified agenda.
+
+        :param exp_dir: Path to experiment directory where to store results.
+        :type exp_dir: str
+
+        :param agenda: Path to the YAML file to be passed to
+            workload-automation. You can pass either an absolute path or a
+            filename. In the latter case the file must be under
+            android/workloads.
+        :type agenda: str
+
+        :param workload_name: Name of the workload
+        :type workload_name: str
+
+        :param collect: Specifies what to collect. Possible values:
+            - 'energy'
+        :type collect: list(str)
+        """
+        _log = logging.getLogger('Wlauto')
+
+        # Check if the specified path is absolute
+        if agenda.startswith('/'):
+            _agenda = agenda
+        else:
+            _agenda = os.path.join(agendas_dir, agenda)
+
+        _log.debug('Using agenda %s', _agenda)
+
+        # Create temprary output directory for WA
+        res_dir = tempfile.mkdtemp(prefix='lisa_wa_')
+        _log.debug('WA output dir is %s', res_dir)
+
+        # Initialize energy meter results
+        nrg_report = None
+        # Start energy collection
+        if 'energy' in collect and Wlauto.te.emeter:
+            Wlauto.te.emeter.reset()
+
+        # Execute workload-automation
+        p = sp.Popen(WA_RUN_CMD.format(_agenda, res_dir).split(),
+                     stdout=sp.PIPE,
+                     stderr=sp.PIPE)
+        out, err = p.communicate()
+
+        # Stop energy collection
+        if 'energy' in collect and Wlauto.te.emeter:
+            nrg_report = Wlauto.te.emeter.report(exp_dir)
+
+        _log.debug(out)
+        _log.debug(err)
+
+        # Copy results to exp_dir
+        # Take into account results from multiple iterations
+        wload_dirs = glob.glob('{}/{}*'.format(res_dir, workload_name))
+
+        # Store relevant results under `exp_dir`
+        for res_dir in wload_dirs:
+            dst_res_dir = os.path.join(exp_dir, os.path.basename(res_dir))
+            if os.path.exists(dst_res_dir):
+                # Remove possible outdated results
+                rmtree(dst_res_dir)
+            os.mkdir(dst_res_dir)
+
+            for f in os.listdir(res_dir):
+                src_path = os.path.join(res_dir, f)
+                if os.path.isfile(src_path):
+                    dst_path = os.path.join(dst_res_dir, f)
+                    copyfile(src_path, dst_path)
+
+        return nrg_report
+
+    def generate_yaml_agenda(self, agenda):
+        """
+        Converts a dictionary into a YAML file which should be used as the
+        agenda for running workload-automation benchmarks.
+
+        The YAML file is put under the results directory specified in the test
+        environment.
+
+        :param agenda: Dictionary to be converted into YAML file
+        :type agenda: dict
+        """
+        if not isinstance(agenda, dict):
+            raise ValueError('Specified agenda must be a dictionary')
+
+        _agenda = os.path.join(self.te.res_dir, 'agenda.yaml')
+
+        with open(_agenda, 'w') as f:
+            f.write(yaml.dump(agenda, default_flow_style=True))
+
+        return _agenda
+
+# vim :set tabstop=4 shiftwidth=4 expandtab
--- a/libs/utils/android/workload.py
+++ b/libs/utils/android/workload.py
@@ -70,7 +70,14 @@ class Workload(object):
         _log.debug('Building list of available workloads...')
         for sc in Workload._subclasses():
             _log.debug('Checking workload [%s]...', sc.__name__)
-            if sc.package in cls._packages or sc.package == '':
+            # Check if it is a WA-wrapped workload
+            if sc.__name__ == 'Wlauto':
+                cls._availables[sc.__name__.lower()] = sc
+            elif 'Wa' in sc.__name__:
+                if sc._is_available(test_env.target):
+                    cls._availables[sc.__name__.lower()] = sc
+            # Check if non-WA-wrapped workload is available
+            elif sc.package in cls._packages or sc.package == '':
                 cls._availables[sc.__name__.lower()] = sc
 
         _log.info('Supported workloads available on target:')
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ sudo: required
 language: python
 install:
   # Use IPython 5.x because 6.0+ only supports Python 3.3
-  - pip install --upgrade "ipython<6.0.0" Cython trappy bart-py devlib psutil wrapt jupyter matplotlib
+  - pip install --upgrade "ipython<6.0.0" Cython trappy bart-py devlib psutil wrapt jupyter matplotlib pyyaml
 script:
   - cd $TRAVIS_BUILD_DIR
   - source init_env && lisa-test tests/lisa/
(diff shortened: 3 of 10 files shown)


libs/utils/android/wlauto.py
```diff
@@ -0,0 +1,141 @@
...
+            filename. In the latter case the file must be under
+            android/workloads.
+        :type agenda: str
+        """
+        # Check if the specified path is absolute
+        if agenda.startswith('/'):
+            path_to_agenda = agenda
+        else:
+            # Assume only file name
+            path_to_agenda = os.path.join(workloads_dir, agenda)
```
derkling: Should be better use a subfolder for agendas?

libs/utils/android/wlauto.py
```diff
@@ -0,0 +1,141 @@
...
+
+        logging.debug('%14s - Using agenda %s',
+                      'Wlauto',
+                      path_to_agenda)
+
+        # Create temprary output directory for WA
+        tmp_dir = tempfile.mkdtemp()
```
derkling: Should we add something like: prefix="lisa_wa_"?

libs/utils/android/wlauto.py
derkling: Should we use a shorter _agenda?

libs/utils/android/wlauto.py
```diff
@@ -0,0 +1,141 @@
...
+        logging.debug('%14s - wa output dir is %s', 'Wlauto', tmp_dir)
+
+        # Execute workload-automation
+        p = sp.Popen(WA_RUN_CMD.format(path_to_agenda, tmp_dir).split(),
+                     stdout=sp.PIPE,
+                     stderr=sp.PIPE)
+        out, err = p.communicate()
+        logging.debug(out)
+        logging.debug(err)
+        return tmp_dir
```
derkling: Should we rename this to results_dir... at the end is where WA preduces it's output, isn't it?

libs/utils/android/wlauto.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def generate_yaml_agenda(self, agenda):
+        """
+        Converts a dictionary into a YAML file which should used as the agenda
```
derkling: nit: which should BE

libs/utils/android/wlauto.py
```diff
@@ -0,0 +1,141 @@
...
+        for running workload-automation benchmarks.
+
+        The YAML file is put under the result directory specified in the test
```
derkling: resultS directory

libs/utils/android/wlauto.py
```diff
@@ -0,0 +1,141 @@
...
+        environment.
+
+        :param agenda: Dictionary to be converted into YAML file
+        :type agenda: dict
+        """
+
+        agenda_file = os.path.join(self.te.res_dir, 'agenda.yaml')
```
derkling: again, just _agenda?

libs/utils/android/wlauto.py
```diff
@@ -0,0 +1,141 @@
...
+
```
derkling: Should we:
```python
if not isinstance(agenda, dict):
    raise ValueError('Specified agenda must be a dictionary')
```

libs/utils/android/workload.py
```diff
@@ -92,7 +92,7 @@ def availables(target):
                     module_name = module.__name__
                     # Check if a package is required and is available on target
                     aw = Workload._AW(module_name, module, class_name, handler)
-                    if (Workload._is_available(target, aw)):
```
derkling: Was this broken since before this patch?!?
In that case we shoud better post this as a separate fix patch.

libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,99 @@
...
+from subprocess import Popen, PIPE
+from android import Screen, Workload
+from time import sleep
+from devlib.utils.android import adb_command
+
+import logging
+
+GET_FRAME_STATS_CMD = 'shell dumpsys gfxinfo {} > {}'
+
+class CameraPreview(Workload):
```
derkling: This is not a Wlauto workload... better add it as a separate patch...

libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def run(self, exp_dir, preview_duration_s, collect=''):
+
+        # Initialize energy meter results
+        nrg_data, nrg_file = None, None
+
+        # Unlock device screen (assume no password required)
+        self.target.execute('input keyevent 82')
+        # Press Back button to be sure we run the video from the start
+        self.target.execute('input keyevent KEYCODE_BACK')
```
derkling: There is a new Android::System module available... it would be great if we can move these previous two commands to be a standard API of that module. Something like:
```
android.system.unlock_screen()
android.system.press_back_button()
```

derkling: Maybe you can add this code as is right now... and than add a patch which adds System functionalities followed by a patch which refactors all the libs/utils/android/ modules which make use of these calls.
For example, there are similar calls in:
https://github.com/ARM-software/lisa/blob/master/libs/utils/android/workloads/jankbench.py#L83

As well as some other APIs:

```python
self.target.execute('am force-stop com.android.benchmark')
self.target.execute('pm clear com.android.benchmark')
```

Which should be better become a:
```Python
System.force_stop(apk_name, clear=True)
```

libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Force screen in LANDSCAPE mode
+        Screen.set_orientation(self.target, portrait=False)
+
+        # Start CameraPreview video on the target device
+        self.target.execute('am start -n {}/{}'.format(self.package,
+                                                       self.activity))
```
derkling: Another candidate for System API

```
System.start(apk_name, activity_name)
```

libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,99 @@
...
+        # Allow the activity to start
+        sleep(3)
+
+        # Reset framestats collection
+        self.target.execute('dumpsys gfxinfo {} reset'.format(self.package))
```
derkling: Yet another candidate...
```python
System.collect_framestats(apk_name)
```

libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Stop energy collection
+        if 'energy' in collect and self.te.emeter:
+            nrg_data, nrg_file = self.te.emeter.report(exp_dir)
+            logging.info("Estimated energy: %7.3f", float(nrg_data['BAT']))
+
+        # Get frame stats
+        db_file = os.path.join(exp_dir, "framestats.txt")
+        adb_command(self.target.adb_name,
+                    GET_FRAME_STATS_CMD.format(self.package, db_file))
```
derkling: Why this adb_command call?
However... another candidate to pair with the previous System.collect_framestats(apk_name)...

libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,99 @@
...
+        nrg_data, nrg_file = None, None
+
+        # Unlock device screen (assume no password required)
+        self.target.execute('input keyevent 82')
+        # Press Back button to be sure we run the video from the start
+        self.target.execute('input keyevent KEYCODE_BACK')
+
+        # Force screen in LANDSCAPE mode
+        Screen.set_orientation(self.target, portrait=False)
+
```
derkling: Here we need also to do two additional thinks:
1. use max backlight (Screen API)
2. set AIRPLANE mode (System API)

libs/utils/android/wlauto.py
mdigiorgio: Yes, it is

libs/utils/android/workloads/camerarecord.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # Initialize energy meter results
+        nrg_data, nrg_file = None, None
+
+        # Setting agenda_yaml properly
+        if type(agenda) is str:
+            agenda_yaml = agenda
+        else:
+            # Convert dictionary to YAML file
+            agenda_yaml = self.generate_yaml_agenda(agenda)
```
derkling: Overall I would prefer something like:
```python
_agenda = agenda
if type(agenda) is str:
    _agenda = self.generate_yaml_agenda(agenda)
```

I.e remove if conditions by using a default assignment...

libs/utils/android/workloads/camerarecord.py
```diff
@@ -0,0 +1,112 @@
...
+        # Start energy collection
+        if 'energy' in collect and self.te.emeter:
+            self.te.emeter.reset()
+
+        wa_out_dir = Wlauto.wa_run(agenda_yaml)
+
+        # Stop energy collection
+        if 'energy' in collect and self.te.emeter:
+            nrg_data, nrg_file = self.te.emeter.report(exp_dir)
+            logging.info("Estimated energy: %7.3f", float(nrg_data['BAT']))
```
derkling: Here we still have that hardcoded BAT lable... have we never decided how to make that more generic?

libs/utils/android/wlauto.py
mdigiorgio: Yes better :)

libs/utils/android/workloads/camerarecord.yaml
```diff
@@ -0,0 +1,7 @@
+config:
+    device: generic_android
+workloads:
+    - id: b01
```
derkling: Maybe use a more meaningful name?
I guess this is reported in logs as well... something like "lisa_camerarecord_20" ?

libs/utils/android/workloads/rtapp.py
```diff
@@ -0,0 +1,103 @@
...
+from android import Wlauto
+from android import Workload
+
+import os
+import glob
+import logging
+
+from shutil import copyfile
+
+class RTApp(Wlauto):
```
derkling: Overall same comments as per the camerarecord case...

One more consideration, should we better name these workloads to identify that they are provided via WA?
This means:
1. rename the workload to something like: WaRTapp, WaCameRarecord
2. rename the file as well: wa_rtapp.py, wa_camerarecord.py

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,167 @@
...
+import os
+import collections
+import copy
+import logging
+
+VELLAMO_DB_PATH = '/data/data/com.quicinc.vellamo/files'
+VELLAMO_SCORE_NAME = 'chapterscores.json'
+VELLAMO_TESTS = ['BROWSER', 'METAL', 'MULTI']
+
+class Vellamo(Workload):
```
derkling: Post this on a separate patch, maybe in the same patchset of CameraPreview.
Let's have a separate PR for these non WA related WL... they are more easy to merge ;-)

libs/utils/android/workload.py
mdigiorgio: No, as far as I remember we introduced this for making WA wrapper work

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,167 @@
...
+
+        if test_name.upper() not in VELLAMO_TESTS:
+            raise ValueError('Vellamo workload [%s] not supported', test_name)
+
+        # Set parameter depending on test
+        self.logger.debug('Start Vellamo Benchmark [%s]', test_name)
+        test_cmd = ''
+        sleep_time = 0
+        if test_name.upper() == 'BROWSER':
+            test_cmd = 'input tap 970 740'
```
derkling: Any chance to use "percentages of screen resolutions" for these tap?
I have a 2K display to cross-check if we want to give it a try... and maybe we can find a 720p or lower as well...

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,167 @@
...
+        self.target.execute('input keyevent 82')
+        # Start Vellamo on the target device
+        self.target.execute('pm clear {}'.format(self.package))
+        self.target.execute('am start -W -n {}/{}'.format(self.package,
+                                                          self.activity))
+        # Wait few seconds to make sure the app is loaded
+        sleep(1)
+
+        # Accept EULA
+        self.target.execute('input tap 864 1646')
```
derkling: And a new API similar to:
```python
System.tap(x_perc, y_perc)
```
would make thinks :100: times more cool

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # Start measuring energy
+        if 'energy' in collect and self.te.emeter:
+            self.te.emeter.reset()
+
+        # Run Workload
+        self.target.execute(test_cmd)
+
+        # Skip instructions
+        self.target.execute('input swipe 900 1000 200 1000 100')
```
derkling: What about?
```python
System.hswipe(x_perc, y_lower_perc, y_upper_perc, swipe_up=True)
```

and the companion
```python
System.hswipe(y_perc, x_left_perc, x_right_perc, swipe_right=True)
```

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # Wait for completion
+        sleep(sleep_time*60)
+
+        # Stop measuring energy
+        if 'energy' in collect and self.te.emeter:
+            nrg_data, nrg_file = self.te.emeter.report(exp_dir)
+            self.logger.info('%s - Estimated energy: %7.3f',
+                             test_name,
+                             float(nrg_data['BAT']))
```
derkling: To Be Fixed...

derkling: Do you think it could be sufficient, at least for the time being, to expose the channel name as a run() method parameter?!? At the end the EMeter is configured by the caller... thus:
1. if we specify a channel: that means we want to collect energy and we know also from which channel.
```python
def wload_run(self, exp_dir, test_name, collect='', emeter_channel=None):
    #....
    if 'energy' in collect and self.te.emeter and emeter_channel:
        #....
```

libs/utils/android/workloads/videopreview.py
```diff
@@ -0,0 +1,98 @@
...
+from subprocess import Popen, PIPE
+from android import Screen, Workload
+from time import sleep
+from devlib.utils.android import adb_command
+
+import logging
+
+GET_FRAME_STATS_CMD = 'shell dumpsys gfxinfo {} > {}'
+
+class VideoPreview(Workload):
```
derkling: Move in the same PR for the non WA workloads...

libs/utils/android/workloads/youtube.py
```diff
@@ -56,6 +56,11 @@ def run(self, exp_dir, video_url, video_duration_s, collect=''):
         # Force screen in LANDSCAPE mode
         Screen.set_orientation(self.target, portrait=False)
 
+        self.target.execute('monkey -p {} 1'.format(self.package))
```
derkling: These seems more a fix to be posted in a separate patch....
BTW: what is that?!?

libs/utils/android/workloads/camerapreview.py
mdigiorgio: Because I want to save the db_file locally. It is similar to what you wrote in `workload.py`, it is was already available in `devlib`

libs/utils/android/workloads/camerarecord.py
mdigiorgio: No, we still haven't discuss that

libs/utils/android/workloads/vellamo.py
mdigiorgio: Not sure, this is a button. I don't know where it will be placed with different screens...

libs/utils/android/workloads/vellamo.py
mdigiorgio: Yes, we can do this

libs/utils/android/workloads/youtube.py
mdigiorgio: This was needed to just start YouTube without any video if I well remember

libs/utils/android/workload.py
derkling: Still I don't remember the purpose and it's not easy to figure it out from the code.
We should:
1. use better var names, let's redefined:
```python
AndroidWorkload = namedtuple('AndroidWorkload',  ['module_name', 'module', 'class_name', 'ctor'])
```
2. add a better comment which explain why we need that check

derkling:
This patch needs to be rebased on master and split in a series providing:
- the base Wlauto class which extends the Workload base class
- one patch for each new workload added

mdigiorgio:
Rebased and split as requested :)

PR #121, opened 2016-05-31
energy: iio-capture integration

mdigiorgio:
This is a continuation of #72.

Fixes/changes requested in that PR have been added.

close #72

### DIFF
--- a/libs/utils/energy.py
+++ b/libs/utils/energy.py
@@ -22,6 +22,8 @@ import time
 import os
 
 from collections import namedtuple
+from subprocess import Popen, PIPE, STDOUT
+from time import sleep
 
 # Default energy measurements for each board
 DEFAULT_ENERGY_METER = {
@@ -66,8 +68,22 @@ DEFAULT_ENERGY_METER = {
             'resistor_values' : [0.033],
             'device_entry'    : '/dev/ttyACM0',
         }
-    }
+    },
 
+    # ACME: by default uses iiocapture
+    'acme' : {
+        'instrument' : 'iiocapture',
+        'conf' : {
+            # Assume it is available in PATH
+            'iiocapturebin' : 'iio-capture',
+            # By default, Use the first connected probe only, i.e. the one
+            # connected to probe n, with n being the lowest probe number with a
+            # device attached
+            'iiodevice' : ['iio:device0'],
+            # BBB default IP
+            'device_ip' : '192.168.7.2'
+        }
+    }
 }
 
 class EnergyMeter(object):
@@ -105,6 +121,8 @@ class EnergyMeter(object):
             EnergyMeter._meter = HWMon(target, emeter, res_dir)
         elif emeter['instrument'] == 'aep':
             EnergyMeter._meter = AEP(target, emeter['conf'], res_dir)
+        elif emeter['instrument'] == 'iiocapture':
+            EnergyMeter._meter = IIOCapture(target, emeter['conf'], res_dir)
 
         logging.debug('%14s - Results dir: %s', 'EnergyMeter', res_dir)
         return EnergyMeter._meter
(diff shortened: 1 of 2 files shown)


mdigiorgio:
I've just pushed a new commit to support multiple channels (i.e. multiple probes)

derkling:
Can you just push a fresh new patch, reviewing this one is quite confusing.
Moreover, we do not need to keep track the previous version of the patch while you can still give credit by adding a "suggested-by" tag in the commit message ;-)

libs/utils/energy.py
```diff
@@ -105,6 +121,8 @@ def getInstance(target, conf, force=False, res_dir=None):
             EnergyMeter._meter = HWMon(target, emeter, res_dir)
         elif emeter['instrument'] == 'aep':
             EnergyMeter._meter = AEP(target, emeter['conf'], res_dir)
+        elif emeter['instrument'] == 'iiocapture':
```
derkling: I would propose to use "Acme" as instrument name, since iio-capture is a more generic tool which we do not currently support.

libs/utils/energy.py
```diff
@@ -325,4 +343,133 @@ def report(self, out_dir, out_energy='energy.json', out_samples='samples.csv'):
...
         return (channels_nrg, nrg_file)
 
+
+class IIOCapture(EnergyMeter):
+    "iio-capture Energy Meter for BayLibre ACME board"
+
+    def __init__(self, target, conf, res_dir):
+        super(IIOCapture, self).__init__(target)
+
+        self._iiocapturebin = conf['iiocapturebin']
```
derkling: Can we make this an optional configuration parameter, so that the user is not forced to express it in the configuration file:
```python
self._iiocapturebin = DEFAULT_ENERGY_METER['acme']['iiocapturebin']
#...
try:
    self._iiocapturebin = conf['iiocapturebin']
except: pass
```

The same should be done for all other configuration properties.

libs/utils/energy.py
```diff
@@ -325,4 +343,133 @@ def report(self, out_dir, out_energy='energy.json', out_samples='samples.csv'):
...
+        self._hostname = conf['device_ip']
+        # Channels represent the connected probes
+        # iio:device<n>
+        self._channels = conf['iiodevice']
+        self._iio = [None] * len(self._channels)
+
```
derkling: Would like to have a section resuming the configuration in use...

libs/utils/energy.py
```diff
@@ -325,4 +343,133 @@ def report(self, out_dir, out_energy='energy.json', out_samples='samples.csv'):
...
+                              self._iiocapturebin, iiodevice)
+                logging.error("%14s - Make sure there are no iio-capture "
+                              "processes connected to %s and device %s",
+                              "IIOCapture", self._hostname, iiodevice)
+                continue
+
+            # Wait few milliseconds to check if there is any output.
+            # If this is the case, there must have been an error, so the user
+            # must be informed
+            sleep(0.02)
```
derkling: Should be better split the starting of all iio-captures from the check that they are running.
That way this 20ms delay should be paid just one time.

libs/utils/energy.py
```diff
@@ -325,4 +343,133 @@ def report(self, out_dir, out_energy='energy.json', out_samples='samples.csv'):
...
+            self._iio[i].poll()
+            if self._iio[i].returncode:
+                logging.error("%14s - Failed to run %s for %s", "IIOCapture",
```
derkling: Just noticed that in the rest of this module we use single quotes for all logging messages... :-/

derkling:
Will push a branch with some patches to back-port, fixing previous comments, and some other to stack on top, adding some new interesting feature...

PR #130, opened 2016-06-17
Freq residency v2

mdigiorgio:
This is v2 of #87.

It includes the requested changes. Also, some comments are added to ease understanding what the code does.

### DIFF
--- a/libs/utils/trace_analysis.py
+++ b/libs/utils/trace_analysis.py
@@ -29,6 +29,8 @@ import trappy
 # Configure logging
 import logging
 
+NON_IDLE_STATE = 4294967295
+
 class TraceAnalysis(object):
 
     def __init__(self, trace, tasks=None, plotsdir=None, prefix=''):
@@ -785,3 +787,166 @@ class TraceAnalysis(object):
         # Save generated plots into datadir
         figname = '{}/{}schedtune_conf.png'.format(self.plotsdir, self.prefix)
         pl.savefig(figname, bbox_inches='tight')
+
+    def _computeNonIdleTime(self, t_prev, t, tasks_time_intervals):
+        """
+        Compute the non-idle time spent by tasks within the time interval
+        [t_prev. t].
+
+        :param t_prev: Time interval init boundary
+        :type t_prev: double
+
+        :param t: Time interval end boundary
+        :type t: double
+
+        :param tasks_time_intervals: DataFrame containing start and end times
+            of tasks
+        :type tasks_time_intervals: :mod:`pandas.DataFrame`
+        """
+        nonidle_time = 0.0
+        for _, row in tasks_time_intervals.iterrows():
+            start = row['start']
+            end = row['end']
+            if start > t:
+                # Task not running at f_prev
+                continue
+            elif start >= t_prev and start <= t:
+                # Task starting at instant of time where frequency is f_prev
+                if end <= t:
+                    # Task runs at f_prev for the whole interval
+                    nonidle_time += end - start
+                else:
+                    # Task continues running at different frequency
+                    nonidle_time += t - start
+            else:
+                # Task started with previous frequency
+                if end > t_prev:
+                    # Task is still running now, at f_prev
+                    if end <= t:
+                        nonidle_time += end - t_prev
+                    else:
+                        # Task will continue running at different frequency
+                        nonidle_time += t - t_prev
+
+        return nonidle_time
+
+    def getCPUFrequencyResidency(self, cpu):
+        """
+        Get a DataFrame with per-CPU frequency residency, i.e. amount of
+        time CPU `cpu` spent at each frequency. Both total and active times
+        will be computed.
+
+        :param cpu: CPU ID
+        :type cpu: int
+        """
+
+        if not self.trace.hasEvents('cpu_frequency'):
+            logging.warn('Events [cpu_frequency] not found, '\
+                         'plot DISABLED!')
+            return None
+        if not self.trace.hasEvents('cpu_idle'):
+            logging.warn('Events [cpu_idle] not found, '\
+                         'plot DISABLED!')
+            return None
+
+        freq_dfr = self.trace.df('cpu_frequency')
+        cpu_freqs = freq_dfr[freq_dfr.cpu == cpu]
+        idle_dfr = self.trace.df('cpu_idle')
+        cpu_idle = idle_dfr[idle_dfr.cpu_id == cpu]
+
+        ### Compute TOTAL time spent at each frequency ###
+        time_intervals = cpu_freqs.index[1:] - cpu_freqs.index[:-1]
+        total_time = pd.DataFrame({
+            'TOTAL Time' : time_intervals,
+            'Frequency [MHz]' : [f/1000 for f in cpu_freqs.iloc[:-1].frequency]
+        })
+        total_time = total_time.groupby(['Frequency [MHz]']).sum()
+
+        ### Compute ACTIVE time spent at each frequency ###
+        # time instants where CPU is idle
+        idle_times = cpu_idle[cpu_idle['state'] != NON_IDLE_STATE]
+        # time instants where CPU is not idle
+        non_idle_times = cpu_idle[cpu_idle['state'] == NON_IDLE_STATE]
+
+        if idle_times.index[0] < non_idle_times.index[0]:
+            idle_times = idle_times.iloc[1:]
+
+        time_intervals = [t2 - t1 for t1, t2 in zip(non_idle_times.index,
+                                                    idle_times.index)]
+
+        times_len = min(len(idle_times), len(non_idle_times))
+        tasks_time_intervals = pd.DataFrame({
+            "start": non_idle_times.index[:times_len],
+            "end": idle_times.index[:times_len]
+        })
+
+        # Compute the non-IDLE time spent at a certain frequency
+        time_intervals = []
+        frequencies = []
+        row_iterator = cpu_freqs.iterrows()
+        # Take first item from row_iterator
+        t_prev, f_prev = row_iterator.next()
+        for t, f in row_iterator:
+            nonidle_time = self._computeNonIdleTime(t_prev,
+                                                    t,
+                                                    tasks_time_intervals)
+            freq = f_prev['frequency']
+            time_intervals.append(nonidle_time)
+            frequencies.append(freq)
+            t_prev = t
+            f_prev = f
+
+        active_time = pd.DataFrame({
+            'ACTIVE Time' : time_intervals,
+            'Frequency [MHz]' : [f/1000 for f in frequencies]
+        })
+        # Sum the time intervals with same frequncy
+        active_time = active_time.groupby(['Frequency [MHz]']).sum()
+        return total_time, active_time
+
+    def plotCPUFrequencyResidency(self, cpus=None):
+        """
+        Plot per-CPU frequency residency.
+
+        :param cpus: List of cpus. By default plot all CPUs
+        :type cpus: list(str)
+        """
+
+        if not cpus:
+            cpus = sorted(self.platform['clusters']['little'] + self.platform['clusters']['big'])
+        else:
+            cpus.sort()
+
+        n_cpus = len(cpus)
+        if n_cpus == 1:
+            gs = gridspec.GridSpec(1, 1)
+        else:
+            gs = gridspec.GridSpec(n_cpus/2 + n_cpus%2, 2)
+
+        fig = plt.figure()
+
+        axis = None
+        for idx, cpu in enumerate(cpus):
+            total_time, active_time = self.getCPUFrequencyResidency(cpu)
+
+            axis = fig.add_subplot(gs[idx])
+            total_time.plot.barh(ax=axis,
+                                 color='g',
+                                 legend=False,
+                                 figsize=(16,8),
+                                 fontsize=16);
+            active_time.plot.barh(ax=axis,
+                                  color='r',
+                                  legend=False,
+                                  figsize=(16,8),
+                                  fontsize=16);
+            axis.set_title('CPU{}'.format(cpu))
+            axis.set_xlabel('Time [s]')
+            axis.set_xlim(0, self.x_max)
+
+        if axis:
+            handles, labels = axis.get_legend_handles_labels()
+            plt.figlegend(handles,
+                          labels,
+                          loc='lower right')
+            plt.subplots_adjust(hspace=0.5)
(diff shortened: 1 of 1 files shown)


libs/utils/trace_analysis.py
```diff
@@ -29,6 +29,8 @@
 # Configure logging
 import logging
 
+IDLE_STATE = 4294967295
```
JaviMerino: Weren't you going to rename it to `NON_IDLE_STATE`?

libs/utils/trace_analysis.py
```diff
@@ -912,54 +810,216 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
         """
 
+        if isinstance(cluster, str):
+            try:
+                _cluster = self.platform['clusters'][cluster.lower()]
+            except KeyError:
+                logging.warn('%s cluster not found, plot DISABLED!', cluster)
+                return None
+        else:
+            _cluster = copy.deepcopy(cluster)
```
JaviMerino: Why the `deepcopy`?

libs/utils/trace_analysis.py
```diff
@@ -970,138 +1030,140 @@ def getClusterFrequencyResidency(self, cluster):
...
+        :type clusters: str ot list(str)
         """
 
-        if len(clusters) == 1:
-            gs = gridspec.GridSpec(1, 1)
+        if isinstance(clusters, str):
+            _clusters = [clusters]
         else:
-            gs = gridspec.GridSpec(1, 2)
+            _clusters = copy.deepcopy(clusters)
```
JaviMerino: Same here, why the `deepcopy()` Can't just just do:
``` python
    clusters = listify(clusters)
```

libs/utils/trace_analysis.py
JaviMerino: I think so, yes.

JaviMerino:
Other than the minor nits, I vote for merge :heart:

libs/utils/trace_analysis.py
mdigiorgio: :+1:

libs/utils/trace_analysis.py
mdigiorgio: You're right in this case I don't need it.

libs/utils/trace_analysis.py
mdigiorgio: And also in this case we can simplify, thanks...

derkling:
I like to optimizations in 30a46c2 but I'm wondering why not squashing these fixes into the previous two patches? Moreover, these "optimization" patch is too complex to be understood... better just have two initial patches with the optimization already done.
Am I missing another reason to keep it separated?

derkling:
The IDLE_STATE constant renaming patch also IMHO should be squashed into the first two...

libs/utils/trace_analysis.py
```diff
@@ -950,3 +950,158 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+        Get a DataFrame with per cluster frequency residency, i.e. amount of
+        time spent at a given frequency in each cluster.
+
+        :param cluster: list of CPU IDs belonging to the cluster
+        :type cluster: list(int)
+        """
+
+        if not self.trace.hasEvents('cpu_frequency'):
+            logging.warn('Events [cpu_frequency] not found, '\
+                         'plot DISABLED!')
```
derkling: If we call this method we are just getting two series of active and total time, thus this warning should not refer to a plot.
What about: "cluster active time computation not possible" ?

libs/utils/trace_analysis.py
```diff
@@ -950,3 +950,158 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+            return None
+        if not self.trace.hasEvents('cpu_idle'):
```
derkling: Same comment as before...

libs/utils/trace_analysis.py
```diff
@@ -970,138 +1030,140 @@ def getClusterFrequencyResidency(self, cluster):
...
 
         freq_dfr = self.trace.df('cpu_frequency')
-        cluster_freqs = freq_dfr[freq_dfr['cpu'].isin(cluster)]
+        # We assume that every CPU in a cluster always run at the same
+        # frequency, i.e. the frequency is scaled per-cluster not per-CPU.
+        # Hence, we can limit the cluster frequencies data to a single CPU
+        #TODO: we could check that this is valid in the trace and return a
+        #      warning in case it is not
+        cluster_freqs = freq_dfr[freq_dfr.cpu == _cluster[0]]
         idle_dfr = self.trace.df('cpu_idle')
```
derkling: It seems we do not use anymore this DataFrame in this function?

libs/utils/trace_analysis.py
```diff
@@ -970,138 +1030,140 @@ def getClusterFrequencyResidency(self, cluster):
             return None
 
         freq_dfr = self.trace.df('cpu_frequency')
-        cluster_freqs = freq_dfr[freq_dfr['cpu'].isin(cluster)]
+        # We assume that every CPU in a cluster always run at the same
+        # frequency, i.e. the frequency is scaled per-cluster not per-CPU.
+        # Hence, we can limit the cluster frequencies data to a single CPU
+        #TODO: we could check that this is valid in the trace and return a
+        #      warning in case it is not
+        cluster_freqs = freq_dfr[freq_dfr.cpu == _cluster[0]]
```
derkling: For consistency within this file, can we rename all the dataframe variables to be either just "df", when only one is in scope, or name_df when needed. For example this should be cluster_df.

libs/utils/trace_analysis.py
```diff
@@ -970,138 +1030,140 @@ def getClusterFrequencyResidency(self, cluster):
             return None
 
         freq_dfr = self.trace.df('cpu_frequency')
```
derkling: For consistency within this file, can we rename all the dataframe variables to be either just "df", when only one is in scope, or name_df when needed. For example this should be freq_df.

libs/utils/trace_analysis.py
```diff
@@ -912,54 +810,216 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+                         fontsize=14)
+        figname = '{}/{}cpu_freq_residency.png'.format(self.plotsdir,
+                                                       self.prefix)
+        pl.savefig(figname, bbox_inches='tight')
+
+    def _getClusterActiveSignal(self, cluster):
+        """
+        Build a square wave representing the active (i.e. non-idle) cluster
+        time, i.e.:
+            cluster_active[t] == 1 if the CPU is reported to be exited from an
```
derkling: This is technically incorrect, there is always a task running in every CPU, eventually the idle task.
Let's try to make it less ambiguous by saying something like:
```python
#     cluster_active[t] == 1 if at least one CPU has been reported to be non-idle by CPUFreq at time t
```

libs/utils/trace_analysis.py
```diff
@@ -970,138 +1030,140 @@ def getClusterFrequencyResidency(self, cluster):
...
-            activation_indexes = sorted(
-                non_idle_times.index.values.tolist() + \
-                idle_times.index.values.tolist()
+        cluster_active = self._getClusterActiveSignal(_cluster)
+
+        # In order to compute the active time spent at each frequency we
+        # multiply 2 square waves:
+        # - cluster_active, a square of the form:
+        #     cluster_active[t] == 1 if there is a task running in the
+        #                          cluster at time t
```
derkling: Same comment on this definition as in _getClusterActiveSignal

libs/utils/trace_analysis.py
```diff
@@ -912,54 +810,216 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+            cluster_active[t] == 0 otherwise
+
+        :param cluster: list of CPU IDs belonging to a cluster
+        :type cluster: list(int)
+        """
+        idle_dfr = self.trace.df('cpu_idle')
+
+        cpu_active = {}
+        for cpu in cluster:
+            cpu_idle = idle_dfr[idle_dfr.cpu_id == cpu]
```
derkling: Still a DataFrame... thus better cpu_idle_df

mdigiorgio:
On 17/06/16 12:31, Patrick Bellasi wrote:
> I like to optimizations in 30a46c2
> <https://github.com/ARM-software/lisa/commit/30a46c2a5e89cda04c1d9ca65f3c20f125802ecb>
> but I'm wondering why not squashing these fixes into the previous two
> patches? Moreover, these "optimization" patch is too complex to be
> understood... better just have two initial patches with the optimization
> already done.
> Am I missing another reason to keep it separated?

It's just to keep the history of how we started thinking about computing
the frequency residency and how we do it now. But if you prefer to have
it squashed I can do that...

>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/ARM-software/lisa/pull/130#issuecomment-226746654>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AGFSei-0QI7McA7OEz7SXVkiEJH-bp1aks5qMoWbgaJpZM4I4Ov0>.
>
IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium. Thank you.

libs/utils/trace_analysis.py
```diff
@@ -912,54 +810,216 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+
+            # time instants where CPU goes to an idle state
+            idle_times = cpu_idle[cpu_idle.state != IDLE_STATE]
+            # time instants where CPU leaves an idle state
+            non_idle_times = cpu_idle[cpu_idle.state == IDLE_STATE]
+
+            active_indexes = sorted(non_idle_times.index.values.tolist() + \
+                                    idle_times.index.values.tolist())
+
```
derkling: It seems to me that these last tree lines are exactly equivalent to:
```python
active_indexes = cpu_idle.index.tolist()
```

mdigiorgio: True, good point!

libs/utils/trace_analysis.py
```diff
@@ -912,196 +810,354 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+            # time instants where CPU leaves an idle state
+            non_idle_times = cpu_idle[cpu_idle.state == NON_IDLE_STATE]
+
+            active_indexes = cpu_idle.index.values.tolist()
+
+            exist_zero = active_indexes[0] == 0.0
+            if not exist_zero:
+                active_indexes.insert(0, 0.0)
+
+            if idle_times.index[0] < non_idle_times.index[0]:
```
derkling: The following block is quite "convoluted"... maybe we can simplify a bit at least to avoid repetition of the assignment to active?... I mean using a var to defined i or i+1...

libs/utils/trace_analysis.py
```diff
@@ -912,196 +810,354 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+                                   non-idle by CPUFreq at time t
+            cluster_active[t] == 0 otherwise
+
+        :param cluster: list of CPU IDs belonging to a cluster
+        :type cluster: list(int)
+        """
+        idle_df = self.trace.df('cpu_idle')
+
+        cpu_active = {}
+        for cpu in cluster:
```
derkling: I've also tried a completely different approach for the body of this loop... working at table level...
The idea is to:
1. inject a fake initial event if required
2. add a new column to the original DF which represents the active state
Here is the code:

```python
        for cpu in cluster:
            idle_cpu_df = idle_df[idle_df.cpu_id == cpu]

            # Inject intial status if required
            if idle_cpu_df.index[0] != 0:
                # Get a copy of the first cpu_idle events
                idle_cpu_0_df = idle_cpu_df.iloc[:1,:].copy(deep=True)
                # Reset its index to be at the origin [0.0]
                idle_cpu_0_df.reset_index(inplace=True)
                idle_cpu_0_df.loc[:,'Time'] = idle_cpu_0_df['Time'] - idle_cpu_0_df['Time']
                idle_cpu_0_df.set_index('Time', inplace=True)
                # The initial state is the opposite of the first cpu_idle event
                idle_cpu_0_df.loc[:,'state'] = idle_cpu_0_df['state'].apply(
                    lambda state : 0 if state == NONIDLE_STATE else NONIDLE_STATE)
                # Append the original set of events
                idle_cpu_0_df = idle_cpu_0_df.append(idle_cpu_df)
                idle_cpu_df = idle_cpu_0_df

            # Create a new CPU active signal
            idle_cpu_df.loc[:,'active'] = idle_cpu_df['state'].apply(
                lambda state : 1 if state == NONIDLE_STATE else 0)

            # Keep track of active states for this CPU
            cpu_active[cpu] = idle_cpu_df['active']
```

I would like to have something similar in the Trace module, as a Sanitize function for the cpu_idle DF.
Having such a column in the DF will allows to have the data on hand for many other plots...

However, for the time being it does not give much more advantages in performances (when measured via %timeti) and maybe the code complexity is not improved sensible.

Thus, for the time being, if we are able to simplify a bit the following if-else nesting... I would be happy to for with your version.

libs/utils/trace_analysis.py
```diff
@@ -912,196 +810,354 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+
+        # Cluster active is the OR between the actives on each CPU
+        # belonging to that specific cluster
+        cluster_active = reduce(
+            operator.or_,
+            [cpu_active.astype(int) for _, cpu_active in
+             active.iteritems()]
+        )
+
+        cluster_active = active[active.columns[0]].astype(int)
```
derkling: Should this block not be necessary after the previous reduce() call?

mdigiorgio: Yes something weird happened while rebasing, I'll fix it

libs/utils/trace_analysis.py
```diff
@@ -912,196 +810,354 @@ def plotCPUFrequencyResidency(self, cpus=None):
...
+        """
+        Build a square wave representing the active (i.e. non-idle) CPU time,
+        i.e.:
+            cpu_active[t] = 1 if there is a task running in the CPU at time t
+            cpu_active[t] = 0 otherwise
+
+        :param cpu: CPU ID
+        :type cpu: int
+        """
+        return self._getClusterActiveSignal([cpu])
```
derkling: I would actually prefer to reverse this code... I mean would it not be better to have this function implementing the code
that you have in the per-cpu loop of the _getClusterActiveSignal() call?
IMO it has more sense to implement something in that function and use it in the loop of the previous one.

This means that the Cluster code depends on the CPU code... which makes also more sense...

derkling:
These modifications have been addressed in #132

PR #131, opened 2016-06-17
[Android] add video preview workload

mdigiorgio:
Three additional workloads for Android targets.

Signed-off-by: Michele Di Giorgio michele.digiorgio@arm.com

### DIFF
--- a/libs/utils/android/workloads/videopreview.py
+++ b/libs/utils/android/workloads/videopreview.py
@@ -1,0 +1,103 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2015, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import logging
+import os
+import re
+
+from time import sleep
+from android import Screen, System
+from android.workload import Workload
+
+class VideoPreview(Workload):
+    """
+    Android VideoPreview workload
+    """
+
+    # Package required by this workload
+    package = 'com.google.android.GoogleCamera'
+    action = 'android.media.action.VIDEO_CAPTURE'
+
+    def __init__(self, test_env):
+        super(VideoPreview, self).__init__(test_env)
+        self._log = logging.getLogger('VideoPreview')
+        self._log.debug('Workload created')
+
+    def run(self, out_dir, duration_s=30, collect='systrace'):
+        """
+        Run a video preview workload
+
+        :param out_dir: Path to experiment directory where to store results.
+        :type out_dir: str
+
+        :param duration_s: Duration of test
+        :type duration_s: int
+
+        :param collect: Specifies what to collect. Possible values:
+            - 'energy'
+            - 'systrace'
+            - 'ftrace'
+            - any combination of the above
+        :type collect: list(str)
+        """
+        self._log.info("Running VideoPreview for {}s and collecting {}".format(duration_s, collect))
+
+        # Keep track of mandatory parameters
+        self.out_dir = out_dir
+        self.collect = collect
+
+        # Unlock device screen (assume no password required)
+        Screen.unlock(self._target)
+
+        # Set airplane mode
+        System.set_airplane_mode(self._target, on=True)
+
+        # Set min brightness
+        Screen.set_brightness(self._target, auto=False, percent=0)
+
+        # Start video capture
+        System.start_action(self._target, self.action)
+
+        # Force screen in LANDSCAPE  mode
+        Screen.set_orientation(self._target, portrait=False)
+
+        # Allow the activity to start
+        sleep(3)
+
+        self.tracingStart()
+
+        sleep(duration_s)
+
+        self.tracingStop()
+
+        # Close the app without clearing the local data to
+        # avoid the dialog to select the account at next start
+        System.force_stop(self._target, self.package, clear=False)
+
+        # Go back to home screen
+        System.home(self._target)
+
+        # Set brightness back to auto
+        Screen.set_brightness(self._target, auto=True)
+
+        # Switch back to screen auto rotation
+        Screen.set_orientation(self._target, auto=True)
+
+        # Switch off airplane mode
+        System.set_airplane_mode(self._target, on=False)
+
+# vim :set tabstop=4 shiftwidth=4 expandtab


libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,98 @@
...
+        super(CameraPreview, self).__init__(test_env)
+        logging.debug('%14s - Workload created', 'CameraPreview')
+
+    def run(self, exp_dir, preview_duration_s, collect=''):
+
+        # Initialize energy meter results
+        nrg_data, nrg_file = None, None
+
+        # Unlock device screen (assume no password required)
+        self.target.execute('input keyevent 82')
```
derkling: These new workloads should make use of the recently merged new android:system APIs...

mdigiorgio: They do, I modify them yesterday. More are needed (like `back` and `home`) to make this workloads even easier to understand.

libs/utils/android/workloads/camerapreview.py
```diff
@@ -0,0 +1,98 @@
...
+import os
+
+from subprocess import Popen, PIPE
+from time import sleep
+from devlib.utils.android import adb_command
+from android import Screen, Workload, System
+
+import logging
+
+GET_FRAME_STATS_CMD = 'shell dumpsys gfxinfo {} > {}'
```
derkling: Can we wrap on an android::system API also this call?
Something like:
```python
# android.system module
@static
def gfxinfo_reset(target):
    GET_FRAME_STATS_CMD = 'shell dumpsys gfxinfo {} > {}'
    target.execute('dumpsys gfxinfo {} reset'.format(self.package))

@static
def gfxinfo_get(package, framestats_file):
    adb_command(target.adb_name,
        GET_FRAME_STATS_CMD.format(package, framestats_file))

# in a Workload file...
android.system.gfxinfo_reset()
#... your experiment
android.system.gfxinfo_get(framestats_file)
```

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # Start measuring energy
+        if 'energy' in collect and self.te.emeter:
+            self.te.emeter.reset()
+
+        # Run Workload
+        System.tap(self.target, test_x, test_y)
+
+        # Skip instructions
+        self.target.execute('input swipe 900 1000 200 1000 100')
```
derkling: This has to be move into a new System API...

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,175 @@
...
+        # Set max brightness
+        Screen.set_brightness(self.target, auto=False, percent=100)
+
+        self.target.execute('pm clear {}'.format(self.package))
+        System.start(self.target, self.package, self.activity)
+        # Wait few seconds to make sure the app is loaded
+        sleep(1)
+
+        # Accept EULA
+        System.tap(self.target, 864, 1646)
```
derkling: This tap call should really be extended to use (by default) a screen percentage instead of these hardcoded pixels values. Can we try to extend this call using the display info returned by devlib?

I would really love to call something like:

```python
# percentage
System.tap(self.target, 50, 80)
#... or absolute:
System.tap(self.target, 864, 1646, absolute=True)
```

libs/utils/android/workloads/videopreview.py
```diff
@@ -0,0 +1,96 @@
...
+        # Unlock device screen (assume no password required)
+        self.target.execute('input keyevent 82')
+        # Press Back button to be sure we run the video from the start
+        self.target.execute('input keyevent KEYCODE_BACK')
+
+        # Force screen in LANDSCAPE mode
+        Screen.set_orientation(self.target, portrait=False)
+
+        # Start VideoPreview video on the target device
+        self.target.execute('am start -a android.media.action.VIDEO_CAPTURE')
```
derkling: we already have an activity start API, isn't it?

mdigiorgio: Yes but that does `start -n`, this is `start -a` because an action is needed to directly start video capture mode. Otherwise we can use the start method of the API and then swipe to select recording mode.

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,175 @@
...
+            test_x, test_y  = (970, 1275)
+            sleep_time = 1
+        elif test_name.upper() == 'MULTI':
+            test_x, test_y = (970, 1555)
+            sleep_time = 3.5
+
+        nrg_data, nrg_file = None, None
+
+        # Unlock device screen (assume no password required)
+        self.target.execute('input keyevent 82')
```
derkling: Use System API...

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if test_name.upper() not in VELLAMO_TESTS:
+            raise ValueError('Vellamo workload [%s] not supported', test_name)
+
+        # Set parameter depending on test
+        self.logger.debug('Start Vellamo Benchmark [%s]', test_name)
+        test_x, test_y = (0, 0)
+        sleep_time = 0
+        if test_name.upper() == 'BROWSER':
+            test_x, test_y = (970, 740)
```
derkling: Absolute coords are quite likely not portable across different screen sizes... should we go for percentage values, maybe on some devices this can be just enough ;-)

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,175 @@
...
+        System.tap(self.target, test_x, test_y, absolute=True)
+
+        # Skip instructions
+        System.hswipe(self.target, 10, 80, duration=100, swipe_right=False)
+        System.hswipe(self.target, 10, 80, duration=100, swipe_right=False)
+        System.hswipe(self.target, 10, 80, duration=100, swipe_right=False)
+        self.logger.debug("Vellamo - {} started!".format(test_name.upper()))
+
+        # Wait for completion
+        sleep(sleep_time*60)
```
derkling: why this *60?

mdigiorgio: To transform it in minutes

libs/utils/android/workloads/vellamo.py
derkling: Maybe we can test it by playing with "adb shell wm size"...

derkling:
The last two paches of this series, once rebased, they are ready for merge.
@mdigiorgio, can you post then in a separate new PR?

libs/utils/android/workloads/vellamo.py
```diff
@@ -0,0 +1,166 @@
...
+        if test_name.upper() == 'BROWSER':
+            test_x, test_y = (91, 33)
+            sleep_time = 3.5
+        elif test_name.upper() == 'METAL':
+            test_x, test_y  = (91, 59)
+            sleep_time = 1
+        elif test_name.upper() == 'MULTI':
+            test_x, test_y = (91, 71)
+            sleep_time = 3.5
+
```
derkling: Maybe we can check the screen density of the current target and, if different that the one used to get these relative values, dump a warning message to the user.

The required command to get the density is:
```shell
adb shell getprop ro.sf.lcd_density
```
which should be integrated as a new screen API: screen.get_density()

derkling:
I would also prefer to have this split in three different patches, one for each workload.

mdigiorgio:
Removed camera preview and vellamo which had already been merged.

valschneider:
This is identical to with the exception of using `android.intent.action.VIDEO_CAPTURE` instead of `android.intent.action.MAIN`, but since we don't actually record anything I'd say the "workload" should be exactly the same.

PR #132, opened 2016-06-17
Freq residency v3

mdigiorgio:
Version 3 of frequency residency. Follows #130 and #87 

Close #130 
Close #87

### DIFF
--- a/libs/utils/trace_analysis.py
+++ b/libs/utils/trace_analysis.py
@@ -25,9 +25,13 @@ import pylab as pl
 import re
 import sys
 import trappy
+import operator
+from devlib.utils.misc import memoized
 
 # Configure logging
 import logging
+
+NON_IDLE_STATE = 4294967295
 
 class TraceAnalysis(object):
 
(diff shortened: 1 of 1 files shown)


libs/utils/trace_analysis.py
```diff
@@ -852,3 +853,109 @@ def getClusterActiveSignal(self, cluster):
...
+            )
+            active_t = cluster_active * freq_active
+            # Sum up the time intervals by computing the integral of the square
+            # wave
+            nonidle_time.append(self._integrate_square_wave(active_t))
+
+        active_time = pd.DataFrame({'time' : nonidle_time},
+                                   index=[f/1000 for f in available_freqs])
+        active_time.index.name = 'frequency'
+        return total_time, active_time
```
derkling: Should we return a single named tuple, e.g.
```
from collections import namedtuple

ResidencyTime = namedtuple('ResidencyTime', ['total', 'active'])
...
return ResidencyTime(total_time, active_time)
```

mdigiorgio: Nice. This makes things more clear. I'm going to push it here...

libs/utils/trace_analysis.py
```diff
@@ -967,3 +967,273 @@ def getCPUFrequencyResidency(self, cpu):
...
+        :param res_type: type of residency, either TOTAL or ACTIVE
+        :type title: str
+        """
+        # Compute sum of the time intervals
+        duration = residency_df.time.sum()
+        residency_pct = pd.DataFrame(
+            {label : residency_df.time.apply(lambda x: x*100/duration)},
+            index=residency_df.index
+        )
+        yrange = 1.5 * n_plots
```
derkling: Better use 3 for higher OPP count

libs/utils/trace_analysis.py
```diff
@@ -918,11 +919,6 @@ def getClusterFrequencyResidency(self, cluster):
             _cluster = listify(cluster)
 
         freq_df = self.trace.df('cpu_frequency')
-        # Assumption: all CPUs in a cluster run at the same frequency, i.e. the
```
derkling: Move this and the following chunk in the previous patch

libs/utils/trace_analysis.py
```diff
@@ -934,6 +930,9 @@ def getClusterFrequencyResidency(self, cluster):
         total_time = total_time.groupby(['frequency']).sum()
 
         ### Compute ACTIVE Time ###
+        # Assumption: all CPUs in a cluster run at the same frequency, i.e. the
```
derkling: Maybe add that this assumption is verified by the Trace analysis module, when the trace is parsed.

derkling: Should not we use the Trace.freq_coherency attribute to raise a warning/exception in case we call this method on a non freq_chorent platform?

mdigiorgio: Right. In which case, should we return `None`?

libs/utils/trace_analysis.py
```diff
@@ -978,3 +977,265 @@ def getCPUFrequencyResidency(self, cpu):
...
+        xmax = 0.0
+        for c in _cpus:
+            r = self.getCPUFrequencyResidency(c)
+            residencies.append(ResidencyData('CPU{}'.format(c), r))
+
+            max_time = r.total.max().values[0]
+            if xmax < max_time:
+                xmax = max_time
+
+        self._generateFrequencyResidencyPlots(residencies, 'cpu', xmax,
```
derkling: I would suggest to rename this to: _plotFrequencyResidency

libs/utils/trace_analysis.py
```diff
@@ -978,3 +977,265 @@ def getCPUFrequencyResidency(self, cpu):
...
+        if not self.trace.hasEvents('cpu_idle'):
+            logging.warn('Events [cpu_idle] not found, '\
+                         'plot DISABLED!')
+            return
+
+        # Assumption: all CPUs in a cluster run at the same frequency, i.e. the
+        # frequency is scaled per-cluster not per-CPU. Hence, we can limit the
+        # cluster frequencies data to a single CPU
+        if not self.trace.freq_coherency:
+            logging.warn('Cluster frequency is not coherent! plot DISABLED!')
```
derkling: For consistency, better use a comma: "Cluster frequency is not coherent, plot DISABLED!"

libs/utils/trace_analysis.py
```diff
@@ -978,3 +977,265 @@ def getCPUFrequencyResidency(self, cpu):
...
+            return
+
+        # Sanitize clusters
+        if clusters is None:
+            _clusters = self.platform['clusters'].keys()
+        else:
+            _clusters = listify(clusters)
+
```
derkling: Previous block, a bit more Pythonic...

```python
try:
    _clusters = listify(clusters)
except:
    _clusters = self.platform['clusters'].keys()
```

libs/utils/trace_analysis.py
```diff
@@ -978,3 +977,265 @@ def getCPUFrequencyResidency(self, cpu):
...
+        # Precompute active and total time for each cluster
+        residencies = []
+        xmax = 0.0
+        for c in _clusters:
+            r = self.getClusterFrequencyResidency(
+                self.platform['clusters'][c.lower()])
```
derkling: indent second line one more stop?

libs/utils/trace_analysis.py
mdigiorgio: `listify(None)` does not raise an exception but rather returns `[None]`

libs/utils/trace_analysis.py
```diff
@@ -978,3 +977,265 @@ def getCPUFrequencyResidency(self, cpu):
...
+                         'plot DISABLED!')
+            return
+
+        if cpus is None:
+            # Generate plots only for available CPUs
+            cpufreq_data = self.trace.df('cpu_frequency')
+            _cpus = range(cpufreq_data.cpu.max()+1)
+        else:
+            _cpus = listify(cpus)
+
```
derkling: Same note as for the block starting at line 1221

derkling:
Well done, it seems we reached a stable and feature complete version on this topic!
Let's brexi it!

PR #133, opened 2016-06-20
libs/utils/android: add more system utility functions

mdigiorgio:

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -37,4 +37,48 @@ class System(object):
                        '--ez state {}'\
                        .format(ap_state))
 
+    @staticmethod
+    def start(target, apk_name, activity_name):
+        """
+        Start an application.
+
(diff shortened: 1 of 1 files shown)


libs/utils/android/system.py
```diff
@@ -37,4 +37,48 @@ def set_airplane_mode(target, on=True):
...
+
+        :param clear: clear application data
+        :type clear: bool
+        """
+        target.execute('am force-stop {}'.format(apk_name))
+        if clear:
+            target.execute('pm clear {}'.format(apk_name))
+
+    @staticmethod
+    def tap(target, x, y):
```
derkling: We should add a percentage version of this call, where we can exploit another (memoaized) screen size to compute the actual coords from a screen percentage coordinate.
Perhaps such a method is good enough to be quite portable... but let's do it on a second time.

PR #134, opened 2016-06-22
Cluster frequency coherency check

mdigiorgio:
Usually we work under the assumption that every CPU in a cluster runs at the
same frequency, i.e. frequency changes per-cluster and not per-CPU.

This adds a simple check to prove this assumption true.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -70,6 +70,9 @@ class Trace(object):
         # List of events available in the parsed trace
         self.available_events = []
 
+        # Cluster frequency coherency flag
+        self.freq_coherency = True
+
         # Folder containing trace
         if not os.path.isdir(datadir):
              self.datadir = os.path.dirname(datadir)
@@ -123,6 +126,8 @@ class Trace(object):
         self._sanitize_SchedOverutilized()
 
         self.__loadTasksNames(tasks)
+
+        self.__checkClusterFrequencyCoherency()
 
         # Compute plot window
         if not normalize_time:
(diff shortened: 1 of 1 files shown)


libs/utils/trace.py
```diff
@@ -377,3 +382,25 @@ def _sanitize_SchedOverutilized(self):
...
+            return
+
+        df = self.df('cpu_frequency')
+        clusters = self.platform['clusters']
+        for c, cpus in clusters.iteritems():
+            cluster_df = df[df.cpu.isin(cpus)]
+            for chunk in self.__chuncker(cluster_df, len(cpus)):
+                f = chunk.iloc[0].frequency
+                if any(chunk.frequency != f):
+                    logging.warn('Cluster Frequency is not coherenti! '\
```
derkling: nit: s/coherenti/coherent/

libs/utils/trace.py
```diff
@@ -377,3 +382,25 @@ def _sanitize_SchedOverutilized(self):
         df['start'] = df.index
         df['len'] = (df.start - df.start.shift()).fillna(0).shift(-1)
         df.drop('start', axis=1, inplace=True)
+
+    def __chuncker(seq, size):
+        return (seq.iloc[pos:pos + size] for pos in range(0, len(seq), size))
```
derkling: :heart:

libs/utils/trace.py
```diff
@@ -377,3 +382,25 @@ def _sanitize_SchedOverutilized(self):
         df['start'] = df.index
         df['len'] = (df.start - df.start.shift()).fillna(0).shift(-1)
         df.drop('start', axis=1, inplace=True)
+
+    def __chuncker(seq, size):
```
JaviMerino: nit: chunker

libs/utils/trace.py
```diff
@@ -70,6 +70,9 @@ def __init__(self, platform, datadir, events,
         # List of events available in the parsed trace
         self.available_events = []
 
+        # Cluster frequency coherency flag
+        self.freq_coherency = True
```
JaviMerino: What are you going to use this variable for?  Maybe we can leave it off until we have a user for it.

mdigiorgio: It is used in the frequency residency code because we assume all CPUs in a cluster run at the same frequency.

PR #136, opened 2016-06-23
libs/utils/trace: fix chunker and mispelling

mdigiorgio:


derkling:
Already fixed  in 47895783a

Issue #84, opened 2016-04-04
Results processor for LISA


raw-bin:
[Disclaimer: The enhancements here may not be applicable to LISA alone but I'll mention them here all the same to gather viewpoints. I've also not been through all the issues in the issue list - so there may be some redundancy].

There's a set of common 'views' of data that can be very useful in diagnosing a given use-case or system to form a first order viewpoint for diagnosis etc. 

For example - in the past cr2 produced a fixed set of plots that were very valuable to form an initial opinion of what's up. It would be good to have a similar setup with one 'report' generator that first confirms that all necessary tokens are available in input trace and if so, produces all the views.

It's kind of easy to get carried away with this sort of stuff admittedly - one type of view can sometimes be inferred from another so you don't always want to plot every type of view. However, in my experience, if the generation cost is minimal, having multiple views of interest - some redundancy notwithstanding - is a very useful correlation tool to diagnose problems.

Some of the views that I can think of are:
1. OPP residency break-down for a given trace input.
   - Per-CPU OPP residency break-down in a suitable plot.
   - Great to know where the compute's going. For example, Audio playback should show low OPPs on the LITTLE side only etc.
   - @JaviMerino has an implementation internally which might be reusable.
2. Idle state residency break-down.
   - Per-CPU/cluster idle residency break-down in a suitable plot.
   - Similar to 1 above, can help correlate problems. 
3. Parallelisation scoring for a given trace input.
   - Aggregate scoring - ie what %age of time were N cpus active (N: {0,1,...NCPUs}) - in a suitable plot.
   - Most obvious use is for benchmarks where you want to verify for example 1 thread per CPU kind of dynamic.
   - WA has a [result processor](https://github.com/ARM-software/workload-automation/blob/master/wlauto/result_processors/cpustate.py) that does this and the above. IMHO those kind of processors are better implemented in LISA.
4. IRQ histogram by CPU
   - Great for confirming who's causing frequent wake-ups on the wrong CPU type for example.

Some more items that came up in recent discussions which are worth including in a 'top level diagnostic notebook':
1. Per CPU task context switch count
   - Could help in contrast analyses for a given use-case between EAS and other strategies (including default mainline behaviour)
2. OPP transition rate indication
   - The idea being that you can see if scheduler driven DVFS is being too trigger happy compared to conventional governors

Please add to the list as ideas come up.

derkling:
The first bullet has been probably covered by Michele in PR #87 which is still under review.
[...]

PR #141, opened 2016-06-24
[TraceAnalysis] Idle state residency plots

mdigiorgio:
Satisfies point 2 of #84

### DIFF
--- a/libs/utils/trace_analysis.py
+++ b/libs/utils/trace_analysis.py
@@ -1303,3 +1303,85 @@ class TraceAnalysis(object):
 
         self._plotFrequencyResidency(residencies, 'cluster', xmax, pct, active)
 
+    def _getIdleStateResidency(self, entity):
+        """
+        Compute time spent by a given entity in each idle state.
+
+        :param entity: cpu ID or cluster name or list of CPU IDs
+        :type entity: int or str or list(int)
+
+        :returns: :mod:`pandas.DataFrame` - idle time dataframe
+        """
+        if not self.trace.hasEvents('cpu_idle'):
+            logging.warn('Events [cpu_idle] not found, '\
+                         'idle state residency computation not possible!')
+            return None
+
+        if isinstance(entity, str):
+            try:
+                _entity = self.platform['clusters'][entity.lower()]
+            except KeyError:
+                logging.warn('%s entity not found!', entity)
+                return None
+        else:
+            _entity = listify(entity)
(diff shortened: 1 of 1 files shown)


JaviMerino:
Looks good to me :+1:

derkling:
Last patch seems can be merged straight on master.
For the first two I'm going to add some comments inline.

libs/utils/trace_analysis.py
```diff
@@ -1303,3 +1303,85 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+            idle_t = entity_idle * idle_state
+            # Compute total time by integrating the square wave
+            idle_time.append(self._integrate_square_wave(idle_t))
+
+        idle_time_df = pd.DataFrame({'time' : idle_time}, index=available_idles)
+        idle_time_df.index.name = 'idle_state'
+        return idle_time_df
+
+    @memoized
+    def getCPUIdleStateResidency(self, cpu):
```
derkling: Here "cpu" can actually be a list... because the internal method uses listify, however it is not clear what is going to return in case a list is specified. Actually it uses only the first element of the list... we should make that API less ambiguous.

Right now these two methods are used as a "service" method for the plots APIs of the next patch.
However, by them self they returns a nicely formatted DataFrame with the idle times for a specified CPU/Cluster.
Would it be possible to transform these to "fist class citizen" and make them return a properly formatted table (aka DataFrame) with the idle residencies for each cpu/cluster specified?

libs/utils/trace_analysis.py
```diff
@@ -1385,3 +1385,135 @@ def getClusterIdleStateResidency(self, cluster):
...
+                              xytext=(-50, 45), textcoords='offset points',
+                              fontsize=18)
+
+        figname = '{}/{}{}_idle_state_residency.png'\
+                  .format(self.plotsdir, self.prefix, entity_name)
+
+        pl.savefig(figname, bbox_inches='tight')
+
+
+    def plotCPUIdleStateResidency(self, cpus=None):
```
derkling: These plots are quite similar to the plotCPUFrequencyResidency, however in this case we do not support the "relative" representation (i.e. using percentages)... could it make sense to add this mode as well?

mdigiorgio:
To be rebased on top of #153

mdigiorgio:
#164 is the new PR for this feature

PR #142, opened 2016-06-27
[Analysis] crop data to user-defined x range

mdigiorgio:
Enhance frequency residency analysis by limiting data within the time range specified by the user (which by default corresponds to the whole duration of the trace).

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -128,14 +128,13 @@ class Trace(object):
                           trace_format)
         self.__computeTimeSpan()
 
-        # Minimum and Maximum x_time to use for all plots
-        self.x_min = 0
-        self.x_max = self.time_range
-
         # Reset x axis time range to full scale
-        t_min = self.window[0]
-        t_max = self.window[1]
-        self.setXTimeRange(t_min, t_max)
+        self.x_min = self.window[0]
+        if self.window[0] is None:
+            self.x_min = 0
+        self.x_max = self.window[1]
+        if self.window[1] is None:
+            self.x_max = self.time_range
 
         self.data_frame = TraceData()
         self._registerDataFrameGetters(self)
@@ -169,16 +168,37 @@ class Trace(object):
         :param t_max: upper bound
         :type t_max: int or float
         """
-        if t_min is None:
+        self.x_min = self.window[0]
+        if self.window[0] is None:
             self.x_min = 0
-        else:
-            self.x_min = t_min
-        if t_max is None:
+        self.x_max = self.window[1]
+        if self.window[1] is None:
             self.x_max = self.time_range
-        else:
-            self.x_max = t_max
+
+        if t_min is not None:
+            if t_min < self.x_min:
+                logging.warning('t_min out of range: '
+                                'capping to trace minimum %.6f [s]',
+                                self.x_min)
+            elif t_min > self.x_max:
+                raise ValueError('t_min out of range: '
+                                 'trace boundaries are (%.6f, %.6f) [s]',
+                                 self.x_min, self.x_max)
+            else:
+                self.x_min = t_min
+        if t_max is not None:
+            if t_max > self.x_max:
+                logging.warning('t_max out of range: '
+                                'capping to trace maximum %.6f [s]',
+                                self.x_max)
+            elif t_max < self.x_min:
+                raise ValueError('t_max out of range: '
+                                 'trace boundaries are (%.6f, %.6f) [s]',
+                                 self.window[0], self.x_max)
+            else:
+                self.x_max = t_max
         logging.info('Set plots time range to (%.6f, %.6f)[s]',
-                     self.x_min, self.x_max)
+                self.x_min, self.x_max)
 
     def __registerTraceEvents(self, events):
         """
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -748,6 +748,43 @@ class Trace(object):
 
         return len(self._functions_stats_df) > 0
 
+    def _cropToXTimeRange(self, data):
+        """
+        Crop a dataframe containing FTrace events to the X time range specified
+        by the user through setXTimeRange().
+
+        When cropping a series it might be necessary to include a first and
+        last element with timestamps equal to the boundaries of the X time
+        range.
+
+        :param data: series to be cropped
+        :type data: :mod:`pandas.Series`
+        """
+        if not isinstance(data, pd.Series):
+            msg = 'Cropping supported only for pandas.Series objects!'
+            raise ValueError(msg)
+
+        first = None
+        # Avoid duplicate indexes in case the user specifies an x_min that
+        # corresponds to the time stamp of an existing event.
+        if self.x_min not in data.index:
+            # data[data.index < self.x_min] may return an empty series if x_min
+            # is lower than the lowest index of the data series. This will
+            # raise an exception that can be caught and leave first to None, as
+            # in this case a first element is not needed
+            try:
+                first = pd.Series([data[data.index < self.x_min].iloc[-1]],
+                                  index=[self.x_min])
+            except: pass
+        last = None
+        if self.x_max not in data.index:
+            try:
+                last = pd.Series([data[data.index <= self.x_max].iloc[-1]],
+                                  index=[self.x_max])
+            except: pass
+        data = data.loc[self.x_min:self.x_max]
+        return pd.concat([first, data, last])
+
 
 class TraceData:
     """ A DataFrame collector exposed to Trace's clients """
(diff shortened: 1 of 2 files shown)


libs/utils/trace_analysis.py
```diff
@@ -921,7 +921,38 @@ def _integrate_square_wave(self, sq_wave):
...
+        first = None
+        if self.x_min > 0.0:
+            first = pd.Series([data[data.index < self.x_min].iloc[-1]],
+                              index=[self.x_min])
+        last = None
+        if self.x_max < self.trace.time_range:
+            last = pd.Series([data[data.index <= self.x_max].iloc[-1]],
+                             index=[self.x_max])
+        data = data.loc[self.x_min:self.x_max]
+        return pd.concat([first, data, last])
```
JaviMerino: Instead of creating a series for the first and last element, you can calculate its indexes and use `data.loc` on them.  I think this function can be simplified to (not tested):
``` python
min_idx = data.loc[:self.x_min].index.max()
max_idx = data.loc[self.x_max:].index.min()
return data.loc[min_idx:max_idx]
```

mdigiorgio: Looks nice! I'll give it a try.

mdigiorgio: This does work, but I had to add a check for the case where `x_min` and `x_max` reflect the entire trace duration. In this case `min_idx` and `max_idx` are set to `nan` and an empty series is returned by the function. Other than that, it looks much better :smile:.

libs/utils/trace_analysis.py
```diff
@@ -987,8 +1009,9 @@ def getClusterFrequencyResidency(self, cluster):
         #     freq_active[t] == 1 if at time t the frequency is f
         #     freq_active[t] == 0 otherwise
         available_freqs = sorted(cluster_freqs.unique())
-        new_idx = sorted(cluster_freqs.index.tolist() + \
-                         cluster_active.index.tolist())
+        # Use set() to remove possible duplicate indexes
+        new_idx = sorted(set(cluster_freqs.index.tolist() + \
```
derkling: Since these two series are generated from two different trace points, I would say it can never happen that we have two equal indexes in these two series. Maybe this set() is not required, or otherwise we need to add either a better comment or raise an exception for a malformed trace file.

libs/utils/trace_analysis.py
mdigiorgio: I tested again the implementation with this simplification and I found an issue.

Basically, if we crop the dataframe this way we might lose lose an event and make wrong computations afterwards. For example, assume we have the following data:

```python
     freq
0.1  600
0.2  1200
0.3  600
0.4  1200
```

If I crop the data in the range `(0.15, 0.35)` I'll get the following:

```python
     freq
0.2  1200
0.3  600
```

Two things happened:

1. I lost the information about the first event and don't know any more that frequency at 0.15 was 1200. As a consequence following computations will be incorrect.

2. I know the there are events after 0.35, but because of what I did I lost the information for the last 0.05 seconds.

Now, what you proposed kind of overcomes this loss of information by adding more information which might be a lot out of the specified range, falling once again into wrong computations later on.

Therefore, we need those tricks that add an entry at the beginning and at the end.

libs/utils/trace_analysis.py
mdigiorgio: You are right. I removed the `set()` instruction and provided a check in the crop function to make sure that no duplicate indexes are generated (it could happen if the user specifies an x range with a value that is exactly the same as a time stamp in the DataFrame).

JaviMerino:
FWIW, looks good to me :+1:

libs/utils/trace_analysis.py
```diff
@@ -966,7 +966,35 @@ def _integrate_square_wave(self, sq_wave):
...
+        if not isinstance(data, pd.Series):
+            raise ValueError('cropping supported only for pandas.Series '
+                             'objects!')
+
+        first = None
+        if self.x_min > 0.0 and self.x_min not in data.index:
+            first = pd.Series([data[data.index < self.x_min].iloc[-1]],
+                              index=[self.x_min])
+        last = None
+        if self.x_max < self.trace.time_range and self.x_max not in data.index:
```
derkling: Mmm.. wondering if this check on "self.x_max < self.trace.time_range" is valid also for "non normalized" traces.
Did you tried to set a range for a non normalized trace?

mdigiorgio:
To be rebased on top of #153

mdigiorgio:
Rebased on top of master after merge of #153...

libs/utils/trace.py
```diff
@@ -128,14 +128,8 @@ def __init__(self, platform, data_dir, events,
                           trace_format)
         self.__computeTimeSpan()
 
-        # Minimum and Maximum x_time to use for all plots
-        self.x_min = 0
-        self.x_max = self.time_range
```
derkling: Should not we keep these two initializations here in __init__ just to better list which attributes are available in this class?

derkling: Actually, should not be better to initialize them with self.trace.window[1 and 2]?

libs/utils/trace.py
```diff
@@ -169,16 +163,22 @@ def setXTimeRange(self, t_min=None, t_max=None):
...
         :type t_max: int or float
         """
-        if t_min is None:
-            self.x_min = 0
-        else:
+        self.x_min = self.trace.window[0]
+        self.x_max = self.trace.window[1]
+        if t_min is not None:
+            if t_min < self.x_min or t_min > self.x_max:
+                raise ValueError('t_min out of range: '
```
derkling: Maybe in the first case we can just dump a warning and cap the value to the trace minimum?

```python
if t_min < self.x_min:
        logging.warning('t_min out of range: capping to trace minimum %.6f', self.x_min)
if t_min > self.x_max:
        raise ValueError('t_min out of range: trace boundaries are ({:.6f}, {:.6f}) [s]',
        .format(self.x_min, self.x_max))
```

libs/utils/trace.py
mdigiorgio: Yeah we can do that without calling `setXTimeRange()` which is more of a method for the user than for `init`.

libs/utils/trace.py
```diff
@@ -169,16 +163,22 @@ def setXTimeRange(self, t_min=None, t_max=None):
...
+                                 'trace boundaries are ({:.6f}, {:.6f}) [s]'
+                                 .format(self.x_min, self.x_max))
             self.x_min = t_min
-        if t_max is None:
-            self.x_max = self.time_range
-        else:
+        if t_max is not None:
+            if t_max < self.x_min or t_max > self.x_max:
```
derkling: Same comment: maybe we can just cap for (t_max > self.x_max) ?

derkling:
Te second patch in this series:
   "libs/utils/analysis: use only relevant data for residency ..."
seems not part of the PR topic... maybe you can just remove it an push straight to master, it's ok for me and it's just a cosmetic/improvement.

libs/utils/trace.py
```diff
@@ -729,6 +729,43 @@ def _loadFunctionsStats(self, path='trace.stats'):
...
+        When cropping a series it might be necessary to include a first and
+        last element with timestamps equal to the boundaries of the X time
+        range.
+
+        :param data: series to be cropped
+        :type data: :mod:`pandas.Series`
+        """
+        if not isinstance(data, pd.Series):
+            raise ValueError('cropping supported only for pandas.Series '
+                             'objects!')
```
derkling: let's keep on same line... or better defined a msg=''cropping..." in the previous line ;-)

libs/utils/trace.py
```diff
@@ -729,6 +729,43 @@ def _loadFunctionsStats(self, path='trace.stats'):
...
+        first = None
+        # Avoid duplicate indexes in case the user specifies an x_min that
+        # corresponds to the time stamp of an existing event.
+        if self.x_min not in data.index:
+            # data[data.index < self.x_min] may return an empty series if x_min
+            # is lower than the lowest index of the data series. This will
+            # raise an exception that can be caught and leave first to None, as
+            # in this case a first element is not needed
+            try:
+                first = pd.Series([data[data.index < self.x_min].iloc[-1]],
```
derkling: Tested with:

```python
import pandas as pd

data = pd.Series([1,2,3], index=[10,15,20])
for x_min in [11, 14, 16, 19]:
    first = pd.Series([data[data.index <= x_min].iloc[-1]], index=[x_min]);
    print "{}".format(first)
```

and it seems to work as expected :+1:

mdigiorgio:
Second commit has been pushed straight to master and the other are updated according to comments.

libs/utils/trace.py
mdigiorgio: Regarding the intialization to `self.window[]`, in order to do that we should make that window to be `(0, self.time_range)` if the user didn't specify a window and `normalize_time` is true.

derkling:
Seems that a rebase is required before merge... can you update the PR please. Cheers

derkling:
This is now traced in #190

PR #149, opened 2016-07-18
[TasksAnalysis] add per-CPU task context switch count

mdigiorgio:

### DIFF
--- a/libs/utils/analysis/cpus_analysis.py
+++ b/libs/utils/analysis/cpus_analysis.py
@@ -39,6 +39,33 @@ class CpusAnalysis(AnalysisModule):
     def __init__(self, trace):
         super(CpusAnalysis, self).__init__(trace)
 
+
+###############################################################################
+# DataFrame Getter Methods
+###############################################################################
+
+    def _dfg_context_switches(self):
+        """
+        Compute number of context switches on each CPU.
+
+        :returns: :mod:`pandas.DataFrame`
+        """
+        if not self._trace.hasEvents('sched_switch'):
+            logging.warn('Events [sched_switch] not found, context switch '
+                         'computation not possible!')
+            return None
+
+        sched_df = self._dfg_trace_event('sched_switch')
+        cpus = range(self._platform['cpus_count'])
+        ctx_sw_df = pd.DataFrame(
+            [len(sched_df[sched_df['__cpu'] == cpu]) for cpu in cpus],
+            index=cpus,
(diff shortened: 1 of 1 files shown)


libs/utils/trace_analysis.py
```diff
@@ -1303,3 +1303,38 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+    def getContextSwitchCount(self):
+        """
+        Compute per-CPU task context switch counts
+
+        :returns: :mod:`pandas.DataFrame`
+        """
+        if not self.trace.hasEvents('sched_switch'):
+            logging.warn('Events [sched_switch] not found, context switch '
+                         'computation not possible!')
+            return
```
derkling: Should we return 0 (or -1) just to propagate an error code?

mdigiorgio: Oh right, we should return `None`

libs/utils/trace_analysis.py
```diff
@@ -1303,3 +1303,38 @@ def plotClusterFrequencyResidency(self, clusters=None,
 
         self._plotFrequencyResidency(residencies, 'cluster', xmax, pct, active)
 
+    def getContextSwitchCount(self):
```
derkling: I would suggest to move this function into Trace, likewise we did for the kernel functions profiling API.
```python
Trace::context_switch_df(self)
```
The idea is to use the Trace module to expose all the APIs which returns a "useful" DataFrame. Indeed this module already has a trace.df() method which exposes the Trappy generated DFs, thus it seems a reasonable solution also to host all other APIs which returns some data.

I would also propose to extend the signature to be a bit more generic:
```python
Trace::context_switch_df(self, pids=None, task_names=None, cpus=None)
```
but this can be eventually done in a second patch ;-)

mdigiorgio: OK for the first thing.

Regarding the second part, this is meant to be "per-CPU" task context switch count, so I just return that information for all CPUs and then the user can pick the CPU he or she is interested in.

derkling: That's true... still a more abstract and generic API can be useful sometimes.
However, agree we can add it later...

mdigiorgio:
To be rebased on top of #153

mdigiorgio:
Rebased on top of master

derkling:
Maybe we should better move this API to be part of the cpu_analysis module?

PR #150, opened 2016-07-20
tests/eas/hmp_parity: Add descriptions for tests

JaviMerino:
The hmp parity tests don't have descriptions and are very cryptic for those that have no previous experience with EAS.  Add a simple description.


derkling:
:heart: thanks!

PR #160, opened 2016-08-12
[EnergyMeter] make the API more generic

mdigiorgio:
Currently available energy meters provide different formats for the results, which makes the API not generic.

This series aims at improving this aspect.


derkling:
As an initial comment, this would be the best series to add a couple of commits covering the two missing examples notebook :kissing_closed_eyes: 
At the end we just need to copy/update the recently merged example for the ACME ;-)

ipynb/sched_dvfs/smoke_test.ipynb
```diff
@@ -324,7 +324,7 @@
     "    rtapp.run(out_dir=exp_dir)\n",
     "    \n",
     "    # Stop FTrace and sample Energy consumption\n",
-    "    nrg = te.emeter.sample()\n",
+    "    nrg = te.emeter.report(exp_dir)[0]\n",
```
derkling: You was right, returning a tuple is at least horrible :-/
Should we better think about some form of named tuple?

```python
EnergyReport = namedtuple ('EnergyReport', ['channels', 'report_file'])
```

which should then allow to consume the data with something like
```python
er = te.emeter.report(exp_dir)
nrg = er.channels
```

libs/utils/energy.py
```diff
@@ -134,41 +142,33 @@ def __init__(self, target, hwmon_conf=None, res_dir=None):
         self.readings = {}
 
         if 'hwmon' not in self._target.modules:
-            logging.info('%14s - HWMON module not enabled',
-                    'EnergyMeter')
+            logging.info('%14s - HWMON module not enabled', 'HWMon')
             logging.warning('%14s - Energy sampling disabled by configuration',
-                    'EnergyMeter')
+                            'HWMon')
```
derkling: I would suggest to keep these cleanup/renaming chunks in a separate "cosmetics patch"... maybe at the beginning of the series

ipynb/sched_dvfs/smoke_test.ipynb
mdigiorgio: Yes, I can add that.. It should make things more clear

libs/utils/energy.py
```diff
@@ -134,41 +142,33 @@ def __init__(self, target, hwmon_conf=None, res_dir=None):
...
-            self.little_channel = hwmon_conf['channel_map']['little']
-            self.big_channel = hwmon_conf['channel_map']['big']
-        logging.info('%14s - Using channel %s as little channel',
-                     'EnergyMeter', self.little_channel)
-        logging.info('%14s - Using channel %s as big channel',
-                     'EnergyMeter', self.big_channel)
+            logging.info('%14s -    %s', 'HWMon', channel.label)
 
+        # record the HWMon channels
+        self._channels = hwmon_conf.get('channel_map', {
```
derkling: Like that, is it possible/worth to have it in the constructor?

ipynb/examples/energy_meter/EnergyMeter_ACME.ipynb
```diff
@@ -92,10 +92,10 @@
     "        \"conf\" : {\n",
     "            #'iio-capture' : '/usr/bin/iio-capture',\n",
     "            #'ip_address' : 'baylibre-acme.local',\n",
-    "            'channels' : {\n",
-    "                'Device0' : 0,\n",
-    "                'Device1' : 1,\n",
-    "            }\n",
+    "        },\n",
```
derkling: Are we messing up with some spaces here?

derkling: Oh not, right... we moved the "channel_map" to be an attribute of "emeter"... should be ok!

libs/utils/energy.py
mdigiorgio: Good point, I think it is possible. Let me try...

libs/utils/energy.py
```diff
@@ -72,9 +73,13 @@
...
         'instrument' : 'aep',
         'conf' : {
-            'labels'          : ['LITTLE'],
             'resistor_values' : [0.033],
             'device_entry'    : '/dev/ttyACM0',
+        },
+        # AEP requires channel_map for compatibility with the generic energy
+        # meters interface
+        "channel_map" : {
+            "LITTLE" : 'LITTLE',
```
derkling: Not sure about that combined with the previous removal of the labels attribute... is it working?

mdigiorgio: Yes, because below I pass `labels=aep_conf["channel_map"]`.... Oh so this means we cannot assign `channels` in the constructor

libs/utils/energy.py
```diff
@@ -34,6 +34,10 @@
         'conf' : {
             'sites' : [ 'A7 Jcore', 'A15 Jcore' ],
             'kinds' : [ 'energy']
+        },
+        'channel_map' : {
+            'little' : 'A7 Jcore',
```
derkling: Kind of boring, sorry... but the current libs/utils/results.py has an hardcoded "LITTLE" value.
We should keep that script working at least for the eas/rfc test.

I guess we can either rename these channels maps, or add a patch to this series which update the previous script as well.

libs/utils/energy.py
```diff
@@ -153,21 +163,15 @@ def __init__(self, target, hwmon_conf=None, res_dir=None):
...
-            self.little_channel = hwmon_conf['channel_map']['little']
-            self.big_channel = hwmon_conf['channel_map']['big']
-        logging.info('%14s - Using channel %s as little channel',
-                     'HWMon', self.little_channel)
-        logging.info('%14s - Using channel %s as big channel',
-                     'HWMon', self.big_channel)
-
+        # record the HWMon channels
+        self._channels = hwmon_conf.get('channel_map', {
+            'little': self._target.little_core.upper(),
```
derkling: Same renaming required according to previous comment

libs/utils/energy.py
```diff
@@ -320,13 +320,13 @@ class ACME(EnergyMeter):
     BayLibre's ACME board based EnergyMeter
     """
 
-    def __init__(self, target, conf, res_dir):
+    def __init__(self, target, acme_conf, res_dir):
```
derkling: Do we really need this renaming? That's a conf param for the ACME class, it's quite non ambiguous to me.

libs/utils/android/workloads/youtube.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, test_env):
     def run(self, exp_dir, video_url, video_duration_s, collect=''):
 
         # Initialize energy meter results
-        nrg_data, nrg_file = None, None
+        er = None
```
derkling: I like this er short for (i guess) energy_deport... maybe it's more readable is we always use something like "nrg_report"?... if we want to go that way it requires an update all over the places where you are using "er" ;-)

libs/utils/energy.py
mdigiorgio: Then we should change it in the other instruments too to be just `conf`.

libs/utils/android/workloads/youtube.py
```diff
@@ -77,8 +77,8 @@ def run(self, exp_dir, video_url, video_duration_s, collect=''):
 
         # Stop energy collection
         if 'energy' in collect and self.te.emeter:
-            nrg_data, nrg_file = self.te.emeter.report(exp_dir)
-            logging.info("Estimated energy: %7.3f", float(nrg_data['BAT']))
+            er = self.te.emeter.report(exp_dir)
+            logging.info("Estimated energy: %7.3f", float(er.channels['BAT']))
```
derkling: well... that's somehow correlated with this channels renaming map.
I think we should find a way to remove this hardcoded "BAT" from workloads which was used to run on a specific AEP configuration ;-)

Maybe we can expose a "channels_map" attribute in the run method and here, for each of the channels listed by this parameter, report the energy consumption?

ipynb/examples/energy_meter/EnergyMeter_AEP.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "<hr>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook shows how to use the `AEP` energy meters.\n",
+    "\n",
+    "*NOTE*: `caiman` is required to collect data from the probe. Instructions on how to install it can be found here https://github.com/ARM-software/caiman"
```
derkling: I would better link:
https://github.com/ARM-software/lisa/wiki/Energy-Meters-Requirements#arm-energy-probe-aep
.. where we have much more info about the installation and usage of the tool.

ipynb/examples/energy_meter/EnergyMeter_AEP.ipynb
```diff
@@ -0,0 +1,413 @@
...
+     "output_type": "stream",
+     "text": [
+      "{\r\n",
+      "    \"BAT\": \"0.026167\"\r\n",
+      "}"
+     ]
+    }
+   ],
+   "source": [
+    "logging.info(\"Returned energy file:\")\n",
```
derkling: Maybe s/Returned/Generated/ ?

ipynb/examples/energy_meter/EnergyMeter_AEP.ipynb
```diff
@@ -0,0 +1,413 @@
...
+   "execution_count": 10,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "02:06:33  INFO    : BAT stats (head)\n"
```
derkling: Maybe something a bit more detailed, e.g. "Samples collected for the BAT channel (only first 10)" ?

derkling:
I :heart: this re-factoring... it's improving a lot the usability of these meters ;-)

mdigiorgio:
@derkling I noticed that `AEP` energy meters return the energy value as a string. Isn't it better to have all the energy meters return a float instead (this is the case for `HWMon` already and in this series I also modify `ACME` to return a float)?

derkling:
Absolutely agree @mdigiorgio, let's use float for all classes, update and merge ;-)

ipynb/examples/energy_meter/EnergyMeter_HWMON.ipynb
```diff
@@ -0,0 +1,357 @@
...
+    "    # Energy Meters Configuration for BayLibre's ACME Cape\n",
+    "    \"emeter\" : {\n",
+    "        \"instrument\" : \"hwmon\",\n",
+    "        \"conf\" : {\n",
+    "            # Prefixes of the HWMon labels\n",
+    "            'sites' : ['a53', 'a57'],\n",
+    "            # Type of hardware monitor to be used\n",
+    "            'kinds' : ['energy']\n",
+    "        },\n",
+    "        'channel_map' : {\n",
```
derkling: Perhaps here we got the wrong mapping? I.e. LITTLE is a53...

ipynb/examples/energy_meter/EnergyMeter_HWMON.ipynb
```diff
@@ -0,0 +1,357 @@
...
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "02:13:52  INFO    : Measured channels energy:\n",
+      "02:13:52  INFO    : {'big': 9.036729999999807, 'little': 2.050003000000288}\n"
```
derkling: Indeed, for the configured ramp... I would expect a bigger energy consumption on the LITTLE cluster...

derkling:
Just a minor fix/update for the HWMON notebook and than that's ready for merge ;-)

ipynb/examples/energy_meter/EnergyMeter_HWMON.ipynb
mdigiorgio: That's right! I'll fix it...

libs/utils/energy.py
```diff
@@ -482,8 +482,8 @@ def report(self, out_dir, out_energy='energy.json'):
                 nrg[key] = float(val)
             channels_stats[channel] = nrg
 
-            logging.info('%14s - %s', 'ACME', self._str(channel))
-            logging.info('%14s - %s', 'ACME', nrg)
+            logging.debug('%14s - %s', 'ACME', self._str(channel))
```
derkling: Well done, but... for consistency shake, should not we add a similar logging.debugs for the AEP class's report().
It should be dumped from within the loop at line 314...

derkling:
:heart:

PR #162, opened 2016-08-26
libs/wlgen/workload: fix background execution for Android targets

JaviMerino:
Commit c83009d8fc9c ("libs/wlgen/workload: fix background execution when CGroups are in use") changed the command for a Workload that runs in the background from `target.background()` to `target.kick_off()`.  Sadly, `target.kick_off()` is broken for Android targets, see ARM-software/devlib#55.

As c83009d8fc9c only talked about problems with cgroups and `target.background()`, probably due to interactions with `cgroup_run_into.sh`.  As a compromise, run `target.kick_off()` for cgroup
and `target.background()` if cgroup is not set.


derkling:
This should be better fixed by #163

JaviMerino:
Fair enough

PR #165, opened 2016-09-05
EAS HMP-parity test improvements

JaviMerino:
A number of improvements to fix tests failing when they shouldn't. The new "Offload migration and idle pull" test needs an updated bart.  Trappy also has a new version, so update both modules while we are at it.

### DIFF
--- a/tests/eas/hmp_parity.py
+++ b/tests/eas/hmp_parity.py
@@ -340,7 +340,7 @@ class OffloadMigrationAndIdlePull(unittest.TestCase):
     def populate_tasks(cls):
         migrator_workload = BIG_WORKLOAD.copy()
         migrator_workload["duration_s"] = 9
-        migrator_workload["delay_s"] = 1
+        migrator_workload["delay_s"] = OFFLOAD_MIGRATION_MIGRATOR_DELAY
 
         for idx in range(cls.num_tasks):
             task = "early_starters" + str(idx)
--- a/tests/eas/hmp_parity.py
+++ b/tests/eas/hmp_parity.py
@@ -565,6 +565,8 @@ class WakeMigration(unittest.TestCase):
             task_name = "{}_{}".format(cls.task_prefix, i)
             cls.params[task_name] = Step(**STEP_WORKLOAD).get()
 
+        cls.phase_duration = STEP_WORKLOAD["time_s"]
+
     @classmethod
     def run_workload(cls):
         wload = RTA(
@@ -602,7 +604,7 @@ class WakeMigration(unittest.TestCase):
 
     def test_little_big_switch1(self):
         """Wake Migration: LITTLE -> BIG: 1"""
-        expected_time = self.offset + 5
+        expected_time = self.offset + self.phase_duration
         switch_window = (
             expected_time -
             SWITCH_WINDOW_HALF,
@@ -632,7 +634,11 @@ class WakeMigration(unittest.TestCase):
     def test_little_big_switch2(self):
         """Wake Migration: LITTLE -> BIG: 2"""
 
-        expected_time = self.offset + 15
+        # little - big - little - big
+        #                       ^
+        # We want to test that this little to big migration happens.  So we skip
+        # the first three phases.
+        expected_time = self.offset + 3 * self.phase_duration
         switch_window = (
             expected_time -
             SWITCH_WINDOW_HALF,
@@ -688,7 +694,11 @@ class WakeMigration(unittest.TestCase):
     def test_big_little_switch2(self):
         """Wake Migration: BIG -> LITLLE: 2"""
 
-        expected_time = self.offset + 10
+        # little - big - little - big
+        #              ^
+        # We want to test that this big to little migration happens.  So we skip
+        # the first two phases.
+        expected_time = self.offset + 2 * self.phase_duration
         switch_window = (
             expected_time -
             SWITCH_WINDOW_HALF,
--- a/tests/eas/hmp_parity.py
+++ b/tests/eas/hmp_parity.py
@@ -109,8 +109,8 @@ class ForkMigration(unittest.TestCase):
     Detailed Description
     ====================
 
-    The test spawns as many threads as there are little cores.  It
-    then checks that all threads started on a big core.
+    The test spawns as many threads as there are cores in the system.
+    It then checks that all threads started on a big core.
 
     Expected Behaviour
     ==================
@@ -138,12 +138,14 @@ class ForkMigration(unittest.TestCase):
 
     @classmethod
     def populate_params(cls):
+        big_prefix = cls.task_prefix + "_big"
         for idx in range(len(cls.env.target.bl.bigs)):
-            task = cls.task_prefix + str(idx)
+            task = big_prefix + str(idx)
             cls.params[task] = Periodic(**BIG_WORKLOAD).get()
 
+        little_prefix = cls.task_prefix + "_little"
         for idx in range(len(cls.env.target.bl.littles)):
-            task = cls.task_prefix + str(idx)
+            task = little_prefix + str(idx)
             cls.params[task] = Periodic(**SMALL_WORKLOAD).get()
 
     @classmethod
--- a/tests/eas/hmp_parity.py
+++ b/tests/eas/hmp_parity.py
@@ -521,10 +521,11 @@ class WakeMigration(unittest.TestCase):
     Detailed Description
     ====================
 
-    This test creates two tasks that alternate between being big and
-    small workload.  They start being small load for 5 seconds, they
-    become big for another 5 seconds, then small for another 5 seconds
-    and finally big for the last 5 seconds.
+    This test creates as many tasks as there are big cpus.  The tasks
+    alternate between high and low utilization.  They start being
+    small load for 5 seconds, they become big for another 5 seconds,
+    then small for another 5 seconds and finally big for the last 5
+    seconds.
 
     Expected Behaviour
     ==================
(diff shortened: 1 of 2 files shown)


mdigiorgio:
We should have done the submodules update on Friday, but we forgot it. Could you please add the update to devlib as well in the third commit? Otherwise we could push a separate PR for submodules' updates...

derkling:
@Michele why don't you just push a patch which progress all the submodules... than this PR from Javi will just skip the submodules update patch.

mdigiorgio:
@derkling yes, that's what I was thinking... Let's do it! :smile:

JaviMerino:
ok, rebased on top of master and dropped the submodules update

tests/eas/hmp_parity.py
```diff
@@ -400,108 +423,92 @@ def test_first_cpu_migrators(self):
...
-                self.env.target.bl.bigs,
-                EXPECTED_RESIDENCY_PCT,
-                operator.ge,
-                percent=True,
-                window=big_residency_window,
-                rank=self.num_tasks),
-            msg="Not all 'early starter' tasks are running on big cores for at least {}% of their execution time"\
-                    .format(EXPECTED_RESIDENCY_PCT))
+    def test_big_cpus_fully_loaded(self):
+        """Offload Migration and Idle Pull: Big cpus are fully loaded as long as there are tasks left to run in the system"""
```
derkling: Can we re-format using 80 columns?

tests/eas/hmp_parity.py
```diff
@@ -400,108 +423,92 @@ def test_first_cpu_migrators(self):
...
+        num_big_cpus = len(self.env.target.bl.bigs)
```
derkling: This will fails on bob bL platforms, or whenever the bL module is not loaded.
Should we better catch that exception and just report the test as passed?
I guess that if the platform is not bL this test is meaningless... thus perhaps if there is a better way to skip it in case the bl module is not loaded, that would make more sense. What do you think?

tests/eas/hmp_parity.py
```diff
@@ -400,108 +423,92 @@ def test_first_cpu_migrators(self):
...
+
+        end_times = sorted(self.end_times.values())
+
+        # Window of time until the first migrator finishes
+        window = (0, end_times[-num_big_cpus])
```
derkling: Looking at how end_times is built in calculate_end_time, are we assuming that the topology lists before LITTLE and after big CPUs? I cannot get if there is an implied ordering in end_times...

tests/eas/hmp_parity.py
```diff
@@ -400,108 +423,92 @@ def test_first_cpu_migrators(self):
...
+        expected_busy_time = window_len * num_offloaded_tasks * \
+                             EXPECTED_BIG_BUSY_TIME_PCT / 100.
+        msg = "Little cpus did not pick up big tasks while big cpus were fully loaded"
+
+        self.assertGreater(busy_time, expected_busy_time, msg=msg)
+
+    def test_all_tasks_run_on_a_big_cpu(self):
+        """Offload Migration and Idle Pull: All tasks run on a big cpu at some point
+
+        Note: this test may fail in "golden/silver" systems where the little
```
derkling: Maybe we should avoid the "golden/silver" nomenclature... I think that "big.LITTLE"  can still work given the following specification (i.e. "where the little cpus are almost as performant as the big ones").
What do you think?

tests/eas/hmp_parity.py
```diff
@@ -400,108 +423,92 @@ def test_first_cpu_migrators(self):
...
+        cpus are almost as performant as the big ones
+        """
+
+        for task in self.params.keys():
+            sa = SchedAssert(self.trace, self.env.topology, execname=task)
+            window = (0, self.end_times[task])
+            big_residency = sa.getResidency("cluster", self.env.target.bl.bigs,
+                                            window=window, percent=True)
+            msg = "Task {} only spent {}% on a big cpu, instead of a minimum of {}%".format(
+                task, big_residency, OFFLOAD_MIGRATION_MIN_BIG_RESIDENCY)
```
derkling: How did we defined this MIN_BIG_RESIDENCY? Do you think is generic enough?

tests/eas/hmp_parity.py
```diff
@@ -400,108 +423,92 @@ def test_first_cpu_migrators(self):
...
+
+            self.assertGreater(big_residency,
+                               OFFLOAD_MIGRATION_MIN_BIG_RESIDENCY, msg=msg)
+
+    def test_all_tasks_finish_on_a_big_cpu(self):
+        """Offload Migration and Idle Pull: All tasks finish on a big cpu
+
+        Note: this test may fail in "golden/silver" systems where the
```
derkling: Same previous comment for the usage of "golden/silver"

tests/eas/hmp_parity.py
JaviMerino: nosetests only prints the first line of the docstring, so if you break the line it looks weird:
```
Offload Migration and Idle Pull: Big cpus are fully loaded as long as ... ok
```

I'd rather leave it like this, even if it breaks the 80 column rule.

tests/eas/hmp_parity.py
JaviMerino: It's not only this test, all test cases in this file use the `bl` module.  The `bl` module is part of `TEST_CONF` so it should be loaded when setting up `TestEnv`.  If it fails, will `TestEnv` fail with some meaningful error?  As the `bl` module is needed for all the test cases, I'd rather have `TestEnv` fail than making all the tests in this file be skipped when you do `lisa-test eas/hmp_parity.py`

mdigiorgio: If there is a failure with `bl` module `TestEnv` will tell you this:

```python
"Unable to initialize [bl] module"
"Fix your target kernel configuration or update your kernel or test configurations"
```

tests/eas/hmp_parity.py
JaviMerino: No, there is no assumption in the ordering of big and LITTLE cpu in `end_times`.  The `end_times` array holds the times at which the tasks finished.  The window is the window from start until the first migrator finishes.  We then make sure that the bigs were fully utilized during this time.  We are not checking which task ran on a big cpu, just that while there are more tasks than bigs still running, the bigs are fully utilized.

tests/eas/hmp_parity.py
JaviMerino: You are right, we shouldn't use the golden/silver terminology.  I've refrased it to:

>         Note: this test may fail in big.LITTLE platforms in which the
>        little cpus are almost as performant as the big ones.

tests/eas/hmp_parity.py
JaviMerino: Agreed.  I've rephrased it as:

>        Note: this test may fail in big.LITTLE systems where the
>        little cpus' performance is comparable to the bigs' and they
>        can take almost the same time as a big cpu to complete a
>        task.

tests/eas/hmp_parity.py
JaviMerino: I defined it as 10% hoping that it would be permissive enough to accomodate most systems.  To make it more generic, we could lower it to the extreme of the definition of the test: "running at some point" can just mean "the task residency on a big cpu is greater than zero".  Do you prefer that and remove the OFFLOAD_MIGRATION_MIN_BIG_RESIDENCY?

tests/eas/hmp_parity.py
JaviMerino: and you get a `RutimeError`.  I think that's more informative than skipping all the tests in the file.

tests/eas/hmp_parity.py
```diff
@@ -304,6 +305,21 @@ class OffloadMigrationAndIdlePull(unittest.TestCase):
...
+    It is possible that when the migrator tasks start they do it on
+    big cpus and they end up displacing the early starters.  This is
+    acceptable behaviour.  As long as big cpus are fully utilized
+    running big tasks, the scheduler is doing a good job.
+
+    That is why this test doesn't test for migrations of the migrator
+    tasks to the bigs when we expect that the early starters have
+    finished.  Instead, it tests that the big cpus were fully loaded
+    as long as there are tasks left to run in the system; that the
+    litlle cpus run tasks while the bigs are busy (offload migration);
```
mdigiorgio: s/litlle/little/

JaviMerino: fixed, thanks!

mdigiorgio:
Could you also remove the hard-coded values in `WakeMigration`? Like this one:

https://github.com/JaviMerino/lisa/blob/2e2242e77ddaefd34e0e10d8d1d7d1703b1766bf/tests/eas/hmp_parity.py#L632

tests/eas/hmp_parity.py
derkling: Perhaps it's better, you properly report the actual residency time in the log message and IMO,  on most platforms, a 9% residency on bigs is as much broken as a 10%.
Thus perhaps better avoiding to overload the meaning of the "running at some point".

derkling:
Few comments but overall is a really good and valuable rework of these tests, thanks @JaviMerino for the effort! Once fixed the two remaining minor nits it's ok for me to merge, @mdigiorgio can you do it?

mdigiorgio:
Sure :+1:

tests/eas/hmp_parity.py
JaviMerino: Ok, I'll remove OFFLOAD_MIGRATION_MIN_BIG_RESIDENCY and output the actual residency in the log.

JaviMerino:
The new upload removes `OFFLOAD_MIGRATION_MIN_BIG_RESIDENCY` and the hardcoded values in `WakeMigration`.

PR #168, opened 2016-09-05
tests: simplify stune's boosted utilization signal test

JaviMerino:
Trappy 5.7 knows how to filter traces.  We can get rid of the `.add_parsed_event()` workaround and let the grammar filter the event for us.


JaviMerino:
Great :heart: !

PR #170, opened 2016-09-08
ipynb: remove the add_parsed_event() hack in the UseCaseExamples notebook

JaviMerino:
The bart check of util estimation in the UseCaseExamples_SchedTuneAnalysis notebook starts with code to add the sched_boost_task event filtered by comm back to the trappy `FTrace` object
which is a hack.  This is needed because trappy's grammar module couldn't do it when the notebook was originally written.  trappy learned to apply filters in its `Parser()` object in version 5.7.0, which led to
the simplification of the similar SchedTune test in b9d488f58ab2 ("tests: simplify stune's boosted utilization signal test").

Apply a similar change to the UseCaseExamples_SchedTuneAnalysis ipython notebook.  This way the test is simpler and we educate our users on the ability to use filters with the Analyzer :heart:

### DIFF
--- a/ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
+++ b/ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
@@ -1628,26 +1628,6 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
-   "metadata": {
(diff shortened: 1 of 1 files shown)


derkling:
Lovely (almost) all red patch ;-)

PR #172, opened 2016-09-09
[Energy Model] notebooks with energy model building flow

mdigiorgio:
This series adds two notebooks that describe the energy model building flow that will be part of LISA in the near future.

These two notebooks target two reference platforms:

- one where energy can be measured at cluster level
- the other one where energy can only be measured at system level, which means either at battery level or as near as possible to the cores

### DIFF
--- a/ipynb/energy/EnergyModel_ClusterEnergy.ipynb
+++ b/ipynb/energy/EnergyModel_ClusterEnergy.ipynb
@@ -1,0 +1,1696 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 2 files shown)


mdigiorgio:
This depends on #160 and should be rebased on top of it.

derkling:
Here is a patch with some of the proposed updates for the JUNO notebook.
https://gist.github.com/derkling/b52d10011a66c1be08d4aad95ec85e67
Similar updates are proposed for the system wide NB

derkling:
Every great story should start somewhere ;-)

PR #176, opened 2016-09-15
tests/eas/acceptance: Increase required busy time for idle pull test

bjackman:
90% is too forgiving for EXPECTED_BIG_BUSY_TIME_PCT, it allows traces to
pass the "Bug cpus are fully loaded" test even when 100% tasks have run
on LITTLE CPUs while big CPUs are idle for significant periods.

Since traces contain a perdiod of inactivity at the beinning, to require
99% busy time it's necessary to adjust the trace window over which
assertions apply. Instead of starting at t=0 we use
SchedMultiAssert.getStartTime to find the time at which the first task
started, and use that as the start of our window.

### DIFF
--- a/tests/eas/acceptance.config
+++ b/tests/eas/acceptance.config
@@ -9,7 +9,7 @@
     "STEP_HIGH_DCYCLE" : 50,
     "EXPECTED_RESIDENCY_PCT" : 85,
     "OFFLOAD_MIGRATION_MIGRATOR_DELAY": 1,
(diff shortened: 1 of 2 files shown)


mdigiorgio:
@JaviMerino to me this makes sense. If you agree I can merge it.

mdigiorgio:
Thanks!

JaviMerino:
@bjackman next time please assign me or mention me in the comments in order for me to get a notification.

PR #177, opened 2016-09-16
tests/eas/acceptance: increase accuracy of little cpus test in the Offload Migration testcase

JaviMerino:
The window for `test_little_cpus_run_tasks()` was loosely defined as just the migrator delay, which doesn't take into account the time it takes for rt-app to spawn and create its tasks.  After fa4bce7359ec ("tests/eas/acceptance: Increase required busy time for idle pull test"), the test is more strict, so increase the accuracy of the window to meet the new restrictions.


mdigiorgio:
Should be enough to just rename it without `BIG`, right? If yes, can we do it in this PR?

JaviMerino:
Sure, just added a commit that does it.

PR #179, opened 2016-09-16
Custom rtapp v2

credp:

There's an example too.


libs/wlgen/wlgen/rta.py
```diff
@@ -266,19 +266,28 @@ def _confCustom(self):
...
             '__DURATION__' : str(self.duration),
             '__PVALUE__'   : str(calibration),
             '__LOGDIR__'   : str(self.run_dir),
             '__WORKDIR__'  : '"'+self.target.working_directory+'"',
         }
 
+	# check for inline config
+	if 'custom_config' in self.params['custom']:
+		# inline config present, turn it into a file repr so we can parse the same
+		tmp_json = json.dumps(self.params['custom']['custom_config'], indent=4, separators=(',', ': '), sort_keys=True)
```
mdigiorgio: Can you change it to not break the 80 columns limit? Should be like this:

```python
tmp_json = json.dumps(self.params['custom']['custom_config'],
                      indent=4,
                      separators=(',', ': '),
                      sort_keys=True)
```

libs/wlgen/wlgen/rta.py
```diff
@@ -266,19 +266,28 @@ def _confCustom(self):
...
+		ifile = tmp_json.split('\n')
+	else:
+		# open the supplied filename
+	        ifile = open(self.params['custom']['custom_filename'], 'r')
+
```
mdigiorgio: Indentation should be 4 spaces.

libs/wlgen/wlgen/rta.py
```diff
@@ -266,19 +266,28 @@ def _confCustom(self):
...
         for line in ifile:
             for src, target in replacements.iteritems():
                 line = line.replace(src, target)
             ofile.write(line)
-        ifile.close()
+        if not 'custom_config' in self.params['custom']:
+		ifile.close()
```
mdigiorgio: Same as above.

libs/wlgen/wlgen/rta.py
```diff
@@ -472,6 +481,30 @@ def conf(self,
...
+	description stored in a file whose name is held in a variable called
+	'filename':
+
+	generated_file_name = rtapp.conf(
+	    kind='custom',
+	    duration=-1,
+	    params={
+		'custom_filename' : "{}".format(filename)
+	    },
+	    run_dir='/data/local/tmp' )
```
mdigiorgio: Indentation needs to be fixed.

PR #180, opened 2016-09-16
Custom rtapp v3

credp:

There's an example too.

Hopefully the indentation is preferable :)


libs/wlgen/wlgen/rta.py
```diff
@@ -266,19 +266,29 @@ def _confCustom(self):
...
+	# check for inline config
+	if 'custom_config' in self.params['custom']:
+        # inline config present, turn it into a file repr so we can parse the same
+        tmp_json = json.dumps(self.params['custom']['custom_config'],
+            indent=4, separators=(',', ': '), sort_keys=True)
+        ifile = tmp_json.split('\n')
+	else:
+        # open the supplied filename
+        ifile = open(self.params['custom']['custom_filename'], 'r')
+
```
mdigiorgio: Still missing 3 spaces in this block :smile:

credp: hahaha, which editor are you using? I can't persuade any of mine to show me the whitespaces.

credp: that's odd, there are 8 spaces before both those lines. what settings are you after?

mdigiorgio: Try to use vim with:

```python
:set tabstop=4
:set softtabstop=4
:set shiftwidth=4
:set textwidth=79
:set expandtab
:set autoindent
:set fileformat=unix
```

Those are my settings for `python`

PR #181, opened 2016-09-16
Updates to custom rt-app workload generation

credp:

There's an example too.

Hopefully the indentation is preferable :)


credp:
Patrick mentioned that there was an existing example of using a json file in the examples - which I inadvertently broke with this change. v5 incoming!

derkling:
Hi @credp , regarding PR's for me it's fine if you keep pushing on the same branch.

Github allows you to refresh a patch set (e.g. moving from v4 to v5 after a revision) by just force pushing the new patchset on an already existing branch (e.g. credp:custom_rtapp).

That way we keep a single PR's ID and keep updating it until we are happy for merge.

PR #182, opened 2016-09-19
Custom rtapp support - now supports multiple styles of links

credp:
I wanted to be able to pass a dictionary in with the configuration, and also to reference a file.


libs/wlgen/wlgen/rta.py
```diff
@@ -472,6 +489,50 @@ def conf(self,
...
+            run_dir='/data/local/tmp' )
+
+
+        Or to write the config into a variable directly:
+
+        generated_file_name = rtapp.conf(
+            kind='custom',
+            duration=-1,
+            params={
+                'custom_config'= { ...my description of the rt-app job... }
```
derkling: Maybe you already considered that, isn't possible to avoid introducing a nested dictionary and just check the value of params?
```python
if ininstace(self.params['custom'], str):
    # we have the filename for a JSON configuraiton
elif ininstace(self.params['custom'], dict):
    # we have an embedded dictionary-cased configuration
else:
    raise ValueError("params can be only a filename or an embedded dictionary")
```

valschneider:
@credp Could we revive this or is it no longer relevant ?

PR #183, opened 2016-09-20
[Android] update YouTube and Jankbench to use android utils

mdigiorgio:


libs/utils/android/workloads/youtube.py
```diff
@@ -56,6 +56,8 @@ def run(self, exp_dir, video_url, video_duration_s, collect=''):
         # Force screen in LANDSCAPE mode
         Screen.set_orientation(self.target, portrait=False)
 
+        self.target.execute('monkey -p {} 1'.format(self.package))
```
derkling: This monkey tool is always available or do we need to add the dependency in LISA?

derkling: In any case, it it's really required, we should add it as a System API, isn't it?

derkling: Add a comment reporting the reason, why we want to start youtube without playing the video?!?
Add the same reason in the commit message.

libs/utils/android/workloads/youtube.py
```diff
@@ -41,58 +39,50 @@ class YouTube(Workload):
...
-                      .format(video_url)
-        logging.info(youtube_cmd)
-        self.target.execute(youtube_cmd)
+        System.start_action(self.target, self.action, video_url)
         # Allow the activity to start
-        sleep(3)
-
-        # Reset framestats collection
-        self.target.execute('dumpsys gfxinfo --reset')
+        sleep(1)
```
derkling: The reduction from 3s to 1s is intended?

libs/utils/android/workloads/youtube.py
mdigiorgio: I used it because it allowed me to start YouTube without playing any video and then reset the frame stats. We can add it to the API, but I'm not sure if we are going to use for other workloads.

mdigiorgio: And no, it's provided by `adb` so no dependencies in LISA.

libs/utils/android/workloads/youtube.py
mdigiorgio: Yes, there wasn't any reason for waiting such a long time

libs/utils/android/workloads/youtube.py
```diff
@@ -41,56 +39,53 @@ class YouTube(Workload):
...
         nrg_report = None
 
         # Unlock device screen (assume no password required)
-        self.target.execute('input keyevent 82')
+        System.menu(self.target)
         # Press Back button to be sure we run the video from the start
-        self.target.execute('input keyevent KEYCODE_BACK')
+        System.back(self.target)
+
+        # Use the monkey tool to start YouTube without playing any video.
```
derkling: I would keep this chunk, where you use monkey, on a separate patch to better describe it's purpose and need.

mdigiorgio: OK, let me update it.

PR #184, opened 2016-09-20
utils: remove comments from platform JSON

bjackman:
Proper JSON doesn't actually allow comments -_-

Python 2.7.6's JSON parser throws a ValueError when trying to parse
these files.

### DIFF
--- a/libs/utils/platforms/juno.json
+++ b/libs/utils/platforms/juno.json
@@ -1,5 +1,4 @@
 {
-    // JUNO Energy Model
     "nrg_model" : {
         "little" : {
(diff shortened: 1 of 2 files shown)


derkling:
That's why we added this class:
https://github.com/ARM-software/lisa/blob/master/libs/utils/conf.py

Is it possibly in you use-case to use the utils's JsonConf to load a platform configuration?

bjackman:
Aha great, will see if I can refactor the relevant code to use that class.

bjackman:
The code that was causing my error was just in a notebook, so closing this.

PR #185, opened 2016-09-20
Add PID file and Vagrant symlink to .gitignore

bjackman:


mdigiorgio:
Since you are at it, we could probably add `vagrant` too, which appears as untracked file.

.gitignore
```diff
@@ -5,3 +5,4 @@ results*
 *.json
 *.log
 tools/android-sdk-linux
+*.pid
```
derkling: A PID file should be generated only to track the ipython server, right?
Thus, maybe better:
```bash
ipynb/*.pid
```

bjackman: Sure, will amend.

bjackman:
@mdigiorgio Ah yes, will add that too. Didn't do it initially because I wasn't sure if it was an artefact of my own setup.

mdigiorgio:
Thanks, I guess it can be merged.

.gitignore
```diff
@@ -6,3 +6,4 @@ results*
 *.log
 tools/android-sdk-linux
 ipynb/*.pid
+/vagrant
```
derkling: maybe without the initial "/" ?

bjackman: I think the `/` should be there - if there's a random file called `vagrant` anywhere else in the tree we shouldn't be ignoring it. 

However I've just noticed it's inconsistent with the PID one I just added and the android-sd-linux one. How about I stick a `/` at the beginning of all of them?

mdigiorgio: Yeah, actually it works both with and without the `/`.

PR #186, opened 2016-09-20
tests/eas/acceptance: fix missed rename of config value

bjackman:
An EXPECTED_BIG_BUSY_TIME_PCT got missed out from the recent renaming commit.


bjackman:
Oops, this is in the wrong test, need a different config name I think.

bjackman:
No, wait, I was right the first time :doughnut:

PR #187, opened 2016-09-21
tests/eas/acceptance: enable sched_is_big_little sysctl

bjackman:
Where this sysctl is present, if it isn't enabled the EAS wake path
views the CPUs as a flat list, so that LITTLE CPUs are not necessarily
preferred over big when selecting CPUs for a low-load task. Set it, so
that the WakeMigration and other tests can pass.

Since this flag isn't used by the mainline-integration version of EAS,
we only attempt to set it if the procfs file is present.

### DIFF
--- a/tests/eas/acceptance.py
+++ b/tests/eas/acceptance.py
@@ -17,6 +17,8 @@
 
 from env import TestEnv
 from wlgen import RTA, Periodic, Step
+from devlib.target import TargetError
+
 import trappy
 import shutil
 import os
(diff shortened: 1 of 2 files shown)


tests/eas/acceptance.py
```diff
@@ -44,6 +44,10 @@ def local_setup(env):
         env.target.execute(
             "echo ENERGY_AWARE > /sys/kernel/debug/sched_features")
 
+    if SET_IS_BIG_LITTLE:
+        is_big_little_sysctl = "/proc/sys/kernel/sched_is_big_little"
+        if (env.target.file_exists(is_big_little_sysctl)):
```
derkling: The best devlib's API to write sysfs attributes is write_value():
   https://github.com/ARM-software/devlib/blob/master/devlib/target.py#L322
which provides also a check for the value being properly set.

Moreover, since it's a kind-of "brute force" setup, maybe we can avoid the two ADB commands (check and write) by just doing:
```python
if SET_IS_BIG_LITTLE:
    try:
        te.target.write_value(is_big_little_sysctl, 1)
    except: pass
```

bjackman: Looks good.
At first I wondered why the sched_features one doesn't use this, looks like that's because `write_value` is intended for single-value sysctls etc, whereas sched_features is a list.

PR #188, opened 2016-09-22
Vagrantfile: install VirtualBox guest additions to access enable USB

mdigiorgio:

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,7 +13,10 @@ Vagrant.configure(2) do |config|
   # Forward ipython notebook's port to the host
   config.vm.network "forwarded_port", guest: 8888, host: 8888
 
-  config.vm.provision "shell", inline: <<-SHELL
+  # Get VirtualBox version installed on the host
+  VBOX_VERSION = `VirtualBox -help | head -1 | cut -d' ' -f5`
+
+  config.vm.provision "shell", args: "#{VBOX_VERSION}", inline: <<-SHELL
     sudo apt-get update
     sudo apt-get install -y autoconf automake build-essential expect git \
         libfreetype6-dev libpng12-dev libtool nmap openjdk-7-jdk \
@@ -47,6 +50,26 @@ Vagrant.configure(2) do |config|
     do
         echo unset $LC  >> /home/vagrant/.bashrc
     done
+
+    # Install VirtualBox guest additions
+    cd /tmp
+    VBOX_VERSION=$1
+    VBOX_ADDITIONS_URL="http://download.virtualbox.org/virtualbox/VER/VBoxGuestAdditions_VER.iso"
+    sudo wget -q -c ${VBOX_ADDITIONS_URL//VER/$VBOX_VERSION} -O VBoxGuestAdditions_$VBOX_VERSION.iso
+    sudo mount VBoxGuestAdditions_$VBOX_VERSION.iso -o loop /mnt
+    expect -c '
+        set timeout -1;
+        spawn sudo sh /mnt/VBoxLinuxAdditions.run --nox11;
+        expect {
+            "Do you wish to continue" { exp_send "yes\r" ; exp_continue }
+            eof
+        }
+    '
+    sudo umount /mnt
+    sudo rm *.iso
+    sudo /etc/init.d/vboxadd setup
+    cd -
+
     echo "export ANDROID_HOME=/vagrant/tools/android-sdk-linux" >> /home/vagrant/.bashrc
     echo 'export PATH=\$ANDROID_HOME/platform-tools:\$ANDROID_HOME/tools:\$PATH' >> /home/vagrant/.bashrc
     echo source init_env >> /home/vagrant/.bashrc


Vagrantfile
```diff
@@ -47,6 +50,26 @@ Vagrant.configure(2) do |config|
...
+    expect -c '
+        set timeout -1;
+        spawn sudo sh /mnt/VBoxLinuxAdditions.run --nox11;
+        expect {
+            "Do you wish to continue" { exp_send "yes\r" ; exp_continue }
+            eof
+        }
+    '
+    sudo umount /mnt
+    sudo rm *.iso
```
JaviMerino: Let us not be lazy here and only remove the `.iso` that we downloaded.  Set `VBoxGuestAdditions_$VBOX_VERSION.iso` to a variable and use it for wget, mount and in here.

Vagrantfile
```diff
@@ -47,6 +50,26 @@ Vagrant.configure(2) do |config|
     do
         echo unset $LC  >> /home/vagrant/.bashrc
     done
+
+    # Install VirtualBox guest additions
+    cd /tmp
+    VBOX_VERSION=$1
+    VBOX_ADDITIONS_URL="http://download.virtualbox.org/virtualbox/VER/VBoxGuestAdditions_VER.iso"
+    sudo wget -q -c ${VBOX_ADDITIONS_URL//VER/$VBOX_VERSION} -O VBoxGuestAdditions_$VBOX_VERSION.iso
```
JaviMerino: Why do you need to do `${VBOX_ADDITIONS_URL//VER/$VBOX_VERSION}`?  Can't you do:

    sudo wget -q -c "http://download.virtualbox.org/virtualbox/$VBOX_VERSION/VBoxGuestAdditions_$VBOX_VERSION.iso" -O VBoxGuestAdditions_$VBOX_VERSION.iso

JaviMerino: Why `wget -q`?  We hide the errors when it fails, which forces you to rerun `vagrant up` without it.  Also, why `-c`?  We can't be continuing anything because this iso can't be in `/tmp` at this point.

Vagrantfile
```diff
@@ -13,7 +13,10 @@ Vagrant.configure(2) do |config|
   # Forward ipython notebook's port to the host
   config.vm.network "forwarded_port", guest: 8888, host: 8888
 
-  config.vm.provision "shell", inline: <<-SHELL
+  # Get VirtualBox version installed on the host
+  VBOX_VERSION = `VirtualBox -help | head -1 | cut -d' ' -f5`
+
+  config.vm.provision "shell", args: "#{VBOX_VERSION}", inline: <<-SHELL
     sudo apt-get update
```
JaviMerino: Should we do `set -e` here to have the script fail when any command in this fails?  I tried running this and it failed to download the file (404).  Afterwards it tried to mount the iso that it hadn't downloaded, and launch VBoxLinuxAdditions.run and,...

With `-e` at least when a command fails we stop the execution.  This doesn't belong to this commit, but I think it merits a new commit.  Note that the output for failures is less than perfect because vagrant sends all the output again for whatever reason, but still, you get the script to fail as soon as there is a filure.

Vagrantfile
mdigiorgio: Yes, it is the same. @derkling thought it was more cool like this :smile:

mdigiorgio: OK `-c` is not necessary. Anyway, `-q` is necessary otherwise you get a long output from `wget`

Vagrantfile
mdigiorgio: :+1:

Vagrantfile
JaviMerino: This fails in Ubuntu 14.04, 

    --2016-09-22 14:43:22--  http://download.virtualbox.org/virtualbox/4.3.36_Ubuntu/VBoxGuestAdditions_4.3.36_Ubuntu.iso
    Resolving download.virtualbox.org (download.virtualbox.org)... 62.254.36.51, 62.254.36.9
    Connecting to download.virtualbox.org (download.virtualbox.org)|62.254.36.51|:80... connected.
    HTTP request sent, awaiting response... 404 Not Found
    2016-09-22 14:43:24 ERROR 404: Not Found.

I think it would be a lot simpler to install the package with the Guest additions:

    apt-get install -y virtualbox-guest-additions-iso

We can then mount the iso from `/usr/share/virtualbox/VBoxGuestAdditions.iso` and we don't need to play games with the version, guessing URLs or anything.

JaviMerino: What you want is `-nv` then.  With `-q` you hide the errors so you don't know why it failed.

mdigiorgio: Cool, thanks.

mdigiorgio: @derkling shall I write it as Javi suggests? I think it's more clear and we are anyway breaking the 80 columns...

mdigiorgio: Let me give it a try.

mdigiorgio:
Actually, when the VM is created it already has guest additions. Hence, this is not needed :smile:

PR #189, opened 2016-09-22
Some refactoring/enhancements to the Executor class

bjackman:
These are some refactors and enhancements to enable the existing EAS acceptance tests to be ported to use the LisaTest class.

This PR might be considered an RFC at this stage, I have only ported the ForkMigration test to use the LisaTest class, doing other tests should provide some validation of the changes made here.


bjackman:
Ref issue: https://github.com/ARM-software/lisa/issues/191

bjackman:
superseded by https://github.com/ARM-software/lisa/pull/197

PR #190, opened 2016-09-23
[Analysis] crop data to user-defined x range - v2

mdigiorgio:
This is v2 of #142. Differences with respect to previous PR:
- Simplified time range computation by getting it from TRAPpy
- Added a method to compute the plot window instead of calculating it in `__parseTrace()`. Also, it avoids the `window` attribute to have `None` values, but rather sets the two plot bounds to the actual plot window. This provides a reference from where to get the range of values that represent the boundaries for the x time range.

close #142

### DIFF
--- a/libs/utils/analysis/frequency_analysis.py
+++ b/libs/utils/analysis/frequency_analysis.py
@@ -446,7 +446,6 @@ class FrequencyAnalysis(AnalysisModule):
 # Utility Methods
 ###############################################################################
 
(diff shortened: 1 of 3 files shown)


valschneider:
I think some of this was addressed in https://github.com/ARM-software/lisa/pull/435. If you think the contents of this PR are still relevant, could you please update it ? Otherwise we could close it.

PR #195, opened 2016-09-26
tests/eas/capacity_capping: Add a capacity capping test

JaviMerino:
Add a simple test to verify that if the maximum frequency of the big cpus is reduced, busy threads are moved from the bigs to the littles.  When the maximum frequency is allowed to go back to the absolute
maximum, check that the threads are migrated back to bigs.

### DIFF
--- a/tests/eas/capacity_capping.config
+++ b/tests/eas/capacity_capping.config
@@ -1,0 +1,15 @@
+{
+    "BIG_DCYCLE": 100,
+    "WORKLOAD_DURATION_S": 9,
+    "WORKLOAD_PERIOD_MS": 100,
+    "MIGRATION_WINDOW": 0.75,
+    "EXPECTED_BUSY_TIME_PCT": 99,
+    "TEST_CONF": {
+        "modules": ["bl", "cpufreq"],
+        "tools": ["rt-app"],
+        "ftrace" : {
+             "events" : ["sched_switch"],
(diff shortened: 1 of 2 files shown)


derkling:
It would be interesting to rebase this test on top of #197... or otherwise we can merge it as it is and than ask @bjackman to include it in his refactoring. Ideas?

bjackman:
@derkling, either is fine with me, I can rework do the rework either way.

JaviMerino:
This test doesn't use the `Executor`, so it doesn't depend on #197.  I decided not to use the `Executor` class because the test has to execute commands while rt-app is running which is not currently supported by `Executor`. I had a quick stab at adding the needed code to the `Executor` class and it looked ugly, so I decided to not use it.  Do you want me to use the `Executor`?

derkling:
Ok, thus for me it's ok to merge this before #197 and let @bjackman check what parts of this new test can be included in it's refactoring/optimization work. Please, @bjackman merge this if you like.

bjackman:
Just thinking, a simple way to do this using `Executor` might be:
- Adding a hook to Executor that gets run before each workload is run
- In that hook, running on the target `sleep $time; echo $freq > /sys/fs/cpufreq/file` in the background for each change. We could probably abstract this as `target.schedule_write_value` or something.

But I don't think it's worth the complexity for one test, let's merge this.

PR #196, opened 2016-09-26
vagrant: do no apply filters to update Android SDK

mdigiorgio:
Let Android SDK Manager decide what to update.

Apparently with filters the build-tools are not installed. This seems to
solve the issue.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -32,8 +32,7 @@ Vagrant.configure(2) do |config|
         wget -qO- $ANDROID_SDK_URL | tar xz -C tools
         expect -c '
             set timeout -1;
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -32,8 +32,7 @@ Vagrant.configure(2) do |config|
         wget -qO- $ANDROID_SDK_URL | tar xz -C tools
         expect -c '
             set timeout -1;
-            spawn ./tools/android-sdk-linux/tools/android update sdk --no-ui \
-                -t tool,platform-tool,platform,build-tools-24.0.1;
```
JaviMerino: When I added `build-tools-24.0.1` it was the current version.  Newer Android SDK include build-tools-24.0.2 and they fail to install 24.0.1 :flushed:

PR #197, opened 2016-09-27
Rework EAS acceptance tests to inherit from LisaTest

bjackman:
- Refactor/enhance the Executor class; most importantly enable the "experiments" to be saved so that test objects can refer back to them in `test_` methods
- Add an EasTest class abstracting the common operations for all EAS acceptance tests. It's likely that some of this functionality (+ documentation which is not written) could go directly into LisaTest, but I haven't thought about that properly.
- Rework each of the acceptance test classes to inherit from EasTest

Also some miscellaneous cleanups that seem to me to be related to this.

### DIFF
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -99,8 +99,9 @@ class Executor():
         self.te = TestEnv(target_conf, tests_conf)
         self.target = self.te.target
 
+        self._iterations = self._tests_conf.get('iterations', 1)
         # Compute total number of experiments
-        self._exp_count = self._tests_conf['iterations'] \
+        self._exp_count = self._iterations \
                 * len(self._tests_conf['wloads']) \
                 * len(self._tests_conf['confs'])
 
@@ -116,7 +117,7 @@ class Executor():
 
         logging.info('%14s -   %3d workloads (%d iterations each)',
                      'Executor', len(self._tests_conf['wloads']),
-                     self._tests_conf['iterations'])
+                     self._iterations)
         wload_confs = ', '.join(self._tests_conf['wloads'])
         logging.info('%14s -       %s', 'Executor', wload_confs)
 
@@ -138,7 +139,7 @@ class Executor():
             for wl_idx in self._tests_conf['wloads']:
                 # TEST: configuration
                 wload = self._wload_init(tc, wl_idx)
-                for itr_idx in range(1, self._tests_conf['iterations']+1):
+                for itr_idx in range(1, self._iterations + 1):
                     # WORKLOAD: execution
                     self._wload_run(exp_idx, tc, wl_idx, wload, itr_idx)
                     exp_idx += 1
@@ -491,7 +492,7 @@ class Executor():
         self._print_title('Executor', 'Experiment {}/{}, [{}:{}] {}/{}'\
                 .format(exp_idx, self._exp_count,
                         tc_idx, wl_idx,
-                        run_idx, self._tests_conf['iterations']))
+                        run_idx, self._iterations))
 
         # Setup local results folder
         self._wload_run_init(run_idx)
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -485,12 +485,6 @@ class Executor():
         logging.debug(r'%14s - out_dir [%s]', 'Executor', self.te.out_dir)
         os.system('mkdir -p ' + self.te.out_dir)
 
-        logging.debug(r'%14s - cleanup target output folder', 'Executor')
-
-        target_dir = self.target.working_directory
-        logging.debug('%14s - setup target directory [%s]',
-                'Executor', target_dir)
-
     def _wload_run(self, exp_idx, tc, wl_idx, wload, run_idx):
         tc_idx = tc['tag']
 
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -130,19 +130,28 @@ class Executor():
     def run(self):
         self._print_section('Executor', 'Experiments execution')
 
+        self.experiments = []
+
         # Run all the configured experiments
-        exp_idx = 1
         for tc in self._tests_conf['confs']:
             # TARGET: configuration
             if not self._target_configure(tc):
                 continue
             for wl_idx in self._tests_conf['wloads']:
                 # TEST: configuration
-                wload = self._wload_init(tc, wl_idx)
+                wload, test_dir = self._wload_init(tc, wl_idx)
                 for itr_idx in range(1, self._iterations + 1):
-                    # WORKLOAD: execution
-                    self._wload_run(exp_idx, tc, wl_idx, wload, itr_idx)
-                    exp_idx += 1
+                    self.experiments.append({
+                        'wload_name' : wl_idx,
+                        'wload'      : wload,
+                        'conf'       : tc,
+                        'iteration'  : itr_idx,
+                        'out_dir'    : os.path.join(test_dir, str(itr_idx))
+                    })
+
+            # WORKLOAD: execution
+            for exp_idx, experiment in enumerate(self.experiments):
+                self._wload_run(exp_idx, experiment)
 
         self._print_section('Executor', 'Experiments execution completed')
         logging.info('%14s - Results available in:', 'Executor')
@@ -464,38 +473,38 @@ class Executor():
         wload = self._wload_conf(wl_idx, wlspec)
 
         # Keep track of platform configuration
-        self.te.test_dir = '{}/{}:{}:{}'\
-            .format(self.te.res_dir, wload.wtype, tc_idx, wl_idx)
-        os.system('mkdir -p ' + self.te.test_dir)
-        self.te.platform_dump(self.te.test_dir)
+        test_dir = '{}/{}:{}:{}'.format(self.te.res_dir, wload.wtype,
+                                        tc_idx, wl_idx)
+        os.system('mkdir -p ' + test_dir)
+        self.te.platform_dump(test_dir)
 
         # Keep track of kernel configuration and version
         config = self.target.config
-        with gzip.open(os.path.join(self.te.test_dir, 'kernel.config'), 'wb') as fh:
+        with gzip.open(os.path.join(test_dir, 'kernel.config'), 'wb') as fh:
             fh.write(config.text)
         output = self.target.execute('{} uname -a'\
                 .format(self.target.busybox))
-        with open(os.path.join(self.te.test_dir, 'kernel.version'), 'w') as fh:
+        with open(os.path.join(test_dir, 'kernel.version'), 'w') as fh:
             fh.write(output)
 
-        return wload
+        return wload, test_dir
 
-    def _wload_run_init(self, run_idx):
-        self.te.out_dir = '{}/{}'\
-                .format(self.te.test_dir, run_idx)
-        logging.debug(r'%14s - out_dir [%s]', 'Executor', self.te.out_dir)
-        os.system('mkdir -p ' + self.te.out_dir)
+    def _wload_run_init(self, experiment):
+        experiment['out_dir'] = _get_experiment_out_dir(experiment)
 
-    def _wload_run(self, exp_idx, tc, wl_idx, wload, run_idx):
+    def _wload_run(self, exp_idx, experiment):
+        tc = experiment['conf']
+        wload = experiment['wload']
         tc_idx = tc['tag']
 
         self._print_title('Executor', 'Experiment {}/{}, [{}:{}] {}/{}'\
-                .format(exp_idx, self._exp_count,
-                        tc_idx, wl_idx,
-                        run_idx, self._iterations))
+                .format(exp_idx, self._exp_count, # todo len(self.experiments?)
+                        tc_idx, experiment['wload_name'],
+                        experiment['iteration'], self._iterations))
 
         # Setup local results folder
-        self._wload_run_init(run_idx)
+        logging.debug(r'%14s - out_dir [%s]', 'Executor', experiment['out_dir'])
+        os.system('mkdir -p ' + experiment['out_dir'])
 
         # FTRACE: start (if a configuration has been provided)
         if self.te.ftrace and self._target_conf_flag(tc, 'ftrace'):
@@ -507,23 +516,23 @@ class Executor():
             self.te.emeter.reset()
 
         # WORKLOAD: Run the configured workload
-        wload.run(out_dir=self.te.out_dir, cgroup=self._cgroup)
+        wload.run(out_dir=experiment['out_dir'], cgroup=self._cgroup)
 
         # ENERGY: collect measurements
         if self.te.emeter:
-            self.te.emeter.report(self.te.out_dir)
+            self.te.emeter.report(experiment['out_dir'])
 
         # FTRACE: stop and collect measurements
         if self.te.ftrace and self._target_conf_flag(tc, 'ftrace'):
             self.te.ftrace.stop()
 
-            trace_file = self.te.out_dir + '/trace.dat'
+            trace_file = experiment['out_dir'] + '/trace.dat'
             self.te.ftrace.get_trace(trace_file)
             logging.info(r'%14s - Collected FTrace binary trace:', 'Executor')
             logging.info(r'%14s -    %s', 'Executor',
                          trace_file.replace(self.te.res_dir, '<res_dir>'))
 
-            stats_file = self.te.out_dir + '/trace_stat.json'
+            stats_file = experiment['out_dir'] + '/trace_stat.json'
             self.te.ftrace.get_stats(stats_file)
             logging.info(r'%14s - Collected FTrace function profiling:', 'Executor')
             logging.info(r'%14s -    %s', 'Executor',
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -376,9 +376,7 @@ class Executor():
         if conf['class'] == 'profile':
             params = {}
             # Load each task specification
-            for task_name in conf['params']:
-                task = conf['params'][task_name]
-                task_name = conf['prefix'] + task_name
+            for task_name, task in conf['params'].items():
                 if task['kind'] not in wlgen.__dict__:
                     logging.error(r'%14s - RTA task of kind [%s] not supported',
                             'RTApp', task['kind'])
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -385,7 +385,12 @@ class Executor():
                         'in RT-App workload specification'\
                         .format(task))
                 task_ctor = getattr(wlgen, task['kind'])
-                params[task_name] = task_ctor(**task['params']).get()
+                task_idxs = self._wload_task_idxs(wl_idx, task['tasks'])
+                for idx in task_idxs:
+                    idx_name = str(idx) if len(task_idxs) > 0 else ""
+                    task_name_idx = conf['prefix'] + task_name + idx_name
+                    params[task_name_idx] = task_ctor(**task['params']).get()
+
             rtapp = wlgen.RTA(self.target,
                         wl_idx, calibration = self.te.calibration())
             rtapp.conf(kind='profile', params=params, loadref=loadref,
--- a/tests/eas/acceptance.py
+++ b/tests/eas/acceptance.py
@@ -20,7 +20,6 @@ from wlgen import RTA, Periodic, Step
 from devlib.target import TargetError
 
 import trappy
-import shutil
 import os
 import unittest
 import logging
--- a/tests/eas/acceptance.py
+++ b/tests/eas/acceptance.py
@@ -54,26 +54,6 @@ def local_setup(env):
             # worry if the file isn't present.
             pass
 
-def between_threshold_pct(a, b):
-    THRESHOLD_PERCENT = 3
-    lower = b - THRESHOLD_PERCENT
-    upper = b + THRESHOLD_PERCENT
-
-    if a >= lower and a <= upper:
-        return True
-    return False
-
-
-def between_threshold_abs(a, b):
-    THRESHOLD = 0.25
-    lower = b - THRESHOLD
-    upper = b + THRESHOLD
-
-    if a >= lower and a <= upper:
-        return True
-    return False
-
-
 SMALL_WORKLOAD = {
 
     "duty_cycle_pct": SMALL_DCYCLE,
--- a/tests/eas/acceptance.py
+++ b/tests/eas/acceptance.py
@@ -355,7 +355,7 @@ class OffloadMigrationAndIdlePull(EasTest):
             self.assertIn(sa.getLastCpu(), self.te.target.bl.bigs, msg=msg)
 
 
-class WakeMigration(unittest.TestCase):
+class WakeMigration(EasTest):
     """
     Goal
     ====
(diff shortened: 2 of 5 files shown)


libs/utils/executor.py
```diff
@@ -130,19 +130,28 @@ def __init__(self, target_conf=None, tests_conf=None):
...
                 continue
             for wl_idx in self._tests_conf['wloads']:
                 # TEST: configuration
-                wload = self._wload_init(tc, wl_idx)
+                wload, test_dir = self._wload_init(tc, wl_idx)
                 for itr_idx in range(1, self._iterations + 1):
-                    # WORKLOAD: execution
-                    self._wload_run(exp_idx, tc, wl_idx, wload, itr_idx)
-                    exp_idx += 1
+                    self.experiments.append({
```
derkling: I like this idea to initialize all the confs and do the actual execution only after.
I'm wondering if it could make more sense here to use a namedtuple to represent one experiment, e.g.

```python
from collections import namedtuple

Experiment = namedtuple('Experiment', ['wload_name', 'wload', 'conf', 'iteration' ,'out_dir'])

self.experiments.append(Experiment(wl_idx, wload, tc, itr_idx, os.path.join(test_dir, str(itr_idx)))
```

This should perhaps simplify the following code cause you can:
1. loop on Experiment tuples, which are passed directly to the _wload_run (with ** expansion if required)
2. remove some dictionary access lines by replacing them with simpler calls, e.g. exp.wload_name instead of exp['wload_name']

libs/utils/executor.py
```diff
@@ -464,38 +473,38 @@ def _wload_init(self, tc, wl_idx):
         wload = self._wload_conf(wl_idx, wlspec)
 
         # Keep track of platform configuration
-        self.te.test_dir = '{}/{}:{}:{}'\
```
derkling: That's just a cosmetics chunk... can you keep them on a separate patch (maybe before)?

libs/utils/executor.py
```diff
@@ -464,38 +473,38 @@ def _wload_init(self, tc, wl_idx):
...
-        os.system('mkdir -p ' + self.te.test_dir)
-        self.te.platform_dump(self.te.test_dir)
+        test_dir = '{}/{}:{}:{}'.format(self.te.res_dir, wload.wtype,
+                                        tc_idx, wl_idx)
+        os.system('mkdir -p ' + test_dir)
+        self.te.platform_dump(test_dir)
 
         # Keep track of kernel configuration and version
         config = self.target.config
-        with gzip.open(os.path.join(self.te.test_dir, 'kernel.config'), 'wb') as fh:
```
derkling: Again, these two chunks are cosmetics... keep in the same separate patch with the previous one.

libs/utils/executor.py
```diff
@@ -464,38 +473,38 @@ def _wload_init(self, tc, wl_idx):
...
-        return wload
+        return wload, test_dir
 
-    def _wload_run_init(self, run_idx):
-        self.te.out_dir = '{}/{}'\
-                .format(self.te.test_dir, run_idx)
-        logging.debug(r'%14s - out_dir [%s]', 'Executor', self.te.out_dir)
-        os.system('mkdir -p ' + self.te.out_dir)
+    def _wload_run_init(self, experiment):
+        experiment['out_dir'] = _get_experiment_out_dir(experiment)
```
derkling: isn't the out_dir already defined for each experiment in line 149?

libs/utils/executor.py
```diff
@@ -464,38 +473,38 @@ def _wload_init(self, tc, wl_idx):
...
+    def _wload_run(self, exp_idx, experiment):
+        tc = experiment['conf']
+        wload = experiment['wload']
         tc_idx = tc['tag']
 
         self._print_title('Executor', 'Experiment {}/{}, [{}:{}] {}/{}'\
-                .format(exp_idx, self._exp_count,
-                        tc_idx, wl_idx,
-                        run_idx, self._iterations))
+                .format(exp_idx, self._exp_count, # todo len(self.experiments?)
```
derkling: Here is where we can simplify a bit using the namedtuple, maybe by also s/experiment/exp/

And the #todo, either do it or remove it from the patch ;-)

libs/utils/executor.py
```diff
@@ -376,9 +376,7 @@ def _wload_rtapp(self, wl_idx, wlspec, cpus):
         if conf['class'] == 'profile':
             params = {}
             # Load each task specification
-            for task_name in conf['params']:
-                task = conf['params'][task_name]
-                task_name = conf['prefix'] + task_name
```
derkling: Here we remove the conf['prefix'], is that intentional? And in that case, why?

tests/eas/acceptance.py
```diff
@@ -85,6 +88,52 @@ def log_result(data, log_fh):
...
+        return SchedMultiAssert(experiment['out_dir'],
+                                self.te.topology,
+                                [t for t in tasks if task_filter in t])
+
+    def get_start_time(self, experiment):
+        start_times_dict = self.get_multi_assert(experiment).getStartTime()
+        return min([t['starttime'] for t in start_times_dict.values()])
+
+    @experiment_test
+    def test_first_cpu(self, experiment, tasks):
```
derkling: Maybe this is overkilling for a base class. I know that in the original implementation KP was checking this condition in each test but, given the opportunity of this refactoring, i'm wondering if it really makes sense to check this condition multiple times.

Moreover, since this is a base class, should not we consider the further addition of new EasTests where instead for example we expect the task to start on a little CPU (e.g. negative boosted tasks)?

Perhaps I would prefer to have a little bit of replicated code in subclasses instead of having mandatory tests enforced in all them.

Maybe a convenient midway solution is to turn this into a local private method, e.g. _test_first_cpu, which can than be simply called by a test_first_cpu defined by a derived class?

tests/eas/acceptance.py
```diff
@@ -85,6 +88,52 @@ def log_result(data, log_fh):
...
 
+@wrapt.decorator
+def experiment_test(wrapped_test, instance, args, kwargs):
+    for experiment in instance.executor.experiments:
+        tasks = experiment['wload'].tasks.keys()
+        try:
+            wrapped_test(experiment, tasks, *args, **kwargs)
+        except AssertionError as e:
+            trace_relpath = os.path.join(experiment['out_dir'], "trace.dat")
+            logging.error("Check trace file: " + os.path.abspath(trace_relpath))
```
derkling: :heart:

derkling: Is this reference to the failing trace printed before or after the test being reported as failed?

tests/eas/fork_migration.config
```diff
@@ -0,0 +1,52 @@
...
+                "class" : "profile",
+                "params" : {
+                    "small" : {
+                        "kind" : "Periodic",
+                        "params" : {
+                            "duty_cycle_pct": 10,
+                            "duration_s": 5,
+                            "period_ms": 10,
+                        },
+                        "tasks" : "little", // Create one task for each LITTLE CPU
```
derkling: That's just a note for the further clean-up of the hardcoded big.LITTLE values... here we should probably provide the possibility to define a string which is used to match the CPU names defined in platform.json.
For example on a X20 platform, which has:
```python
	"board" : {
		"cores" : [
			"A53_0", "A53_0", "A53_0", "A53_0",
			"A53_1", "A53_1", "A53_1", "A53_1",
			"A72_0", "A72_0",
		],
	},
```
we could use:
- "A53_0" to generate as many tasks as the cores of the first cluster
- "A53" to generate as many tasks as the cores in the first two clusters

But that's just a note for another rework ;-)

tests/eas/acceptance.py
```diff
@@ -240,141 +240,49 @@ class OffloadMigrationAndIdlePull(unittest.TestCase):
...
-                self.env.target.bl.bigs,
-                rank=self.num_tasks),
-            msg="Not all the new 'migrator' tasks started on a big CPU")
-
-    def test_big_cpus_fully_loaded(self):
+    @experiment_test
+    def test_big_cpus_fully_loaded(self, experiment, tasks):
         """Offload Migration and Idle Pull: Big cpus are fully loaded as long as there are tasks left to run in the system"""
-        num_big_cpus = len(self.env.target.bl.bigs)
+        num_big_cpus = len(self.te.target.bl.bigs)
```
derkling: Noticing only now, maybe we can use the base class EasTest, or LisaTest directly, to track an alias for target?

```python
self.target = self.te.target
```

tests/eas/acceptance.py
```diff
@@ -240,141 +240,49 @@ class OffloadMigrationAndIdlePull(unittest.TestCase):
...
 
-        end_times = sorted(self.end_times.values())
+        s_assert = self.get_multi_assert(experiment, "early")
```
derkling: This seems more a refactoring chunk which change the way the test is implemented.
Can we move this in a separate patch, which follows the one which transform the original test into an EasTest?

tests/eas/acceptance.py
```diff
@@ -240,141 +240,49 @@ class OffloadMigrationAndIdlePull(unittest.TestCase):
...
+        m_assert = self.get_multi_assert(experiment, "migrators")
+        a_assert = self.get_multi_assert(experiment)
```
derkling: I cannot quite get the meaning of the prefixes: s_, m_ and a_. Can we use a better naming?

tests/eas/acceptance.py
```diff
@@ -383,17 +291,27 @@ def test_big_cpus_fully_loaded(self):
...
+
+        window = (migrator_activation_time, first_task_finish_time)
+
+        a_assert = self.get_multi_assert(experiment)
 
-        window = (self.get_migrator_activation_time(), first_task_finish_time)
-        busy_time = self.a_assert.getCPUBusyTime("cluster",
-                                                 self.env.target.bl.littles,
-                                                 window=window)
+        busy_time = a_assert.getCPUBusyTime("cluster",
```
derkling: Again, these last two chunks seems more an update of the test code instead of a refactoring to be an EasTest.
Can we move them in a following patch?

tests/eas/offload_idle_pull.config
```diff
@@ -0,0 +1,50 @@
+{
+    "modules"  : [ "bl" ],
+    "tools"    : [ "rt-app" ],
+    "ftrace" : { // TODO make these inherited
```
derkling: How do you want to inherit these?
In any case, if we do not do it now let's remove the comment as well from the committed code.

tests/eas/acceptance.py
```diff
@@ -378,193 +378,65 @@ class WakeMigration(unittest.TestCase):
...
-            calibration=cls.env.calibration())
-        wload.conf(kind="profile", params=cls.params)
-        cls.env.ftrace.start()
-        wload.run(
-            out_dir=cls.env.res_dir,
-            background=False)
-        cls.env.ftrace.stop()
-        trace = cls.env.ftrace.get_trace(cls.trace_file)
+    conf_basename = "wake_migration.config"
+    # TODO...
```
derkling: What?

tests/eas/acceptance.py
```diff
@@ -378,193 +378,65 @@ class WakeMigration(unittest.TestCase):
...
+    phase_duration = WORKLOAD_DURATION_S
+
+    def _assert_switch(self, experiment, expected_switch_to, phases):
```
derkling: Again, this seems more an update to the test than a refactoring, isn't it?
Maybe we can still use one additional patch with these changes?

tests/eas/acceptance.py
```diff
@@ -378,193 +378,65 @@ class WakeMigration(unittest.TestCase):
...
+        if expected_switch_to == "big":
```
derkling: Cannot we really avoid this big.LITTLE-ism? Perhaps not right now... :-/

tests/eas/acceptance.py
```diff
@@ -378,193 +378,65 @@ class WakeMigration(unittest.TestCase):
...
-    def test_big_little_switch2(self):
+    @experiment_test
+    def test_big_little_switch2(self, experiment, tasks):
         """Wake Migration: BIG -> LITLLE: 2"""
 
         # little - big - little - big
         #              ^
         # We want to test that this big to little migration happens.  So we skip
         # the first two phases.
-        expected_time = self.offset + 2 * self.phase_duration
```
derkling: That patch is changing so much... we should really try to split it in two, or three if required, to make more easy to understand what it is doing...

libs/utils/executor.py
bjackman: Nice, will give that a go.

libs/utils/executor.py
bjackman: Agreed.

libs/utils/executor.py
bjackman: Agreed on the change to the format call, but not on the self.te.test_dir -> test_dir:

Before this change, `self.te.test_dir` is global state that is picked up later in `_wload_run_init`. I could put this change in a previous patch which also modifies _wload_run and _wload_init to take a test_dir argument, but it seems unnecessary given that it will immediately be undone...

libs/utils/executor.py
bjackman: Yep, looks like this is leakage from another version of the commit.

libs/utils/executor.py
bjackman: That's a mistake. I think I'll drop this commit.

tests/eas/acceptance.py
bjackman: It's printed next to the failure report, the output ends up looking like this:
```
...
04:49:03  INFO    :       Executor -    <res_dir>/rtapp::wake_migration/1/trace_stat.json
04:49:03  INFO    : --------------------------------------------------------------------------------
04:49:03  INFO    : 
04:49:03  INFO    : ################################################################################
04:49:03  INFO    :       Executor - Experiments execution completed
04:49:03  INFO    : ################################################################################
04:49:03  INFO    :       Executor - Results available in:
04:49:03  INFO    :       Executor -       /vagrant/results/20160927_164800
Wake Migration: BIG -> LITLLE: 1 ... 04:49:07  ERROR   : Check trace file: /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
ok
Wake Migration: BIG -> LITLLE: 2 ... 04:49:11  ERROR   : Check trace file: /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
ok
Test First CPU ... ok
Wake Migration: LITTLE -> BIG: 1 ... 04:49:17  ERROR   : Check trace file: /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
ok
Wake Migration: LITTLE -> BIG: 2 ... 04:49:20  ERROR   : Check trace file: /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
ok

----------------------------------------------------------------------
Ran 5 tests in 91.952s

OK
```

I've just realised that's different from the original output (it shouldn't say "OK" at the end) so I'll go back and take a look. This `raise` might not have the semantics I thought it did.

tests/eas/acceptance.py
bjackman: Good point. I like the idea of the `_test_first_cpu` method; that way we don't duplicate the actual test logic but it's explicit that each test class includes that test.

tests/eas/fork_migration.config
bjackman: Agreed; I'll look at this after this rework is done.

I'm actually a little concerned that JSON representation might end up adding more complexity than simplicity because ideally we don't want to have any platform knowledge ("A53_0" or even "bigs"/"mediums") in the test. We can discuss this separately.

tests/eas/acceptance.py
bjackman: Seems sensible.

tests/eas/acceptance.py
bjackman: Agreed, how about `earlystarters_assert`, `migrators_assert` and `all_assert`?

tests/eas/acceptance.py
bjackman: I'm not sure this can really be done since using LisaTest the `SchedMultiAssert`s need to be per-experiment and we don't have access to the experiment or tasks until the `test_` method. I'll have a think about it.

tests/eas/acceptance.py
```diff
@@ -85,6 +88,52 @@ def log_result(data, log_fh):
...
+        except AssertionError as e:
+            trace_relpath = os.path.join(experiment['out_dir'], "trace.dat")
+            logging.error("Check trace file: " + os.path.abspath(trace_relpath))
+
+class EasTest(LisaTest):
+    """
+    Base class for EAS tests
+    """
+
+    set_is_big_little = True
```
JaviMerino: This is not used anywhere in the commit.  Should it be moved to the commit where it is actually used?

tests/eas/acceptance.py
bjackman: Yes, I think this can be split up.

tests/eas/offload_idle_pull.config
bjackman: Sorry, commit leakage again here. I was thinking about the fact that we want to specify the required events in order to avoid enormous traces, but they are the same for all the EAS tests. This is something to improve later, I think.

tests/eas/acceptance.py
bjackman: Oops, more commit leakage.

tests/eas/acceptance.py
bjackman: Agreed, I'll split this up.

tests/eas/acceptance.py
bjackman: Yeah, I think this is something to remove later.

tests/eas/acceptance.py
bjackman: Agreed, will split this patch into one that ports to EasTest and one that factors out the commonality between the switch tests.

tests/eas/acceptance.py
bjackman: This isn't used at all (it's left over from before I spotted the "sched_features" in the Executor class' config). WIll remove

tests/eas/acceptance.py
```diff
@@ -177,82 +177,24 @@ class SmallTaskPacking(unittest.TestCase):
...
+                self.te.target.bl.littles,
                 EXPECTED_RESIDENCY_PCT,
                 operator.ge,
                 percent=True,
-                rank=self.num_tasks),
+                rank=len(tasks)),
             msg="Not all tasks are running on LITTLE cores for at least {}% of their execution time"\
-                    .format(EXPECTED_RESIDENCY_PCT))
-
+            .format(EXPECTED_RESIDENCY_PCT))
```
JaviMerino: Why remove the indentation?  It's a continuation line and pep8 says "Make sure to indent the continued line appropriately".

bjackman:
Since this PR is enormous and I'm rapidly discovering that the GitHub interface is fair from ideal I will rework this and **submit a separate PR as v2** unless anyone objects.

tests/eas/acceptance.py
JaviMerino: The EAS acceptance tests are all about big.LITTLE I'm afraid.  They use the "bl" module of target!

tests/eas/acceptance.py
```diff
@@ -378,193 +378,65 @@ class WakeMigration(unittest.TestCase):
...
+
+    def _assert_switch(self, experiment, expected_switch_to, phases):
+        if expected_switch_to == "big":
+            switch_from = self.te.target.bl.littles
+            switch_to   = self.te.target.bl.bigs
+        elif expected_switch_to == "little":
+            switch_from = self.te.target.bl.bigs
+            switch_to   = self.te.target.bl.littles
+        else:
+            raise Exception("Invalid expected_switch_to")
```
JaviMerino: `ValueError` is more specific than `Exception`

tests/eas/acceptance.py
```diff
@@ -378,193 +378,65 @@ class WakeMigration(unittest.TestCase):
...
+
+        expected_time = (self.get_start_time(experiment) +
+                         WORKLOAD_DURATION_S * phases)
```
JaviMerino: I know we are almost never consistent with this part of pep8, but it recommends breaking line before the `+` and removing the spaces around the `*` because it takes higher preference

``` python
        expected_time = (self.get_start_time(experiment)
                         + WORKLOAD_DURATION_S*phases)
```

https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
https://www.python.org/dev/peps/pep-0008/#other-recommendations

tests/eas/acceptance.py
bjackman: Oops, this is editor noise, will undo.

tests/eas/acceptance.py
bjackman: Agreed.

tests/eas/acceptance.py
bjackman: Ah, I didn't know about that. Will change.

tests/eas/acceptance.py
bjackman: (oh, the `raise` isn't there at all :$)

tests/eas/acceptance.py
JaviMerino: Bear in mind that the variables are short because the 80-character line restriction.  We are dropping the `self.` from the beginning, which gives you 5 extra characters, but still, some of the asserts may become too hard to read with longer variable names.

JaviMerino:
While you are at it, you could fix "targt" in the log output in `libs/utils/executor.py`, line 112

libs/utils/executor.py
```diff
@@ -130,19 +130,28 @@ def __init__(self, target_conf=None, tests_conf=None):
...
                 for itr_idx in range(1, self._iterations + 1):
-                    # WORKLOAD: execution
-                    self._wload_run(exp_idx, tc, wl_idx, wload, itr_idx)
-                    exp_idx += 1
+                    self.experiments.append({
+                        'wload_name' : wl_idx,
+                        'wload'      : wload,
+                        'conf'       : tc,
+                        'iteration'  : itr_idx,
+                        'out_dir'    : os.path.join(test_dir, str(itr_idx))
```
JaviMerino: Is it possible to give the `out_dir` a more meaningful name?  Before the traces had relevant names (e.g. `result_latest/fork_migration.dat`.  Now the trace for that experiment would be in `results_latest/1/trace.dat` which is far less readable.

tests/eas/acceptance.py
```diff
@@ -378,193 +378,65 @@ class WakeMigration(unittest.TestCase):
...
-            "wake_migration",
-            calibration=cls.env.calibration())
-        wload.conf(kind="profile", params=cls.params)
-        cls.env.ftrace.start()
-        wload.run(
-            out_dir=cls.env.res_dir,
-            background=False)
-        cls.env.ftrace.stop()
-        trace = cls.env.ftrace.get_trace(cls.trace_file)
+    conf_basename = "wake_migration.config"
```
JaviMerino: You forgot to `git add tests/eas/wake_migration.config`

JaviMerino:
I've tried running this, but it explodes in a ball of flames.  Some of the errors that I see, other than what I have pointed out in the review:
```
acceptance.experiment_test ... ERROR

======================================================================
ERROR: test suite for <class 'acceptance.EasTest'>
----------------------------------------------------------------------
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 208, in run
    self.setUp()
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 291, in setUp
    self.setupContext(ancestor)
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 314, in setupContext
    try_run(context, names)
  File "/usr/lib/python2.7/dist-packages/nose/util.py", line 471, in try_run
    return func()
  File "/vagrant/tests/eas/acceptance.py", line 66, in setUpClass
    cls.conf_basename)
AttributeError: type object 'EasTest' has no attribute 'conf_basename'

======================================================================
ERROR: test suite for <class 'acceptance.ForkMigration'>
----------------------------------------------------------------------
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 208, in run
    self.setUp()
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 291, in setUp
    self.setupContext(ancestor)
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 314, in setupContext
    try_run(context, names)
  File "/usr/lib/python2.7/dist-packages/nose/util.py", line 471, in try_run
    return func()
  File "/vagrant/tests/eas/acceptance.py", line 68, in setUpClass
    super(EasTest, cls)._init(conf_file, *args, **kwargs)
  File "/home/vagrant/lisa/libs/utils/test.py", line 53, in _init
    cls._runExperiments()
  File "/home/vagrant/lisa/libs/utils/test.py", line 72, in _runExperiments
    cls.executor.run()
  File "/home/vagrant/lisa/libs/utils/executor.py", line 154, in run
    self._wload_run(exp_idx, experiment)
  File "/home/vagrant/lisa/libs/utils/executor.py", line 526, in _wload_run
    self.te.emeter.report(experiment['out_dir'])
  File "/home/vagrant/lisa/libs/utils/energy.py", line 206, in report
    nrg_total = nrg[label]['total']
KeyError: 'a57'
-------------------- >> begin captured logging << --------------------
```

libs/utils/executor.py
bjackman: Removed the todo; it would be nice to get rid of the calculation of `self._exp_count` in `__init__` but for now I think it's more trouble than it's worth.

Re: s/experiment/exp/: why? We aren't short on horizontal space (the `format` call has to be broken up, but it would be anyway), so why not be totally clear on what the variable refers to?

tests/eas/acceptance.py
bjackman: This is actually already there in LisaTest :)

bjackman:
@JaviMerino Re: hwmon error: It looks at hwmon because it's in the default Devlib modules. I thought that that error was due to the way I'd set up my kernel (the hwmon module doesn't produce any errors but it produces empty data, hence they key errors) and disabled hwmon in my target.config. My plan was in a separate PR to a) make hwmon disabled by default (only enabled on tests that need it) and b) make it fail more gracefully when the board doesn't support it as expected. Since you're seeing the error too, I'll take a look at it before submitting v2 of this.

Re: missing wake_migration.config, I noticed that shortly after submitting the PR :$ will be back in v2.

Re: s/targt/target/: will do.

bjackman:
The HWMon issue is present in master too, but it doesn't manifest because the existing acceptance tests don't call `report()` (whereas the Executor class does this after running the workload).

I'm going to disable hwmon in the test configs (unfortunately I can't see a way to disable using energy meters in general, only hwmon specifically), and raise a separate issue to figure out why HWMon isn't working. Disabling it in the test configs is reasonably sensible anyway since they don't need energy readings, having hwmon just slows things down.

libs/utils/executor.py
derkling: I agree on using clear names... just wondering is exp can be a shorter still clear name.
No main concerns if you prefer to keep experiment ;-)

tests/eas/acceptance.py
derkling: I like it... maybe we can improve a bit readability by moving the trace link in a new line?
For example:
```
...
04:49:03  INFO    :       Executor -    <res_dir>/rtapp::wake_migration/1/trace_stat.json
04:49:03  INFO    : --------------------------------------------------------------------------------
04:49:03  INFO    : 
04:49:03  INFO    : ################################################################################
04:49:03  INFO    :       Executor - Experiments execution completed
04:49:03  INFO    : ################################################################################
04:49:03  INFO    :       Executor - Results available in:
04:49:03  INFO    :       Executor -       /vagrant/results/20160927_164800
Wake Migration: BIG -> LITLLE: 1 ... 04:49:07  ERROR   : Check trace file:
   /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
FAILED
Wake Migration: BIG -> LITLLE: 2 ... 04:49:11  ERROR   : Check trace file:
   /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
FAILED
Test First CPU ... ok
Wake Migration: LITTLE -> BIG: 1 ... 04:49:17  ERROR   : Check trace file:
   /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
FAILED
Wake Migration: LITTLE -> BIG: 2 ... 04:49:20  ERROR   : Check trace file:
   /vagrant/results/20160927_164800/rtapp::wake_migration/1/trace.dat
FAILED

----------------------------------------------------------------------
Ran 5 tests in 91.952s

OK
```

tests/eas/acceptance.py
derkling: Right.... I usually use alias here and there ;-)

tests/eas/acceptance.py
derkling: What about shrinking the assert to a single "_a" suffix?
i.e. `earlystarters_a`, `migrators_a` and `all_a`

tests/eas/fork_migration.config
derkling: Agree, let's have a brainstorming session one I'm back... that topic we have to got it right this next time. ;-)

tests/eas/acceptance.py
JaviMerino: yeah, that would work

bjackman: There are actually no instances where the long names force us to split lines we wouldn't otherwise have to split so I'm going to go with the longer ones for now, feel free to object on the v2 PR.

Full disclosure: I find 2-dimensional code where function call arguments are split across lines _easier_ to read so I'm generally biased in favour of [long names](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition/blob/master/src/main/java/com/seriouscompany/business/java/fizzbuzz/packagenamingpackage/impl/Main.java)

JaviMerino: Fair enough, I must admit I didn't check whether longer names would make more splits, but I did remember struggling with line length and variable names.  Maybe it was in another part of this file.

A little bit of a digression following on your full disclosure: I'm happy with long variable names, but I don't think we want to look at FizzBuzzEnterpriseEdition for inspiration regarding coding practices :wink: .  Code is "read-mostly", so I'm all for descriptive variable names.  I'm just not happy when we hit the 70th column and all the code starts going vertical instead of horizontal, if you know what I mean.

bjackman:
superseded by v2: https://github.com/ARM-software/lisa/pull/202

PR #202, opened 2016-10-03
Rework EAS acceptance tests to inherit from LisaTest  v2

bjackman:
This is a successor to https://github.com/ARM-software/lisa/pull/197

Differences from v1:
- Numerous minor changes and cleanups I won't list
- Split the changes into more commits
- In Executor, the `experiments` dict was changed to be a NamedTuple-
- Removed `test_first_cpu` from `EasTest` and replaced it with a helper method, subclasses must explicitly call it
- `experiment_test` decorator moved to `utils.test` (where `LisaTest` is)
- Added missing `is_big_little` logic to `EasTest`
- Fixed the logic `EasTest` that attaches trace file locations to test failure messages

----

- Refactor/enhance the Executor class; most importantly enable the "experiments" to be saved so that test objects can refer back to them in `test_` methods
- Add an EasTest class abstracting the common operations for all EAS acceptance tests. It's likely that some of this functionality (+ documentation which is not written) could go directly into LisaTest at a later date
- Rework each of the acceptance test classes to inherit from EasTest

Also some miscellaneous cleanups that seem to me to be related to this.

### DIFF
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -108,7 +108,7 @@ class Executor():
 
         logging.info('%14s - Configured to run:', 'Executor')
 
-        logging.info('%14s -   %3d targt configurations:',
+        logging.info('%14s -   %3d target configurations:',
                      'Executor', len(self._tests_conf['confs']))
         target_confs = [conf['tag'] for conf in self._tests_conf['confs']]
         target_confs = ', '.join(target_confs)
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -484,7 +484,7 @@ class Executor():
         # Keep track of platform configuration
         test_dir = '{}/{}:{}:{}'\
             .format(self.te.res_dir, wload.wtype, tc_idx, wl_idx)
-        os.system('mkdir -p ' + test_dir)
+        os.makedirs(test_dir)
         self.te.platform_dump(test_dir)
 
         # Keep track of kernel configuration and version
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -381,9 +381,7 @@ class Executor():
         if conf['class'] == 'profile':
             params = {}
             # Load each task specification
-            for task_name in conf['params']:
-                task = conf['params'][task_name]
-                task_name = conf['prefix'] + task_name
+            for task_name, task in conf['params'].items():
                 if task['kind'] not in wlgen.__dict__:
                     logging.error(r'%14s - RTA task of kind [%s] not supported',
                             'RTApp', task['kind'])
@@ -392,7 +390,13 @@ class Executor():
                         'in RT-App workload specification'\
                         .format(task))
                 task_ctor = getattr(wlgen, task['kind'])
-                params[task_name] = task_ctor(**task['params']).get()
+                num_tasks = task.get('tasks', 1)
+                task_idxs = self._wload_task_idxs(wl_idx, num_tasks)
+                for idx in task_idxs:
+                    idx_name = str(idx) if len(task_idxs) > 0 else ""
+                    task_name_idx = conf['prefix'] + task_name + idx_name
+                    params[task_name_idx] = task_ctor(**task['params']).get()
+
             rtapp = wlgen.RTA(self.target,
                         wl_idx, calibration = self.te.calibration())
             rtapp.conf(kind='profile', params=params, loadref=loadref,
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -17,6 +17,7 @@
 
 from bart.common.Analyzer import Analyzer
 import collections
+from collections import namedtuple
 import datetime
 import gzip
 import json
@@ -40,6 +41,9 @@ from env import TestEnv
 from conf import JsonConf
 
 import wlgen
+
+Experiment = namedtuple('Experiment', ['wload_name', 'wload',
+                                       'conf', 'iteration', 'out_dir'])
 
 class Executor():
 
@@ -130,19 +134,28 @@ class Executor():
     def run(self):
         self._print_section('Executor', 'Experiments execution')
 
+        self.experiments = []
+
         # Run all the configured experiments
-        exp_idx = 1
         for tc in self._tests_conf['confs']:
             # TARGET: configuration
             if not self._target_configure(tc):
                 continue
             for wl_idx in self._tests_conf['wloads']:
                 # TEST: configuration
-                wload = self._wload_init(tc, wl_idx)
+                wload, test_dir = self._wload_init(tc, wl_idx)
                 for itr_idx in range(1, self._iterations + 1):
-                    # WORKLOAD: execution
-                    self._wload_run(exp_idx, tc, wl_idx, wload, itr_idx)
-                    exp_idx += 1
+                    exp = Experiment(
+                        wload_name=wl_idx,
+                        wload=wload,
+                        conf=tc,
+                        iteration=itr_idx,
+                        out_dir=os.path.join(test_dir, str(itr_idx)))
+                    self.experiments.append(exp)
+
+            # WORKLOAD: execution
+            for exp_idx, experiment in enumerate(self.experiments):
+                self._wload_run(exp_idx, experiment)
 
         self._print_section('Executor', 'Experiments execution completed')
         logging.info('%14s - Results available in:', 'Executor')
@@ -479,24 +492,21 @@ class Executor():
         with open(os.path.join(self.te.test_dir, 'kernel.version'), 'w') as fh:
             fh.write(output)
 
-        return wload
+        return wload, self.te.test_dir
 
-    def _wload_run_init(self, run_idx):
-        self.te.out_dir = '{}/{}'\
-                .format(self.te.test_dir, run_idx)
-        logging.debug(r'%14s - out_dir [%s]', 'Executor', self.te.out_dir)
-        os.system('mkdir -p ' + self.te.out_dir)
-
-    def _wload_run(self, exp_idx, tc, wl_idx, wload, run_idx):
+    def _wload_run(self, exp_idx, experiment):
+        tc = experiment.conf
+        wload = experiment.wload
         tc_idx = tc['tag']
 
         self._print_title('Executor', 'Experiment {}/{}, [{}:{}] {}/{}'\
                 .format(exp_idx, self._exp_count,
-                        tc_idx, wl_idx,
-                        run_idx, self._iterations))
+                        tc_idx, experiment.wload_name,
+                        experiment.iteration, self._iterations))
 
         # Setup local results folder
-        self._wload_run_init(run_idx)
+        logging.debug(r'%14s - out_dir [%s]', 'Executor', experiment.out_dir)
+        os.system('mkdir -p ' + experiment.out_dir)
 
         # FTRACE: start (if a configuration has been provided)
         if self.te.ftrace and self._target_conf_flag(tc, 'ftrace'):
@@ -508,23 +518,23 @@ class Executor():
             self.te.emeter.reset()
 
         # WORKLOAD: Run the configured workload
-        wload.run(out_dir=self.te.out_dir, cgroup=self._cgroup)
+        wload.run(out_dir=experiment.out_dir, cgroup=self._cgroup)
 
         # ENERGY: collect measurements
         if self.te.emeter:
-            self.te.emeter.report(self.te.out_dir)
+            self.te.emeter.report(experiment.out_dir)
 
         # FTRACE: stop and collect measurements
         if self.te.ftrace and self._target_conf_flag(tc, 'ftrace'):
             self.te.ftrace.stop()
 
-            trace_file = self.te.out_dir + '/trace.dat'
+            trace_file = experiment.out_dir + '/trace.dat'
             self.te.ftrace.get_trace(trace_file)
             logging.info(r'%14s - Collected FTrace binary trace:', 'Executor')
             logging.info(r'%14s -    %s', 'Executor',
                          trace_file.replace(self.te.res_dir, '<res_dir>'))
 
-            stats_file = self.te.out_dir + '/trace_stat.json'
+            stats_file = experiment.out_dir + '/trace_stat.json'
             self.te.ftrace.get_stats(stats_file)
             logging.info(r'%14s - Collected FTrace function profiling:', 'Executor')
             logging.info(r'%14s -    %s', 'Executor',
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -99,8 +99,9 @@ class Executor():
         self.te = TestEnv(target_conf, tests_conf)
         self.target = self.te.target
 
+        self._iterations = self._tests_conf.get('iterations', 1)
         # Compute total number of experiments
-        self._exp_count = self._tests_conf['iterations'] \
+        self._exp_count = self._iterations \
                 * len(self._tests_conf['wloads']) \
                 * len(self._tests_conf['confs'])
 
@@ -116,7 +117,7 @@ class Executor():
 
         logging.info('%14s -   %3d workloads (%d iterations each)',
                      'Executor', len(self._tests_conf['wloads']),
-                     self._tests_conf['iterations'])
+                     self._iterations)
         wload_confs = ', '.join(self._tests_conf['wloads'])
         logging.info('%14s -       %s', 'Executor', wload_confs)
 
@@ -138,7 +139,7 @@ class Executor():
             for wl_idx in self._tests_conf['wloads']:
                 # TEST: configuration
                 wload = self._wload_init(tc, wl_idx)
-                for itr_idx in range(1, self._tests_conf['iterations']+1):
+                for itr_idx in range(1, self._iterations + 1):
                     # WORKLOAD: execution
                     self._wload_run(exp_idx, tc, wl_idx, wload, itr_idx)
                     exp_idx += 1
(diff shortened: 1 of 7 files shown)


libs/utils/test.py
```diff
@@ -100,4 +102,24 @@ def _experimentsFinalize(cls):
...
+    The method will be passed the experiment object and a list of the names of
+    tasks that were run as the experiment's workload.
+    """
+    for experiment in instance.executor.experiments:
+        tasks = experiment.wload.tasks.keys()
+        try:
+            wrapped_test(experiment, tasks, *args, **kwargs)
+        except AssertionError as e:
+            trace_relpath = os.path.join(experiment.out_dir, "trace.dat")
+            e.args += ("Check trace file: " + os.path.abspath(trace_relpath),)
```
bjackman: This produces failure messages that look like this (copy-pasted from my terminal)

```
======================================================================
FAIL: Wake Migration: LITTLE -> BIG: 1
----------------------------------------------------------------------
  File "/home/vagrant/lisa/libs/utils/test.py", line 116, in experiment_test
    wrapped_test(experiment, tasks, *args, **kwargs)
  File "/vagrant/tests/eas/acceptance.py", line 381, in test_little_big_switch1
    self._assert_switch(experiment, "big", 1)
  File "/vagrant/tests/eas/acceptance.py", line 376, in _assert_switch
    msg=msg)
AssertionError: ('Not all tasks wake-migrated to big cores in the expected window: (5.3126789999660105, 6.3126789999660105)', 'Check trace file: /vagrant/results/20161003_112951/rtapp::wake_migration/1/trac)

----------------------------------------------------------------------
Ran 5 tests in 47.696s
```

Note that in this case, where the original failure message is long, the trace file path is truncated by the width of my terminal; in order to see it I have to scroll to the right :disappointed: . 

The previous version of this logic didn't actually work; it prevented the test failures from being recognised by `nose` (it swallowed the `AssertionError`). When I fixed that mistake, the "Check trace file" message was no longer reported next to the test failure. So this isn't ideal.

tests/eas/fork_migration.config
```diff
@@ -0,0 +1,53 @@
...
+        // the number of big CPUs and M is the number of LITTLE CPUs.
+        "fmig" : {
+            "type" : "rt-app",
+            "conf" : {
+                "class" : "profile",
+                "params" : {
+                    "small" : {
+                        "kind" : "Periodic",
+                        "params" : {
+                            "duty_cycle_pct": 10,
```
bjackman: Note that these values used to be configured by values in acceptance.config, but those values can no longer be used here. I think this can be something that we discuss later along with workload description in general, along with what @derkling mentioned [previously](https://github.com/ARM-software/lisa/pull/197#discussion_r80811949)

tests/eas/acceptance.py
```diff
@@ -84,6 +86,52 @@ def log_result(data, log_fh):
...
+                cls.target.write_value(
+                    "/proc/sys/kernel/sched_is_big_little", 1)
+            except TargetError:
+                # That flag doesn't exist on mainline-integration kernels, so
+                # don't worry if the file isn't present.
+                pass
+
+    def get_start_time(self, experiment):
+        start_times_dict = self.get_multi_assert(experiment).getStartTime()
+        return min([t["starttime"] for t in start_times_dict.values()])
```
bjackman: Amended and pushed.

bjackman:
@JaviMerino  Re `--nologcapture` and `os.makedirs`: I'll submit separate PRs for those (I know I already put the "fix typo" and "use `min`" commit in here, but now it feels like going to far!).

libs/utils/test.py
derkling: Maybe a possible alternative solution could be to split the 'Check trace...' message as a WARNING log message right before raising the exception.

I would expect in that case we should get the trace path right before the Traceback.

You can also split some useful "\n\n\n" right at the end of the warning message to keep some space before the traceback...

libs/utils/executor.py
```diff
@@ -478,24 +491,21 @@ def _wload_init(self, tc, wl_idx):
...
                 .format(exp_idx, self._exp_count,
-                        tc_idx, wl_idx,
-                        run_idx, self._iterations))
+                        tc_idx, experiment.wload_name,
+                        experiment.iteration, self._iterations))
 
         # Setup local results folder
-        self._wload_run_init(run_idx)
+        logging.debug(r'%14s - out_dir [%s]', 'Executor', experiment.out_dir)
+        os.system('mkdir -p ' + experiment.out_dir)
```
derkling: No problems for me with having many commits on a PR ;-)

libs/utils/test.py
JaviMerino: See my previous message regarding capturing logging.  If you see the output it gets really messy: you first see the normal logging and then when there is a failure, you see the all the logging again with the traceback.  If you have multiple failures, it is very hard to see them all.  

My personal preference is what this patch does and run nose with `--nologcapture`.

bjackman: OK, pushed a new version that ends up looking like this, which I think is ideal:
```
10:16:15  INFO    : --------------------------------------------------------------------------------
10:16:15  INFO    : 
10:16:15  INFO    : ################################################################################
10:16:15  INFO    :       Executor - Experiments execution completed
10:16:15  INFO    : ################################################################################
10:16:15  INFO    :       Executor - Results available in:
10:16:15  INFO    :       Executor -       /vagrant/results/20161004_101543
Wake Migration: BIG -> LITLLE: 1 ... FAIL
Wake Migration: BIG -> LITLLE: 2 ... ok
Wake Migration: Test First CPU ... ok
Wake Migration: LITTLE -> BIG: 1 ... FAIL
Wake Migration: LITTLE -> BIG: 2 ... ok

======================================================================
FAIL: Wake Migration: BIG -> LITLLE: 1
----------------------------------------------------------------------
  File "/home/vagrant/lisa/libs/utils/test.py", line 116, in experiment_test
    wrapped_test(experiment, tasks, *args, **kwargs)
  File "/vagrant/tests/eas/acceptance.py", line 396, in test_big_little_switch1
    self._assert_switch(experiment, "little", 0)
  File "/vagrant/tests/eas/acceptance.py", line 376, in _assert_switch
    msg=msg)
AssertionError: Not all tasks wake-migrated to little cores in the expected window: (0.29281700000137789, 1.2928170000013779)
	Check trace file: /vagrant/results/20161004_101543/rtapp::wake_migration/1/trace.dat

======================================================================
FAIL: Wake Migration: LITTLE -> BIG: 1
----------------------------------------------------------------------
  File "/home/vagrant/lisa/libs/utils/test.py", line 116, in experiment_test
    wrapped_test(experiment, tasks, *args, **kwargs)
  File "/vagrant/tests/eas/acceptance.py", line 381, in test_little_big_switch1
    self._assert_switch(experiment, "big", 1)
  File "/vagrant/tests/eas/acceptance.py", line 376, in _assert_switch
    msg=msg)
AssertionError: Not all tasks wake-migrated to big cores in the expected window: (5.2928170000013779, 6.2928170000013779)
	Check trace file: /vagrant/results/20161004_101543/rtapp::wake_migration/1/trace.dat

----------------------------------------------------------------------
Ran 5 tests in 46.439s

FAILED (failures=2)
```

bjackman:
OK the `lisa-test` change felt like it should go in with the trace file location change so I thought "since @derkling has hinted that he enjoys large PRs, let's put the `makedirs` thing in here too". No v3 because the changes are minor.

libs/utils/executor.py
```diff
@@ -477,21 +477,21 @@ def _wload_init(self, tc, wl_idx):
         wload = self._wload_conf(wl_idx, wlspec)
 
         # Keep track of platform configuration
-        self.te.test_dir = '{}/{}:{}:{}'\
+        test_dir = '{}/{}:{}:{}'\
             .format(self.te.res_dir, wload.wtype, tc_idx, wl_idx)
-        os.system('mkdir -p ' + self.te.test_dir)
-        self.te.platform_dump(self.te.test_dir)
+        os.system('mkdir -p ' + test_dir)
```
derkling: Should we better use Python's APIs?
```python
try:
    os.stat(test_dir)
except:
    os.mkdir(test_dir)
```
I guess this should take care of potential spaces in the path.

tests/eas/acceptance.py
```diff
@@ -84,6 +86,52 @@ def log_result(data, log_fh):
...
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        conf_file = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                                 cls.conf_basename)
+
+        super(EasTest, cls)._init(conf_file, *args, **kwargs)
+
+    @memoized
+    def get_multi_assert(self, experiment, task_filter=""):
```
derkling: I :heart: that function! And actually defining it as non private enables auto-completion in derived classes...
:+1:

tests/eas/acceptance.py
```diff
@@ -84,6 +86,52 @@ def log_result(data, log_fh):
...
+        if SET_IS_BIG_LITTLE:
+            try:
+                cls.target.write_value(
+                    "/proc/sys/kernel/sched_is_big_little", 1)
+            except TargetError:
+                # That flag doesn't exist on mainline-integration kernels, so
+                # don't worry if the file isn't present.
+                pass
+
+    def get_start_time(self, experiment):
```
derkling: :+1: 
Although not required not, maybe we can add a dual `get_end_time(self. experiment)` just for consistency?

tests/eas/fork_migration.config
```diff
@@ -0,0 +1,54 @@
...
+                            "duration_s": 5,
+                            "period_ms": 10,
+                        },
+                        "tasks" : "little", // Create one task for each LITTLE CPU
```
derkling: Can we just move the comment above to save columns?
```python
// Create one task for each LITTLE CPU
"tasks" : "little",
```

tests/eas/acceptance.py
```diff
@@ -133,7 +133,7 @@ def _do_test_first_cpu(self, experiment, tasks):
                 rank=len(tasks)),
             msg="Not all the new generated tasks started on a big CPU")
 
-class ForkMigration(unittest.TestCase):
+class ForkMigration(EasTest):
```
derkling: As @JaviMerino used to say: I :heart_eyes: these massively red coloured patches ;-)

tests/eas/acceptance.py
```diff
@@ -252,130 +252,39 @@ class OffloadMigrationAndIdlePull(unittest.TestCase):
...
-        task_start_times = cls.a_assert.getStartTime().values()
-        return min([t['starttime'] for t in task_start_times])
+    @experiment_test
+    def test_first_cpu(self, experiment, tasks):
+        """Offload Migration and Idle Pull: Test First CPU"""
+        self._do_test_first_cpu(experiment, tasks)
 
     @classmethod
-    def calculate_end_times(cls):
+    def calculate_end_times(cls, experiment):
```
derkling: Isn't that another candidate functions to be moved into the base class?
Maybe making it memorized as well ;-)

derkling: I think in general it would be lovely to have derived classes which exposes only "test_" methods.

tests/eas/acceptance.py
```diff
@@ -395,17 +304,27 @@ def test_big_cpus_fully_loaded(self):
...
+    @experiment_test
+    def test_little_cpus_run_tasks(self, experiment, tasks):
         """Offload Migration and Idle Pull: Little cpus run tasks while bigs are busy"""
-        tasks = self.params.keys()
+
         num_offloaded_tasks = len(tasks) / 2
 
-        first_task_finish_time = min(self.end_times.values())
+        end_times = self.calculate_end_times(experiment).values()
+        first_task_finish_time = min(end_times)
```
derkling: What about having the end_times function to return a:

```python
EndTimes = namedtuple("EndTimes", ['values', 'first', 'last'])
```
where values is populated as `self.calculate_end_times(experiment).values()`, while `first` and `last` are already populated to be min and max of the end_times?

Such a tuple can be returned by the new base class's  method `LisaTest::end_times(self, experiment)`.

tests/eas/acceptance.py
derkling: Look at the following comment related to the proposed `EndTimes` namedtuple.

tests/eas/acceptance.py
```diff
@@ -395,17 +304,27 @@ def test_big_cpus_fully_loaded(self):
...
+
+        migrators_assert = self.get_multi_assert(experiment, "migrator")
+        start_time = min(t["starttime"]
```
derkling: For start times we can probably have the same concept of EndTime.
Although I'm not completely sure we can populate it in a generic way...

If it should be possible than perhaps we should defined a more generic
```python
Times = namedtuple("Times", ['values', 'first', 'last'])
```
to be used both for start and end times ;-)

tests/eas/acceptance.py
```diff
@@ -393,34 +393,45 @@ class WakeMigration(EasTest):
...
-    @experiment_test
-    def test_little_big_switch1(self, experiment, tasks):
-        """Wake Migration: LITTLE -> BIG: 1"""
-        expected_time = self.get_start_time(experiment) + self.phase_duration
-        switch_window = (
-            expected_time -
-            SWITCH_WINDOW_HALF,
-            expected_time +
-            SWITCH_WINDOW_HALF)
+    def _assert_switch(self, experiment, expected_switch_to, phases):
```
derkling: Not completely sure... but perhaps which can be another candidate method to move into the base class?
I guess it can be quite usual to check for tasks switches in many different other test scenarios...

src/shell/lisa_shell
```diff
@@ -256,7 +256,7 @@ EOF
 
 function _lisa-test {
 CLASS=${1}
-nosetests -v tests/$CLASS
+nosetests --nocapture --nologcapture -v tests/$CLASS
```
derkling: Seems an interesting feature :+1:

libs/utils/executor.py
bjackman: This is taken care of in a separate [commit](https://github.com/ARM-software/lisa/pull/202/commits/334055c952058598f8c2706f0a0bc15cc8a75dcb)

libs/utils/executor.py
```diff
@@ -130,19 +134,28 @@ def __init__(self, target_conf=None, tests_conf=None):
...
+                    exp = Experiment(
+                        wload_name=wl_idx,
+                        wload=wload,
+                        conf=tc,
+                        iteration=itr_idx,
+                        out_dir=os.path.join(test_dir, str(itr_idx)))
+                    self.experiments.append(exp)
+
+            # WORKLOAD: execution
+            for exp_idx, experiment in enumerate(self.experiments):
```
bjackman: I didn't want to do this because it doesn't actually tell you anything useful about the experiment, we only use it to show execution progress in the log output. I think we could add it in if necessary later, but for now it would bloat the data IMO.

tests/eas/acceptance.py
bjackman: Also the `phases` argument is a specific notion for this test. Would definitely support moving it into a base class later if we had some existing commonality though.

tests/eas/acceptance.py
bjackman: Good point, I will have a think about making get_start_time and get_end_time symmetrical, factoring them into a base class, and using a namedtuple. This is probably something lots of tests could use.

libs/utils/executor.py
JaviMerino: surely you mean 92440b8875612b0f7eb6fd7a82d3b9939cbf28c8 ("Use os.makedirs instead of os.system('mkdir ...)") (thanks github :disappointed: )

tests/eas/acceptance.py
JaviMerino: I still say it!

src/shell/lisa_shell
derkling: Actually while running this code on a (misconfiguration) target to test it, I realized that this feature can be counter productive in case we have multiple failures. For example, if the testenv setup fails for whatever reasons, usually we split a WARNING statement explaining the reason for the failure right before raising the exception.
Unfortunately,  since with this patch all the test failures are grouped at the end, it becomes difficult to correlate an error with its explanation.

I would propose to remove this patch from this PR and post-pone the logs clean-up into a separate PR once we investigated a better solution.

tests/eas/acceptance.py
```diff
@@ -141,68 +141,12 @@ class ForkMigration(unittest.TestCase):
...
-
-        log_result(
-            f_assert.getFirstCpu(), self.log_fh)
-
-        self.assertTrue(
-            f_assert.assertFirstCpu(
-                self.env.target.bl.bigs,
-                rank=self.num_tasks),
-            msg="Not all the new generated tasks started on a big CPU")
+    conf_basename = "fork_migration.config"
```
bjackman: Have made this change, but
- Haven't added the "acceptance" prefix to the `capacity_capping` config as that test is in a separate Python file...
- The ordering isn't quite what you said, it's like this:
```
acceptance.config
acceptance.py
acceptance_fork_migration.config
acceptance_offload_idle_pull.config
acceptance_small_task_packing.config
acceptance_wake_migration.config
capacity_capping.config
capacity_capping.py
rfc.config
rfc.py
```

bjackman:
Pushed a new version with @derkling's comments addressed and without the `lisa-test` changes.

src/shell/lisa_shell
JaviMerino: Up to you.  One of the problems that I had with the way we are doing logging right now is that if the first test of a testcase fails, then all the output from the test up to the point of failure is added as "captured stdout", including board and test set up.  If you have multiple test failures, it is very hard to see them all and differentiate the original stdout from the duplicated one.

Maybe what we should do is to remove the reloading of the logging module in the `Executor`.

bjackman:
🙌

PR #204, opened 2016-10-05
libs/utils/platforms: Add topology info for Juno

bjackman:


bjackman:
Although this info is present in the nexus5.json it shouldn't be necessary where the info is already provided by Devlib

PR #209, opened 2016-10-11
Add support for writing into files as part of the executor config

credp:
Hi there,

I wanted to be able to change the values of sched_is_big_little in different sections of tests_conf used with the Executor class in the kernel function profiling workbook. I added a 'files' config section so I could express the file and content I wanted to write. It looks like this in my workbook:

```
# Platform configurations to test
"confs" : [
    {
        "tag"            : "base",
        "flags"          : "ftrace",
        "sched_features" : "NO_ENERGY_AWARE",
        "cpufreq"        : {
            "governor" : "performance",
        },
    },
    {
        "tag"            : "eas",
        "flags"          : "ftrace",
        "sched_features" : "ENERGY_AWARE",
        "cpufreq"        : {
              "governor" : "performance",
        },
        "files"          : {
            '/proc/sys/kernel/sched_is_big_little' : '1',
        },
    },
    {
        "tag"            : "eas_SMP",
        "flags"          : "ftrace",
        "sched_features" : "ENERGY_AWARE",
        "cpufreq"        : {
              "governor" : "performance",
        },
        "files"          : {
            '/proc/sys/kernel/sched_is_big_little' : '0',
        },
    },
],
```

What do you guys think?


credp:
Just added a little error handling to log write failures.

derkling:
I think it's a quite useful feature... although it overlaps a bit with WA.
However, I understand it's a feature which makes sense to have in our really simple executor class.
+1 for me... I'll have a look at the code as well.

libs/utils/executor.py
```diff
@@ -272,13 +272,29 @@ def _setup_group(self, tc, group):
...
+        if 'files' not in tc:
+            logging.debug('%14s - Configuration not provided', 'Files')
+            return True
+        for name in tc['files']:
+            logging.info('%14s - Write value into file: \'%s\' -> \'%s\'',
+                         'Files', tc['files'][name], name)
+            try:
+                self.target.write_value(name, tc['files'][name], True)
+            except Exception:
+                logging.debug('%14s - Write value into file failed: \'%s\' -> \'%s\'',
```
derkling: Maybe better to split a warning here?

libs/utils/executor.py
```diff
@@ -272,13 +272,29 @@ def _setup_group(self, tc, group):
         # Configure each required attribute
         group.set(**tc['cgroups']['conf'][kind][name])
 
+    def _setup_files(self, tc):
+        errors = False
+        if 'files' not in tc:
+            logging.debug('%14s - Configuration not provided', 'Files')
+            return True
+        for name in tc['files']:
+            logging.info('%14s - Write value into file: \'%s\' -> \'%s\'',
```
derkling: Since the filepath is already quite long and we risk a line overflow
Can we keep this shorter?

Maybe just
```python
 logging.info('%14s - Write: \'%s\' -> \'%s\'',
              'Files', tc['files'][name], name)
```

libs/utils/executor.py
```diff
@@ -272,13 +272,29 @@ def _setup_group(self, tc, group):
...
+                         'Files', tc['files'][name], name)
+            try:
+                self.target.write_value(name, tc['files'][name], True)
+            except Exception:
```
bjackman: I think we should only catch `TargetError` here.

libs/utils/executor.py
bjackman: I think this should be `logging.error`. It's actually quite likely that errors will occur here, e.g. the if user is running the wrong kernel version/config, and we should complain loudly.

libs/utils/executor.py
```diff
@@ -272,13 +272,29 @@ def _setup_group(self, tc, group):
...
         group.set(**tc['cgroups']['conf'][kind][name])
 
+    def _setup_files(self, tc):
+        errors = False
+        if 'files' not in tc:
+            logging.debug('%14s - Configuration not provided', 'Files')
+            return True
+        for name in tc['files']:
+            logging.info('%14s - Write value into file: \'%s\' -> \'%s\'',
+                         'Files', tc['files'][name], name)
```
bjackman: It would be clearer to use `for name, value in tc['files'].iteritems()` above then use `value` instead of `tc['files'][name]` here and in the `write_value` call.

libs/utils/executor.py
```diff
@@ -272,13 +272,29 @@ def _setup_group(self, tc, group):
...
+            try:
+                self.target.write_value(name, tc['files'][name], True)
```
derkling: Forcing the value check can fails for files like sched_features.

In devlib a prefix token is used to mark files for which we do not want to check the value, i.e. append-mode files.

Maybe we can use something like:

```python
check_value = True
tc['files'][name].startswith('+/'):
    check_value = False
self.target.write_value(name, tc['files'][name], check_value)
```

which should allow to have something like:

```json
    "files" : {
        "+/sys/kernel/debug/sched_features" : "ENERGY_AWARE"
    }
```

libs/utils/executor.py
```diff
@@ -272,13 +272,29 @@ def _setup_group(self, tc, group):
...
+            except Exception:
+                logging.debug('%14s - Write value into file failed: \'%s\' -> \'%s\'',
+                         'Files', tc['files'][name], name)
+        return not errors
```
credp: I've changed it a little, we always check the write (to cause an exception to be thrown if it fails), and always catch the exception and log the failed write. If we are interested in a failure, then I re-raise the same exception. This gives me a nice log, plus the ability to completely fail if a required file isn't writeable for some reason.

ipynb/profiling/kernel_functions_profiling.ipynb
```diff
@@ -119,13 +119,43 @@
     "            \"cpufreq\"        : {\n",
     "                \"governor\" : \"performance\",\n",
     "            },\n",
+    "            # provide a set of files and values to write into them\n",
```
derkling: The idea of the example notebooks (btw this should be moved under examples/profiling), it's to show the bear minimum required to use a certain API.

This notebook's goal is to show how to profile kernel functions and for that in general it's not required to write files.
Thus, I would probably prefer to keep this notebook as much generic as possible.

Maybe we can add a simple example of files write in this other notebook:
   https://github.com/ARM-software/lisa/blob/master/ipynb/utils/testenv_example.ipynb
which is devoted to show how the test env can be configured.
Moreover, this notebook as well has to be generic... thus perhaps we can just add a couple of examples:
```python
"files"          : {
     # If filenames start with '/' then we do not check if the content
     # matches what we wrote, or even if the write was allowed
     # (i,e, the file existed, we have write permission, etc.)
     # this means we can use this function for sched_features as well
     # as for sysctls whose presence or permissions might depend
     # upon kernel config or OS support.
     # If filenames start with '!/' then we do check that the file exists
     # and the configured value has been properly written.
     '/this_file_doesnt_exist_and_we_dont_care'     : 1
     '!/proc/sys/kernel/sched_latency_ns'           : 16000000
}
```

libs/utils/executor.py
```diff
@@ -272,13 +273,34 @@ def _setup_group(self, tc, group):
         # Configure each required attribute
         group.set(**tc['cgroups']['conf'][kind][name])
 
+    def _setup_files(self, tc):
+        if 'files' not in tc:
+            logging.debug('%14s - Configuration not provided', 'Files')
+            return True
+        for name, value in tc['files'].iteritems():
+            check = True
+            if name.startswith('+/'):
```
derkling: When I proposed the "+" prefix I was mainly thinking about an "append" feature.

A syntax like this one:
```python
# By default we do not check for file to exists and/or written value
check = False
if name.startswith('!/'):
    # If the filename has a "!" at the beginning, than we want to enforce
    check = True
```
should not be more intuitive?

credp: I'll change all these but not until Ionella has done her move stuff.

derkling: Ok, thus let's remember that this depends on #219

libs/utils/executor.py
```diff
@@ -321,13 +322,34 @@ def _setup_group(self, tc, group):
         # Configure each required attribute
         group.set(**tc['cgroups']['conf'][kind][name])
 
+    def _setup_files(self, tc):
+        if 'files' not in tc:
+            logging.debug('%14s - Configuration not provided', 'Files')
```
derkling: We updated the way logging is used since #237, you should use this syntax now:

```python
self._log.debug('Configuration not provided')
```
Notice we also removed the odd %14s namespece field... ;-)

derkling: Since we removed "Files", this will appear as an "Executor" message. Thus, maybe a better log message can be something like: "No files to configure provided"...

PR #214, opened 2016-10-18
Vagrant: Fix Pandas installation

bjackman:
**Don't merge yet**, I haven't tested this on a clean box, will do tonight. Just putting this here in case anyone has hit the same issue.


bjackman:
OK, seems to work. Ready to merge if it looks good to you.

PR #220, opened 2016-11-02
libs/utils/trace: Fix getCPUActive signal for fully-idle CPUs

bjackman:
If no cpu_idle events are reported for a particular CPU (but there are
cpu_idle events for other CPUs), assume that the CPU was never used
during the period of trace event collection.


mdigiorgio:
Haven't tested it, but LGTM :+1:

PR #221, opened 2016-11-02
Allow test classes to be loaded from Python code

bjackman:
- Add the project root to PYTHONPATH so that `tests` can be imported
  N.B. this will prevent importing the existing `tests` module from
  IPython.
- Add \_\_init__.py to tests/ directories so they are recognised
  as packages


derkling:
Maybe better to add this patch to the first notebook which shows I to use tests from IPython... otherwise it's also difficult to test the patch ;-)

derkling:
@bjackman does it make sens to stall this waiting for you to add an example notebook? Or this patch is required by some other more recent PRs?... in the last case I'm ok also to merge it as it is... while if we can wait for an example Notebook I would really prefer to wait.

bjackman:
Yep, let's wait for an example notebook. I have been planning to submit one along with the new tests.

PR #222, opened 2016-11-02
Miscellaneous new features for LisaTest

bjackman:
- Require root connection on Android targets
- Add `get_sched_assert`. Currently no users for this - should we merge it anyway (since I've written it) or wait until we need it?
- Add `get_end_time` and `get_window`. Also no users yet.


derkling:
Ok, AFAIR the /data/local/tmp is W+A folder, isn't it?
In that case I wondering why you get a permission error while trying to push on a subfolder.
Maybe we forecely create this path as root while it's not really required?

bjackman:
Pushed a version without the root permissions thing

Issue #224, opened 2016-11-03
LISA energy model notebook fails to execute in Juno R0 board 


cloverpop:
KeyErrorTraceback (most recent call last)
<ipython-input-15-9b1ffaa2f234> in <module>()
      2 loop_cnt = 5
      3 
----> 4 power_perf_df = compute_power_perf(clusters, loop_cnt, sysbench)

<ipython-input-14-0543b4f623b4> in compute_power_perf(clusters, loop_cnt, benchmark, bkp_file)
     77                         # Run benchmark into the target cgroup
     78                         perf += benchmark.run(target_cg.name, cnt + 1)
---> 79                         nrg = te.emeter.report(te.res_dir).channels
     80                         energy += nrg[cl.emeter_ch]
     81                         sleep(10)

/data/lisa_clean/lisa/libs/utils/energy.py in report(self, out_dir, out_file)
    204         for channel in self._channels:
    205             label = self._channels[channel]
--> 206             nrg_total = nrg[label]['total']
    207             logging.debug('%14s - Energy [%16s]: %.6f', 'HWMon', label, nrg_total)
    208             clusters_nrg[channel] = nrg_total

KeyError: 'a57'

cloverpop:
This issue can be assigned to me, please.

bjackman:
"juno" and "juno2" are actually to do with kernel versions i.e. the hwmon setup. The way these notebooks are intended to be used is that each user modifies some sections of it so that it works on their board - it's more of a guideline/example than a generic automated tool. `my_conf` is an example of this - we expect that each user will have to change the hwmon bits in my_conf.

This stuff isn't very user friendly at all, we definitely have some work to do on documentation. There's also a related issue here: https://github.com/ARM-software/lisa/issues/11 

But ultimately this is an issue with documentation and clarity, not with the notebook being broken, so I'm going to close this issue.

PR #225, opened 2016-11-03
issue #224: LISA energy model notebook fails to run in Juno

cloverpop:
solution: fix hwmon settings for Juno r0 board


mdigiorgio:
I don't think this patch is needed. The solution should be to simply use `juno2` instead of `juno` as `board` in the `my_conf` dictionary.

cloverpop:
The problem is that there are two kinds of Juno boards: Juno R0 and Juno R2.
In my_conf,  "juno" is for Juno R0 board ( 4 x A53 + 2 x A57 ) and "juno2" is for Juno R2 board ( 4 x A53 + 2 x A72 ).
For "juno2" settings, its configuration is correct. However, for "juno", its configuration is wrong.
You can see the whole notebook file for that.

ipynb/energy/EnergyModel_ClusterEnergy.ipynb
```diff
@@ -106,12 +106,12 @@
     "    \"emeter\" : {\n",
     "        'instrument' : 'hwmon',\n",
     "        'conf' : {\n",
-    "            'sites' : [ 'a53', 'a57' ],\n",
+    "            'sites' : [ 'BOARDLITTLE', 'BOARDBIG' ],\n",
```
derkling: This is the same sites naming used by HWMON on JunoR2 boards... is that a JunoR0 or a JunoR2 board?
Because, as @mdigiorgio previously commented, if you are using a JunoR2 board you should better set
```
"board" = "juno2" 
```
in the previous line 92, at that point you can avoid the "emeter" section here because is automatically specified for each supported platform.

cloverpop: I use Juno R0 board and Linux kernel 3.18 ( Linaro deliverables for Juno 16.06 or 16.09 )

I ever thought "juno" is for Juno R0 board and "juno2" is for Juno R2 board.
If it is not that case, we can use "juno2" in my_conf to bypass this issue. 
But it may cause some confusion for other customers if they have same misunderstanding.

derkling: So I guess the issue is that on JunoR0 boards, running a v3.18 LSK kernel, HWMON channels are named BOARDLITTLE and BOARDBIG... which means that some patches have been backported from mainline. Can you please check this? If that's the case we should actually get this patch in. ;-)

bjackman: I don't know how hwmon works, but I'm using a Juno r0 with `lsk-3.18-armlt-20161028` and the hwmon setup works with `"board" : 'juno'` and `'sites' : [ 'a53', 'a57' ]`

derkling: Thanks @bjackman for testing on R0! :+1: 
@cloverpop the real issue is that in most recent kernel there is not more need of the  juno/juno2 distinction, can you refresh this PR to be a patch which removes that distinction?

bjackman:
Fixed by https://github.com/ARM-software/lisa/pull/241

PR #219, opened 2016-11-01
Notebook restructure

ionela-voinescu:
The changes here represent initial work related to https://github.com/ARM-software/lisa/issues/211.
The first of the patches moves the example notebooks in one location and the second one adds more details to the testenv_example notebook, work which will also be done for other example notebooks in future commits.

PR #227, opened 2016-11-07
executor: add flag to freeze userspace while running experiments

bjackman:
Note that if you manually stopped services before running the LISA tests, this will re-start them after running the workload. We could modify this to actually look at the services that are running before we run `stop` to determine if we should re-`start`, but I don't want to add that complexity unless it's definitely necessary.

### DIFF
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -40,6 +40,18 @@ Experiment = namedtuple('Experiment', ['wload_name', 'wload',
                                        'conf', 'iteration', 'out_dir'])
 
 class Executor():
+    critical_tasks = {
+        'linux': ['init', 'systemd', 'sh', 'ssh'],
+        'android': [
+            'sh', 'adbd', 'init',
+            'usb', 'transport',
+            # We don't actually need this task but on Google Pixel it apparently
+            # cannot be frozen, so the cgroup state gets stuck in FREEZING if we
+            # try to freeze it.
+            'thermal-engine'
+        ]
+    }
+    """Tasks in the system that we can't afford to freeze"""
 
     def __init__(self, test_env, experiments_conf):
         """
@@ -532,6 +544,11 @@ class Executor():
         self._log.debug('out_dir set to [%s]', experiment.out_dir)
         os.system('mkdir -p ' + experiment.out_dir)
 
+        # Freeze all userspace tasks that we don't need for running tests
+        need_thaw = False
+        if self._target_conf_flag(tc, 'freeze_userspace'):
+            need_thaw = self._freeze_userspace()
+
         # FTRACE: start (if a configuration has been provided)
         if self.te.ftrace and self._target_conf_flag(tc, 'ftrace'):
             self._log.warning('FTrace events collection enabled')
@@ -564,7 +581,32 @@ class Executor():
             self._log.info('   %s',
                            stats_file.replace(self.te.res_dir, '<res_dir>'))
 
+        # Unfreeze the tasks we froze
+        if need_thaw:
+            self._thaw_userspace()
+
         self._print_footer()
+
+    def _freeze_userspace(self):
+        if 'cgroups' not in self.target.modules:
+            raise RuntimeError(
+                'Failed to freeze userspace. Ensure "cgroups" module is listed '
+                'among modules in target/test configuration')
+        controllers = [s.name for s in self.target.cgroups.list_subsystems()]
+        if 'freezer' not in controllers:
+            self._log.warning('No freezer cgroup controller on target. '
+                              'Not freezing userspace')
+            return False
+
+        exclude = self.critical_tasks[self.te.target.os]
+        self._log.info('Freezing all tasks except: %s', ','.join(exclude))
+        self.te.target.cgroups.freeze(exclude)
+        return True
+
+
+    def _thaw_userspace(self):
+        self._log.info('Un-freezing userspace tasks')
+        self.te.target.cgroups.freeze(thaw=True)
 
 ################################################################################
 # Utility Functions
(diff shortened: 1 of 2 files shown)


libs/utils/test.py
```diff
@@ -98,12 +98,18 @@ def _experimentsInit(cls):
         """
         Code executed before running the experiments
         """
+        if cls.target.os == "android":
+            # Stop all the system services that introduce noise
```
derkling: Maybe it's worth to log a warning about that?
Also, maybe we can add a target configuration feature to enable/disable this behaviour?
I'm thinking about the possibility to have, in the future, android based tests which requires the runtime to be up and running...

bjackman:
OK I pushed a new version, this version is in `Excecutor` instead of `LisaTest`. There is an option in the test conf, i.e. you can set `test_conf["confs"][i]["pause_android_services"] = False` to prevent pausing services for a particular test. That means you can easily run a workload both with and without services paused.

This is sorely lacking documentation.. but looking around I've realised that `LisaTest`/`Executor`/`TestEnv`/`target_conf`/`test_conf`(a.k.a `tests_conf`) could do with a big documentation effort of which https://github.com/ARM-software/lisa/pull/219 is the first part, so I'm going to leave that documentation out for the moment and come back to it.

libs/utils/executor.py
```diff
@@ -512,6 +512,12 @@ def _wload_run(self, exp_idx, experiment):
         logging.debug(r'%14s - out_dir [%s]', 'Executor', experiment.out_dir)
         os.system('mkdir -p ' + experiment.out_dir)
 
+        # ANDROID: Stop all the Android system services that introduce noise
+        pause_android_services = tc.get("pause_android_services", True)
```
derkling: Maybe it's worth to use one of the features flags instead of adding a new configuration option?
I mean something similar to what we do here:
https://github.com/ARM-software/lisa/blob/master/libs/utils/env.py#L853

bjackman: Do you mean a target conf flag? I don't think this belongs in the TestEnv target configuration - it should be configurable per-experiment (also TestEnv doesn't have an "end" where the change can be undone). Will try with a target configuration flag.

bjackman:
Pushed a new version that uses a target conf flag.. not sure if that's what you meant?

bjackman:
Pushed a new version that uses the cgroups freezer. We may want to wait until https://github.com/ARM-software/devlib/pull/76 is merged in devlib before merging this.

derkling:
That's a way better solution and much more portable :+1: 
Agree we should wait for ARM-software/devlib#76 as well as give another run of tests on different platforms... before merging it we should also *cross-check the energy-model building flow*

bjackman:
Dependencies (+ [this fix](https://github.com/ARM-software/devlib/pull/77)) are merged but having issues getting this to work on Google Pixel

bjackman:
Alrighty, rebased with the logging biz and included necessary excludes for Pixel to work. Github is showing the commits in the wrong order, log actually looks like:

```
498cb07 tests/eas/acceptance: Use freeze_userspace flag
4c5ba74 executor: Add flag to freeze userspace while running experiments
b6c6840 submodules: Update devlib
```

libs/utils/executor.py
```diff
@@ -551,8 +569,32 @@ def _wload_run(self, exp_idx, experiment):
...
+            raise RuntimeError(
+                "freeze_userspace: devlib cgroups module not loaded. "
+                "Ensure it's listed in your target/test modules")
+        controllers = [s.name for s in self.target.cgroups.list_subsystems()]
+        if 'freezer' not in controllers:
+            self._log.warning('freeze_userspace: no freezer cgroup controller')
+            return False
+
+        exclude = self.critical_tasks[self.te.target.os]
+        self._log.info('Freezing all tasks except %s', exclude)
```
derkling: Maybe better:
```python
self._log.info('Freezing all tasks except: %s', ','.join(exclude))
```

libs/utils/executor.py
```diff
@@ -519,6 +531,12 @@ def _wload_run(self, exp_idx, experiment):
         self._log.debug('out_dir set to [%s]', experiment.out_dir)
         os.system('mkdir -p ' + experiment.out_dir)
 
+        # Freeze all userspace tasks that we don't need for running tests
+        if self._target_conf_flag(tc, "freeze_userspace"):
```
derkling: I would prefer (mainly cosmetics) to avid this dummy else branches... can we write this as:
```python
need_thaw = False
if self._target_conf_flag(tc, "freeze_userspace"):
    need_thaw = self._freeze_userspace()
```

libs/utils/executor.py
```diff
@@ -551,8 +569,32 @@ def _wload_run(self, exp_idx, experiment):
...
+        # Unfreeze the tasks we froze
+        if need_thaw:
+            self._thaw_userspace()
+
         self._print_footer()
 
+    def _freeze_userspace(self):
+        if 'cgroups' not in self.target.modules:
+            raise RuntimeError(
+                "freeze_userspace: devlib cgroups module not loaded. "
```
derkling: This message is not completely clear on which module we are missing. What about:
```python
raise RuntimeError(
    "Userspace freezing failed. Ensure the 'cgroups' module is listed "
    "among your target/test devlib's modules")
```

Notice that I would also avoid the "freeze_userspace, we recently removed all these kind-of namespace prefixes.

libs/utils/executor.py
bjackman: Sure

bjackman:
Rebased and addressed comments.

derkling:
Not sure if it's my github visualization... but it seems to me that two of the three comments (first and last) have not been addressed. Is that a push error or intentional... I'm ok to merge also as it is, the missing comments was not major concerns ;-)

PR #229, opened 2016-11-08
libs/utils/tests: Don't look at Executor config in LisaTest

bjackman:
We don't need to validate the config; Executor already does
that. Therefore we don't even need to parse it - just pass it blindly on
to Executor. That way we can use raw dicts instead of JSON files.

### DIFF
--- a/libs/utils/test.py
+++ b/libs/utils/test.py
@@ -24,14 +24,13 @@ from bart.sched.SchedMultiAssert import SchedMultiAssert
 from devlib.utils.misc import memoized
 import wrapt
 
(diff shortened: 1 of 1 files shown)


bjackman:
There are obviously other ways to allow us to configure LisaTest with raw dicts but I thought I'd first propose the one that results in deleting code rather than writing it :stuck_out_tongue:  open to suggestions though.

derkling:
Seems interesting... we are actually doing the same checks two times.
Lovely red coloured patch... let's merge it!

PR #230, opened 2016-11-09
tests/eas/acceptance: set sched_initial_test_util

bjackman:
This tunable is required to use the "sched" cpufreq governor (schedfreq)

Note that this isn't required right now - the acceptance tests use the performance governor - but we should expect people to want to change that.

### DIFF
--- a/tests/eas/acceptance.py
+++ b/tests/eas/acceptance.py
@@ -60,6 +60,12 @@ class EasTest(LisaTest):
             cls.target.write_value(
                 "/proc/sys/kernel/sched_is_big_little", 1, verify=False)
 
+        if SET_INITIAL_TASK_UTIL:
+            # This flag doesn't exist on all kernels, so don't worry if the file
+            # isn't present (hence verify=False)
+            cls.target.write_value(
+                "/proc/sys/kernel/sched_initial_task_util", 1024, verify=False)
+
     def _do_test_first_cpu(self, experiment, tasks):
         """Test that all tasks start on a big CPU"""
 
--- a/tests/eas/acceptance.py
+++ b/tests/eas/acceptance.py
@@ -55,13 +55,10 @@ class EasTest(LisaTest):
         super(EasTest, cls)._experimentsInit(*args, **kwargs)
 
         if SET_IS_BIG_LITTLE:
-            try:
-                cls.target.write_value(
-                    "/proc/sys/kernel/sched_is_big_little", 1)
-            except TargetError:
-                # That flag doesn't exist on mainline-integration kernels, so
-                # don't worry if the file isn't present.
-                pass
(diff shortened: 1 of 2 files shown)


tests/eas/acceptance.py
```diff
@@ -54,15 +54,20 @@ def setUpClass(cls, *args, **kwargs):
...
+        # Try writing a value but do not report an error if it fails; these
+        # flags do not exist on all kernels.
+        def write_maybe(path, value):
             try:
-                cls.target.write_value(
-                    "/proc/sys/kernel/sched_is_big_little", 1)
-            except TargetError:
-                # That flag doesn't exist on mainline-integration kernels, so
-                # don't worry if the file isn't present.
+                cls.target.write_value(path, value)
```
derkling: Isn't just enough to use
```python
cls.target.write_value(path, value, verify=False)
```

Both target.write_value and target.executes provides verification flags which should silently ignore errors in case the specified path does not exist or is not writable (e.g. because of permissions).

tests/eas/acceptance.py
```diff
@@ -54,15 +54,20 @@ def setUpClass(cls, *args, **kwargs):
...
+            except TargetError as e:
                 pass
 
+        if SET_INITIAL_TASK_UTIL:
+            write_maybe("/proc/sys/kernel/sched_initial_task_util", 1024)
```
derkling: Given the previous comment we can probably avoid to add this write_maybe function...

tests/eas/acceptance.py
```diff
@@ -54,15 +54,20 @@ def setUpClass(cls, *args, **kwargs):
...
+
+        if SET_IS_BIG_LITTLE:
```
derkling: This option has not been added by this patch, maybe an error on chunks selection?

derkling:
Moreover: is this not something which is overlapping with the #209 PR by @credp... maybe we should merge this as well...

tests/eas/acceptance.py
bjackman: Agreed

tests/eas/acceptance.py
bjackman: This `if` was already there, it just got reformatted :)

bjackman:
Pushed a new version addressing your review comments.

Re: The other PR - good point, I had forgotten about that! However I think this is still useful - I am now leaning towards making things explicit in Python (so you can see a call to `write_value`) rather than adding more stuff to our sparsely-documented JSON format. Do you and @credp agree with that suggestion?

derkling:
Ok, let's merge this and still pause the #209 since we are heading toward a more programmatic definition of tests. @credp, I'll keep #209 still open just to remember that we need these functions as well ;-)

PR #231, opened 2016-11-10
Fixes for Executor

bjackman:
I looked [here](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Security-Enhanced_Linux/sect-Security-Enhanced_Linux-Working_with_SELinux-Mounting_File_Systems.html) to try and figure out if there's a mount option that could let us fix this properly. It sounds like adding `context=u:object_r:shell_data_file:s0` to the mount options should fix the issue but it gives me:

```
'schedtest' is read-only
mount: 'schedtest'->'/data/local/tmp/devlib-target/run_dir': Permission denied
```

I decided to be pragmatic about it - rather than reading great dusty tomes of SELinux documentation to figure out incantations I suggest we just put it in permissive mode.

### DIFF
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -139,6 +139,7 @@ class Executor():
         self.experiments = []
 
         # Run all the configured experiments
+        exp_idx = 0
         for tc in self._tests_conf['confs']:
             # TARGET: configuration
             if not self._target_configure(tc):
@@ -155,10 +156,9 @@ class Executor():
                         out_dir=os.path.join(test_dir, str(itr_idx)))
                     self.experiments.append(exp)
 
-            # WORKLOAD: execution
-            for exp_idx, experiment in enumerate(self.experiments):
-                self._wload_run(exp_idx, experiment)
-
+                    # WORKLOAD: execution
+                    self._wload_run(exp_idx, exp)
+                    exp_idx += 1
             self._target_cleanup(tc)
 
         self._print_section('Executor', 'Experiments execution completed')
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -42,6 +42,8 @@ from conf import JsonConf
 
 import wlgen
 
+from devlib import TargetError
+
 Experiment = namedtuple('Experiment', ['wload_name', 'wload',
                                        'conf', 'iteration', 'out_dir'])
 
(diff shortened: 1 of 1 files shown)


libs/utils/executor.py
```diff
@@ -212,12 +212,17 @@ def _setup_rootfs(self, tc):
...
                 'TargetSetup', self.te.run_dir)
         self.target.execute('[ -d {0} ] || mkdir {0}'\
-                .format(self.te.run_dir), as_root=True)
+                .format(self.te.run_dir))
         self.target.execute(
                 'grep schedtest /proc/mounts || '\
                 '  mount -t tmpfs -o size=1024m {} {}'\
                 .format('schedtest', self.te.run_dir),
                 as_root=True)
+        # tmpfs mounts have a context with "tmpfs" as the type (while other
```
derkling: This concept of "context" is not really clear to me, is it referring to SELinux?
Can you please try to rephrase the issue to be clear to SELinux-idiots :-)

libs/utils/executor.py
```diff
@@ -212,12 +212,17 @@ def _setup_rootfs(self, tc):
...
+        # files we create have "shell_data_file"). That prevents non-root
+        # users from creating files in tmpfs mounts. For now, just put SELinux
+        # in permissive mode to get around that.
+        self.target.execute('setenforce 0', as_root=True, check_exit_code=False)
```
derkling: Maybe worth to dump a warning to notify the user about this setting?

Moreover, did you considered the possibility to set "permissive mode" only across the execution of a test?
Here we are in the executor, thus maybe it's possible to enable this mode only before the setup of a new test and than reset it once the test complete.

libs/utils/executor.py
bjackman: Good point - I was obviously in a narrow SELinux-focussed mindset when I wrote this :tongue: .
Also, copying this comment to the commit message but also leaving it here.

libs/utils/executor.py
bjackman: Sounds good, there's a `getenforce` command so should be easy to avoid clobbering other settings too.

bjackman:
Pushed a new version with
- Additional fix for multiple confs bug I introduced previously
- Better commit message
- Logging
- Restores old SELinux mode after finishing workload execution
More testing needed, will report back once that's done.

bjackman:
OK, ran some tests and pushed a new version.

derkling:
Ok, I'll merge this and than refresh #237 since here we add some more logging statements.
Than we should close that PR as well... otherwise we risk to keep refreshing it each time we merge something new. ;-)

PR #233, opened 2016-11-11
Trace: Automatically populate the task/PID map when sched_switch enabled

bjackman:
I'm not really sure what the original intention for these methods in `Trace` were so I might be off the mark here, but this seems to basically be the behaviour I'd expect from trace. That is I would expect to be able to have a trace parsed and then simply call `getTasks` to see which tasks were found.

This should make it more convenient for tests to figure out PIDs of tasks run during workload experiments so that they can do analysis. 

For example: https://github.com/ARM-software/lisa/blob/2f1eef3bdbd539380684472eb2e02b58584551cd/tests/eas/load_tracking.py#L111

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -295,17 +295,17 @@ class Trace(object):
         :param tasks: list of task names
         :type tasks: list(str)
         """
+        def load(tasks, event, name_key, pid_key):
+            df = self._dfg_trace_event(event)
+            self.getTasks(df, tasks, name_key=name_key, pid_key=pid_key)
+            self._scanTasks(df, name_key=name_key, pid_key=pid_key)
+
         if 'sched_switch' in self.available_events:
-            self.getTasks(self._dfg_trace_event('sched_switch'), tasks,
-                          name_key='next_comm', pid_key='next_pid')
-            self._scanTasks(self._dfg_trace_event('sched_switch'),
-                            name_key='next_comm', pid_key='next_pid')
-            return
-        if 'sched_load_avg_task' in self.available_events:
-            self.getTasks(self._dfg_trace_event('sched_load_avg_task'), tasks)
-            self._scanTasks(self._dfg_trace_event('sched_load_avg_task'))
-            return
-        logging.warning('Failed to load tasks names from trace events')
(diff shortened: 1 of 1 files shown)


libs/utils/trace.py
```diff
@@ -148,6 +148,8 @@ def __init__(self, platform, data_dir, events,
 
         self.analysis = AnalysisRegister(self)
 
+        self._populate_tasks()
+
```
bjackman: Looks like this should already be done by __loadTaskNames. Todo.

libs/utils/trace.py
```diff
@@ -429,11 +429,11 @@ def getTasks(self, dataframe=None,
         :param pid_key: The name of the dataframe columns containing task PIDs
         :type pid_key: str
         """
-        if dataframe is None:
-            return self.tasks
-        df = dataframe
         if task_names is None:
             task_names = self.tasks.keys()
+        if dataframe is None:
+            return {k: v for k, v in  self.tasks.iteritems() if k in task_names}
```
derkling: We should fix the docstring for task_names to say "(by default all tasks in the trace)"

libs/utils/trace.py
```diff
@@ -292,20 +293,22 @@ def __loadTasksNames(self, tasks):
...
         Try to load tasks names using one of the supported events.
 
-        :param tasks: list of task names
-        :type tasks: list(str)
+        :param tasks: list of task names. If None, load all tasks found.
+        :type tasks: list(str) or NoneType
         """
+        def load(tasks, event, name_key, pid_key):
+            df = self._dfg_trace_event(event)
+            if tasks is None:
```
derkling: Can we split this patch into a refactoing one which adds the load method and a "Populate all task names when no tasks passed" which adds this "if task is None:" bits?

bjackman: Sounds good, will do

libs/utils/trace.py
bjackman: Done in the previous commit - will move that change to here.

bjackman:
OK, pushed a new version. BTW it seems like GitHub shows the commits in order of author date instead of ancestry order :expressionless:

PR #234, opened 2016-11-11
idle_analysis: Fix idle residency when idle events are sparse

bjackman:
Not too sure if this is the best approach.. maybe we should come up with a more generic solution (`Trace._sanitizeCpuIdle`?). I haven't investigated whether `_dfg_cluster_idle_state_residency` is affected by the same issue (in my brief tests it didn't seem to be but good chance that's coincidence).

This is kind of a similar issue to this: https://github.com/ARM-software/lisa/pull/220

### DIFF
--- a/libs/utils/analysis/idle_analysis.py
+++ b/libs/utils/analysis/idle_analysis.py
@@ -76,6 +76,12 @@ class IdleAnalysis(AnalysisModule):
         cpu_idle = cpu_idle.join(cpu_is_idle.to_frame(name='is_idle'),
                                  how='outer')
         cpu_idle.fillna(method='ffill', inplace=True)
+
+        # Extend the last cpu_idle event to the end of the time window under
+        # consideration
+        final_entry = pd.DataFrame([cpu_idle.iloc[-1]], index=[self._trace.x_max])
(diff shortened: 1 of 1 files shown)


bjackman:
@mdigiorgio not demanding review, just pinging you because you commented on the related PR :)

mdigiorgio:
Yes, I wanted to ask you if you have an example trace with which I can test this...

One thing you might consider is to use `append()` instead of `concat()` since you are inserting a row at the end of the dataframe, I'm not sure if it's more efficient (we could test it with `%timeit`) but it looks like:

```python
final_entry = pd.Dataframe([cpu_idle.iloc[-1]], index=[self._trace.x_max])
cpu_idle = cpu_idle.append(final_entry)
```

bjackman:
Here's a trace and a notebook to illustrate the issue (check the `sum` at the bottom)

bjackman:
Re `append` vs `concat` - timed it and difference is in the noise. Switching for `append` though, it makes more sense, cheers.

mdigiorgio:
OK, LGTM :+1: thanks!

mdigiorgio:
Let's merge this! Thanks for the fix.

PR #235, opened 2016-11-15
[RFC] Preliminary tests to see if a platform is suitable for EAS

bjackman:
Taken from @credp w/ some cleanups.
Posting this as a repository for discussion of these tests. 
TODOs:
- Choose a descriptive name for this suite. "preliminary"?
- Some more coding style cleanups (CamelCase vs snake_case etc)
- @derkling to look at converting the SchedTune test to use devlib's cgroups module
- @credp to add some docstrings to explain what is being tested


bjackman:
Made some more changes - will squash this eventually but leaving it as separate commits so people can comment if desired.

tests/eas/basic_check.py
```diff
@@ -0,0 +1,306 @@
...
+                    '{}{}/{}'.format(mount_location, group, name))
+        self.assertTrue(schedtune['groups']['root']['schedtune.boost'] == 0,
+                        "SchedTune Root Group Boost != 0")
+        groups_with_nonzero_boost = 0
+        for groups in schedtune['groups']:
+            if groups == 'root':
+                continue
+            if schedtune['groups'][groups]['schedtune.boost'] != 0:
+                groups_with_nonzero_boost += 1
+        self.assertTrue(groups_with_nonzero_boost != 0,
```
bjackman: Don't think unittest has a warning facility, so I suggest we just keep the test and add some wordiness to the docstring to make it clear that this is not a problem with the kernel or platform but just some configuration that hasn't been done. Maybe eventually there could be separate suites here: one that says "your platform and kernel have the required features" and one that says "you've done the configuration you probably need to do"?

derkling:
On 16-Nov 02:18, Brendan Jackman wrote:
> bjackman commented on this pull request.
> 
> 
> 
> > +        # now read the values
> +        files = [ 'schedtune.boost' ]
> +        for group in paths:
> +            schedtune['groups'][paths[group]] = {}
> +            for name in files:
> +                schedtune['groups'][paths[group]][name] = self.target.read_int(
> +                    '{}{}/{}'.format(mount_location, group, name))
> +        self.assertTrue(schedtune['groups']['root']['schedtune.boost'] == 0,
> +                        "SchedTune Root Group Boost != 0")
> +        groups_with_nonzero_boost = 0
> +        for groups in schedtune['groups']:
> +            if groups == 'root':
> +                continue
> +            if schedtune['groups'][groups]['schedtune.boost'] != 0:
> +                groups_with_nonzero_boost += 1
> +        self.assertTrue(groups_with_nonzero_boost != 0,
> 
> Don't think unittest has a warning facility, so I suggest we just
> keep the test and add some wordiness to the docstring to make it
> clear that this is not a problem with the kernel or platform but
> just some configuration that hasn't been done. Maybe eventually
> there could be separate suites here: one that says "your platform
> and kernel have the required features" and one that says "you've
> done the configuration you probably need to do"?

Right, perhaps the configuration tests can be Android version
specific. Thus, let's keep this "preliminary" test as simple as
possible, i.e. it should check for knobs being available and usable
but do not verify any specific value.

-- 
#include <best/regards.h>

Patrick Bellasi

PR #236, opened 2016-11-16
platforms: Add JSON config for Google Pixel

bjackman:
Source of info: arch/arm/boot/dts/qcom/msm8996.dtsi in Android MSM
kernel, tag android-7.1.0_r0.3 (76cdcfd5b0c3)

https://android.googlesource.com/kernel/msm.git/+/android-7.1.0_r0.3/arch/arm/boot/dts/qcom/msm8996.dtsi#157

### DIFF
--- a/libs/utils/platforms/pixel.json
+++ b/libs/utils/platforms/pixel.json
@@ -1,0 +1,29 @@
+{
+    "board" : {
+        "cores" : [
+            "Silver", "Silver", "Gold", "Gold"
+        ],
+        "big_core" : "Gold",
+        "modules" : ["bl", "cpufreq"]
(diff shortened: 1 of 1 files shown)


libs/utils/platforms/pixel.json
```diff
@@ -0,0 +1,29 @@
+{
+    "board" : {
+        "cores" : [
+            "Sliver", "Sliver", "Gold", "Gold"
```
mdigiorgio: nit: s/Sliver/Silver/g

bjackman: Good catch, thanks. Amended.

PR #238, opened 2016-11-17
Executor & LisaTest interface change

bjackman:
The end goal of this PR is to be able to write LisaTest subclasses that generate workloads in a way that depends on target information. It contains:

"Actual" code changes..
- Modifications to Executor to enable this
- The key commit which adds `_getTestConf` and `_getExperimentsConf` to `LisaTest`
And changes that illustrate how the above should be used
- Updates to examples in notebooks
- Updates to acceptance tests to follow the style suggested by the changes made here

I haven't "tagged" it but this PR is probably an RFC.


libs/utils/test.py
```diff
@@ -42,20 +42,31 @@ def _init(cls, conf, *args, **kwargs):
...
         Default experiments execution engine
         """
 
         cls.logger.info("%14s - Setup tests execution engine...", "LisaTest")
-        test_env = TestEnv()
+        test_env = TestEnv(test_conf=cls._getTestConf())
 
-        cls.executor = Executor(test_env, experiments_conf=cls.conf);
+        exp_conf = cls._getExperimentsConf(test_env)
+        cls.executor = Executor(test_env, experiments_conf=exp_conf)
```
derkling: Why not s/exp_conf/experiments_conf/ and pass it straight to the Executor?

```python
experiments_conf = cls._getExperimentsConf(test_env)
cls.executor = Executor(test_env, experiments_conf)
```

Thus avoiding to mix named and positional parameters in Executor ctor which also gets only these two params.

libs/utils/executor.py
```diff
@@ -44,7 +44,7 @@
 
 class Executor():
 
-    def __init__(self, test_env, tests_conf=None):
+    def __init__(self, test_env, experiments_conf=None):
```
derkling: Should not be mandatory?
Don't remember anymore why we used optional values at the beginning but they seems not being more an option... form the checks on lines 80:91 we do not allow a None value, isn't it?

ipynb/utils/executor_example.ipynb
```diff
@@ -68,10 +69,41 @@
...
+      "05:06:17  INFO    :        TestEnv -    /home/bjackman/sources/lisa/results/ExecutorExample\n",
+      "05:06:17  INFO    :        TestEnv - Experiment results available also in:\n",
+      "05:06:17  INFO    :        TestEnv -    /home/bjackman/sources/lisa/results_latest\n"
+     ]
+    }
+   ],
    "source": [
     "# Setup a target configuration\n",
-    "my_target_conf = {\n",
+    "env = TestEnv({\n",
```
derkling: Interesting... we should probably use this inline version every time we do not need to access the configuration from somewhere else code in the notebook...

ipynb/utils/executor_example.ipynb
```diff
@@ -203,50 +233,29 @@
...
+      "05:06:17  INFO    :       Executor -       rta, perf\n",
+      "05:06:17  INFO    :       Executor - Total: 4 experiments\n",
+      "05:06:17  INFO    :       Executor - Results will be collected under:\n",
+      "05:06:17  INFO    :       Executor -       /home/bjackman/sources/lisa/results/ExecutorExample\n"
      ]
     }
    ],
    "source": [
-    "executor = Executor(my_target_conf, my_tests_conf)"
+    "executor = Executor(env, my_tests_conf)"
```
derkling: Should not for consistency be s/my_tests_conf/my_experiments_conf/?

libs/utils/executor.py
```diff
@@ -70,6 +70,10 @@ def __init__(self, test_env, experiments_conf=None):
         - <wload>   : the identified of one of the specified workload
         - <run_id>  : the progressive execution number from 1 up to the
                       specified iterations
+
+        After the workloads have been run, the Executor object's `experiments`
```
derkling: Cannot this chunk be squashed in the patch which introduces this list of Experiments?

libs/utils/test.py
```diff
@@ -42,20 +42,31 @@ def _init(cls, conf, *args, **kwargs):
             cls.logger.setLevel(kwargs['loglevel'])
             kwargs.pop('loglevel')
 
-        cls.conf = conf
-
         cls._runExperiments()
 
     @classmethod
+    def _getTestConf(cls):
+        if not hasattr(cls, "test_conf"):
```
derkling: Here is not clear that the "test_conf" is expected to be defined by one of the derived classes.

Why not adding a NoneType initialized `cls.test_conf` attribute and check here that the user has overridden its definition to be a dictionary?

Dose it make sense to use this method also to provide a sanity check of the provided configuration dictionary?

libs/utils/test.py
```diff
@@ -42,20 +42,31 @@ def _init(cls, conf, *args, **kwargs):
...
+            raise NotImplementedError("Please add `test_conf` attribute")
+        return cls.test_conf
+
+    @classmethod
+    def _getExperimentsConf(cls, test_env):
```
derkling: It is not clear why we expect test_env or what a derived class can use it for.

We should probably add a bit of documentation at least to this two methods.

libs/utils/test.py
```diff
@@ -42,20 +42,31 @@ def _init(cls, conf, *args, **kwargs):
...
+
+    @classmethod
     def _runExperiments(cls):
         """
         Default experiments execution engine
         """
 
         cls.logger.info("%14s - Setup tests execution engine...", "LisaTest")
-        test_env = TestEnv()
+        test_env = TestEnv(test_conf=cls._getTestConf())
```
derkling: Here I think we can avoid named parameters...

tests/eas/acceptance.py
```diff
@@ -28,15 +28,15 @@
...
-# Read the config file and update the globals
-CONF_FILE = os.path.join(
-    os.path.dirname(
-        os.path.abspath(__file__)),
-    "acceptance.config")
-
-with open(CONF_FILE, "r") as fh:
-    conf_vars = json.load(fh)
-    globals().update(conf_vars)
+WORKLOAD_DURATION_S = 5
```
derkling: Let's add at least a single line of comment to mark that this are global test configuration parameters...

tests/eas/acceptance.py
```diff
@@ -160,7 +160,24 @@ class SmallTaskPacking(EasTest):
     All tasks run on little cpus.
     """
 
-    conf_basename = "acceptance_small_task_packing.config"
```
derkling: I cannot see the removal of the acceptance_small_task_packing.config file in this patch.

tests/eas/acceptance_small_task_packing.config
```diff
@@ -1,42 +0,0 @@
-{
```
derkling: This chunk belongs to this path:
tests/eas/acceptance: SmallTaskPacking: Use Python dict not JSON

tests/eas/acceptance.py
```diff
@@ -28,15 +28,15 @@
 from env import TestEnv
 from test import LisaTest, experiment_test
 
-# Read the config file and update the globals
```
derkling: Let's keep this patch before the previous one... so that we have all the tests re-factoring one after the other...

tests/eas/acceptance_wake_migration.config
```diff
@@ -1,48 +0,0 @@
...
-        // Create one small task for each CPU
-        "wake_migration" : {
-            "type" : "rt-app",
-            "conf" : {
-                "class" : "profile",
-                "params" : {
-                    "wmig" : {
-                        "kind" : "Step",
-                        "params" : {
-                            "start_pct": 10, // STEP_LOW_DCYCLE
```
derkling: Why we discarded the comments? Should these be translated into global variables for consistency?

libs/utils/test.py
bjackman: Agreed, that would be clearer. I don't think we should put a sanity check in this method since subclasses might override this method without calling `super..._getTestConf`.

bjackman: First point: agreed, that would be clearer.

Second point: I don't think we should put a sanity check in this method since subclasses might override this method without calling super..._getTestConf.

bjackman: First point: agreed, that would be clearer. However let's not check for it being a dictionary, let's just check it against `None` since it could in theory be a string (file path) or anything that implements `__getitem__`.

Second point: I don't think we should put a sanity check in this method since subclasses might override this method without calling super..._getTestConf.

libs/utils/test.py
bjackman: Sounds good.

libs/utils/executor.py
bjackman: That patch is already in master

ipynb/utils/executor_example.ipynb
bjackman: Yeah, it's probably clearer.

ipynb/utils/executor_example.ipynb
bjackman: Yes. But I'm going to put the dict inline here as you suggested above.

tests/eas/acceptance.py
bjackman: Good spot, will fix.

tests/eas/acceptance.py
bjackman: I think this is Github being silly, my git log looks like this:

```
0600010 * lisa-test-interface-change tests/eas/acceptance: WakeMigration: Use Python dict not JSON
688e012 * tests/eas/acceptance: Offload+IdlePull: Use Python dict not JSON
1fa29f4 * tests/eas/acceptance: SmallTaskPacking: Use Python dict not JSON
675b792 * tests/eas/acceptance: ForkMigration: use Python dict not JSON
3320677 * tests/eas/acceptance: Remove JSON config
ea0b921 * LisaTest: Flesh out docstrings
6fd0f93 * LisaTest: Add methods to override for test configuration
4d62d29 * ipynb/tutorial: Update Executor tutorial notebook
051a6e6 * ipynb/utils: Update Executor example notebook
1a44422 * executor: Extend docstring
394991f * executor: Rename tests_conf to experiments_conf
acd8c57 * executor: Take a TestEnv instead of a target_conf
```

tests/eas/acceptance.py
bjackman: Added

tests/eas/acceptance_small_task_packing.config
bjackman: Fixed

tests/eas/acceptance_wake_migration.config
bjackman: Good point - I'll make them global variables

bjackman:
OK, pushed a new version.

libs/utils/test.py
bjackman: Sounds good.

libs/utils/test.py
bjackman: Nope, we need the named parameter because we're using the default for `target_conf`.

bjackman:
And another version addressing the comments I missed, also updating an additional notebook (kernel_function_profiling.ipynb)

PR #239, opened 2016-11-19
libs/utils/trace: avoid interleaved cpu_frequency events between devl…

mdigiorgio:
devlib injects fake cpu_frequency events at the beginning of a trace to get
information about the frequency. However, in some cases those events get mixed
with real cpu_frequency trace events generated by the cpufreq governor. This
leads to getting a dataframe with with wrong events and also makes the cluster
frequency coherency checks fail, whereas they should not.

For this reason, LISA checks whether the fake events for all CPUs happen before
the first event generated by the cpufreq governor and if this is the case they
are inserted as first events in the cpu_frequency dataframe.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -202,6 +202,9 @@ class Trace(object):
             self.events = events
         else:
             raise ValueError('Events must be a string or a list of strings')
+        # Register devlib fake cpu_frequency events
+        if 'cpu_frequency' in events:
+            self.events.append('cpu_frequency_devlib')
 
     def __parseTrace(self, path, tasks, window, normalize_time, trace_format):
         """
@@ -670,10 +673,38 @@ class Trace(object):
         Verify that all platform reported clusters are frequency coherent (i.e.
         frequency scaling is performed at a cluster level).
         """
-        if not self.hasEvents('cpu_frequency'):
+        if not self.hasEvents('cpu_frequency_devlib'):
             return
+
+        devlib_freq = self._dfg_trace_event('cpu_frequency_devlib')
+        devlib_freq.rename(columns={'cpu_id':'cpu'}, inplace=True)
+        devlib_freq.rename(columns={'state':'frequency'}, inplace=True)
+
         df = self._dfg_trace_event('cpu_frequency')
         clusters = self.platform['clusters']
+
+        # devlib always introduces fake cpu_frequency events, in case the
+        # OS has not generated cpu_frequency envets there are the only
+        # frequency events to report
+        if len(df) == 0:
+            # Register devlib injected events as 'cpu_frequency' events
+            setattr(self.ftrace.cpu_frequency, 'data_frame', devlib_freq)
+            df = devlib_freq
+            self.available_events.append('cpu_frequency')
+
+        # make sure fake cpu_frequency events are never interleaved with
+        # OS generated events
+        else:
+            if len(devlib_freq) > 0:
+                first_devlib_freq = devlib_freq.iloc[:self.platform['cpus_count']]
+                if df.index[0] > first_devlib_freq.index[-1]:
+                    df = pd.concat([first_devlib_freq, df])
+                last_devlib_freq = devlib_freq.iloc[self.platform['cpus_count']:]
+                if df.index[-1] < last_devlib_freq.index[0]:
+                    df = pd.concat([df, last_devlib_freq])
+            setattr(self.ftrace.cpu_frequency, 'data_frame', df)
+
+        # Frequency Coherency Check
         for _, cpus in clusters.iteritems():
             cluster_df = df[df.cpu.isin(cpus)]
             for chunk in self._chunker(cluster_df, len(cpus)):


mdigiorgio:
If this gets merged I will close my PR in devlib (https://github.com/ARM-software/devlib/pull/66)

bjackman:
I'm concerned about using the comm to detect the fake events. If (and this is a big "if") I understand correctly, real cpu_frequency events could, depending on driver/governor combo, be traced from process context, which means we could get real cpu_frequency events with `__comm = "shutils"`. 

A 3rd alternative: perhaps we could get trappy to expose the actual name of the event that gave rise to a DataFrame row; then we could check `df["event_name"].isin(["print", "bprint"])` or something?

derkling:
Not sure to get @bjackman's comment. All cpu_frequency events generated from kernel-space, which is what normally CPUFreq tracepoints does, do not have `__comm` set to "shutils". The only events which have this `__comm` value are the fake events we inject from userspace via the devlib's shutils library.

Thus, to me what @mdigiorgio proposes it seems a safe way to distinguish among user and kernel space generated events.

@bjackman  are you saying that there can be user-space triggered `cpu_frequency` events which are not generated via shutils?

bjackman:
Just trying to reiterate a F2F discussion: my concern is that if we used schedutil/schedfreq and a cpufreq driver with `CPUFREQ_DRIVER_FAST` then we may still have a race condition. In that case the event may be triggered from in the scheduler path in process context (not a workqueue or an interrupt). If the currently scheduled process is shutils (which is likely, since we're currently looking at clashes between the shutils invokation and frequency changes) then with this patch we will wrongly identify the event as fake.

I haven't observed this condition, it is very unlikely and perhaps impossible on ARM platforms we have access to (and maybe totally impossible if I've misunderstood the kernel), but if we're fixing race conditions let's fix them all no matter how unlikely.

mdigiorgio:
Rebased on top of master and updated the implementation according to ARM-software/devlib#66.

libs/utils/trace.py
```diff
@@ -202,6 +202,8 @@ def __registerTraceEvents(self, events):
             self.events = events
         else:
             raise ValueError('Events must be a string or a list of strings')
+        # Register devlib fake cpu_frequency events
+        self.events.append('cpu_frequency_devlib')
```
bjackman: I'd say only do this `if 'cpu_frequency' in events`

libs/utils/trace.py
```diff
@@ -674,6 +676,19 @@ def _sanitize_CpuFrequency(self):
             return
         df = self._dfg_trace_event('cpu_frequency')
         clusters = self.platform['clusters']
+
```
bjackman: Can you put a comment here explaining where the 'cpu_frequency_devlib' events come from (I know it's obviously devlib but could mention the `trace_frequencies` method) and one below explaining the interleaving problem (as in the commit message)?

mdigiorgio: Sure.

mdigiorgio:
Applied the requested changes and pushed a new version.

libs/utils/trace.py
```diff
@@ -674,6 +677,23 @@ def _sanitize_CpuFrequency(self):
...
+            devlib_freq.rename(columns={'cpu_id':'cpu'}, inplace=True)
+            devlib_freq.rename(columns={'state':'frequency'}, inplace=True)
+            # Only use fake events if introduced before any cpufreq governor
+            # did, otherwise they may end up being interleaved with real events
+            # and make the the frequency coherency check fail
+            if len(devlib_freq) > 0:
+                first_devlib_freq = devlib_freq.iloc[:self.platform['cpus_count']]
+                if df.index[0] > first_devlib_freq.index[-1]:
+                    df = pd.concat([first_devlib_freq, df])
+            setattr(self.ftrace.cpu_frequency, 'data_frame', df)
```
JaviMerino: In general I prefer this alterations to happen in trappy. Most of the `_sanitize_*` in `Trace()` should be moved to trappy.

However, in this particular case it's operating on stuff coming from devlib, so I'd rather leave it in `lisa`.

derkling: Well, most of the `_sanitize_*` methods are required to add custom columns and compute values to a TRAPpy table. Thus we are in the domain of mangling dataframes which is exactly what we do here.

The main issue for me to move certain bits into TRAPpy is that, sometimes we need platform data which are available only at the LISA side. Thus, in general I would avoid mixed approaches where we have somethings done in LISA others in TRAPpy.

IMHO, perhaps considering TRAPpy as a "raw" trace events parser while additional logic is added on top of LISA.

JaviMerino: Fair enough.  I agree on not moving stuff that needs platform information that is only available in LISA.  It's just that we have some very basic sanitizing in trappy (e.g., in `cpu_capacity` we change the column names, https://github.com/ARM-software/trappy/blob/master/trappy/sched.py#L83 ) and it looked like some of the sanitization done in LISA is similar to that.

mdigiorgio:
Thanks for the fix, I've just pushed a new version of the commit that has your changes.

PR #240, opened 2016-11-22
executor: Add freqs param to confs['cpufreq']

bjackman:
This is to be used for https://github.com/bjackman/lisa/blob/fie-test/tests/eas/load_tracking.py which I will submit as a separate PR once the dependencies are sorted.

I appreciate that there are some piles of documentation for the config dicts piling up here.. I will have to spend some time on it soon.

### DIFF
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -242,6 +242,14 @@ class Executor():
 
         self.target.cpufreq.set_all_governors(cpufreq['governor'])
 
+        if 'freqs' in cpufreq:
+            if cpufreq['governor'] != 'userspace':
+                raise ValueError('Must use userspace governor to set CPU freqs')
+            self._log.info(r'%14s - CPU frequencies: %s',
+                    'CPUFreq', str(cpufreq['freqs']))
+            for cpu, freq in cpufreq['freqs'].iteritems():
+                self.target.cpufreq.set_frequency(cpu, freq)
+
         if 'params' in cpufreq:
             self._log.info('governor params: %s', str(cpufreq['params']))
             for cpu in self.target.list_online_cpus():


libs/utils/executor.py
```diff
@@ -252,6 +252,14 @@ def _setup_cpufreq(self, tc):
 
         self.target.cpufreq.set_all_governors(cpufreq['governor'])
 
+        if 'freqs' in cpufreq:
+            if cpufreq['governor'] != 'userspace':
+                raise ValueError('Must use userspace governor to set CPU freqs')
+            logging.info(r'%14s - CPU frequencies: %s',
```
derkling: This seems not rebased on the most recent logging refactored code, isn't it?

libs/utils/executor.py
```diff
@@ -252,6 +252,14 @@ def _setup_cpufreq(self, tc):
...
+                    'CPUFreq', str(cpufreq['freqs']))
+            for cpu, freq in cpufreq['freqs'].iteritems():
```
derkling: Should not we add a check like:
```python
if len(cpufreq['freqs']) < len(self.target.list_online_cpus()):
    self._log.warning('Frequencies not defined for all CPUs')
```

Maybe the warning should be instead a raise ValueError?

bjackman: I think it's totally valid to not set frequencies for all CPUs - in fact that's how I've used it in my [use case](https://github.com/bjackman/lisa/blob/fie-test/tests/eas/load_tracking.py#L106)

bjackman:
Rebased and fixed logging

libs/utils/executor.py
derkling: Ok, I did not noticed `cpufreq['freqs']` was a map... then that makes much more sense! ;-)

PR #241, opened 2016-11-25
Remove juno/juno2 distinction

bjackman:
All Juno boards now have the same hwmon setup, so we don't need the
juno/juno2 distinction any more.

---
This supersedes https://github.com/ARM-software/lisa/pull/225 as it's quite urgent to get this fixed.
Note I haven't included backwards compatibility here, people who currently use 'juno2' will need to change their config. Any objections there?


bjackman:
Need to check if notebooks need updating

bjackman:
Nope, all is good.

cloverpop:
Per what you said "people who currently use 'juno2' will need to change their config", is there a method to make all Juno R0/R2 board users happy at the same time in LISA source code? 
You know some R0/R1/R2 boards were sold to users out of ARM. Complains from R2 board users may come back accordingly.

bjackman:
This should make all Juno revision users happy, it means that they can use "juno" for any revision/kernel config. Anyone who is *already* using the "juno2" option will just have to change their target.config to say "juno" instead.

I thought about treating "juno" and "juno2" as equivalent but decided to avoid introducing the first "legacy cruft" into LISA.

derkling:
Agree, let's keep it simple... however, this patch should fix at least update this notebook:
https://github.com/ARM-software/lisa/blob/master/ipynb/tutorial/04_ExecutorUsage.ipynb
to use "board" : 'juno'.

bjackman:
Ah yeah thanks for spotting that (it wasn't there in my branch, only saw it once I rebased on master). Fixed and pushed.

derkling:
Nice and neat clean-up! :+1:

PR #242, opened 2016-11-28
executor/env: Install rt-app automatically if required

bjackman:
With this commit, tests that use rt-app workloads need not explicitly
specify 'rt-app' in the 'tools' field of the TestEnv test configuration.

This adds an `install_tools` method to the TestEnv class.

---

This might be considered a bit of a hack? The generic solution would probably be to fix this in the workload class(es) by passing a TestEnv (rather than a Target) to their `__init__.py` that they can use to install the required binary.

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -99,6 +99,7 @@ class TestEnv(ShareState):
         self.ftrace = None
         self.workdir = WORKING_DIR_DEFAULT
         self.__tools = []
+        self.__installed_tools = set()
         self.__modules = []
         self.__connection_settings = None
         self._calib = None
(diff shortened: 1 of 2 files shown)


bjackman:
Rebased with logging biz

libs/utils/env.py
```diff
@@ -99,6 +99,7 @@ def __init__(self, target_conf=None, test_conf=None, wipe=True,
         self.ftrace = None
         self.workdir = WORKING_DIR_DEFAULT
         self.__tools = []
+        self.__installed_tools = set()
```
derkling: We need to push a clean-up patch which removes these horrible double underscores... will do after merging this patch ;-)

libs/utils/env.py
```diff
@@ -503,6 +497,25 @@ def _init_target(self, force = False):
...
+        tools_to_install = []
+        for tool in tools:
+            if tool in self.__installed_tools:
+                continue
+            binary = '{}/tools/scripts/{}'.format(basepath, tool)
+            if not os.path.isfile(binary):
+                binary = '{}/tools/{}/{}'\
+                         .format(basepath, self.target.abi, tool)
+                tools_to_install.append(binary)
+        self.target.setup(tools_to_install)
```
derkling: I've kind-of liked the usage of devlib's setup() tool, which is mandatory at least one time, but here...
This is an on-demand method... is there any risk on calling it multiple times on the same target?

derkling: But still we expose Env to a strict dependency on a devlib method implementation.
Not sure it's safe to assume that that function will never integrate code meant to be executed just one time...

bjackman: Good point.

bjackman:
Agree that a generic tools installation method is a good thing, probably in this case for example the `RTA` class should say "I need the rt-app tool installed". For now though I think this ad-hoc method in the `Executor` is fine.

Pushed a new version that just calls `setup` once.

derkling:
The usage of install_tools is way better and should be safer from the devlib integration stadpoint too.
Thanks for the update! :+1:

PR #243, opened 2016-11-29
LisaTest: Add get_trace method

bjackman:
It's a good idea to unify the creation of Traces in one place, because
passing a different sent of events will result in different time
indexes due to the time normalization being based on the time of the
first event.

This is also a nice convenience method.

### DIFF
--- a/libs/utils/test.py
+++ b/libs/utils/test.py
@@ -26,6 +26,7 @@ import wrapt
 
 from env import TestEnv
 from executor import Executor
+from trace import Trace
 
 
 class LisaTest(unittest.TestCase):
@@ -123,7 +124,8 @@ class LisaTest(unittest.TestCase):
         """
         Return a SchedAssert over the task provided
         """
-        return SchedAssert(experiment.out_dir, self.te.topology, execname=task)
+        return SchedAssert(
+            self.get_trace(experment), self.te.topology, execname=task)
 
     @memoized
     def get_multi_assert(self, experiment, task_filter=""):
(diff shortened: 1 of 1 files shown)


libs/utils/test.py
```diff
@@ -134,10 +136,27 @@ def get_multi_assert(self, experiment, task_filter=""):
...
         """
         tasks = experiment.wload.tasks.keys()
-        return SchedMultiAssert(experiment.out_dir,
+        return SchedMultiAssert(self.get_trace(experiment).ftrace,
                                 self.te.topology,
                                 [t for t in tasks if task_filter in t])
 
+    def get_trace(self, experiment):
+        if not hasattr(self, "__traces"):
+            self.__traces = {}
```
derkling: Is that "__" intentional, a private member is not enough?
Or maybe you really want to avoid subclasses to overwrite this member?

libs/utils/test.py
```diff
@@ -134,10 +136,27 @@ def get_multi_assert(self, experiment, task_filter=""):
...
+        if experiment.out_dir in self.__traces:
+            return self.__traces[experiment.out_dir]
+
+        if ('ftrace' not in experiment.conf['flags']
+            or 'ftrace' not in self.test_conf):
+            raise ValueError('get_trace only supported when ftrace enabled')
```
derkling: Maybe better:
```python
raise ValueError(
    'Tracing not enabled. If this experiment needs a trace '
    'you should add "ftrace" to your test/experiment configuration flags.')
```

libs/utils/test.py
bjackman: Yeah it was intentional, I thought if subclasses want to override it they should just override this method entirely. Not a strong opinion though, happy to change to a single underscore.

derkling: Ok, makes sense.

libs/utils/test.py
bjackman: Sure

bjackman:
Pushed a new version with the better error message, left the `__` though.

PR #245, opened 2016-11-30
lisa_shell: Reset terminal colours in PS1

bjackman:
So we don't mess up the terminal too much for people with pale backgrounds.


bjackman:
Don't merge, doesn't work properly on $colleague's machine

bjackman:
This should be fixed now and works on my machine.. couldn't see any problems on $colleague's but should probably have someone test it

PR #247, opened 2016-12-01
libs/utils/android: reset airplane mode for Jankbench

jlelli:
Jankbench activates airplaine mode before executing.
Set airplane mode back to OFF after execution finishes.


libs/utils/android/workloads/jankbench.py
```diff
@@ -154,6 +154,8 @@ def run(self, exp_dir, test_name, iterations, collect=''):
 
         # Switch back to screen auto rotation
         Screen.set_orientation(self.target, auto=True)
+
```
derkling: That's a good one! :+1: 
Can you remove this empty line and change the previous comment to be more "generic", e.g. "Reset initial setup"?

PR #249, opened 2016-12-07
Update trappy and fix cpu_idle

bjackman:
Some people have been using a later trappy version than the one specified in the submodules. That version contains a commit by me that breaks LISA's idle analysis module, so here's a fix for it.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -33,7 +33,7 @@ from devlib.utils.misc import memoized
 from trappy.utils import listify
 
 
-NON_IDLE_STATE = 4294967295
+NON_IDLE_STATE = -1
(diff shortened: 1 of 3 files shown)


libs/utils/trace.py
```diff
@@ -33,7 +33,6 @@
 from trappy.utils import listify
 
 
-NON_IDLE_STATE = 4294967295
```
mdigiorgio: I'd still keep `NON_IDLE_STATE = -1`, it makes the code more readable.

bjackman: Agreed.

bjackman:
Pushed a new version w/ comment addressed

mdigiorgio:
LGTM, let's merge it.

PR #250, opened 2016-12-08
submodules: Update devlib to latest master

bjackman:
This includes a fix that allows file_exists to find file in directories
readable only by root, such as /sys/kernel/debug/tracing/available_events.


derkling:
Ok, let's pull in this one... which I need it also to push another fix on our side...

PR #251, opened 2016-12-09
tests/eas/preliminary: Add initial EAS suitability check

bjackman:
This introduces a test suite that provides early-warnings that the
platform does not have the preliminary setup necessary to support EAS

---
supersedes https://github.com/ARM-software/lisa/pull/235

CC @credp

### DIFF
--- a/tests/eas/preliminary.py
+++ b/tests/eas/preliminary.py
@@ -1,0 +1,164 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2016, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import json
+import time
+import re
+import pandas
+import StringIO
+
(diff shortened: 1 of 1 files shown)


tests/eas/preliminary.py
```diff
@@ -0,0 +1,241 @@
...
+        ]
+
+        fail_list = [c for c in necessary_configs
+                     if not kernel_config.is_enabled(c)]
+
+        if len(fail_list):
+            message = 'Mising kernel configs: ' + ', '.join(fail_list)
+            self.assertTrue(len(fail_list) == 0, msg=message)
+
+class TestCpufreqAccounting(BasicCheckTest):
```
bjackman: Doesn't work on mainline or without `CONFIG_CPU_FREQ_STAT`. Not sure if we need this given that we test cpufreq below.

tests/eas/preliminary.py
```diff
@@ -0,0 +1,241 @@
...
+        # NB: We don't test for schedtune/schedutil, that's tested by
+        # TestSchedGovernor.
+        necessary_configs = [
+            # 'CONFIG_CPU_FREQ_STAT',
+            'CONFIG_CGROUPS',
+            'CONFIG_SMP',
+            'CONFIG_SCHED_MC',
+            'CONFIG_CPU_FREQ',
+            'CONFIG_CPU_IDLE',
+            'CONFIG_SCHED_DEBUG',
```
bjackman: Note that I've removed Schedtune stuf here too, let's test that separately since we'd probably like to be able to test kernels for EAS-related stuff without schedtune.

derkling:
I :heart: this test... simple yet a fundamental important initial step.
Tested on a broken Juno, where all the tests fails and the backtrace allows to identify the missing featrues.
Tested on Pixel, where the configuration test has been properly skipped because of the missing Kernel config.

To me that's ok for merge, let's give to @credp the honour ;-)

Issue #252, opened 2016-12-13
Define consistent documentation style/format


msrasmussen:
Documentation of many functions/classes is a bit sparse, but before that can be properly addressed a decision has to made on the in-code docstring documentation style. Looking through a few random python files I came across at least three different styles of documenting function arguments. In some cases there is even multiple styles used within a single file.
Could we have a decision on the in-code documentation style so it can be cleaned up and people help writing useful documentation?

bjackman:
Seems to me that [Sphinx autodoc](http://www.sphinx-doc.org/en/1.4.9/tutorial.html#autodoc) is the way forward for this. It's what Trappy and Bart use. I'll spend a little time investigating.

bjackman:
> we'll need to list the modules we want to document and how to structure that documentation (probably just mirror the Python package structure)

Actually seems we can just use `sphinx-apidoc` to structure this for us, that generates documentation for entire packages at a time. We'll still be able to add our own structure to things if we like.

Issue #198, opened 2016-09-28
LisaTest: Conveniently run tests on previously-gathered workload traces


bjackman:
While developing tests it would be useful to be able to run a workload once, then run the same set of tests against the trace results multiple times.

valschneider:
Another one taken care of by https://github.com/valschneider/lisa/tree/tests/executor-rework

PR #254, opened 2016-12-13
Add frequency invariant load tracking test 

bjackman:
See [commit message](https://github.com/ARM-software/lisa/commit/3224f5a344ddd8f16967fd6163289bd160d3f63e).

This test is quite simple so I've tried to make it an example of how we could design tests so that they can be used both from the command line as automated tests, and via notebooks as diagnostic/analysis tools. That means I've also added a couple of commits to make instantiating tests from notebooks easier. The interface for doing that is a bit bizarre (instantiate then `setUpClass`) but it works. The test/target configuration is also not inline in the notebook, instead the config comes from `target.config`. I personally like that, but it might be confusing invisible magic to someone new to LISA.

### DIFF
--- a/ipynb/tests/Frequency_Invariance_Test.ipynb
+++ b/ipynb/tests/Frequency_Invariance_Test.ipynb
@@ -1,0 +1,677 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Frequency Invariant Load Tracking Test\n",
+    "\n",
+    "FreqInvarianceTest is a LisaTest class for automated testing of frequency invariant load tracking. This notebook uses the methods it provides to perform the same analysis as the automated test and plot some results.\n",
+    "\n",
+    "The test class runs the same workload at a selection of frequencies, each entry in `t.experiments` represents a run at a different frequency."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Setup"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib inline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import json\n",
+    "\n",
+    "from trace import Trace\n",
+    "from trappy.plotter import plot_trace\n",
+    "from trappy.stats.grammar import Parser\n",
+    "from trappy import ILinePlot"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "2017-02-13 18:36:59,553 INFO    : root         : Using LISA logging configuration:\n",
+      "2017-02-13 18:36:59,554 INFO    : root         :   /home/brendan/sources/lisa/logging.conf\n"
+     ]
+    }
+   ],
+   "source": [
+    "import logging\n",
+    "from conf import LisaLogging\n",
+    "LisaLogging.setup()\n",
+    "logging.getLogger('Analysis').setLevel(logging.ERROR)\n",
+    "logging.getLogger('Trace').setLevel(logging.ERROR)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Run test workload\n",
+    "\n",
+    "There's currently no way to pass a `TestEnv` or configuration to automated test classes. Instead the target information comes from the `target.config` file (in the root of the LISA source tree), so you'll need to edit that to configure LISA to connect to your target."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\n",
+      "    Goal\n",
+      "    ====\n",
+      "    Basic check for frequency invariant load tracking\n",
+      "\n",
+      "    Detailed Description\n",
+      "    ====================\n",
+      "    This test runs the same workload on the most capable CPU on the system at a\n",
+      "    cross section of available frequencies. The trace is then examined to find\n",
+      "    the average activation length of the workload, which is combined with the\n",
+      "    known period to estimate an expected mean value for util_avg for each\n",
+      "    frequency. The util_avg value is extracted from scheduler trace events and\n",
+      "    its mean is compared with the expected value (ignoring the first 300ms so\n",
+      "    that the signal can stabilize). The test fails if the observed mean is\n",
+      "    beyond a certain error margin from the expected one. load_avg is then\n",
+      "    similarly compared with the expected util_avg mean, under the assumption\n",
+      "    that load_avg should equal util_avg when system load is light.\n",
+      "\n",
+      "    Expected Behaviour\n",
+      "    ==================\n",
+      "    Load tracking signals are scaled so that the workload results in roughly the\n",
+      "    same util & load values regardless of frequency.\n",
+      "    \n"
+     ]
+    }
+   ],
+   "source": [
+    "from tests.eas.load_tracking import FreqInvarianceTest\n",
+    "\n",
+    "t = FreqInvarianceTest()\n",
+    "print t.__doc__"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To run automated tests from within a notebook we instantiate the test class and call `runExperiments` on it."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {
+    "collapsed": false,
+    "scrolled": true
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "2017-02-13 18:36:59,608 INFO    : LisaTest     : Setup tests execution engine...\n",
+      "2017-02-13 18:36:59,609 INFO    : TestEnv      : Using base path: /home/brejac01/sources/lisa\n",
+      "2017-02-13 18:36:59,610 INFO    : TestEnv      : Loading default (file) target configuration\n",
+      "2017-02-13 18:36:59,611 INFO    : TestEnv      : Loading target configuration [/home/brejac01/sources/lisa/target.config]...\n",
+      "2017-02-13 18:36:59,613 INFO    : TestEnv      : Loading custom (inline) test configuration\n",
+      "2017-02-13 18:36:59,614 INFO    : TestEnv      : Devlib modules to load: ['bl', u'cpuidle', 'cgroups', 'cpufreq']\n",
+      "2017-02-13 18:36:59,615 INFO    : TestEnv      : Connecting linux target:\n",
+      "2017-02-13 18:36:59,616 INFO    : TestEnv      :   username : brendan\n",
+      "2017-02-13 18:36:59,617 INFO    : TestEnv      :       host : 192.168.2.2\n",
+      "2017-02-13 18:36:59,618 INFO    : TestEnv      :   password : password\n",
+      "2017-02-13 18:36:59,618 INFO    : TestEnv      : Connection settings:\n",
+      "2017-02-13 18:36:59,619 INFO    : TestEnv      :    {'username': u'brendan', 'host': u'192.168.2.2', 'password': u'password'}\n",
+      "2017-02-13 18:37:07,367 INFO    : TestEnv      : Initializing target workdir:\n",
+      "2017-02-13 18:37:07,369 INFO    : TestEnv      :    /home/brendan/devlib-target\n",
+      "2017-02-13 18:37:11,034 INFO    : CGroups      : Available controllers:\n",
+      "2017-02-13 18:37:12,775 INFO    : CGroups      :   cpuset       : /home/brendan/devlib-target/cgroups/devlib_cgh2\n",
+      "2017-02-13 18:37:14,512 INFO    : CGroups      :   cpu          : /home/brendan/devlib-target/cgroups/devlib_cgh3\n",
+      "2017-02-13 18:37:16,249 INFO    : CGroups      :   cpuacct      : /home/brendan/devlib-target/cgroups/devlib_cgh3\n",
+      "2017-02-13 18:37:17,988 INFO    : CGroups      :   blkio        : /home/brendan/devlib-target/cgroups/devlib_cgh4\n",
+      "2017-02-13 18:37:19,726 INFO    : CGroups      :   memory       : /home/brendan/devlib-target/cgroups/devlib_cgh5\n",
+      "2017-02-13 18:37:21,465 INFO    : CGroups      :   devices      : /home/brendan/devlib-target/cgroups/devlib_cgh6\n",
+      "2017-02-13 18:37:23,205 INFO    : CGroups      :   perf_event   : /home/brendan/devlib-target/cgroups/devlib_cgh7\n",
+      "2017-02-13 18:37:24,944 INFO    : CGroups      :   hugetlb      : /home/brendan/devlib-target/cgroups/devlib_cgh8\n",
+      "2017-02-13 18:37:26,682 INFO    : CGroups      :   pids         : /home/brendan/devlib-target/cgroups/devlib_cgh9\n",
+      "2017-02-13 18:37:29,594 INFO    : TestEnv      : Topology:\n",
+      "2017-02-13 18:37:29,596 INFO    : TestEnv      :    [[0, 3, 4, 5], [1, 2]]\n",
+      "2017-02-13 18:37:32,227 INFO    : TestEnv      : Loading default EM:\n",
+      "2017-02-13 18:37:32,229 INFO    : TestEnv      :    /home/brejac01/sources/lisa/libs/utils/platforms/juno.json\n",
+      "2017-02-13 18:37:33,996 WARNING : LinuxTarget  : Event [sched_load_avg_task] not available for tracing\n",
+      "2017-02-13 18:37:33,999 WARNING : LinuxTarget  : Event [sched_load_avg_cpu] not available for tracing\n",
+      "2017-02-13 18:37:34,001 INFO    : TestEnv      : Enabled tracepoints:\n",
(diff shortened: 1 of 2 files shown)


tests/eas/load_tracking.py
```diff
@@ -0,0 +1,243 @@
...
+    ====
+    Basic check for frequency invariant load tracking
+
+    Detailed Description
+    ====================
+    This test runs the same workload on the most capable CPU on the system at a
+    cross section of available frequencies. The trace is then examined to find
+    the average activation length of the workload, which is combined with the
+    known period and frequency to estimate an expected mean value for util_avg
+    for each frequency. The util_avg value is extract from scheduler trace
```
ionela-voinescu: :%s/extract/extracted

ipynb/tests/Frequency Invariance Test.ipynb
```diff
@@ -0,0 +1,791 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Setup"
```
ionela-voinescu: The notebook starts a bit abrupt. It would be best to start with:
# Frequency Invariant Test

"FreqInvarianceTest is a LisaTest class for automated testing of frequency invariant load tracking. This notebook uses the methods it provides to perform the same analysis as the automated test and plot some results."

And maybe add here:
"Target setup is done based on the information in target.config."

## Setup
[snip]

ipynb/tests/Frequency Invariance Test.ipynb
```diff
@@ -0,0 +1,791 @@
+{
```
ionela-voinescu: It might be worth changing the name of the notebook to Frequency_Invariant_Test.ipynb to avoid having spaces in a file's name.

ipynb/tests/Frequency Invariance Test.ipynb
```diff
@@ -0,0 +1,791 @@
...
+      "    Run workloads providing a known utilization and test that the PELT\n",
+      "    (Per-entity load tracking) util_avg and load_avg signals reflects that\n",
+      "    utilization at various CPU frequencies.\n",
+      "    \n"
+     ]
+    }
+   ],
+   "source": [
+    "t = FreqInvarianceTest()\n",
+    "print t.__doc__"
```
ionela-voinescu: Love the documentation in the python code and how it's shown here :+1: .

bjackman:
Thanks for the review @ionela-voinescu 

> On the previous idea, one thing that might be worth having is a way to point to an existing trace and plot and present the results from a previous run of the test. A use case for this can be to visualise a previously run automatic test and understand what went wrong.
Agreed. I think we should also be able to run automated pass/fail tests on previously-gathered traces. I have an [issue](https://github.com/ARM-software/lisa/issues/198) for that.

ipynb/tests/Frequency Invariance Test.ipynb
bjackman: Good point.

ipynb/tests/Frequency Invariance Test.ipynb
bjackman: OK, I added a header text at the top but kept the info down here that I think is clearest next to the test class instantiation.

bjackman:
OK, pushed a new version.

bjackman:
Just noticed I haven't used the documentation style I suggested in https://github.com/ARM-software/lisa/issues/252 but I think it's OK to fix that at a later date, shouldn't take too long.

tests/eas/load_tracking.py
```diff
@@ -0,0 +1,236 @@
...
+    Goal
+    ====
+    Basic check for frequency invariant load tracking
+
+    Detailed Description
+    ====================
+    This test runs the same workload on the most capable CPU on the system at a
+    cross section of available frequencies. The trace is then examined to find
+    the average activation length of the workload, which is combined with the
+    known period and duty cycle to estimate an expected mean value for util_avg
```
bjackman: > ... and duty cycle ...
no

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,799 @@
...
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import json\n",
+    "\n",
+    "from tests.eas.load_tracking import FreqInvarianceTest\n",
```
derkling: Maybe this is better to be moved in cell 4 where you actually create the FreqInvarianceTest?
This should make more clear where the test class comes from.

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,799 @@
...
+   "source": [
+    "t = FreqInvarianceTest()\n",
+    "print t.__doc__"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To run automated tests from within a notebook we instantiate the test class and call `setUpClass` on it. This is what `nosetests` does internally when running tests from the command line."
```
derkling: What about masking from the Notebook what nosetests does?

I'm kind of afraid by the idea that "exposing nosetests details" can scare users... what do you think?
Maybe a wrapper method, e.g. a `runExperiments()` which basically just call `setUpClass()`, can make more clean the execution of tests from a notebook?

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,799 @@
...
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "# Get the frequency an experiment was run at\n",
+    "def experiment_freq(exp):\n",
+    "    [cpu] = exp.wload.cpus\n",
+    "    freq = exp.conf['cpufreq']['freqs'][cpu]\n",
+    "    return freq\n",
+    "freqs = [experiment_freq(e) for e in t.executor.experiments]"
```
derkling: on the same line of the previous comment, maybe the base class can expose a `experiments` param which actually return `t.executor.experiments`?

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,799 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "def plot_signal_against_freq(signal):\n",
+    "    means = [t.get_signal_mean(e, signal) for e in t.executor.experiments]\n",
```
derkling: the `experiments` param can be use all around to simplify the code by exposing just a simple "experiments" attribute of the "test" class created in cell 4.

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,799 @@
...
+   "cell_type": "code",
+   "execution_count": 13,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "signals = ['util_avg', 'load_avg']\n",
+    "def examine_experiment(idx):\n",
+    "    experiment = t.executor.experiments[idx]\n",
```
derkling: Looking at this other access to the experiments, maybe a better wrapper can be `t.experiment(idx)`?

ipynb/tests/Frequency_Invariance_Test.ipynb
bjackman: Yeah, that sounds like a good idea. The whole way we use `unittest` is a bit iffy IMO but maybe at least we can hide it. I'll see what I can do.

ipynb/tests/Frequency_Invariance_Test.ipynb
bjackman: Yeah, I initially didn't do that because I heard $colleague bemoaning the number of aliases LISA has and didn't want to add another! But I think in this case if we can document the `LisaTest` API properly then it would be useful to actually hide `Executor` completely.

ipynb/tests/Frequency_Invariance_Test.ipynb
bjackman: OK, you think it should be a method (rather than raw attribute)? In that case it should be `get_experiment` not `experiment`.

But actually I prefer a raw attribute to a method, because then you can easily iterate over it.

ipynb/tests/Frequency_Invariance_Test.ipynb
bjackman: OK

bjackman:
Pushed a new version. Responded to your comments and also simplified `get_expected_util_avg` by using `SchedAssert`.

tests/eas/load_tracking.py
```diff
@@ -0,0 +1,225 @@
...
+            })
+
+        return {
+            'wloads': wloads,
+            'confs': confs,
+        }
+
+    def get_expected_util_avg(self, experiment):
+        """
+        Examine trace to figure out an expected mean for util_avg
```
ionela-voinescu: This does not examine the trace yet, right?
It uses the duty cycle from the workload and frequencies from the selected ones, as far as I understand.

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,827 @@
...
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "2017-01-12 18:44:49,210 INFO    : root         : Using LISA logging configuration:\n",
+      "2017-01-12 18:44:49,211 INFO    : root         :   /home/brejac01/sources/lisa/logging.conf\n"
```
bjackman: Good point, will replace.

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,827 @@
...
+    "collapsed": false,
+    "scrolled": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Experiment ran at frequency 307200\n",
+      "Trace plot:\n"
```
bjackman: I can see them On My Machine. I don't have the solution here but this is a trappy issue. Have raised a related Issue about how the trappy JS gets installed here: https://github.com/ARM-software/trappy/issues/239. I think we'll need to investigate this properly at some stage but for now I propose not fixing it here.

tests/eas/load_tracking.py
bjackman: It examines the trace via `SchedAssert.getDutyCycle`. 

You're right about the frequency though, it just checks the workload and assumes that all the cpufreq setup worked fine (which I'm OK with, since this isn't testing cpufreq, devlib or LISA). The motivation for that is that we don't currently have the `target.cpufreq.trace_frequencies` functionality working and so I don't expect there to be any cpu_frequency events in the trace.

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,827 @@
...
+      "text/plain": [
+       "<matplotlib.figure.Figure at 0x7f9dcb170750>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "plot_signal_against_freq('util_avg')"
```
ionela-voinescu: As discussed, it would be best if the load/util_avg axes had a wider range as to show that you should have almost the same value for all frequencies.

bjackman:
OK pushed a new version with your comments addressed

ipynb/tests/Frequency_Invariance_Test.ipynb
```diff
@@ -0,0 +1,827 @@
...
+    "collapsed": false,
+    "scrolled": true
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "2017-01-18 18:34:40,898 INFO    : LisaTest     : Setup tests execution engine...\n",
+      "2017-01-18 18:34:40,899 INFO    : TestEnv      : Using base path: /home/bjackman/sources/lisa\n",
```
ionela-voinescu: I can still see the logs having "/home/bjackman/" instead of "/home/vagrant/". Is this something you missed, or wanted to keep like this?

bjackman: Yeah I changed it from "brejac01" (my username on my local machine) to "bjackman" (my github username).. I don't really have any backup for that reasoning it's just what I saw in some of the other notebooks.

ionela-voinescu: With the restructure of the example notebooks we've discussed it might be better to stick to a generic "/home/vagrant/", but yes, not all notebooks are consistent in that regard at the moment.

bjackman:
Pushed a new version with s/bjackman/vagrant/

ionela-voinescu:
Awesome! That was all from my side. Thank you for this!

libs/utils/test.py
```diff
@@ -191,6 +191,12 @@ def get_end_times(self, experiment):
 
         return end_times
 
+    # Defining a default, empty runTest allows test classes to be conveniently
+    # instantiated. Since we run tests via nose, runTest isn't used. Instead all
```
derkling: Are you saying that this method is required by the base class `unittest.TestCase`?
Perhaps I'm still under jet-la, but... I cannot really quite understand, from this comment, why we need to add this empty method?!

libs/utils/test.py
```diff
@@ -125,7 +125,7 @@ def get_sched_assert(self, experiment, task):
         Return a SchedAssert over the task provided
         """
         return SchedAssert(
-            self.get_trace(experment), self.te.topology, execname=task)
```
derkling: woww... that was a bold error... :-O

libs/utils/test.py
```diff
@@ -54,7 +54,7 @@ class LisaTest(unittest.TestCase):
     """Override this with a dictionary to configure the Executor"""
 
     @classmethod
-    def _init(cls, *args, **kwargs):
+    def init(cls, *args, **kwargs):
```
derkling: Here perhaps I see a better usage for the currently empty runTest() method you added in the first patch.

Could it make sense to avoid this patch and call this _init() from the runTest()?
It would make also a more consistent interface from the Notebooks side... cause we will have:

```
from tests.eas.load_tracking import FreqInvarianceTest
t = FreqInvarianceTest()
r.runTests()
```

which makes more sense instead of calling a `t.init()` which is actually running tests.

ipynb/tests/Frequency_Invariance_Test.ipynb
derkling: The [solution proposed by Javi](ARM-software/trappy/issues/239)  is actually something we should better document/advertise on the LISA side as well. ;-)

Perhaps we can wrap the required export commands into a new `lisa-export` command?

tests/eas/load_tracking.py
derkling: Right, AFAIU we run these tests at a constant userspace defined frequency.
Thus there are not os-generated CPUFreq events... but the devlib injected ones are now available... perhaps we can add a cross-check in the future to verify that the trace was actually collected while running at the expected frequency.

Can you add an issue for this addition and link this discussion as a reference?

tests/eas/load_tracking.py
```diff
@@ -0,0 +1,225 @@
...
+                  the experiment's workload task
+        """
+        [task] = experiment.wload.tasks.keys()
+        trace = self.get_trace(experiment)
+
+        # There are two different scheduler trace events that expose the load
+        # tracking signals. Neither of them is in mainline. Eventually they
+        # should be unified but for now we'll just check for both types of
+        # event.
+        # TODO: Add support for this parsing in Trappy and/or tasks_analysis
```
derkling: Why not doing it now? :-)

I would rather prefer to merge before a patch which fixes all the PELT signals. We already discussed this morning about that being require before next release. Thus, let's do it now... what do you think?

tests/eas/load_tracking.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # 10% periodic RTApp workload:
+        wloads = {
+            'fie_10pct' : {
+                'type' : 'rt-app',
+                'conf' : {
+                    'class' : 'periodic',
+                    'params' : {
+                        'duty_cycle_pct': 10,
+                        'duration_s': 3,
```
derkling: Could this be reduce to 1 or max 2 [s]... it should be more than enough for such a short period task and it could save some time while running multiple tests.

tests/eas/load_tracking.py
```diff
@@ -0,0 +1,225 @@
...
+                        'period_ms': 10,
```
derkling: Would probably suggest to stay aligned with main PELT time constants, 16 or 32 [ms] should produce in general a more smooth signal...

tests/eas/load_tracking.py
```diff
@@ -0,0 +1,225 @@
...
+            },
+        }
+
+        # Create a set of confs with different frequencies
+        # We'll run the 10% workload under each conf (i.e. at each frequency)
+        confs = []
+
+        all_freqs = test_env.target.cpufreq.list_frequencies(cpu)
+        # If we have loads of frequencies just test a cross-section so it
+        # doesn't take all day
```
derkling: smart :+1:

tests/eas/load_tracking.py
bjackman: [Done](https://github.com/ARM-software/lisa/issues/291)

derkling:
Few comments but overall ok for me, although I would suggest to keep the actual code which adds the test on a separate PR. This would allow to easily reference it in the future and point people interesting in writing new tests to some good examples of what need to be provided.

tests/eas/load_tracking.py
bjackman: The reason I haven't done it here is that the format of the sched_pelt_cfs_rq event is going to change soon

tests/eas/load_tracking.py
bjackman: Should be fine

tests/eas/load_tracking.py
bjackman: Sounds good, will test this out.

libs/utils/test.py
bjackman: Yeah, this comment is pretty unclear clear but it's because when you directly instantiate a `TestCase` it looks for a method called `runTest` (unless you specify a different method, which AFAICT is what `nosetests` does.

Anyway I tried out your suggestion of using `runTest` instead of `init` and it works find so let's do that.

bjackman:
I've separated out the dependency patches as https://github.com/ARM-software/lisa/pull/292. Once that's merged I'll remove the relevant commits from this PR so it's just the load_tracking stuff

bjackman:
OK, rebased.

tests/eas/load_tracking.py
derkling: Ok, let's progress this 17.02 and fix/update this later.
At the very least, we will have automated tests remembering us about broken events. ;-)

PR #261, opened 2016-12-19
Some Sphinx autodoc compatible docstrings

bjackman:
Some docstrings I wrote on the train. I'll submit infrastructure to actually build this in the future.

I also got halfway through documenting `Executor` but it's still rather messy so I'll try to get round to submitting it properly.

### DIFF
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -41,6 +41,133 @@ Experiment = namedtuple('Experiment', ['wload_name', 'wload',
                                        'conf', 'iteration', 'out_dir'])
 
 class Executor():
+    """
+    Abstraction for running sets of experiments and gathering data from targets
+
+    An executor can be configured to run a set of workloads (wloads) in each
+    different target configuration of a specified set (confs). These wloads and
+    confs can be specified by the "experiments_conf" input dictionary. Each
+    (workload, conf, iteration) tuple is called an "experiment".
+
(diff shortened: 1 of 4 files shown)


bjackman:
ref issue https://github.com/ARM-software/lisa/issues/252

bjackman:
Added docs for `Executor`. Somewhat incomplete but hopefully still useful.

libs/utils/executor.py
```diff
@@ -52,40 +161,12 @@ class Executor():
...
-
-        The executor module can be configured to run a set of workloads (wloads)
-        in each different target configuration of a specified set (confs). These
-        wloads and confs can be specified by the "experiments_conf" input
-        dictionary. Each (workload, conf, iteration) tuple is called an
-        "experiment".
-
-        All the results generated by each experiment will be collected a result
-        folder which is named according to this template:
-            results/<test_id>/<wltype>:<conf>:<wload>/<run_id>
```
derkling: We lost this info... should we not add it in the description of the Executor class?

bjackman: Yeah, I replaced this with the doc for `Experiment.out_dir` because I didn't think the output directory structure a hard-defined part of `Executor`'s interface (in fact this is already wrong if you set 'results_dir' in your `test_conf`). I can put it back though.

bjackman: Pushed a new version with the info put back in.

PR #263, opened 2016-12-21
Energy model-based tests, RFC v2

bjackman:
See the commit messages for the [first](https://github.com/ARM-software/lisa/commit/399cdfff3c112836408f03346832df52008779d7) and [fourth](https://github.com/ARM-software/lisa/commit/37918fade84d389a6df2183ac02892fe3fef898d) commits for more detail on what this PR aims to introduce. In short, this aims to allow automatically running tests for any given synthetic workload and platform. 

This looks like an enormous PR, and it is... but a lot of the added lines are `.ipynb` noise or boilerplate in the [provided](https://github.com/ARM-software/lisa/commit/399cdfff3c112836408f03346832df52008779d7#diff-6a86319666706547b6bb03baf38e4916) [platform](https://github.com/ARM-software/lisa/commit/399cdfff3c112836408f03346832df52008779d7#diff-ac2a5c8c78e008f3791e81447c911ca6) [descriptions](https://github.com/ARM-software/lisa/commit/399cdfff3c112836408f03346832df52008779d7#diff-b622e9d205b02984946d5cdb1cd96a96). I've also added pretty verbose docstrings which bumps up the line count a lot. 

I've siloed myself a bit on this so please highlight wherever I haven't made the high-level intention clear, It's important that someone can understand what this stuff is trying to do without reading 2k lines of inscrutable bjackman Python :smile: . Hopefully docstrings are a good starting point...

### DIFF
--- a/ipynb/tests/Generic_EAS_Tests.ipynb
+++ b/ipynb/tests/Generic_EAS_Tests.ipynb
@@ -1,0 +1,697 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": true
+   },
+   "source": [
+    "# Run Generic Automated EAS tests\n",
+    "This is a starting-point notebook for running tests from the generic EAS suite in `tests/eas/generic.py`. The test classes that are imported here provide helper methods to aid analysis of the cause of failure. You can use Python's `help` built-in to find those methods (or you can just read the docstrings in the code).\n",
+    "\n",
+    "These tests make estimation of the energy efficiency of task placements, without directly examining the behaviour of cpufreq or cpuidle. Several test classes are provided, the only difference between them being the workload that is used."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Setup"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "2017-01-23 18:50:17,847 INFO    : root         : Using LISA logging configuration:\n",
+      "2017-01-23 18:50:17,847 INFO    : root         :   /home/brendan/sources/lisa/logging.conf\n"
+     ]
+    }
+   ],
+   "source": [
+    "%load_ext autoreload\n",
+    "%autoreload 2\n",
+    "%matplotlib inline\n",
+    "\n",
+    "import logging\n",
+    "from conf import LisaLogging\n",
+    "LisaLogging.setup()#level=logging.WARNING)\n",
+    "\n",
+    "import pandas as pd\n",
+    "\n",
+    "from perf_analysis import PerfAnalysis\n",
+    "\n",
+    "import trappy\n",
+    "from trappy import ILinePlot\n",
+    "from trappy.stats.grammar import Parser"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Run test workload\n",
+    "\n",
+    "If you simply want to run all the tests and get pass/fail results, use this command in the LISA shell: `lisa-test tests/eas/generic.py`. This notebook is intended as a starting point for analysing what scheduler behaviour was judged to be faulty.\n",
+    "\n",
+    "Target configuration is taken from `$LISA_HOME/target.config` - you'll need to edit that file to provide connection details for the target you want to test."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "from tests.eas.generic import TwoBigTasks, TwoBigThreeSmall, RampUp, RampDown, EnergyModelWakeMigration, OneSmallTask"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "By default we'll run the EnergyModelWakeMigration test, which runs a workload alternating between high and low-intensity. All the other test classes shown above have the same interface, but run different workloads. To run the tests on different workloads, change this line below:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\n",
+      "    Test EAS for tasks alternating beetween 10% and 50%\n",
+      "    \n"
+     ]
+    }
+   ],
+   "source": [
+    "t = EnergyModelWakeMigration(methodName=\"test_task_placement\")\n",
+    "print t.__doc__"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false,
+    "scrolled": true
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "2017-01-23 18:50:18,176 INFO    : LisaTest     : Setup tests execution engine...\n",
+      "2017-01-23 18:50:18,178 INFO    : TestEnv      : Using base path: /home/brejac01/sources/lisa\n",
+      "2017-01-23 18:50:18,178 INFO    : TestEnv      : Loading default (file) target configuration\n",
+      "2017-01-23 18:50:18,179 INFO    : TestEnv      : Loading target configuration [/home/brejac01/sources/lisa/target.config]...\n",
+      "2017-01-23 18:50:18,181 INFO    : TestEnv      : Loading custom (inline) test configuration\n",
+      "2017-01-23 18:50:18,182 INFO    : TestEnv      : Devlib modules to load: ['bl', u'cpuidle', 'cgroups', 'cpufreq']\n",
+      "2017-01-23 18:50:18,183 INFO    : TestEnv      : Connecting linux target:\n",
+      "2017-01-23 18:50:18,184 INFO    : TestEnv      :   username : brendan\n",
+      "2017-01-23 18:50:18,185 INFO    : TestEnv      :       host : 192.168.2.2\n",
+      "2017-01-23 18:50:18,186 INFO    : TestEnv      :   password : password\n",
+      "2017-01-23 18:50:18,187 INFO    : TestEnv      : Connection settings:\n",
+      "2017-01-23 18:50:18,188 INFO    : TestEnv      :    {'username': u'brendan', 'host': u'192.168.2.2', 'password': u'password'}\n",
+      "2017-01-23 18:50:25,803 INFO    : TestEnv      : Initializing target workdir:\n",
+      "2017-01-23 18:50:25,805 INFO    : TestEnv      :    /home/brendan/devlib-target\n",
+      "2017-01-23 18:50:29,349 INFO    : CGroups      : Available controllers:\n",
+      "2017-01-23 18:50:31,076 INFO    : CGroups      :   schedtune    : /home/brendan/devlib-target/cgroups/devlib_cgh2\n",
+      "2017-01-23 18:50:31,749 INFO    : TestEnv      : Topology:\n",
+      "2017-01-23 18:50:31,751 INFO    : TestEnv      :    [[0, 3, 4, 5], [1, 2]]\n",
+      "2017-01-23 18:50:34,373 INFO    : TestEnv      : Loading default EM:\n",
+      "2017-01-23 18:50:34,374 INFO    : TestEnv      :    /home/brejac01/sources/lisa/libs/utils/platforms/juno.json\n",
+      "2017-01-23 18:50:36,104 INFO    : TestEnv      : Enabled tracepoints:\n",
+      "2017-01-23 18:50:36,106 INFO    : TestEnv      :    sched_overutilized\n",
+      "2017-01-23 18:50:36,107 INFO    : TestEnv      :    sched_energy_diff\n",
+      "2017-01-23 18:50:36,108 INFO    : TestEnv      :    sched_load_avg_task\n",
+      "2017-01-23 18:50:36,109 INFO    : TestEnv      :    sched_load_avg_cpu\n",
+      "2017-01-23 18:50:36,110 INFO    : TestEnv      :    sched_migrate_task\n",
+      "2017-01-23 18:50:36,111 INFO    : TestEnv      :    sched_switch\n",
+      "2017-01-23 18:50:36,112 INFO    : TestEnv      :    cpu_frequency\n",
+      "2017-01-23 18:50:36,114 INFO    : EnergyMeter  : HWMON module not enabled\n",
+      "2017-01-23 18:50:36,115 WARNING : EnergyMeter  : Energy sampling disabled by configuration\n",
+      "2017-01-23 18:50:36,116 INFO    : TestEnv      : Set results folder to:\n",
+      "2017-01-23 18:50:36,117 INFO    : TestEnv      :    /home/brejac01/sources/lisa/results/20170123_185036\n",
+      "2017-01-23 18:50:36,118 INFO    : TestEnv      : Experiment results available also in:\n",
+      "2017-01-23 18:50:36,119 INFO    : TestEnv      :    /home/brejac01/sources/lisa/results_latest\n",
+      "2017-01-23 18:50:36,121 INFO    : Executor     : Loading custom (inline) test configuration\n",
+      "2017-01-23 18:50:36,121 INFO    : Executor     : \n",
+      "2017-01-23 18:50:36,123 INFO    : Executor     : ################################################################################\n",
+      "2017-01-23 18:50:36,124 INFO    : Executor     : Experiments configuration\n",
+      "2017-01-23 18:50:36,125 INFO    : Executor     : ################################################################################\n",
+      "2017-01-23 18:50:36,126 INFO    : Executor     : Configured to run:\n",
+      "2017-01-23 18:50:36,127 INFO    : Executor     :      1 target configurations:\n",
+      "2017-01-23 18:50:36,128 INFO    : Executor     :       energy_aware\n",
+      "2017-01-23 18:50:36,129 INFO    : Executor     :      1 workloads (1 iterations each)\n",
+      "2017-01-23 18:50:36,130 INFO    : Executor     :       wake_migration\n",
+      "2017-01-23 18:50:36,131 INFO    : Executor     : Total: 1 experiments\n",
+      "2017-01-23 18:50:36,132 INFO    : Executor     : Results will be collected under:\n",
+      "2017-01-23 18:50:36,133 INFO    : Executor     :       /home/brejac01/sources/lisa/results/20170123_185036\n",
+      "2017-01-23 18:50:36,134 INFO    : Executor     : rt-app workloads found, installing tool on target\n",
+      "2017-01-23 18:50:38,252 INFO    : LisaTest     : Experiments execution...\n",
+      "2017-01-23 18:50:38,254 INFO    : Executor     : \n",
+      "2017-01-23 18:50:38,255 INFO    : Executor     : ################################################################################\n",
+      "2017-01-23 18:50:38,256 INFO    : Executor     : Experiments execution\n",
+      "2017-01-23 18:50:38,257 INFO    : Executor     : ################################################################################\n",
+      "2017-01-23 18:50:38,258 INFO    : Executor     : \n",
+      "2017-01-23 18:50:38,259 INFO    : Executor     : ================================================================================\n",
+      "2017-01-23 18:50:38,260 INFO    : Executor     : configuring target for [energy_aware] experiments\n",
+      "2017-01-23 18:50:39,976 INFO    : Executor     : Set scheduler feature: ENERGY_AWARE\n",
+      "2017-01-23 18:50:40,881 WARNING : Executor     : cpufreq governor not specified, using currently configured governor\n",
+      "2017-01-23 18:50:40,883 INFO    : Workload     : Setup new workload wake_migration\n",
+      "2017-01-23 18:50:40,884 INFO    : Workload     : Workload duration defined by longest task\n",
+      "2017-01-23 18:50:40,885 INFO    : Workload     : Default policy: SCHED_OTHER\n",
+      "2017-01-23 18:50:40,886 INFO    : Workload     : ------------------------\n",
+      "2017-01-23 18:50:40,888 INFO    : Workload     : task [task_wmig0], sched: using default policy\n",
+      "2017-01-23 18:50:40,889 INFO    : Workload     :  | calibration CPU: 1\n",
+      "2017-01-23 18:50:40,890 INFO    : Workload     :  | loops count: 2\n",
+      "2017-01-23 18:50:40,891 INFO    : Workload     : + phase_000001: duration 1.000000 [s] (10 loops)\n",
+      "2017-01-23 18:50:40,892 INFO    : Workload     : |  period   100000 [us], duty_cycle  10 %\n",
+      "2017-01-23 18:50:40,893 INFO    : Workload     : |  run_time  10000 [us], sleep_time  90000 [us]\n",
+      "2017-01-23 18:50:40,894 INFO    : Workload     : + phase_000002: duration 1.000000 [s] (10 loops)\n",
+      "2017-01-23 18:50:40,895 INFO    : Workload     : |  period   100000 [us], duty_cycle  50 %\n",
+      "2017-01-23 18:50:40,896 INFO    : Workload     : |  run_time  50000 [us], sleep_time  50000 [us]\n",
+      "2017-01-23 18:50:40,897 INFO    : Workload     : ------------------------\n",
+      "2017-01-23 18:50:40,897 INFO    : Workload     : task [task_wmig1], sched: using default policy\n",
+      "2017-01-23 18:50:40,898 INFO    : Workload     :  | calibration CPU: 1\n",
+      "2017-01-23 18:50:40,899 INFO    : Workload     :  | loops count: 2\n",
+      "2017-01-23 18:50:40,900 INFO    : Workload     : + phase_000001: duration 1.000000 [s] (10 loops)\n",
+      "2017-01-23 18:50:40,901 INFO    : Workload     : |  period   100000 [us], duty_cycle  10 %\n",
+      "2017-01-23 18:50:40,902 INFO    : Workload     : |  run_time  10000 [us], sleep_time  90000 [us]\n",
+      "2017-01-23 18:50:40,903 INFO    : Workload     : + phase_000002: duration 1.000000 [s] (10 loops)\n",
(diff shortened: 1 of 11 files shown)


bjackman:
Another example use: https://github.com/bjackman/lisa/blob/cie-test/tests/eas/load_tracking.py#L247

bjackman:
Today I ran the "generic" tests I've added here and compared the energy estimation from `EnergyModel` with the energy measured with an ACME cape on Pixel and hwmon on Juno. 

The only numerical analysis I did was to compare the estimated and measured energy for the *entirety of each workload*. There's a strong correlation (>95% coefficient) there.

As for finer grained predictions (i.e. predicting& measuring the variation in energy consumption *during the course of the workload*), I plotted energy measurements over time (using ACME) and compared visually with plots of energy predictions over time. They look to match nicely but like I say this wasn't very scientific.

ipynb/tests/Generic_EAS_Tests.ipynb
```diff
@@ -0,0 +1,533 @@
...
...996781, 7.1290729999982432], [7.1311389999991661, 7.1311669999995502], [7.1334329999990587, 7.1334889999998268], [7.1335029999991093, 7.1335199999994074], [7.1351799999993091, 7.1351979999999458], [7.1353299999991577, 7.1353419999995822], [7.1392589999995835, 7.1392769999984012], [7.1751179999992019, 7.1751319999984844], [7.1753509999998641, 7.1753609999996115], [7.1855259999992995, 7.185550999998668], [7.1870129999988421, 7.1870339999986754], [7.1873479999994743, 7.1873819999982516], [7.1874099999986356, 7.187432999999146], [7.188732999999047, 7.1887699999988399], [7.1901979999984178, 7.1902069999996456], [7.1902429999991, 7.1902529999988474], [7.2023279999993974, 7.2023379999991448], [7.2029539999984991, 7.2029599999987113], [7.205298999999286, 7.2053079999986949], [7.2077349999999569, 7.2077469999985624], [7.2151149999990594, 7.2151239999984682], [7.2404409999999189, 7.2404649999989488], [7.242086999998719, 7.242127999999866], [7.2422169999990729, 7.2422439999991184], [7.3629449999989447, 7.3629849999997532], [7.3646639999988111, 7.3646809999991092], [7.3649589999986347, 7.3649699999987206], [7.4134959999992134, 7.4135049999986222], [7.4150779999999941, 7.4150989999998274], [7.4
```
ionela-voinescu: It would be more clear if these would print a success message when the test is passed as there is no output from it otherwise. The same for test_task_placement.

ipynb/tests/Generic_EAS_Tests.ipynb
```diff
@@ -0,0 +1,533 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "utils = t.get_task_utils_df(experiment)"
```
ionela-voinescu: While I understand why you decided not to print the data frames for this call and the ones below, I for one would like to have more information in these test notebooks as it gives me more confidence in the test. If you have some calls to some functions that do not return anything it might make people think that something is wrong and maybe the test is not doing what it's suppose to be doing. I think an option would be to print the traces and plots first and maybe have a debug section after that where the data frames are printed and the process is explained in a few sentences.

ipynb/tests/Generic_EAS_Tests.ipynb
```diff
@@ -0,0 +1,533 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": true
+   },
+   "source": [
+    "# Run Generic Automated EAS tests\n",
```
ionela-voinescu: It would be good to have more information about what the tests report and how the evaluation is done. As you say below "These tests make estimation of the energy efficiency of task placements.. " + "by.. ". It would also be nice to make clear the distinction between the estimated observed power consumption (estimated based on the energy model provided and the observed task placement decisions made) and the expected power consumption (estimated based on the energy model provided and the expected optimal task placements). This can be done either here at the beginning or further in the notebook where the estimated and expected data frames are obtained.

ipynb/tests/Generic_EAS_Tests.ipynb
bjackman: The functions do return something, it's just assigned to a variable, I wouldn't expect to see any output from running that code. But you're right that this is unclear.

How about if we print the docstring for the function before calling it, then you basically get a detailed description of what's going on.

ipynb/tests/Generic_EAS_Tests.ipynb
bjackman: True, will add this.

ipynb/tests/Generic_EAS_Tests.ipynb
bjackman: True. I will try to clarify in the notebook.

tests/eas/generic.py
```diff
@@ -0,0 +1,433 @@
...
+        est_power = self.get_power_df(experiment)
+
+        exp_energy = area_under_curve(exp_power['power'])
+        est_energy = area_under_curve(est_power['power'])
+
+        msg = 'Estimated {} bogo-Joules to run workload, expected {}'.format(
+            est_energy, exp_energy)
+        self.assertLess(est_energy, exp_energy * 1.2, msg=msg)
+
+class OneSmallTask(EnergyModelTest):
```
bjackman: Why? In this case "generic" is supposed to mean "works on any platform".

tests/eas/generic.py
```diff
@@ -0,0 +1,433 @@
...
+        """
+        exp_power = self.get_expected_power_df(experiment)
+        est_power = self.get_power_df(experiment)
+
+        exp_energy = area_under_curve(exp_power['power'])
+        est_energy = area_under_curve(est_power['power'])
+
+        msg = 'Estimated {} bogo-Joules to run workload, expected {}'.format(
+            est_energy, exp_energy)
+        self.assertLess(est_energy, exp_energy * 1.2, msg=msg)
```
bjackman: Yep, will do.

tests/eas/generic.py
```diff
@@ -0,0 +1,433 @@
...
+        def est_power(row):
+            cpu_utils = [0 for cpu in nrg_model.cpus]
+            for task in tasks:
+                cpu = row['cpus'][task]
+                util = row['utils'][task]
+                if not isnan(cpu):
+                    cpu_utils[int(cpu)] += util
+            power = nrg_model.estimate_from_cpu_util(cpu_utils)
+            columns = power.keys()
+            return pd.Series([power[c] for c in columns], index=columns)
```
bjackman: Yep. I think I'll remove the "power" column and sort the node columns by depth-first order.

tests/eas/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+            cls.target.write_value(
+                "/proc/sys/kernel/sched_is_big_little", 1, verify=False)
+
+        if SET_INITIAL_TASK_UTIL:
+            # This flag doesn't exist on all kernels, so don't worry if the file
+            # isn't present (hence verify=False)
+            cls.target.write_value(
+                "/proc/sys/kernel/sched_initial_task_util", 1024, verify=False)
+
+    def _do_test_first_cpu(self, experiment, tasks):
```
bjackman: Yes. This function actually just being moved from acceptance.py so I'll leave it here for now. None of the new tests should use it.

tests/eas/acceptance.py
```diff
@@ -393,7 +408,10 @@ def test_all_tasks_finish_on_a_big_cpu(self, experiment, tasks):
...
 
             msg = "Task {} did not finish on a big cpu".format(task)
-            self.assertIn(sa.getLastCpu(), self.target.bl.bigs, msg=msg)
+
+            self.assertIn(sa.getLastCpu(),
+                          self.te.nrg_model.biggest_cpus,
+                          msg=msg)
 
 
 class WakeMigration(EasTest):
```
bjackman: Will take a look.

bjackman:
OK, thanks for the review. I've responded to the comments to remind myself what changes I need to make, and I'll come back once I have time to make the code changes.

ipynb/tests/Generic_EAS_Tests.ipynb
ionela-voinescu: Sorry! I did not make myself clear. I did not mean to say that the calls don't return anything but that you can't see anything being printed as a result of them being run. I suppose printing the docstrings could work but what I meant to say is that it would be nice to have a section where you can see more details on the test being run. Details on the workload, the data frames obtained, something that could represent the basis of a debugging process.

tests/eas/generic.py
ionela-voinescu: I thought generic meant more that it would hold the base EnergyModelTest class, which then can be extended in specific test classes. I find it more clear to have the EnergyModelTest class in its own file but it's just a preference.

ipynb/tests/Generic_EAS_Tests.ipynb
bjackman: I don't intend for this to be for debugging the *tests*, after all they are absolutely flawless in every single way. Rather it's intended as a starting point for looking at what element of the kernel behaviour was judged to be faulty (since when you have a failure from a lisa-test run it's pretty much "computer says no").

Basically the only thing I mean for this to show is a plot of the workload task residency and a plot of the energy estimations, so you can see which part of the workload run was judged to be mis-scheduled. I don't imagine users diving python in order to do that, but rather just taking a look and seeing "oh yeah this 5% task ran on a big CPU at t=3.5" then going back to the kernel to figure out why that happened.

Nonetheless I'll add some more english to explain what is going on here.

bjackman:
OK I've pushed a new version where I've tried to address your comments:

- Fixing the minor stuff
- I added some more text to the notebook to explain what's going on. Also some of the calls that you commented on actually weren't necessary so I removed them. Hopefully it's a bit more easy to understand (assuming people actually read my walls of text :grimacing:)
- Fixed the order of the columns in power DataFrames (For juno you get [(1), (2), (1, 2), (0), (3), (4), (5), (0, 3, 4, 5)] i.e. LITTLEs followed by LITTLE cluster then bigs followed by big cluster).
- Fixed the remaining `bl` references in `acceptance.py`.

ipynb/tests/Generic_EAS_Tests.ipynb
```diff
@@ -0,0 +1,585 @@
...
+   "metadata": {
+    "collapsed": false,
+    "scrolled": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Maximum estimated system energy: 1725\n",
```
ionela-voinescu: These are no longer printed (based on a commit merged a short while ago).

ipynb/tests/Generic_EAS_Tests.ipynb
```diff
@@ -0,0 +1,585 @@
...
...000076741, 6.8794780000025639], [6.9073580000112997, 6.907457000008435], [6.9077930000057677, 6.9078500000032363], [6.907881000006455, 6.9084830000065267], [6.9091150000022026, 6.9091620000108378], [6.9093140000040876, 6.9095340000058059], [6.9097400000027847, 6.9104690000094706]], \"2\": [[6.7729960000142455, 6.7788480000017444], [6.7816750000056345, 6.7918920000083745], [6.7936480000062147, 6.7936680000129854], [6.793736000006902, 6.8166610000043875], [6.8243340000044554, 6.824374000003445], [6.9012460000085412, 6.9013400000112597]]}, \"sshd-3277\": {\"1\": [[6.9033190000045579, 6.9057820000016363]], \"2\": [[6.9023140000062995, 6.9032010000082664]]}, \"kworker/u12:2-3279\": {\"2\": [[6.9329410000063945, 6.9329740000102902]], \"4\": [[6.9330630000040401, 6.9348960000061197]]}, \"kworker/u12:2-3276\": {\"1\": [[6.7792940000072122, 6.7793430000019725]], \"3\": [[6.7795090000145137, 6.7814940000098431]]}, \"sshd-3278\": {\"0\": [[7.0153290000016568, 7.015495000014198], [7.0179130000033183, 7.0384260000137147], [7.0387300000002142, 7.0392610000126297], [7.0395130000106292, 7.0397250000096392], [7.0401260000071488, 7.0594920000003185], [7.0596900000091409, 7.0597150000103284], [7.0604
```
ionela-voinescu: .. is higher than the **ideal energy plot**.. 
Although a reword of it might work: "Where the values for the observed energy are higher than the ideal energy values..". The same for the following sentence. It's strange to say "where the plot is higher". Maybe a better wording would be "where the plot line is higher.. ".

ionela-voinescu: Awesome! :D

ipynb/tests/Generic_EAS_Tests.ipynb
```diff
@@ -0,0 +1,585 @@
...
+   ],
+   "source": [
+    "%load_ext autoreload\n",
+    "%autoreload 2\n",
+    "%matplotlib inline\n",
+    "\n",
+    "import logging\n",
+    "from conf import LisaLogging\n",
+    "LisaLogging.setup()\n",
+    "logging.getLogger().setLevel(logging.WARNING)\n",
```
ionela-voinescu: Isn't this the same as saying: LisaLogging.setup(level=logging.WARNING)?

tests/eas/generic.py
```diff
@@ -0,0 +1,450 @@
...
+    @experiment_test
+    def test_slack(self, experiment, tasks):
+        self._test_slack(experiment, tasks)
+    @experiment_test
+    def test_task_placement(self, experiment, tasks):
+        self._test_task_placement(experiment, tasks)
+
+class EnergyModelWakeMigration(EnergyModelTest):
+    """
+    Test EAS for tasks alternating beetween 10% and 50%
```
ionela-voinescu: "Test EAS for tasks alternating beetween 10% and 50% over 4 seconds". It would be good to specify the duration of the entire workload as for the time axis of the plot to be better understood. The same for the other classes. Some of them include this information but some of them do not.

ipynb/tests/Generic_EAS_Tests.ipynb
bjackman: Yep, will change.

ipynb/tests/Generic_EAS_Tests.ipynb
bjackman: Good spot. Good point about "plot" vs "values" will reword this.

bjackman:
Updated

libs/utils/platforms/hikey_energy.py
```diff
@@ -0,0 +1,68 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from energy_model import (ActiveState, EnergyModelNode, EnergyModelRoot,
+                          PowerDomain, EnergyModel)
+
+from collections import OrderedDict
+
+cluster_active_states = OrderedDict([
+    (208000, ActiveState(capacity=178, power=16)),
```
bjackman: Sure. I'm a little bit awkward about doing this because for some reason my PEP8 religion forbids it but TBH it is clearer so let's apply it here.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+        return max(s.capacity for s in self.active_states.values())
+
+class EnergyModelRoot(EnergyModelNode):
+    """
+    Convenience class for root of an EnergyModelNode tree.
+
+    Just like EnergyModelNode except that ``active_states`` and ``idle_states``
+    aren't required.
+    """
+    def __init__(self, active_states=None, idle_states=None, *args, **kwargs):
```
bjackman: Sure, it seems clearer.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+        3. Temperature is ignored entirely
+
+    .. _util-distrib:
+
+    .. admonition:: Utilization Distributions
+
+        Used throughout this module: A ``util_distrib`` is a list ``u`` where
+        ``u[N]`` is the sum of the frequency-invariant, capacity-invariant
+        utilization of tasks placed on CPU N. That is, the quantity represented
+        by a CPU runqueue's util_avg in the Linux kernel scheduler's
```
bjackman: Good point, will add this note.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+
+        - 4 idle states: "WFI", "cpu-sleep", "cluster-sleep-0" and "cluster-sleep-1"
+
+        Then here are some example inputs and outputs:
+
+        ::
+
+          # All CPUs idle:
+          [0, 0, 0, 0] -> ["cluster-sleep-0", "cluster-sleep-0",
+                           "cluster-sleep-0", "cluster-sleep-0"]
```
bjackman: Yep, good catch.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+        enter that domain's deepest idle state. If any CPU in a domain has work,
+        no CPUs in that domain are assumed to enter any domain shared state.
+
+        e.g. Consider a system with
+
+        - two power domains PD0 and PD1
+
+        - 4 CPUs, with CPUs [0, 1] in PD0 and CPUs [2, 3] in PD1
+
+        - 4 idle states: "WFI", "cpu-sleep", "cluster-sleep-0" and "cluster-sleep-1"
```
bjackman: Good point, will add this.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+          [1, 1, 1, 1] -> ["WFI","WFI","WFI", "WFI"]
+
+          # One power domain active, the other idle
+          [0, 0, 1, 1] -> ["cluster-sleep-1", "cluster-sleep-1", "WFI","WFI"]
+
+          # One CPU active.
+          # Note that CPU 2 has no work but is assumed to never be able to enter
+          # any "cluster" state.
+          [0, 0, 0, 1] -> ["cluster-sleep-1", "cluster-sleep-1",
+                           "cpu-sleep","WFI"]
```
bjackman: No:

> If a CPU has any tasks it is estimated that it may only enter its
        shallowest idle state in between task activations

And in line 346 all the CPUs have tasks so can only enter WFI

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+            [cpu] = node.cpus
+            required_cap = util_distrib[cpu]
+
+            possible_freqs = [f for f, s in node.active_states.iteritems()
+                              if s.capacity >= required_cap]
+
+            if possible_freqs:
+                ideal_freqs[cpu] = min(possible_freqs)
+            else:
+                # CPU cannot provide required capacity, use max freq
```
bjackman: No, I don't agree that this can never happen - the `util_distrib` is not capped (and IIUC `util_avg` in a trace can be greater than the current capacity).

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+                ideal_freqs[cpu] = max(node.active_states.keys())
+                overutilized = True
```
bjackman: Hmm.. `overutilized` here isn't really the same thing as it is in EAS. I think adding an 80% limit would be a bit arbitrary as this is supposed to be  a theoretical estimation, whereas EAS's 80% is a limit to respect the fact that our util estimation is imperfect. I would prefer to add it as a parameter somewhere, but in a later iteration.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+
+        If combine=False, return idle and active power as separate components.
+        """
+        power = 0
+        ret = {}
+
+        assert all(0.0 <= a <= 1.0 for a in cpu_active_time)
+
+        for node in self.root.iter_nodes():
+            if not node.active_states or not node.idle_states:
```
bjackman: Yep, will add this.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+                continue
+
+            cpus = tuple(node.cpus)
+            # For now we assume topology nodes with energy models do not overlap
+            # with frequency domains
```
bjackman: Sure, I've added a check to `__init__` but I'm also going to leave this comment and assert here, if that sounds good to you.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+            freq = freqs[cpus[0]]
+            assert all(freqs[c] == freq for c in cpus[1:])
+
+            # The active time of a node is estimated as the max of the active
+            # times of its children.
+            # This works great for the synthetic periodic workloads we use in
+            # LISA (where all threads wake up at the same time) but is probably
+            # no good for real workloads.
+            active_time = max(cpu_active_time[c] for c in cpus)
```
bjackman: Yeah, I actually have that on my other branch: https://github.com/bjackman/lisa/blob/energy-model-rfc-v2-evaluation/libs/utils/energy_model.py#L527 (ctrl+F util_aggregator) but decided to avoid the complexity here.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+            active_power = node.active_states[freq].power * active_time
+
+            _idle_power = max(node.idle_states[idle_states[c]] for c in cpus)
+            idle_power = _idle_power * (1 - active_time)
+
+            if combine:
```
bjackman: Hmm, I actually thought I had removed this parameter completely, it seemed useful at first but in the end I never ended up using it, would that be OK with you?

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+        :returns: Dict with power in bogo-Watts (bW), with contributions from
+                  each system component keyed with a tuple of the CPUs
+                  comprising that component (i.e. :attr:EnergyModelNode.cpus)
+
+                  ::
+
+                    {
+                        (0,)    : 10,
+                        (1,)    : 10,
+                        (1, 2)  : 5,
```
bjackman: You are not wrong. Thanks again for all the checking of these examples.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+                    }
+
+                  This represents CPUs 0 and 1 each using 10bW and their shared
+                  resources using 5bW for a total of 25bW.
+        """
+        if len(util_distrib) != len(self.cpus):
```
bjackman: > What about `cpus_utilizations`?
Hmm, yeah I think that would be a sensible name.

> it seems that you want to support the concept of variable workloads using "a series of util_distrib" ... but from this check this seems not to be supported right now, sin't it?
The implication of using a series of `util_distrib`s was that you'd call this function for each of them; that's what is done in generic.py (see `get_power_df`). Perhaps I can reword to make that clearer.

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+            This is a brute force search taking time exponential wrt. the number
+            of tasks.
+
+        :param capacities: Dict mapping tasks to expected utilization
+                           values. These tasks are assumed not to change; they
+                           have a single static utilization value. A set of
+                           single-phase periodic RT-App tasks is an example of a
+                           suitable workload for this model.
+        :returns: List of ``util_distrib`` s representing distribution of work
+                  under optimal task placments see
```
bjackman: mm.. now I'm thinking about....

:balloon:  [**CI FOR SPELLING**](https://sphinxcontrib-spelling.readthedocs.io/en/latest/) :balloon:

libs/utils/energy_model.py
```diff
@@ -0,0 +1,572 @@
...
+
+        candidates = {}
+        for cpus in product(self.cpus, repeat=len(tasks)):
+            placement = {task: cpu for task, cpu in zip(tasks, cpus)}
+
+            util = [0 for _ in self.cpus]
+            for task, cpu in placement.items():
+                util[cpu] += capacities[task]
+            util = tuple(util)
+
```
bjackman: Nice, thanks!

tests/lisa/test_energy_model.py
```diff
@@ -0,0 +1,253 @@
...
+little_cluster_active_states = OrderedDict([
+    (1000, ActiveState(power=10)),
+    (2000, ActiveState(power=20)),
+])
+
+little_cluster_idle_states = OrderedDict([
+    ('WFI',              5),
+    ('cpu-sleep-0',      5),
+    ('cluster-sleep-0',  1),
+])
```
bjackman: Sounds good.

I did also implement a [method to print various observations about the energy model](https://github.com/bjackman/lisa/blob/energy-model-rfc-v2-evaluation/libs/utils/energy_model.py#L793) but didn't include it here.

tests/lisa/test_energy_model.py
```diff
@@ -0,0 +1,253 @@
...
+        s2 = set([tuple(l) for l in l2])
+        self.assertSetEqual(s1, s2)
+
+    def test_single_small(self):
+        placements = em.get_optimal_placements({'task0': 1})
+        self.assertPlacementListEqual(placements, [[1, 0, 0, 0],
+                                                   [0, 1, 0, 0]])
+
+    def test_single_big(self):
+        placements = em.get_optimal_placements({'task0': 350})
```
bjackman: Yeah, this was a deliberate decision at the beginning of writing it, because I thought if I expressed it in terms of the data I'd probably make the exact same mistakes as I'd made in the code under test. So I wanted to work them out by hand. 

However I agree that as the code grew it didn't start to look like such a smart idea. I'm going to keep it as it is for now, but add a comment. Probably if we need to extend these tests it will be faster to add new ones in terms of the data rather than go back and re-factor the hard coded stuff.

derkling:
On 23-Feb 07:20, Brendan Jackman wrote:
> bjackman commented on this pull request.
> 
> 
> 
> > +            # with frequency domains
> +            freq = freqs[cpus[0]]
> +            assert all(freqs[c] == freq for c in cpus[1:])
> +
> +            # The active time of a node is estimated as the max of the active
> +            # times of its children.
> +            # This works great for the synthetic periodic workloads we use in
> +            # LISA (where all threads wake up at the same time) but is probably
> +            # no good for real workloads.
> +            active_time = max(cpu_active_time[c] for c in cpus)
> +            active_power = node.active_states[freq].power * active_time
> +
> +            _idle_power = max(node.idle_states[idle_states[c]] for c in cpus)
> +            idle_power = _idle_power * (1 - active_time)
> +
> +            if combine:
> 
> Hmm, I actually thought I had removed this parameter completely, it seemed useful at first but in the end I never ended up using it, would that be OK with you?

You mean if it's ok to remove it?
I personally like the idea to have a split version... that's why I was
just proposing to use a namedtuple to get rid of the parameter but
still keep the active and idle contributions available.

> 
> -- 
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/ARM-software/lisa/pull/263#discussion_r102735588

-- 
#include <best/regards.h>

Patrick Bellasi

libs/utils/energy_model.py
derkling: Right: actually the tracepoint reports the "raw" RQ's value:
https://android.googlesource.com/kernel/common/+/android-3.18/include/trace/events/sched.h#690

... thus yes, for example on task migrations, it can be bigger than the current capacity ;-)

libs/utils/energy_model.py
derkling: OK, let's add it in future if we need to breakdown of the contributions...

PR #265, opened 2016-12-23
gem5 integration in lisa

avanlaer:
gem5: Added an if-statement in board configuration, to allow for the setup of a gem5 platform

This needs to be combined with changes in devlib (right now, commit 9b41c47 in github.com/avanlaer/devlib but more commits might follows).
The extra code needed in lisa might change, if the underlying code in devlib changes but this might give you an idea.

 Changes to be committed:
	modified:   libs/utils/env.py

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -379,6 +379,14 @@ class TestEnv(ShareState):
             platform = Platform(model='MT8173')
             self.__modules = ['bl', 'cpufreq']
 
+        # Initialize gem5 simulation as a board
+        elif self.conf['board'].upper() == 'GEM5':
+            platform = devlib.platform.gem5.Gem5SimulationPlatform(name='gem5',
+                       gem5_bin=self.conf['gem5_bin'],
+                       gem5_args=self.conf['gem5_args'],
+                       gem5_virtio=self.conf['gem5_virtio'],
+                       host_output_dir=self.conf['host_output_dir'],
+                       gem5_telnet_port=self.conf['gem5_telnet_port'] if 'gem5_telnet_port' in self.conf else None)
         elif self.conf['board'] != 'UNKNOWN':
             # Initilize from platform descriptor (if available)
             board = self._load_board(self.conf['board'])
@@ -860,7 +868,6 @@ class TestEnv(ShareState):
                 self._log.warning('Using pre-installed DTB')
             else:
                 self.tftp_deploy(tc['dtb'])
-
         else:
             raise ValueError('Kernel installation method not supported')
 


libs/utils/env.py
```diff
@@ -379,6 +379,14 @@ def _init_target(self, force = False):
             platform = Platform(model='MT8173')
             self.__modules = ['bl', 'cpufreq']
 
+        # Initialize gem5 simulation as a board
+        elif self.conf['board'].upper() == 'GEM5':
+            platform = devlib.platform.arm.Gem5SimulationPlatform(name='gem5',
```
ionela-voinescu: This has now changed to be `devlib.platform.gem5.Gem5SimulationPlatform`.

avanlaer:
I've updated the commit to account for the change in devlib.

libs/utils/env.py
```diff
@@ -379,6 +379,14 @@ def _init_target(self, force = False):
             platform = Platform(model='MT8173')
             self.__modules = ['bl', 'cpufreq']
 
+        # Initialize gem5 simulation as a board
+        elif self.conf['board'].upper() == 'GEM5':
+            platform = devlib.platform.gem5.Gem5SimulationPlatform(name='gem5',
+                       gem5_bin=self.conf['gem5_bin'],
+                       gem5_args=self.conf['gem5_args'],
+                       gem5_virtio=self.conf['gem5_virtio'],
+                       host_output_dir=self.conf['host_output_dir'],
```
derkling: Cannot this be set to the already defined `res_dir`?

libs/utils/env.py
```diff
@@ -379,6 +379,14 @@ def _init_target(self, force = False):
...
+                       gem5_telnet_port=self.conf['gem5_telnet_port'] if 'gem5_telnet_port' in self.conf else None)
```
derkling: Better:

```python
self.conf.get('gem5_telnet_port', None)
```

libs/utils/env.py
```diff
@@ -860,7 +868,6 @@ def install_kernel(self, tc, reboot=False):
                 self._log.warning('Using pre-installed DTB')
             else:
                 self.tftp_deploy(tc['dtb'])
-
```
derkling: Not part of this commit?

libs/utils/env.py
```diff
@@ -379,6 +379,14 @@ def _init_target(self, force = False):
             platform = Platform(model='MT8173')
             self.__modules = ['bl', 'cpufreq']
 
+        # Initialize gem5 simulation as a board
+        elif self.conf['board'].upper() == 'GEM5':
+            platform = devlib.platform.gem5.Gem5SimulationPlatform(name='gem5',
+                       gem5_bin=self.conf['gem5_bin'],
```
derkling: What abut using a dictionary in target.conf

```python
my_conf = {
    ....
    "board" = "gem5",
    "gem5" = {
        "bin" = "...",
        "virtio" = "...",
        "args" = "...",
    }
```

And perhaps we can also "force" the things by assuming that if you have 'gem5' in conf then this is the board to use... but let see what @bjackman think about that ;-)

libs/utils/env.py
bjackman: Or just `self.conf.get('gem5_telnet_port')` as `dict.get` has a `None` default by default

libs/utils/env.py
bjackman: Agree that we should have a 'gem5' subdict and not 'gem5_*'.

It's unfortunate to have to have a generic "gem5" board as this board is special and violates assumptions that you can make about the other boards (e.g. if you see "pixel" you know it's an arm64 board with topology XYZ etc, but if you see "gem5" it could be anything, you need to examine the config further). 

It would be simpler to instead have a fixed set of gem5 boards like "gem5_myplatform", "gem5_myotherplatform" and require adding their parameters somewhere in the LISA source. However gem5 boards are naturally going to be a) ephemeral experimental HW designs or b) confidential future HW designs and neither of these cases plays nicely with coding simulation parameters into the LISA source. 

So let's keep it this way - I just wanted to share my thinking; we'll need to be aware that there's a new "special" type of board and try to find any assumptions in the code that this violates.

ionela-voinescu:
Thank  you @avanlaer for this pull request. We've created another pull request at https://github.com/ARM-software/lisa/pull/382 and we'll continue reviewing gem5 support there.

PR #269, opened 2017-01-12
Factor out dependency installation script

bjackman:
This provides an installation script that can be used to set up VMs and containers on Ubuntu-like systems. Then modifies the Vagrantfile to use that script.

I've kept this as 3 commits so that I can keep track of the changes from the script I was given by another team, but I'm happy to squash before merging.

### DIFF
--- a/install_base_ubuntu.sh
+++ b/install_base_ubuntu.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+# Script to install the depenencies for LISA on an Ubuntu-like system.
+
+# This is intended to be used for setting up containers and virtual machines to
+# run LISA (e.g. for CI infrastructure or for Vagrant installation). However for
+# the brave, it could be used to set up LISA directly on a real machine.
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+usage() {
+    echo Usage: "$0" [--install-android-sdk]
+}
+
+set -eu
+
+install_android_sdk=n
+
+for arg in "$@"; do
+    if [ "$arg" == "--install-android-sdk" ]; then
+        install_android_sdk=y
+    else
+        echo "Unrecognised argument: $arg"
+        usage
+        exit 1
+    fi
+done
+
+apt-get update
+
+apt-get -y remove ipython ipython-notebook
+
+apt-get -y install build-essential autoconf automake libtool pkg-config \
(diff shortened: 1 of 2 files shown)


install_base_ubuntu.sh
```diff
@@ -0,0 +1,128 @@
...
+    # add more package names here, separated by space
+    pip_pkgs="Cython trappy bart-py devlib psutil wrapt ipython jupyter"
+
+
+    # check root permision
+    # No need to do this on LAVAv2 LXC container
+    # check_root
+
+    # remove some packages
+    for package in $remove_pkgs
```
derkling: Are not these loops a bit overkilling?

Why not just a simple:

```bash
sudo apt-get -y remove $remove_pkgs
sudo apt-get -y install $pre_requisite_pkgs
```

The same for pip installed packages...

bjackman: Agreed.

bjackman:
Pushed a new version, I simplified it a bit but then also complicated it a bit...

derkling:
Can we squash the first two patches together?

bjackman:
OK, pushed a squashed version

bjackman:
Uhh.. meant to only squash 2. Take two...

install_base_ubuntu.sh
```diff
@@ -0,0 +1,56 @@
...
+apt-get -y remove $remove_pkgs
+
+apt-get -y install $pre_requisite_pkgs
+
+yes w | pip install $pip_pkgs
+
+if [ "$install_android_sdk" == y ]; then
+    apt-get -y install openjdk-7-jre openjdk-7-jdk
+    ANDROID_SDK_URL="https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz"
+    if [ ! -e ./tools/android-sdk-linux ]; then
```
ionela-voinescu: Using this as a relative path does not work for calls like: sudo ./lisa/install_base_ubuntu.sh.

bjackman:
OK pushed a version with the `pip install --upgrade packaging appdirs`.

bjackman:
:weary: 

New version with

- The `pip install --upgrade packaging appdirs` nonsense
- Got rid of the `--fast` thing and just set it to use the latest pip, this avoids compiling anything. But to make this work in the Vagrant provisioning environment where the `$PATH` is not suitable, we need to explicitly use `/usr/local/bin/pip`.

PR #271, opened 2017-01-18
tests/eas: Add heavy load spreading test

bjackman:


tests/eas/heavy_load.py
```diff
@@ -0,0 +1,96 @@
...
+    Runs N 100% RT-App threads where N is the number of CPUs, and checks that
+    those tasks were spread across all CPUs in the system by asserting that all
+    CPUs were fully utilized up until the first task completed.
+    """
+
+    test_conf = {
+        'ftrace' : {
+            'events' : ['cpu_idle', 'sched_switch', 'cpu_frequency'],
+        },
+        'modules' : ['cgroups'],
```
derkling: Do we really need that module?

derkling: By removing it, I've noticed that it's required by the executor to freeze the userspace.
Are we know enforcing that behaviour for all tests?... I think we should have a fallback mode for all the cases where, for example, a platform does not support CGroups and/or the freezer is not available...

Otherwise, if it's really required, we should find a way to enable this module by default at executor level (i.e. adding it to the list of modules to be loaded) thus avoiding some "mandatory" lines in the tests sources.

tests/eas/heavy_load.py
```diff
@@ -0,0 +1,96 @@
...
+                            # Create one task for each cpu
+                            "tasks" : "cpus",
+                        },
+                    },
+                },
+            },
+        },
+        "confs" : [{
+            'tag' : 'energy_aware',
+            'flags' : ['ftrace', 'freeze_userspace'],
```
derkling: ok... here is where we know that we need the cgroups module.

tests/eas/heavy_load.py
derkling: Perhaps we can use the `freeze_userspace` flag to automatically add the cgroups module from within the executor?

tests/eas/heavy_load.py
```diff
@@ -0,0 +1,96 @@
...
+            cpu_active = cpu_active.reindex(new_index, method='ffill')
+
+            active_time = trace.integrate_square_wave(cpu_active[start:end])
+            active_proportions.append(active_time / duration)
+
+        if any(a < (REQUIRED_CPU_ACTIVE_TIME_PCT / 100.)
+               for a in active_proportions):
+            raise AssertionError(
+                "Some CPUs were less than {}% utilized\n"
+                " CPU active proportions: {}".format(
```
derkling: What about a more user-friendly formatting?

```python
        if any(a < (REQUIRED_CPU_ACTIVE_TIME_PCT / 100.)
               for a in active_proportions):
            proportions = ""
            for cpu, _ in enumerate(self.target.core_names):
                proportions += " {:3d} {:5.1f}%\n"\
                        .format(cpu, active_proportions[cpu]*100)
            raise AssertionError(
                "Some CPUs were less than {}% utilized\n"
                " CPU active proportions:\n{}".format(
                    REQUIRED_CPU_ACTIVE_TIME_PCT, proportions))
```
Which prints:

```bash
AssertionError: Some CPUs were less than 99% utilized
 CPU active proportions:
   0 100.0%
   1 100.0%
   2 100.0%
   3  99.9%
   4  99.9%
   5  12.2%

        Check trace file: /data/Code/lisa/results/20170118_113343/rtapp:energy_aware:n_heavy_tasks/1/trace.dat
```

tests/eas/heavy_load.py
```diff
@@ -0,0 +1,96 @@
...
+    Test an EAS system under heavy load
+
+    Runs N 100% RT-App threads where N is the number of CPUs, and checks that
+    those tasks were spread across all CPUs in the system by asserting that all
+    CPUs were fully utilized up until the first task completed.
+    """
+
+    test_conf = {
+        'ftrace' : {
+            'events' : ['cpu_idle', 'sched_switch', 'cpu_frequency'],
```
derkling: Perhaps we do not need `cpu_frequency`?
It seems to work without for me...

bjackman: Yep, will remove.

tests/eas/heavy_load.py
bjackman: Yeah, I've been thinking about this, I think there might be a problem because when the Executor is constructed the devlib target has already been setup (via TestEnv), and I don't know how easy it is to add devlib modules at a later time.

I will investigate, but in the meantime how about I just add a comment here saying "Required for freeze_userspace"?

tests/eas/heavy_load.py
bjackman: Looks great, I'll add it.

tests/eas/heavy_load.py
bjackman: The reason I added this is I wanted to use this test from a notebook and plot the frequencies. At some point I'll have to find a way to run a LisaTest from a notebook but provide additional events when you set it up.

bjackman:
OK thanks for the quick review. Pushed a version with the comments addressed.

derkling:
Let's keep the idea to use the `freeze_userspace` flag as a way to force the load of the cgroup module from the executor... if it makes sense, maybe we should open an issue to keep track of that idea?

bjackman:
Issue: https://github.com/ARM-software/lisa/issues/273

PR #277, opened 2017-01-24
Env cleanup

bjackman:
Some cleanups and simplifications for TestEnv.

Generally:

- Simplify a load of conditionals by 
  - Removing the unnecessary case where `conf`/`test_conf` is `None`
  - Using `d.get('foo')` instead of `if 'foo' in d and...`
- Use set operations to simplify stuff

I think this also has a sneaky bugfix for installing rt-app dependencies via `install_tools` but I haven't replicated that presumed bug.

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -92,8 +92,8 @@ class TestEnv(ShareState):
         if self._initialized and not force_new:
             return
 
-        self.conf = None
-        self.test_conf = None
+        self.conf = {}
+        self.test_conf = {}
         self.res_dir = None
         self.target = None
         self.ftrace = None
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -97,7 +97,6 @@ class TestEnv(ShareState):
         self.target = None
         self.ftrace = None
         self.workdir = WORKING_DIR_DEFAULT
-        self.__tools = []
         self.__installed_tools = set()
         self.__modules = []
         self.__connection_settings = None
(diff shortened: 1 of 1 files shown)


bjackman:
Have replicated the bug that this fixes.

libs/utils/env.py
```diff
@@ -395,29 +395,18 @@ def _init_target(self, force = False):
...
-                if module in self.__modules:
-                    self.__modules.remove(module)
-        # Remove tests specific modules
-        if self.test_conf and 'exclude_modules' in self.test_conf:
-            for module in self.test_conf['exclude_modules']:
-                if module in self.__modules:
-                    self.__modules.remove(module)
+        remove_modules = set(self.conf.get('exclude_modules', []) +
+                             self.test_conf.get('exclude_modules', []))
+        modules = modules.difference(remove_modules)
```
ionela-voinescu: Here it might be worth using modules.difference_update as it updates modules to have the difference between the two and it does not return a new set with the difference.

libs/utils/env.py
```diff
@@ -493,10 +476,16 @@ def install_tools(self, tools):
         :param tools: The list of names of tools to install
         :type tools: list(str)
         """
+
+        # Add tools dependencies
+        if 'rt-app' in tools:
+            tools += ['taskset', 'trace-cmd', 'perf', 'cgroup_run_into.sh',]
+
+        # Remove duplicates and already-instaled tools
+        tools = set(tools).difference(self.__installed_tools)
```
ionela-voinescu: The same here, by making tools a set first and then updating it with the difference.

libs/utils/env.py
bjackman: Yep

bjackman:
Pushed a version using `difference_update`.

PR #278, opened 2017-01-25
trace: Handle new field format for sched_energy_diff

bjackman:
There is a new format for the sched_energy_diff trace event in EAS code in the
Android common kernel:

https://android.googlesource.com/kernel/common/+/6aab3e4a9266fb661638e37ad4745573aff6de5c%5E%21/

Add support for parsing this format in Trace, by translating equivalent field
names and making unmatched fields optional.

-----

I have only done the bare minimum to allow me to use LISA to parse traces from kernels containing these events. I haven't checked that the result of this method makes any sense, neither with the existing events nor the new.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -610,6 +610,8 @@ class Trace(object):
         """
         If a energy model is provided, some signals are added to the
         sched_energy_diff trace event data frame.
+
+        Also convert between existing field name formats for sched_energy_diff
         """
         if not self.hasEvents('sched_energy_diff') \
            or 'nrg_model' not in self.platform:
@@ -629,6 +631,10 @@ class Trace(object):
             "Maximum estimated system energy: {0:d}".format(power_max))
 
         df = self._dfg_trace_event('sched_energy_diff')
+
+        translations = {'nrg_d': 'nrg_diff', 'utl_d': 'usage_delta'}
+        df.rename(columns=translations, inplace=True)
+
         df['nrg_diff_pct'] = SCHED_LOAD_SCALE * df.nrg_diff / power_max
 
         # Tag columns by usage_delta


bjackman:
Don't merge yet, I tested it on the wrong trace.

bjackman:
OK tested it properly (and fixed it because it was of course broken)

libs/utils/trace.py
```diff
@@ -629,6 +631,13 @@ def _sanitize_SchedEnergyDiff(self):
             "Maximum estimated system energy: {0:d}".format(power_max))
 
         df = self._dfg_trace_event('sched_energy_diff')
+
+        translations = {'nrg_d': 'nrg_diff', 'utl_d': 'usage_delta'}
+        if any(column in df for column in translations):
+            for new_name, old_name in translations.items():
+                df[old_name] = df[new_name]
+                del df[new_name]
+
```
derkling: Here you are copying data... talking about optimising DataFrame usage, perhaps something like that should work:  

```python
translations = {'nrg_diff' : 'nrg_d', 'usage_delta' : 'utl_d'}
if any(column in df for column in translations):
    df.rename(columns=translations, inplace=True)
```

libs/utils/trace.py
```diff
@@ -638,11 +647,12 @@ def _sanitize_SchedEnergyDiff(self):
             ['< 150', '< 400', '< 600'], '>= 600')
 
         # Tag columns by nrg_payoff
-        ccol = df.nrg_payoff
-        df['nrg_payoff_group'] = np.select(
-            [ccol > 2e9, ccol > 0, ccol > -2e9],
-            ['Optimal Accept', 'SchedTune Accept', 'SchedTune Reject'],
-            'Suboptimal Reject')
+        if 'nrg_payoff' in df:
```
derkling: Can `nrg_payoff` really be missing?
It should be available on both the old and new format, isn't it?

In any case, if we don't add the `nrg_payoff_group` columns, the plot for the "PE filtering" space will be broken.

bjackman: Ah yeah looks like this hunk leaked in by accident, will remove it.

bjackman:
OK pushed a new version using `df.rename` - thanks!

PR #281, opened 2017-01-30
Workload: fix docstring for `ftrace` param

bjackman:
- This parameter should be a Devlib object for collecting traces, not a Trappy
  object for parsing them.

- Add note to say where you can get one of these things from.


bjackman:
Trivial fix, merging directly

PR #287, opened 2017-02-06
lisa_shell: Pass all arguments to nosetests in lisa-test

bjackman:
This allows extra arguments to be passed to nosetests to customise its behaviour
beyond what LISA provides.


ionela-voinescu:
Looks good, although you might want to update `_lisa-test-usage` too.

bjackman:
Pushed version with `usage` udpate

PR #289, opened 2017-02-10
tests/eas: Delete empty 'rfc' test file

bjackman:


derkling:
This test is still something we want to keep on hand, I've just fixed it to match the new executor interface with a patch in #295

PR #285, opened 2017-02-01
Cleanups, fixes, documentation and tests for wlgen

bjackman:
- Provides some cleanups and minor fixes for the `wlgen` package.
- Adds some docstrings describing the API.
- Adds very simple tests for the `RTA` and `PerfMessaging` classes, using a LocalLinuxTarget to avoid requiring a connected external target. This approach has limitations because we don't want to give LISA root on a real machine (which is required e.g. for calibrating rt-app), but useful tests can be done that work around this limitation. Hopefully this could allow us to start self-testing parts of LISA in a CI loop.

PR #293, opened 2017-02-14
Cleanups & Fixes for Workload class

bjackman:


bjackman:
This is split out from https://github.com/ARM-software/lisa/pull/285

On that PR, there was [comment](https://github.com/ARM-software/lisa/pull/285#discussion_r99907146) about the use of `self.target.execute('mkdir -p ...` which I agreed to replace with `os.makedirs`, but actually there isn't a `makedirs` method for devlib targets.

ionela-voinescu:
Looks good to me, but I did not got yet thought the changes at #285 .

PR #296, opened 2017-02-14
Cleanups for RTA

bjackman:


bjackman:
This is split out from https://github.com/ARM-software/lisa/pull/285#discussion_r99907146

PR #297, opened 2017-02-14
Sphinx documentation for Workload & RTA classes

bjackman:


bjackman:
This is split out from https://github.com/ARM-software/lisa/pull/285#discussion_r99907146

libs/wlgen/wlgen/rta.py
```diff
@@ -483,32 +498,41 @@ def conf(self,
...
 
-          params := {task, ...}
-          task   := NAME : {SCLASS, PRIO, [phase, ...]}
-          phase  := (PTIME, PRIOD, DCYCLE)
+        :param kind: Either 'custom' or 'profile' - see above.
+        :param params: RT-App parameters - see above.
+        :param duration: Maximum duration of the workload in seconds. Any
+                         remaining tasks are killed by rt-app when this time has
+                         elapsed.
+        :param cpus: CPUs to restrict this workload to, using ``taskset``.
```
derkling: Is it not possible with sphinx syntax to specify the type of a parameter.
For example in this case it would be useful to tell the user which format has to be used to represent the CPUs mask, e.g. a str or a list(int) ?

libs/wlgen/wlgen/rta.py
```diff
@@ -709,30 +737,25 @@ def __init__(self, start_pct=100, end_pct=0, time_s=1, period_ms=100,
...
-        Configure a periodic load.
+    This class defines a task which load is periodic with a configured
+    period and duty-cycle.
 
-        This class defines a task which load is periodic with a configured
-        period and duty-cycle.
+    :param duty_cycle_pct: the load percentage.
+    :param duration_s: the total duration in seconds of the task.
+    :param period_ms: the period used to define the load in milliseconds.
+    :param delay_s: the delay in seconds before starting the workload.
```
derkling: Technically the workload is started immediately... this is better:
"the delay in seconds before starting the periodic phase"

libs/wlgen/wlgen/rta.py
```diff
@@ -43,11 +43,20 @@ class Phase(_Phase):
...
+    Class for creating RT-App workloads
+    """
 
     def __init__(self,
                  target,
                  name,
                  calibration=None):
+        """
+        :param target: Devlib target to run workload on.
+        :param name: Human-readable name for the workload
```
derkling: Just to be picky... you almost always use a full stop at the end of each param description.
It's up to you if you want to be consistent... otherwise for me it's ok like that as well.
You already did a lot on the documentation side ;-)

derkling:
Few comments... waiting just for a response and than it's ok to merge for me ;-)

libs/wlgen/wlgen/rta.py
bjackman: I will try to add full stops where I've missed them (not going to spend too much time worrying about this though TBH). I guess the more correct thing would actually be no full stop, because they aren't really sentences. Anyway consistency is nicer than correctness here.

libs/wlgen/wlgen/rta.py
bjackman: Yep, will add this here. I've not added it everywhere though because it's often inaccurate (e.g. you say `int` but actually any number is valid).

libs/wlgen/wlgen/rta.py
bjackman: Good point, updating.

libs/wlgen/wlgen/rta.py
derkling: If you want to remove them all around, I'm all it for that better me too ;-)
However, as I said... don't spent too much time on that ;-)

bjackman:
OK, updated responding to comments.

I guess I should note that this isn't complete documentation and also is likely to contain ReST syntax errors (I have built it but have since changed some things). I'm not worried about the latter, I'll just submit fixes when I get round to submitting the doc-build infrastructure.

PR #298, opened 2017-02-14
Wglen self-tests v2

bjackman:
This provides simple tests that use a localhost target to test running workloads and ensuring that the required output was produced. Although this certainly isn't an in-depth test of the wlgen system it should provide us at least a basic canary to check things aren't totally broken.

### DIFF
--- a/tests/lisa/test_wlgen.py
+++ b/tests/lisa/test_wlgen.py
@@ -1,0 +1,95 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2017, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Base classes and utilities for self-testing LISA's wlgen packages"""
+
+import os
+import shutil
+from unittest import TestCase
+
+from devlib import LocalLinuxTarget, Platform
+
+dummy_calibration = {}
+
+class TestTarget(LocalLinuxTarget):
+    """
+    Devlib target for self-testing LISA
+
+    Uses LocalLinuxTarget configured to disallow using root.
+    Adds facility to record the commands that were executed for asserting LISA
+    behaviour.
+    """
+    def __init__(self):
+        self.execute_calls = []
+        super(TestTarget, self).__init__(platform=Platform(),
+                                         load_default_modules=False,
+                                         connection_settings={'unrooted': True})
+
+    def execute(self, *args, **kwargs):
+        self.execute_calls.append((args, kwargs))
+        return super(TestTarget, self).execute(*args, **kwargs)
+
+    @property
+    def executed_commands(self):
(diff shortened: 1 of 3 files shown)


bjackman:
Looks like Travis can run these tests for us: https://travis-ci.org/bjackman/lisa/builds/201596684 (here you can see the wlgen self-tests running successfully and my WIP Executor self-tests failing due to some patches that aren't merged yet)

tests/lisa/test_wlgen.py
```diff
@@ -0,0 +1,379 @@
...
+
+        os.makedirs(self.host_out_dir)
+
+    def assert_output_file_exists(self, path):
+        """Assert that a file was created in host_out_dir"""
+        path = os.path.join(self.host_out_dir, path)
+        self.assertTrue(os.path.isfile(path),
+                        'No output file {} from rt-app'.format(path))
+
+class TestRTAProfile(RTABase):
```
derkling: I can see this source becoming bigger and bigger in short time... maybe it's worth to setup it since the beginning to be a more scalable solution?

What about moving RTA* and Perf* classes into dedicated classes in the same folder?
Do you think it could be possible?

I would propose:
```
tests/lisa/test_wlgen.py
tests/lisa/test_wlgen_rtapp.py
tests/lisa/test_wlgen_perf.py
```

bjackman: Yep, that works.

tests/lisa/test_wlgen.py
```diff
@@ -0,0 +1,379 @@
...
+        rtapp = RTA(self.target, name='test', calibration=self.calibration)
+
+        # Configure this RTApp instance to:
+        rtapp.conf(kind='custom', params=json_path, duration=5,
+                   run_dir=self.target_run_dir)
+
+        with open(rtapp.json) as f:
+            conf = json.load(f)
+
+        # Convert to str because unicode
```
bjackman: Reworded to "Convert k to str because the json loader gives us unicode strings"

tests/lisa/test_wlgen.py
```diff
@@ -0,0 +1,379 @@
...
+        tasks = set([str(k) for k in conf['tasks'].keys()])
+        self.assertSetEqual(
+            tasks,
+            set(['AudioTick', 'AudioOut', 'AudioTrack',
+                 'mp3.decoder', 'OMXCall']))
+
+        # Would like to try running the workload but mp3-short.json has nonzero
+        # 'priority' fields, and we probably don't have permission for that
+        # unless we're root.
```
bjackman: Agreed. Priorities is one of these things that we just won't be able to test using a localhost target.

tests/lisa/test_wlgen.py
```diff
@@ -0,0 +1,379 @@
...
+        for field in fields:
+            msg = 'Perf performance report missing "{}" field'.format(field)
+            self.assertIn(field, perf_json, msg)
+
+class TestHackBench(PerfBenchBase):
+    tools = ['perf']
+
+    def test_hackbench_smoke(self):
+        """
+        Test PerfMessaging hackbench workload
```
bjackman: Done

tests/lisa/test_wlgen.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def test_perf_pipe_smoke(self):
+        """
+        Test PerfPipe workload
+
+        Runs a 'PerfPipe' workload and tests that the expected output was
+        produced.
+        """
+        perf = PerfPipe(self.target, 'perfpipe')
+        perf.conf(loop=100)
```
bjackman: Done

bjackman:
Pushed a new version with

- Files split up as @derkling's suggestion
- perf loop count increased to 100k
- "hackbench" just called "perf messaging"
- "because unicode" comment reworded

PR #300, opened 2017-02-15
Add support for Travis CI

bjackman:
This adds a barebones YAML config for Travis and a fix that allows travis to detect failures from `lisa-test`. You can see an example run against my repository here: https://travis-ci.org/bjackman/lisa/builds/201980373 (running the wlgen tests that were recently meged).


bjackman:
Once this is enabled in the Travis interface we'll get a pass/fail at the bottom of the pr thread like this: https://github.com/bjackman/lisa/pull/1

PR #308, opened 2017-02-23
Add infrastructure to build Sphinx API documentation

bjackman:
This is still slightly rough; the structure of the documentation can probably be improved and I haven't reviewed the entire HTML output. However this should be a proof of concept and provide a base for more complete documentation and automated hosting.

Included:
- All the boring configuration for sphinx, generated by `sphinx-quickstart` and `sphinx-apidoc`. Unfortunately I didn't record the details of the process I used and it was quite a while ago. If we need that info I will just do it again. **No need to really review this**, just skim the files and see what got added and where.
- Some miscellaneous bits, as separate commits
- Mods to the Travis conf so we get notifications if PRs break the doc build.


bjackman:
Fixed the warnings that were causing Travis to say no.

The `doc:` patches could probably be squashed together but I think it might be useful to have them separate so we can see the changes I made to the sphinx config that `sphinx-quickstart` produced.

bjackman:
You can see the results here: http://lisa-linux-integrated-system-analysis.readthedocs.io/en/latest/

Again, I haven't reviewed that output properly, that's a separate task.

doc/conf.py
```diff
@@ -0,0 +1,270 @@
...
+# General information about the project.
+project = u'LISA'
+copyright = u'2016, ARM-Software'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '16.10'
```
derkling: Should we keep updating this by hand... or can it be automatically generated out of some git tag?

doc/conf.py
```diff
@@ -0,0 +1,270 @@
...
+
+# The encoding of source files.
+#source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'LISA'
+copyright = u'2016, ARM-Software'
```
derkling: This has to be updated to 2017... the first release can be 17.04

derkling:
We should probably merge these bits... sooner better then later...
I've give it a run on my machine and I've noticed that in the generated documentation we don't get pages for the wlgen module. Is that expected?

It would be also nice if, in the main page, we can have a sort of index section which points at least to:
1. the tutorial notebooks
2. the examples notebooks
3. the release notes notebooks
4. the tests folder
5. the benchmarks folder
6. (in the future) the scripts folder

derkling:
Just noticed that this series should also add a patch which updates `.gitignore` to exclude `doc/_build`.

doc/conf.py
bjackman: I'd say we can generate it from the git tag. That means you won't get the version if you're using a tarball, but TBH if anyone does do that I wouldn't expect them to be compiling documentation themselves.

bjackman:
Sorry I should have mentioned that I have a massively changed version of this with something like the index section you mentioned. I also copied in the content from the Github wiki - does that sound sensible to you?

valschneider:
As @derkling said, I think it would be nice to have this merged into LISA, even if it is not entirely finished. Maybe simply update the .gitignore and give it a go ?

bjackman:
Yep, I actually already updated the gitignore. The result here is far from perfect but I think we can still merge it and tweak afterwards. I've also done most of the work getting this build automatically and hosted on readthedocs. Due to the way readthedocs works, that will be much easier to finish once this PR is merged.

valschneider:
Automatic version string LGTM, though Travis fails on something weirdly not doc-related:
Unusual max capacity (1023), overriding capacity_scale

valschneider:
The story continues in https://github.com/ARM-software/lisa/pull/534

PR #310, opened 2017-02-24
Add test for CPU invariant load tracking

bjackman:
- Adds some handy members to the `EnergyModel` class
- Make the existing load_tracking test code more generic
- Add a test that runs a workload on each different type of CPU in the system and checks that the utilisation signal from the scheduler is scaled according to each CPU's capacity.

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -349,6 +349,23 @@ class EnergyModel(object):
         states = self.cpu_nodes[0].active_states
         return any(c.active_states != states for c in self.cpu_nodes[1:])
 
+    @property
+    @memoized
+    def cpu_groups(self):
+        """
+        List of lists of CPUs who share the same active state values
+        """
+        groups = []
+        for node in self.cpu_nodes:
+            for group in groups:
+                group_states = self.cpu_nodes[group[0]].active_states
+                if node.active_states == group_states:
+                    group.append(node.cpu)
+                    break
+            else:
+                groups.append([node.cpu])
+        return groups
+
     def _guess_idle_states(self, cpus_active):
         def find_deepest(pd):
             if not any(cpus_active[c] for c in pd.cpus):
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -378,6 +378,17 @@ class EnergyModel(object):
 
         return [find_deepest(pd) for pd in self.cpu_pds]
 
+    def get_cpu_capacity(self, cpu, freq=None):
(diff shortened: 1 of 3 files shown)


libs/utils/energy_model.py
```diff
@@ -378,6 +378,18 @@ def find_deepest(pd):
...
+    def get_cpu_capacity(self, cpu, freq=None):
+        """Convenience method to get the capacity of a CPU at a given frequency
+
+        :param cpu: CPU to get capacity for
+        :param freq: Frequency to get the CPU capacity at. Default is max
+                     capacity.
+        """
+        if freq is None:
+            return self.cpu_nodes[cpu].max_capacity
+        else:
```
derkling: nit-pick: this else is not necessary

tests/eas/load_tracking.py
```diff
@@ -216,3 +229,65 @@ def test_task_load_avg(self, experiment, tasks):
         this test does the same as test_task_util_avg but for load_avg.
         """
         return self._test_signal(experiment, tasks, 'load_avg')
+
+class CpuInvarianceTest(_LoadTrackingBase):
+    """
+    Goal
+    ====
+    Basic check for CPU invariant load tracking
```
derkling: You are checking also `util_avg`... should not call this:

```
Basic check for CPU invariant load and utilization tracking
```

tests/eas/load_tracking.py
```diff
@@ -216,3 +229,65 @@ def test_task_load_avg(self, experiment, tasks):
...
+    events and its mean is compared with the expected value (ignoring the first
+    300ms so that the signal can stabilize). The test fails if the observed mean
+    is beyond a certain error margin from the expected one. load_avg is then
+    similarly compared with the expected util_avg mean, under the assumption
+    that load_avg should equal util_avg when system load is light.
+
+    Expected Behaviour
+    ==================
+    Load tracking signals are scaled so that the workload results in roughly the
+    same util & load values regardless of compute power of the CPU used.
```
derkling: I would also add here again something like
_"Moreover. the load signal is verified to be similar to the utilization one when running on light load conditions."_

tests/eas/load_tracking.py
bjackman: Hmm, I have always thought of "load tracking" as including both `load` and `util` (since they're both features of the "Per-entity load tracker"), but fair enough, I'll change it.

bjackman:
OK, thanks for review, pushed a version addressing your comments.

PR #312, opened 2017-02-24
Android YouTube

ionela-voinescu:
Modify YouTube benchmark to make use of the new android benchmark API and provide an example of an android YouTube benchmark test.

### DIFF
--- a/libs/utils/android/workloads/youtube.py
+++ b/libs/utils/android/workloads/youtube.py
@@ -21,9 +21,7 @@ import logging
 
 from time import sleep
 
-from android import Screen, System
-from android.workload import Workload
-
+from android import Screen, System, Workload
 
 class YouTube(Workload):
     """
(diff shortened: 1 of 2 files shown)


libs/utils/android/workloads/youtube.py
```diff
@@ -42,31 +40,46 @@ def __init__(self, test_env):
...
+        :type exp_dir: str
+
+        :param video_url: Video URL to be played
+        :type video_url: str
+
+        :param duration_s: Play video for this required number of seconds
+        :type duration_s: int
+
+        :param collect: Specifies what to collect. Possible values:
+            - 'energy'
```
derkling: Why only energy?

There are not limitations from the Workload's module standpoint, which now supprots ftrace/systrace and energy:
https://github.com/ARM-software/lisa/blob/master/libs/utils/android/workload.py#L90

BTW thanks for the documentation! ;-)

libs/utils/android/workloads/youtube.py
```diff
@@ -42,31 +40,46 @@ def __init__(self, test_env):
...
 
         # Use the monkey tool to start YouTube without playing any video.
         # This allows to subsequently set the screen orientation to LANDSCAPE
         # and to reset the frame statistics.
-        System.monkey(self.target, self.package)
+        System.monkey(self._target, self.package)
 
         # Force screen in LANDSCAPE mode
-        Screen.set_orientation(self.target, portrait=False)
+        Screen.set_orientation(self._target, portrait=False)
```
derkling: I forgot in my jankbench example to tune the screen brightness, can you add it at least in these new worklaods?

libs/utils/android/workloads/youtube.py
```diff
@@ -78,16 +91,16 @@ def run(self, out_dir, collect,
...
-        System.force_stop(self.target, self.package, clear=False)
+        System.force_stop(self._target, self.package, clear=False)
 
         # Go back to home screen
-        System.home(self.target)
+        System.home(self._target)
 
         # Switch back to screen auto rotation
-        Screen.set_orientation(self.target, auto=True)
+        Screen.set_orientation(self._target, auto=True)
```
derkling: And brightness must be reset here...

tests/benchmarks/android_youtube.py
```diff
@@ -0,0 +1,189 @@
...
+        # Tools required by the experiments
+        "tools"   : [ 'trace-cmd' ],
+
+		# Default results folder
+		"results_dir" : "AndroidYouTube",
+
+    }
+
+	# Android Workload to run
+    bm_name = 'YouTube'
```
derkling: Is there some spaces issue here, before the comment?

tests/benchmarks/android_youtube.py
```diff
@@ -0,0 +1,189 @@
...
+video_duration_s = 60
+
+governors = [
+    'performance',
+    'ondemand',
+    'interactive',
+    'sched',
+    'schedutil',
+    'powersave',
+]
```
derkling: Perhaps we should think about adding a new benchmark command line parameter here:
   https://github.com/ARM-software/lisa/blob/master/libs/utils/android/benchmark.py#L78
which allows to specify at command line which governors we want to test.

Let's keep for another time, but what do you think?... most of the time maybe we do not want to test with really all these governors... do you think it should be possible/useful to expose such a tunables in Jenkins?

libs/utils/android/workloads/youtube.py
```diff
@@ -102,5 +105,7 @@ def run(self, out_dir, video_url, video_duration_s, collect=''):
 
         # Switch back to screen auto rotation
         Screen.set_orientation(self._target, auto=True)
+        # Set brightness back to auto
+        Screen.set_brightness(self._target, auto=True)
```
derkling: Oh... here is the patch... just squash with the first one ;-)

libs/utils/android/workloads/youtube.py
ionela-voinescu: Just omission due to hurry. I'll modify the others too as they will suffer from the copy-paste syndrome.

libs/utils/android/workloads/youtube.py
ionela-voinescu: Squashed! :imp:

derkling: But down here I still see two patches... ?!?

tests/benchmarks/android_youtube.py
ionela-voinescu: Yes, it was seeing them alright with `:set tabstop=4 shiftwidth=4 expandtab`. It's fixed now.

libs/utils/android/workloads/youtube.py
ionela-voinescu: It's done now! I posted the comment a short while before I got the chance to push the branch. You move fast!

PR #314, opened 2017-02-24
Android Geekbench

ionela-voinescu:
Add Geekbench benchmark and provide an example of an android Geekbench benchmark test.

### DIFF
--- a/libs/utils/android/workloads/geekbench.py
+++ b/libs/utils/android/workloads/geekbench.py
@@ -1,0 +1,154 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2015, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import re
+import os
+import logging
+from time import sleep
+
+from subprocess import Popen, PIPE
+
+from android import Screen, Workload, System
+
+# Regexps for benchmark synchronization
+GEEKBENCH_BENCHMARK_START_RE = re.compile(
+    r'ActivityManager: Start.* com.primatelabs.geekbench'
+)
+GEEKBENCH_BENCHMARK_END_RE = re.compile(
+    r'GEEKBENCH_RESULT: (?P<results_file>.+)'
+)
+
+class Geekbench(Workload):
+    """
+    Android Geekbench workload
+    """
+
+    # Package required by this workload
+    package = 'com.primatelabs.geekbench'
+    activity = '.HomeActivity'
+
+    def __init__(self, test_env):
+        super(Geekbench, self).__init__(test_env)
+        self._log = logging.getLogger('Geekbench')
+        self._log.debug('Workload created')
+
+    def run(self, out_dir, test_name, collect=''):
+        """
+        Run single Geekbench workload.
+
+        :param out_dir: Path on host to experiment directory where
+                        to store results.
+        :type out_dir: str
+
+        :param test_name: Name of the test to run
+        :type test_name: str
+
+        :param collect: Specifies what to collect. Possible values:
+            - 'energy'
+            - 'systrace'
+            - 'ftrace'
+            - any combination of the above in a space-separated string.
+        :type collect: list(str)
+        """
+
+        # Initialize energy meter results
+        self.out_dir = out_dir
+        self.collect = collect
+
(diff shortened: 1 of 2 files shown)


ionela-voinescu:
There is an issue with the devices powering off during a CPU benchmark run, so I recommend keeping this a while longer until we understand what the problem is.
I've just created the pull request for it to be available for review.

libs/utils/android/workloads/geekbench.py
```diff
@@ -0,0 +1,154 @@
...
+        super(Geekbench, self).__init__(test_env)
+        self._log = logging.getLogger('Geekbench')
+        self._log.debug('Workload created')
+
+    def run(self, out_dir, test_name, collect=''):
+        """
+        Run single Geekbench workload.
+
+        :param out_dir: Path to experiment directory where to store results.
+        :type out_dir: str
```
bjackman: Can you clarify that this is the directory on the host. I guess it's obvious when you think about it but it's nice not to have to think about it :P

ionela-voinescu:
Thank you for the review! I've updated the branch according to your comments, but I would still not merge this yet.

libs/utils/android/workloads/geekbench.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+
+        # Initialize energy meter results
+        self.out_dir = out_dir
+        self.collect = collect
+
+        # Regexps for benchmark synchronization
+        start_logline = r'ActivityManager: Start.*'\
+                         'com.primatelabs.geekbench'
+        GEEKBENCH_BENCHMARK_START_RE = re.compile(start_logline)
```
derkling: I would have preferred to keep these "globals" defined outside the class... like we did for jankbench.
Otherwise using such a long name with "GEEKBENCH_BENCHMARK" as prefix does not make sense.

Unfortunately I just verified that I've merged your other workloads without pointing that out... so, it's up to you: either move them outside (like we do for jankbench) or just use "START_RE" and "END_RE".

I personally would prefer the first option which avoid to clobber the function code with funky REs :-)

libs/utils/android/workloads/geekbench.py
```diff
@@ -0,0 +1,155 @@
...
+        # Press the 'RUN <test_name> BENCHMARK' button
+        System.tap(self._target, 73, 72)
+
+        while True:
+
+            # read next logcat line (up to max 1024 chars)
+            message = logcat.stdout.readline(1024)
+
+            # Benchmark start trigger
+            if GEEKBENCH_BENCHMARK_START_RE.search(message):
```
derkling: nit-pick: I would prefer a sequence of:

```python
match = GEEKBENCH_BENCHMARK_START_RE.search(message)
if match:
    ....
match = GEEKBENCH_BENCHMARK_END_RE.search(message)
if match:
   ....
```

Which makes the code slightly less optimized but that's not running on target and makes also the code more simple to read. That's a pattern we want users to easily copy-adapt to their workloads. In case they should need to match multiple REs (like we do with for jankbenck) we can easily end up with many horrible indentation levels.

libs/utils/android/workloads/geekbench.py
```diff
@@ -0,0 +1,155 @@
...
+                # Start tracing
+                self.tracingStart()
+                self._log.debug("Benchmark started!")
+
+            else:
+                match = GEEKBENCH_BENCHMARK_END_RE.search(message)
+                if match:
+                    # Stop tracing
+                    self.tracingStop()
+                    remote_result_file = match.group(1)
```
derkling: Can you use match group names in your REs?
This would make the code more documented.
Here is an example, of the defintion (using the "?P<name_of_the_field>"):
   https://github.com/ARM-software/lisa/blob/master/libs/utils/android/workloads/jankbench.py#L45
and of the usage (using the specified "name_of_the_field"):
   https://github.com/ARM-software/lisa/blob/master/libs/utils/android/workloads/jankbench.py#L144

libs/utils/android/workloads/geekbench.py
```diff
@@ -0,0 +1,155 @@
...
+        GEEKBENCH_BENCHMARK_START_RE = re.compile(start_logline)
+        self._log.debug("START string [%s]", start_logline)
+
+        end_logline = r'GEEKBENCH_RESULT: (.*)'
+        GEEKBENCH_BENCHMARK_END_RE = re.compile(end_logline)
+        self._log.debug("END string [%s]", end_logline)
+
+        # Clear the stored data for the application, so we always start with
+        # an EULA to clear
+        self._target.execute('pm clear {}'.format(self.package))
```
derkling: maybe we can avoid inlined adb commands by using the same:
```python
System.force_stop(self._target, self.package, clear=False)
```
we use at the end of the benchmark... but this time with `clear` set to `True`?

libs/utils/android/workloads/geekbench.py
ionela-voinescu: I'll move them outside for all the benchmarks that I have in progress and most probably I'll have some cleaning up patches for the other ones also (based on your comments and the ones from Brendan).

libs/utils/android/workloads/geekbench.py
ionela-voinescu: Right!

PR #321, opened 2017-03-02
Android PCMark

ionela-voinescu:
Add PCMark benchmark and provide an example of an android PCMark benchmark test.

### DIFF
--- a/libs/utils/android/workloads/pcmark.py
+++ b/libs/utils/android/workloads/pcmark.py
@@ -1,0 +1,182 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2015, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import re
+import os
+import logging
+import time
+
+from subprocess import Popen, PIPE
+
+from android import Screen, System
+from android.workload import Workload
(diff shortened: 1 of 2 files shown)


ionela-voinescu:
@derkling , @bjackman : I've modified these commits also according to your comments on other benchmarks.

libs/utils/android/workloads/pcmark.py
```diff
@@ -0,0 +1,182 @@
...
+        logcat_cmd = self._adb(
+                'logcat com.futuremark.pcmandroid.AbstractMainActivity:I *:S'\
+                .format(self._target.adb_name))
+        self._log.info("%s", logcat_cmd)
+
+        logcat = Popen(logcat_cmd, shell=True, stdout=PIPE)
+
+        # Run performance workload
+        if test_name.upper() == 'PERFORMANCE':
+            self._target.execute('input tap 750 1450')
```
derkling: Here we should use relative coords... but it's a pain.

On a Pixel XL, where the screen is 2k (1440,2560), the run button has this coords:
```
   top-left:     ( 914,1630)
   bottom-right: (1340,1730)
```

libs/utils/android/workloads/pcmark.py
```diff
@@ -0,0 +1,182 @@
...
+            logcat = Popen(self._adb(
+                     'logcat com.futuremark.pcmandroid.VirtualMachineState:* *:S'\
+                     .format(self._target.adb_name)),
+                     stdout=log,
+                     shell=True)
+            time.sleep(2)
+            logcat.kill()
+
+        # Parse logcat file and create db with scores
+        os.popen('grep -o "PCMA_.*_SCORE .*" {} | sed "s/ = / /g" | sort -u > {}'
```
derkling: Here we should use some more portable python code, i.e. something like:
```python

SCORE_RE.compile(r'here is the regexp with the capture groups you are interested into')

results = []
with open(db_file, 'r') as fh:
    for line in fh:
        match = SCORE_RE.match(line)
        if not match:
            continue
        score = int(match.group('score'))
        results.append(score)
results.sort()
```

valschneider:
@ionela-voinescu Do you mind if we close this PR ? If I'm missing some functionality in #475 I'll implement it free of charge.

PR #325, opened 2017-03-06
Gmaps workload

valschneider:
Gmaps workloaded, benchmark & notebook


libs/utils/android/workloads/gmaps.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def run(self, out_dir, location_search, swipe_count=5, collect=''):
+        """
+        Run single Gmaps workload.
+
+        :param out_dir: Path to experiment directory where to store results.
+        :type out_dir: str
+
+        :param location_search: Search string to be used in GMaps
+        :type video_url: str
```
ionela-voinescu: This needs to be changed to `location_search`.

libs/utils/android/workloads/gmaps.py
```diff
@@ -0,0 +1,123 @@
...
+
+        :param swipe_count: Number of sets of (left, right, up, down) swipes to do
+        :type video_duration_s: int
```
ionela-voinescu: This needs to be changed to `swipe_count`.

libs/utils/android/workloads/gmaps.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # Keep track of mandatory parameters
+        self.out_dir = out_dir
+        self.collect = collect
+
+        # Unlock device screen (assume no password required)
+        System.menu(self._target)
+        # Press Back button to be sure we run the video from the start
+        System.back(self._target)
+
```
ionela-voinescu: We also set here the brightness to 0 for the display to not have too much impact on the energy measurements:

```
        # Set min brightness
        Screen.set_brightness(self._target, auto=False, percent=0)
```

and restore it to auto at the end:

```
        # Set brightness back to auto
        Screen.set_brightness(self._target, auto=True)
```

valschneider:
Updated the code taking into account Ionela's reviews

libs/utils/android/workloads/gmaps.py
```diff
@@ -0,0 +1,123 @@
...
+    # Package required by this workload
+    package = 'com.google.android.apps.maps'
+    action = 'android.intent.action.VIEW'
+
+    def __init__(self, test_env):
+        super(GMaps, self).__init__(test_env)
+        self._log = logging.getLogger('GMaps')
+        self._log.debug('Workload created')
+
+        # Set of output data reported by Jankbench
```
ionela-voinescu: GMaps here instead of Jankbench.

valschneider:
Last corrections added and fixes squashed into original commits

derkling:
Before merging it, I would have preferred to fix some issues we discovered when I've wrote the really first-and-dirt prototype of this test.

The main issue here is that all swipe actions are send from the host to the target *while* the test is running. This results in a set of new tasks being generated for each and every swipe action, tasks which affects the test itself. They are quite evident in the collected trace and they can potentially impact quite a lot in task placement and ultimately in the overall test performance.

There are two possible solutions:
1. we should find a way to ensure that the ADB server is pinned in CPU0, thus forcing all the commands we send to run on that CPU. Which should mitigate the side-effect of the test instrumentation on the scheduling behaviours.
2. we can think about a possible API extension for the Benchmark class which allows to generate on-the-fly scripts to be kicked-off on target and executed on background without requiring to generate more task.

Since now this test is merged, fixing these issues is kind-of high priority.

PR #328, opened 2017-03-10
uibench: Added parameter to allow action(s)

mcgeagh:
By default when running uibench, no user action is performed during the
tests. This adds an optional parameter called 'action' and performs the
actions as input events. This parameter can take the form of a list and
allows you to perform multiple actions in a set order too.

E.g. When performing ActivityTransition test, action=['tap', 'tap'] will
perform two taps in the centre of the screen which will maximise a
thumbnail then minimise the resulting fullscreen image.

### DIFF
--- a/libs/utils/android/workloads/uibench.py
+++ b/libs/utils/android/workloads/uibench.py
@@ -34,21 +34,33 @@ class UiBench(Workload):
 
     # Supported activities list, obtained via:
     # adb shell dumpsys package | grep -i uibench | grep Activity
-    test_BitmapUpload = 'BitmapUploadActivity'
-    test_DialogList = 'DialogListActivity'
-    test_EditTextType = 'EditTextTypeActivity'
-    test_FullscreenOverdraw = 'FullscreenOverdrawActivity'
-    test_GlTextureView = 'GlTextureViewActivity'
-    test_InflatingList = 'InflatingListActivity'
-    test_Invalidate = 'InvalidateActivity'
-    test_ShadowGrid = 'ShadowGridActivity'
-    test_TextCacheHighHitrate = 'TextCacheHighHitrateActivity'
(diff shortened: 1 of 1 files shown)


bjackman:
LGTM but I'm not that familiar with the benchmark stuff so will wait for @derkling or @ionela-voinescu.

libs/utils/android/workloads/uibench.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, test_env):
         # Set of output data reported by UiBench
         self.db_file = None
 
-    def run(self, out_dir, test_name, duration_s, collect=''):
+    def run(self, out_dir, test_name, duration_s, collect='', action=''):
```
derkling: Since tests are meant to be reproducible and shareable among users, would not be better to use a pre-defined a set of sensible actions for each test?

In that case it would probably be better to encode these reasonable set of actions in this test itself.

For example, we can use a dictionary `actions[TEST_NAME] = [action1, action2, ...]` to map the actions required by `test_TEST_NAME`.

libs/utils/android/workloads/uibench.py
```diff
@@ -162,4 +178,15 @@ def run(self, out_dir, test_name, duration_s, collect=''):
         System.set_airplane_mode(self._target, on=False)
         Screen.set_brightness(self._target, auto=True)
 
+    def _perform_action(self, action, delay_s=0):
```
derkling: This way to perform actions is currently "functionally broken" because it generates tasks on the target side which can affect the test behaviours.

More details in #329, we should fix the System API before adding interactivity features to existing tests.

libs/utils/android/workloads/uibench.py
mcgeagh: Done. Now part of a dictionary with 'sensible' default actions that by default will be performed when given a test that matches the key.

PR #330, opened 2017-03-14
uibench: Set the duration when using systrace

mcgeagh:
uibench by default collects with systrace but the duration_s parameter
is not used correctly. This is because of a limitation that we cannot
CTRL+C stop systrace captures so the workload class expects duration to
be appended to the collect parameter. This fixes uibench workload to
correctly parse the collect parameter to include the given duration_s
parameter.

### DIFF
--- a/libs/utils/android/workloads/uibench.py
+++ b/libs/utils/android/workloads/uibench.py
@@ -84,6 +84,11 @@ class UiBench(Workload):
         # Keep track of mandatory parameters
         self.out_dir = out_dir
         self.collect = collect
+        # For systrace collection, the duration is expected to be part of `collect` parameter
+        # This is because systrace capturing currently doesnt support CTRL+C stopping
+        # so tracingStop() call just waits instead.
+        if collect == 'systrace':
+            self.collect = 'systrace_' + str(duration_s)
 
         # Press Back button to be sure we run the video from the start
         System.menu(self._target)


libs/utils/android/workloads/uibench.py
```diff
@@ -84,6 +84,11 @@ def run(self, out_dir, test_name, duration_s, collect=''):
         # Keep track of mandatory parameters
         self.out_dir = out_dir
         self.collect = collect
+        # For systrace collection, the duration is expected to be part of `collect` parameter
+        # This is because systrace capturing currently doesnt support CTRL+C stopping
```
derkling: Actually it seems that systrace can be "stopper on demand"... by just running it on background and than sending an ENTER when we are done:
https://github.com/catapult-project/catapult/blob/master/systrace/systrace/run_systrace.py#L169

I've just added this as an issue: #331

libs/utils/android/workloads/uibench.py
```diff
@@ -84,6 +84,11 @@ def run(self, out_dir, test_name, duration_s, collect=''):
...
+        # so tracingStop() call just waits instead.
+        if collect == 'systrace':
+            self.collect = 'systrace_' + str(duration_s)
```
derkling: Since the API is already broken, i.e. systrace expect the duration to be defined as part of the `capture` string, should not be better to just remove this non-sense `duration_s` parameter and encode the duration in the `systrace_DURATION` parameter from the caller itself.

This would make the current code more aligned with the new API once we have the proper "on demand" stop support (see #331)

mcgeagh: That would make sense if only systrace was supported.
What happens when the user chooses ftrace instead? That does support stopping after n seconds.
By removing duration_s, I assume you then expect them to format their 'collect' parameter to be <type>_<duration> for both ftrace and systrace (and any future supported methods)?

derkling: Looks like `duration_s` was actually required just because we needed to support systrace when there was not on-demand stop control.

Once we have #332 merged, we should probably completely remove `duration_s` and add instead a RE to capture the end of the test, as we already do for many other tests.
Look here for an example:
https://github.com/ARM-software/lisa/blob/master/libs/utils/android/workloads/jankbench.py#L152

Using RE we can stop on-demand both ftrace and systrace and that should be the "optimal" final solution.

derkling: And #332 has now been merged ;-)

PR #332, opened 2017-03-14
libs/utils/android: Add systrace timeless support

valschneider:
Systrace expects either a time argument to know how long it has
to record traces, or an <enter> keystroke to end the tracing.
Added a 'fake' <enter> keystroke so tracing may be started
and stopped without a predefined duration.

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -30,7 +30,7 @@ class System(object):
     """
 
     @staticmethod
-    def systrace_start(target, trace_file, time,
+    def systrace_start(target, trace_file, time=None,
                        events=['gfx', 'view', 'sched', 'freq', 'idle']):
 
         log = logging.getLogger('System')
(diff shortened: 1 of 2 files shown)


libs/utils/android/system.py
```diff
@@ -46,10 +46,13 @@ def systrace_start(target, trace_file, time,
...
                 return None
 
-        #  Format the command according to the specified time and events
-        systrace_pattern = "{} -e {} -o {} {} -t {}"
+        #  Format the command according to the specified arguments
+        systrace_pattern = "{} -e {} -o {} {}"
         trace_cmd = systrace_pattern.format(systrace_path, target.conf['device'],
-                                            trace_file, " ".join(events), time)
+                                            trace_file, " ".join(events))
+        if (time is not None):
```
derkling: just for consistency, you can avoid parenthesis here

libs/utils/android/workload.py
```diff
@@ -104,9 +104,7 @@ def tracingStart(self):
             if match:
                 self._trace_time = match.group(1)
             else:
-                # TODO: must implement a CTRL+C based systrace stopping
-                self._log.warning("Systrace time NOT defined, tracing for 10[s]")
-                self._trace_time = 10
+                self._trace_time = None
```
derkling: What about:

```python
self._trace_time = match.group(1) if match else None
```

libs/utils/android/workload.py
```diff
@@ -132,6 +130,9 @@ def tracingStop(self):
             else:
                 self._log.info('Waiting systrace report [%s]...',
                                  self.trace_file)
+                if (self._trace_time is None):
```
derkling: Here also you can remove the parentesis

libs/utils/android/workload.py
```diff
@@ -132,6 +130,9 @@ def tracingStop(self):
...
+                    # Systrace expects <enter>
+                    self._systrace_output.sendline("")
```
derkling: Not big deal since we are both ' and "... but perhaps we can try to be consistent at lest within functions and, thanks to your previous removal, that's now the only usage or "... let's switch to '

valschneider:
Branch updated regarding previous comments. It's been tested with the GMaps workload.

PR #333, opened 2017-03-15
lisa_shell: Check for missing submodules and update them

joelagnel:

### DIFF
--- a/src/shell/lisa_shell
+++ b/src/shell/lisa_shell
@@ -245,6 +245,16 @@ echo
 echo
 }
 
+function lisa-check-submods {
+local missing=0
+if [ ! -f ./libs/devlib/setup.py ] ||
+   [ ! -f ./libs/bart/setup.py ] ||
+   [ ! -f ./libs/trappy/setup.py ]; then
+	echo "One or more submodules missing, updating";
+	lisa-update submodules
+fi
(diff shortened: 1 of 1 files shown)


src/shell/lisa_shell
```diff
@@ -245,6 +245,24 @@ echo
...
+local missing=0
+if [ ! -f ./libs/devlib/setup.py ]; then
+	missing=1
+fi
+if [ ! -f ./libs/bart/setup.py ]; then
+	missing=1
+fi
+if [ ! -f ./libs/trappy/setup.py ]; then
+	missing=1
+fi
```
bjackman: Can we reduce this to 

```
function lisa-check-submods {
if [ ! -f ./libs/devlib/setup.py ] ||
   [ ! -f ./libs/bart/setup.py ] ||
   [ ! -f ./libs/trappy/setup.py ]; then
	echo "One or more submodules missing, updating"
	lisa-update submodules
fi
}
```

derkling: I like that solution... I've suggested Jeol to check for just on library but it's actually more safe to check for all the main dependencies. ;-)

joelagnel:
Done, I updated based on your suggestion, please pull, thanks

PR #334, opened 2017-03-15
lisa_shell: Check for missing submodules and update them

joelagnel:


joelagnel:
Could this be pulled? thanks

PR #335, opened 2017-03-17
[RFC] Add a monsoon meter for devlib's monsoon instrument

joelagnel:
Needed some clarification about report(), on Monsoon there is only 1 channel (atleast on my personal monsoon), in this case what should the channel number of the output be? Do we pick a default channel name?

Also I have written code in the Instrument to generate Pandas frames from the collected monsoon data, in this case is there a proposed interface to access the pandas data frames from the meter?

Most of the work is done inside the instrument which pushed to devlib separately in this PR:
https://github.com/ARM-software/devlib/pull/100

### DIFF
--- a/get_wlat.py
+++ b/get_wlat.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python
+
+from collections import namedtuple
+import trappy
+
+EventRow = namedtuple('EventRow', ['time', 'event', 'row'])
+LatRow = namedtuple('LatRow', ['pid', 'switch_row', 'wake_row', 'last_wake_row', 'running', 'wake_pend', 'latency'])
+
+def print_event_rows(rows, start, end):
+	print "time".ljust(20) + "\t" + "event".ljust(10) + "\tpid" + "\tprevpid" + "\tnextpid"
+	for r in rows:
+		if r.time < start or r.time > end:
+			continue
+		if r.event == "switch":
+			nextpid =  r.row['next_pid']
+			prevpid = r.row['prev_pid']
+			pid = -1
+		elif r.event == "wake":
+			pid = r.row['pid']
+			nextpid = -1
+			prevpid = -1
+		else:
+			raise RuntimeError("unknown event seen")
+		print str(r.time).ljust(20)[:20] + "\t" + r.event.ljust(10) + "\t" + str(pid) + "\t" + str(prevpid) + "\t" + str(nextpid)
+
+path_to_html = "/home/joelaf/repo/catapult/systrace/trace_sf-dispsync_rt_pull_fail_1504ms.html"
+
+systrace_obj = trappy.SysTrace(name="systrace", path=path_to_html, \
+			       scope="sched", events=["sched_switch",
+			       "sched_wakeup", "sched_waking"])
+
+df = systrace_obj.sched_switch.data_frame
+# df = df[(df.next_prio < 100) | (df.prev_prio < 100)]
+switch_df = df
+
+df = systrace_obj.sched_wakeup.data_frame
+# df = df[df.prio < 100]
+wake_df = df
+
+# Build list of rows
+switch_rows = [EventRow(time=i, event="switch", row=row) for i, row in switch_df.iterrows()]
+wake_rows = [EventRow(time=i, event="wake", row=row) for i, row in wake_df.iterrows()]
+rows = switch_rows + wake_rows
+rows =  sorted(rows, key=lambda r: r.time)
+
+# Hash table of pid -> LatRow named tuple
+latpids = {}
+
+for r in rows:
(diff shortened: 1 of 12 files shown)


libs/utils/energy.py
```diff
@@ -118,6 +121,58 @@ def reset(self):
...
+	if exception == 1:
+		raise(Exception(errmsg))
+
+    def __init__(self, target, conf=None):
+	super(Monsoon, self).__init__(target, '')
+
+	# Read config
+	# { 'config': { 'device_entry' : '/dev/ttyACM1', 'voltage' : 4.3 } }
+	try:
+		device_entry = conf['config']['device_entry']
```
derkling: You can avoid the try/catch by using:

```python
self.device = conf['config'].get('device', None)
self.voltage = conf['config'].get('voltage', 4.3)
```

I would also prefer to `s/device_entry/device/`

libs/utils/energy.py
```diff
@@ -118,6 +121,58 @@ def reset(self):
...
+	m.get_status()
+	# Provide power to the target
+	m.set_voltage(0)
+	m.set_current(8)
+	m.set_start_current(8)
+	m.set_voltage(self.voltage)
+	self._log.info("Monsoon ready for action")
+
+    # Start a monsoon run asynchronously
+    def reset(self):
```
derkling: why not using an optional parameter to get rid of the `reset_only` function?

```python
def reset(self, start=True):
    # ... here is the reset_only code ...
    if not start:
        return
    m.start()
```

libs/utils/energy.py
```diff
@@ -118,6 +121,58 @@ def reset(self):
...
+	m = self._monsoon
+	self.reset_only()
+	m.start()
+
+    # Kill an already running test and report in panda dataframes
+    def report(self):
+	return self._monsoon.report()
+
+    # This is synchronous (block until duration)
+    def get_samples(self, duration):
```
derkling: LISA's EnergyMeters API expect to provide a really simple interface where the user needs only two methods:
- reset: which start a new energy measurement session
- report: which reports the energy measured since last reset and _all the additional data produced by the specific EM_

This means that you should better put whatever this code produce at the end of the report call.

You should have a look at the ACME's report function at the end of this module.

NOTE: perhaps this can require to have a different implementation of the devlib's Monsoon module's get_samples_sync(), which BTW I'm not sure it is an "expected" interface for the devlib's Instrument module.
We should implement if using either the "instantaneous" or the "continuous" APIs:
https://github.com/ARM-software/devlib/blob/master/devlib/instrument/__init__.py#L218

get_wlat.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python
```
derkling: Not entirely sure what you wanna get with this script, but I have the impression that some of this code can be simplified quite a lot by using this existing API:
https://github.com/ARM-software/lisa/blob/master/libs/utils/analysis/latency_analysis.py

A usage example is reported here:
https://github.com/ARM-software/lisa/blob/master/ipynb/examples/trace_analysis/TraceAnalysis_TasksLatencies.ipynb

Notice that you can get the DF of latency events for each task you want and then just describe() it to get the stats you are after.

joelagnel:
Very sorry for the screw-ups, I normally use topic branches but in this case I screwed up. Sorry for the confusion. I will abandon this PR since me/Brendan have to discuss doing this properly. Will abandon once I reply to your comments above. Thanks for your comments anyway, I didn't want to reveal or push this since the script is very WIP, but anyway since you commented, I will reply shortly :)

joelagnel:
Hi, regarding your comment "Not entirely sure what you wanna get with this script, but I have the impression that some of this code can be simplified quite a lot by using this existing API:"

There isn't an existing API to support _all_ tasks (more on that later).
Also, I initially used data frames to do this, but its too slow to do it using data frames, my first script totally took around 22s to run to completion on a 40MB trace file. With this updated scripts, its now down to 7s. I posted a new feature to trappy for support event-callbacks [1]. Before it can even build data frames, it will call the callbacks, which is faster for this particular script. I am not fully sure how we can add support for this to the latency analysis library, but I certainly prefer the speed up from using [1] than using the library which will be sub-optimal for this usecase. However, if you have any ideas to speed up the library we can discuss.

The motivation to write the script is to automate finding all long runnable times of RT tasks. I did look into your analysis library you pointed to, however I want it to operate all tasks and show me the results, not give it which task to find latency for. Even if you can add support to your library to support all tasks, I feel building data frames for this particular script is overkill and slow, and is not very helpful - once we process the data, we don't care about the data frames anymore.

[1] https://github.com/joelagnel/trappy/commit/b1974af446b843f199d3a538081bfe553e3ff8a6?diff=unified

joelagnel:
Brendan will supersede this with his version

PR #336, opened 2017-03-21
libs/utils/energy: Add 'power' kind for default VExpress hwmon config

bjackman:
It seems like some hwmon instances use 'power' instead of 'energy'. It looks like we can just enable both.

I'm not merging this unilaterally because there's a lot of hwmon-related code between LISA and devlib that I don't understand. It would also be fantastic if at least one other person could test this as it seems likely I'm breaking something for somebody.

BTW, at some point we really need to improve the way the hwmon stuff reports errors. This error was reported as a `KeyError` in `HwMon.report`. I spent a little time to try and make it a bit more helpful but there's really a lot of complexity and it seems like this complexity is necessary.

### DIFF
--- a/libs/utils/energy.py
+++ b/libs/utils/energy.py
@@ -195,6 +195,10 @@ class HWMon(EnergyMeter):
         # Reformat data for output generation
         clusters_nrg = {}
         for channel in self._channels:
+            if channel not in nrg:
+                raise RuntimeError('hwmon channel "{}" not available. '
+                                   'Selected channels: {}'.format(
+                                       channel, nrg.keys()))
             label = self._channels[channel]
             nrg_total = nrg[label]['total']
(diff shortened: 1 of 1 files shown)


derkling:
I would not be surprised if that depends on the kernel version in use... I'm pretty sure it used to work in the past using "energy".
Regarding `kinds`, AFAIR you can specify multiple channels and devlib does its best (sloooowly) to figure out which one are available. But, if there should be only one of the two, the code we have should sill work since we index just on `sites`... but I'm not completely sure and yes, that should be tested by more people... perhaps also using an older kernel.

bjackman:
[-1] don't merge

bjackman:
This is not the correct solution. Will separately submit the error message improvement.

PR #338, opened 2017-03-21
examples/android/benchmarks: Add Geekbench parsing notebook

valschneider:
Adds a notebook to run Geekbench & parse the .gb4 results

### DIFF
--- a/ipynb/examples/android/benchmarks/Android_Geekbench.ipynb
+++ b/ipynb/examples/android/benchmarks/Android_Geekbench.ipynb
@@ -1,0 +1,1105 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "# Geekbench benchmark on Android\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


ipynb/examples/android/benchmarks/Android_Geekbench.ipynb
```diff
@@ -0,0 +1,572 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "It can be interesting to compare Geekbench results with different parameters (kernel, drivers) and even different devices to gauge the impact of these parameters. As Geekbench results can vary a bit from one run to another, having a set of repeated runs is preferable.\n",
```
ionela-voinescu: The comparison setup is difficult to follow here, especially as the code above has a result directory with another structure than the one required by the comparison code (the directory structure below is based on the tests/benchmarks/android_geekbench.py structure). The user running this example notebook might not know anything about that.

I do like the possibility to compare different results but I would eliminate the requirement for  a base/cpu/governor/iteration/* structure and simply compare results from Geekbench_example folders. You can modify the results directory above to have a timestamp added (Geekbench_example_timestamp). When you run the cells above multiple times you'll get multiple folders. Than the user can reference those as baseline and experiment folders and compare only those. I think this would provide some comparison code that can be extended for more complex cases, while not going over the scope of this example. What do you think?

valschneider: Yes, you're absolutely right. I ended up having around a hundred runs to compare, and wanted to add that into the example notebook, but it is probably overkill for a simple example. I've modified it so it simply pulls stats from the Geekbench_example_* directories - this functionality can be used in another notebook to compare runs on different devices.

ionela-voinescu:
Perfect! ;)

PR #342, opened 2017-03-24
Trace: allow platform=None

bjackman:
This should be considered an RFC, I haven't very carefully checked what else this might break. However we need to get this sorted so I'm sending a PR to highlight it.

### DIFF
--- a/libs/utils/analysis/frequency_analysis.py
+++ b/libs/utils/analysis/frequency_analysis.py
@@ -133,13 +133,13 @@ class FrequencyAnalysis(AnalysisModule):
 
         # Extract LITTLE and big clusters frequencies
         # and scale them to [MHz]
-        if len(self._platform['clusters']['little']):
-            lfreq = df[df.cpu == self._platform['clusters']['little'][-1]]
+        if self._little_cpus:
+            lfreq = df[df.cpu == self._little_cpus[-1]]
             lfreq['frequency'] = lfreq['frequency']/1e3
         else:
             lfreq = []
-        if len(self._platform['clusters']['big']):
-            bfreq = df[df.cpu == self._platform['clusters']['big'][-1]]
+        if self._big_cpus:
+            bfreq = df[df.cpu == self._big_cpus[-1]]
             bfreq['frequency'] = bfreq['frequency']/1e3
         else:
             bfreq = []
@@ -296,14 +296,17 @@ class FrequencyAnalysis(AnalysisModule):
             axes.axhline(_avg, color='r', linestyle='--', linewidth=2)
 
             # Set plot limit based on CPU min/max frequencies
-            for cluster,cpus in self._platform['clusters'].iteritems():
-                if cpu_id not in cpus:
-                    continue
-                axes.set_ylim(
-                        (self._platform['freqs'][cluster][0] - 100000)/1e3,
-                        (self._platform['freqs'][cluster][-1] + 100000)/1e3
-                )
-                break
+            if 'clusters' in self._platform:
+                for cluster,cpus in self._platform['clusters'].iteritems():
+                    if cpu_id not in cpus:
+                        continue
+                    freqs = self._platform['freqs'][cluster]
+                    break
+            else:
+                freqs = df['frequency'].unique()
+
+            axes.set_ylim((min(freqs) - 100000) / 1e3,
+                          (max(freqs) + 100000) / 1e3)
 
             # Plot CPU frequency transitions
             _df['frequency'].plot(style=['r-'], ax=axes,
@@ -367,9 +370,8 @@ class FrequencyAnalysis(AnalysisModule):
 
         # Split between big and LITTLE CPUs ordered from higher to lower ID
         _cpus.reverse()
-        big_cpus = [c for c in _cpus if c in self._platform['clusters']['big']]
-        little_cpus = [c for c in _cpus if c in
-                       self._platform['clusters']['little']]
+        big_cpus = [c for c in _cpus if c in self._big_cpus]
+        little_cpus = [c for c in _cpus if c in self._little_cpus]
         _cpus = big_cpus + little_cpus
 
         # Precompute active and total time for each CPU
@@ -412,6 +414,9 @@ class FrequencyAnalysis(AnalysisModule):
             return
         if not self._trace.hasEvents('cpu_idle'):
             self._log.warning('Events [cpu_idle] not found, plot DISABLED!')
+            return
+        if 'clusters' not in self._platform:
+            self._log.warning('No platform cluster info. Plot DISABLED!')
             return
 
         # Assumption: all CPUs in a cluster run at the same frequency, i.e. the
--- a/libs/utils/analysis/tasks_analysis.py
+++ b/libs/utils/analysis/tasks_analysis.py
@@ -450,14 +450,10 @@ class TasksAnalysis(AnalysisModule):
 
             # Get per cluster wakeup events
             df = self._dfg_trace_event('sched_wakeup_new')
-            big_frequent = (
-                    (df.target_cpu.isin(self._big_cpus))
-                    )
+            big_frequent = df.target_cpu.isin(self._big_cpus)
             ntbc = df[big_frequent]
             ntbc_count = len(ntbc)
-            little_frequent = (
-                    (df.target_cpu.isin(self._little_cpus))
-                    )
+            little_frequent = df.target_cpu.isin(self._little_cpus)
             ntlc = df[little_frequent];
             ntlc_count = len(ntlc)
 
@@ -567,14 +563,11 @@ class TasksAnalysis(AnalysisModule):
         # Add column of expected cluster depending on:
         # a) task utilization value
         # b) capacity of the selected cluster
-        bu_bc = ( \
-                (df['util_avg'] > self._little_cap) & \
-                (df['cpu'].isin(self._big_cpus))
-            )
-        su_lc = ( \
-                (df['util_avg'] <= self._little_cap) & \
-                (df['cpu'].isin(self._little_cpus))
-            )
+        bu_bc = ((df['util_avg'] > self._little_cap) &
+                 (df['cpu'].isin(self._big_cpus)))
+        su_lc = ((df['util_avg'] <= self._little_cap) &
+                 (df['cpu'].isin(self._little_cpus)))
+
         # The Cluster CAPacity Matches the UTILization (ccap_mutil) iff:
         # - tasks with util_avg  > little_cap are running on a BIG cpu
         # - tasks with util_avg <= little_cap are running on a LITTLe cpu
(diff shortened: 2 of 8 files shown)


ionela-voinescu:
I suppose the purpose of this PR is to nudge a bigger effort to clean up the use of bigLITTLE data when it's not there. My primary question is how big should the effort be? A simpler solution is to constrain the use of bigLITTLE data for bigLITTLE platforms, as you've done here, but do it for all files that use it (freq_analysis, cpu_analysis, etc.). A more complicated but better solution is to drop all bigLITTLE references and start referring to them as generic (asymmetric) clusters. This would prepare the ground for tri clusters and beyond. Am I over-complicating the problem?

bjackman:
No, not over-complicating the problem IMO, we do need to use a more generic approach wherever possible. This is just some short-term fixes to stop people getting `KeyError`s all the time :D

ionela-voinescu:
Okay, then from my point of view you can merge this as it is. Looking over the code I don't believe it will break anything, it just will not be a complete solution and other PRs will have to be submitted either implementing the generic solution or fixing the other uses of bigLITTLE data.

bjackman:
Updated this with some more goodies. Def needs a cleanup and probably duplicating some effort.. CC @joelagnel

libs/utils/trace.py
```diff
@@ -519,14 +519,19 @@ def _dfg_functions_stats(self, functions=None):
 ###############################################################################
 # Trace Events Sanitize Methods
 ###############################################################################
+    @staticmethod
```
derkling: Why static?
Why not using `self.platform` instead of an explicit param?

derkling: I would also make this a public API... it can probably be useful in some analysis methods.

bjackman: Sounds good.

libs/utils/trace.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, platform, data_dir, events,
                  plots_prefix=''):
 
         # The platform used to run the experiments
-        self.platform = platform
+        self.platform = platform or {}
```
bjackman: Yep, that's what this is trying to do (i.e. allow `Trace(None, path, events)` in which case we set `self.platform = {}`)

libs/utils/trace.py
```diff
@@ -693,7 +697,8 @@ def _sanitize_CpuFrequency(self):
         Verify that all platform reported clusters are frequency coherent (i.e.
         frequency scaling is performed at a cluster level).
         """
-        if not self.hasEvents('cpu_frequency_devlib'):
+        if not self.hasEvents('cpu_frequency_devlib') \
+           or 'clusters' not in (self.platform or {}):
```
bjackman: Yep, good spot.

libs/utils/analysis_module.py
```diff
@@ -55,14 +55,16 @@ def __init__(self, trace):
...
-        if 'big' in self._platform['clusters']:
-            self._log.debug('Parsing big.LITTLE system clusters')
-            self._big_cpus = self._platform['clusters']['big']
-            self._little_cpus = self._platform['clusters']['little']
-        else:
-            self._log.debug('Parsing SMP clusters')
-            for cid in self._platform['clusters']:
-                self._big_cpus.append(self._platform['clusters'][cid])
+
+        if 'clusters' in self._platform:
```
bjackman: Yeah, sounds good, as I attempted to do in tasks_analysis.

libs/utils/analysis/tasks_analysis.py
```diff
@@ -708,16 +708,24 @@ def _plotTaskResidencies(self, axes, tid, signals, is_last=False):
...
+                cdata = data[data.cluster == clabel]
+                if len(cdata) > 0:
+                    cdata.plot(ax=axes, style=[ccolor+'+'], legend=False)
+
         # Y Axis - placeholders for legend, acutal CPUs. topmost empty lane
-        cpus = [str(n) for n in range(self._platform['cpus_count'])]
+        if 'cpus_count' in self._platform:
+            cpus = range(self._platform['cpus_count'])
+        else:
+            cpus = util_df.cpu.unique()
```
bjackman: Yep, good idea.

bjackman:
Thanks for the review. Hopefully can make time to clean up this week.

libs/utils/trace.py
bjackman: OH.. do you mean actually re-order the parameters to have `platform` at the end? I do agree that would be a better API but I dunno about making such a big breaking change in arguably the most important function in LISA...

joelagnel:
Nice!! Still on the road but will post comment soon. Thanks for the effort.

libs/utils/trace.py
derkling: I would say we can warn the use by checking that the first parameter, which will be `data_dir` is a string and not a dict... if it should be a dict we can show a DEPRECATED warning which notify the user about the new required interface and raise a ValueError exception asking for fixing the call.

Do you think this can work?

I think it's worth the change... in any case we wanna update that API to mandatory require only a trace file.

bjackman: Good idea that could definitely mitigate the cost. Anyway I think let's keep it for a separate PR, here I'll just try to allow the first param to be None

bjackman:
OK, I updated again with hopefully a slightly cleaner stack but def still needs cleanup and testing :( sorry can't spend more time on this right now

bjackman:
I have some half-baked unit test to provide some basic coverage here but need to make it cleaner

bjackman:
More progress! Cleaned up the stack a bit and started adding some ~~~tests~~~, it's good up to the "idle_analysis" one, after that the commits need a little more time. But, close!

libs/utils/analysis/frequency_analysis.py
```diff
@@ -133,13 +133,13 @@ def plotClusterFrequencies(self, title='Clusters Frequencies'):
 
         # Extract LITTLE and big clusters frequencies
         # and scale them to [MHz]
-        if len(self._platform['clusters']['little']):
-            lfreq = df[df.cpu == self._platform['clusters']['little'][-1]]
+        if self._little_cpus:
+            lfreq = df[df.cpu == self._little_cpus[-1]]
             lfreq['frequency'] = lfreq['frequency']/1e3
         else:
             lfreq = []
```
mdigiorgio: In the if branch, `lfreq` is a dataframe while here is set to an empty list. Sholdn't we set it to an empty dataframe? Just to avoid confusion...

libs/utils/analysis/frequency_analysis.py
```diff
@@ -133,13 +133,13 @@ def plotClusterFrequencies(self, title='Clusters Frequencies'):
...
-        if len(self._platform['clusters']['big']):
-            bfreq = df[df.cpu == self._platform['clusters']['big'][-1]]
+        if self._big_cpus:
+            bfreq = df[df.cpu == self._big_cpus[-1]]
             bfreq['frequency'] = bfreq['frequency']/1e3
         else:
             bfreq = []
```
mdigiorgio: Same here

libs/utils/analysis/frequency_analysis.py
```diff
@@ -296,14 +296,17 @@ def plotCPUFrequencies(self, cpus=None):
...
-                )
-                break
+            if 'clusters' in self._platform:
+                for cluster,cpus in self._platform['clusters'].iteritems():
+                    if cpu_id not in cpus:
+                        continue
+                    freqs = self._platform['freqs'][cluster]
+                    break
+            else:
+                freqs = df['frequency'].unique()
```
mdigiorgio: Why don't we always use this last line without the above checks? Frequencies specified in `platform` should be the the same as the ones you get in the trace or more, in which case it means you do not have events for the other frequencies so you don't plot them.

tests/lisa/test_trace.py
```diff
@@ -21,6 +21,23 @@
...
+# shouldn't matter.
+#
+
+def _event_common(_comm, _pid, _cpu, timestamp):
+    return "{_comm}-{_pid}  [{_cpu:0>3}] {timestamp}: ".format(**locals())
+
+def sched_switch(timestamp, cpu,
+                 prev_comm, prev_pid, prev_state, next_comm, next_pid):
+    return (_event_common(prev_comm, prev_pid, cpu, timestamp) + "sched_switch: "
+            "prev_comm={prev_comm} prev_pid={prev_pid} prev_prio=120 prev_state={prev_state} "
```
mdigiorgio: Shouldn't `prev_prio` and `next_prio` be parameters too?

libs/utils/analysis/frequency_analysis.py
bjackman: Maybe, but that's nothing to do with this PR. AFAICT it doesn't affect the return value/plot output.

libs/utils/analysis/frequency_analysis.py
bjackman: Hmm, I disagree - if you know the platform goes up to 2GHz and the trace was always at 1GHz then you'd like the plot to be around 50% up the axes (`freqs` is used to set the y-axis limits).

tests/lisa/test_trace.py
bjackman: I didn't include it because it doesn't make any difference to the code under test. If it became necessary we could always add it earlier.

The main question about this code is.. do you think the use of this `sched_switch` helper function actually makes the code easier or harder to read? I cannot decide.

bjackman:
Pushed an updated version. Could do with more test coverage but I don't think I'm gona have time for that.

libs/utils/analysis/frequency_analysis.py
mdigiorgio: Mmm, right. Yeah, in that case we can't simplify this code...

tests/lisa/test_trace.py
mdigiorgio: Since we are talking about tests, it's probably a lot more clear to just have the string with the events in the test itself. I think it makes it much easier to read the test, otherwise one would have to go through the helpers to see what they produce.

bjackman: OK, you're probably right - this is over-complicating things. I'll revert to just having the traces as text.

mdigiorgio: Can you also merge together the last two? They have the same commit message and they modify the same file.

bjackman:
Pushed a new update, now the trace data is just a string in the test file instead of using the convoluted helper functions. Also squashed stuff as necessary.

mdigiorgio:
LGTM :+1:

valschneider:
Had a look over the changes, everything looks sane to me.

bjackman:
having fun with matplotlib

bjackman:
repro'd the travis error by running the tests in a vagrant box with this Vagrantfile

```
Vagrant.configure("2") do |config|
  config.vm.box = "ubuntu/trusty64"

  config.vm.provider "virtualbox" do |vb|
    vb.memory = "2048"
  end

  config.vm.provision "shell", inline: <<-SHELL
    apt-get update
    apt-get install -y python-pip pkg-config python-dev
    pip install --upgrade pip
    pip install --upgrade trappy bart-py devlib psutil wrapt matplotlib nose
  SHELL
end
```

bjackman:
fixed it

PR #343, opened 2017-03-24
libs/wlgen: Fix governor save/restore in rtapp calibration

valschneider:
Governor was set to "performance" for all cpus during the rtapp
calibration, but the previous governors weren't restored after
that.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -98,7 +98,15 @@ class RTA(Workload):
         # Setup logging
         log = logging.getLogger('RTApp')
 
-        # target.cpufreq.save_governors()
+        # Save previous governors
+        cpus = target.list_online_cpus()
+        old_governors = {}
+
+        for cpu in cpus:
+            domain = tuple(target.cpufreq.get_domain_cpus(cpu))
+            if domain not in old_governors:
+                old_governors[domain] = target.cpufreq.get_governor(cpu)
+
         target.cpufreq.set_all_governors('performance')
 
         for cpu in target.list_online_cpus():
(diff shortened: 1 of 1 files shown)


libs/wlgen/wlgen/rta.py
```diff
@@ -133,7 +133,8 @@ def calibrate(target):
                 pload[cpu] = int(pload_match.group(1))
                 log.debug('>>> cpu%d: %d', cpu, pload[cpu])
 
-        # target.cpufreq.load_governors()
+        for cpu, governor in old_governors.iteritems():
+            target.cpufreq.set_governor('cpu{}'.format(cpu), governor)
```
derkling: Here you can use the `int` directly, perhaps the documentation should be updated in devlib but it does already that conversion inside.

libs/wlgen/wlgen/rta.py
```diff
@@ -98,7 +98,7 @@ def calibrate(target):
         # Setup logging
         log = logging.getLogger('RTApp')
 
-        # target.cpufreq.save_governors()
+        old_governors = target.cpufreq.get_all_governors()
```
derkling: nit pick: would save save chars by removing the `old_` prefix

bjackman: I disagree! There's no shortage of characters here and `old_governors` is significantly more accurate IMO.

libs/wlgen/wlgen/rta.py
bjackman: Possible suggestion: Iterating over all the CPUs can take quite a while. The cpufreq module has `get_domain_cpus` which exposes the frequency domains (cpufreq policies). If you set the governor for 1 CPU in a domain it's also set for all the other CPUs on the domain. So we can avoid some devlib calls.

Not really sure if it's worth the hassle, up to you guys.

valschneider: @derkling I actually had to add the string prefix because I was getting an error - I thought devlib wasn't doing the conversion but as you say it actually does it. The value stored in the dict probably is a string, so the conversion is skipped devlib-side.

bjackman: I guess we should fix this in devlib then, i.e. make `get_all_governors` return a dict with `int`s for keys. I suggest submitting a PR and seeing what @setrofim says

valschneider:
As per @bjackman's comment, cpu domains are used to reduce the amount of set_governor() calls.

bjackman:
Nice, LGTM!

bjackman:
Actually, tiny nitpick.. could you add a comment saying something like "To save time, only save and restore the governor for one CPU in each clock domain"? I could imagine this code looking slightly obscure after 6 months not thinking about it, a little pointer like that would instantly make it simple.

valschneider:
Right, I agree with you on that. Comment added right before the governor restoration:
"Setting a governor for a cpu will set it for all cpus in the same clock domain, so only restoring the governor of one cpu per domain is enough to restore all of the previous governors"

bjackman:
OK this looks great to me. @derkling I'll leave it to you to hit merge if you're happy with it.

PR #344, opened 2017-03-26
android: add automatic display on and use in benchmarks

joelagnel:
- Add Screen on/off and unlock support
- Use it in uibench to unlock and turn on screen

### DIFF
--- a/libs/utils/android/screen.py
+++ b/libs/utils/android/screen.py
@@ -16,6 +16,7 @@
 #
 
 import logging
+from system import System
 
 class Screen(object):
     """
@@ -135,5 +136,21 @@ class Screen(object):
         Get screen density of the device.
         """
         return target.execute('getprop ro.sf.lcd_density')
+
+    @staticmethod
+    def set_screen(target, on=True):
+        log = logging.getLogger('Screen')
+        if not on:
+            log.info('Setting screen OFF')
+            System.sleep(target)
+            return
+        log.info('Setting screen ON')
+        System.wakeup(target)
+
+    @staticmethod
+    def unlock(target):
+       Screen.set_screen(target, on=True)
+       System.menu(target)
+       System.home(target)
 
 # vim :set tabstop=4 shiftwidth=4 expandtab
--- a/libs/utils/android/workloads/uibench.py
+++ b/libs/utils/android/workloads/uibench.py
@@ -85,9 +85,8 @@ class UiBench(Workload):
         self.out_dir = out_dir
         self.collect = collect
 
-        # Press Back button to be sure we run the video from the start
-        System.menu(self._target)
-        System.back(self._target)
+        # Unlock device screen (assume no password required)
+        Screen.unlock(self._target)
 
         # Close and clear application
         System.force_stop(self._target, self.package, clear=True)
--- a/libs/utils/android/workloads/youtube.py
+++ b/libs/utils/android/workloads/youtube.py
@@ -66,9 +66,10 @@ class YouTube(Workload):
         self.collect = collect
 
         # Unlock device screen (assume no password required)
-        System.menu(self._target)
-        # Press Back button to be sure we run the video from the start
-        System.back(self._target)
+        Screen.unlock(self._target)
+
+        # Stop youtube if already running
+        System.force_stop(self._target, self.package)
 
         # Use the monkey tool to start YouTube without playing any video.
         # This allows to subsequently set the screen orientation to LANDSCAPE
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -255,6 +255,26 @@ class System(object):
         target.execute('input keyevent KEYCODE_BACK')
 
     @staticmethod
+    def wakeup(target):
+        """
+        Wake up the system if its sleeping
+
+        :param target: instance of devlib Android target
+        :type target: devlib.target.AndroidTarget
+        """
+        target.execute('input keyevent KEYCODE_WAKEUP')
+
+    @staticmethod
+    def sleep(target):
+        """
+        Make system sleep if its awake
+
+        :param target: instance of devlib Android target
+        :type target: devlib.target.AndroidTarget
+        """
+        target.execute('input keyevent KEYCODE_SLEEP')
+
+    @staticmethod
     def gfxinfo_reset(target, apk_name):
         """
         Reset gfxinfo frame statistics for a given app.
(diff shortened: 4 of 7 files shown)


bjackman:
Would prefer one commit for libs/utils/android and one commit for UiBench. Other than that LGTM, but will wait for review from @ionela-voinescu or @derkling as they are more familiar with this code.

joelagnel:
Guys, another comments here?

libs/utils/android/screen.py
```diff
@@ -136,4 +137,25 @@ def get_screen_density(target):
         """
         return target.execute('getprop ro.sf.lcd_density')
 
+    @staticmethod
+    def is_screen_on(target):
+	cmdout = target.execute('dumpsys input_method')
+	return ('mInteractive=true' in cmdout)
+
+    @staticmethod
+    def set_screen_on(target):
```
derkling: Moreover, I've found these two guys:
https://developer.android.com/reference/android/view/KeyEvent.html#KEYCODE_WAKEUP
https://developer.android.com/reference/android/view/KeyEvent.html#KEYCODE_SLEEP

Thus perhaps we can avoid the "expensive" check with dumpsys and go straight with these key events depending on the value of the `on` parameter.

libs/utils/android/screen.py
```diff
@@ -136,4 +137,25 @@ def get_screen_density(target):
...
+        log = logging.getLogger('Screen')
+        if Screen.is_screen_on(target) == False:
+            log.info('Screen is off, turning on')
+            System.power(target)
+
+    @staticmethod
+    def unlock(target):
+       Screen.set_screen_on(target)
+       target.execute('input keyevent 82')
```
derkling: This is `System.menu()`

libs/utils/android/screen.py
```diff
@@ -136,4 +137,25 @@ def get_screen_density(target):
...
+       # Incase screen was already unlocked, code 82
+       # may cause it zoom out, in this case just issue
+       # back key to restore it
+       System.back(target)
```
derkling: Perhaps this is not working in case the scree is already on and an app runnning, we are not granted to be in the home screen.

I think this sequence should be more robust:
```python
System.set_screen(on=True)
System.menu()
System.home()
```

This should grant to unlock the device and move back to the home screen... at least it did for all the starting conditions I've tested.

joelagnel:
Patrick, thanks for the great review, I made changes, please review and merge, thanks

libs/utils/android/screen.py
```diff
@@ -136,4 +137,20 @@ def get_screen_density(target):
         """
         return target.execute('getprop ro.sf.lcd_density')
 
+    @staticmethod
+    def set_screen(target, on=True):
+        log = logging.getLogger('Screen')
+        if on == True:
+            target.execute('input keyevent KEYCODE_WAKE')
```
derkling: I would really like to have this keyevent (and the following one) exposed as System APIs.
Can you add these two calls to System in a patch before this one?

```python
System.wake()
System.sleep()
```

libs/utils/android/screen.py
```diff
@@ -136,4 +137,20 @@ def get_screen_density(target):
...
+            log.info('Setting screen ON')
+        else:
```
derkling: I would prefer:

```python
if not on:
    System.sleep()
    log.info('Setting screen OFF')
    return

System.wake()
log.info('Setting screen ON')
```

Which better match the API signature where also the non default case returns immediately.

libs/utils/android/workloads/uibench.py
```diff
@@ -85,7 +85,7 @@ def run(self, out_dir, test_name, duration_s, collect=''):
         self.out_dir = out_dir
         self.collect = collect
 
-        # Press Back button to be sure we run the video from the start
+        Screen.unlock(self._target)
         System.menu(self._target)
```
derkling: A think you can make even more nicely red coloured this patch by removing this and the next call... aren't they already done by the new `Screen.unlock()` call?

derkling: Can you also please check if there are similar patterns in the other `libs/utils/android/workloads/*.py` and do the same change in all of them in this patch (which will became a more generic "android: use a more generic display unlocking function"

libs/utils/android/screen.py
joelagnel: I thought you preferred to send the event directly from Screen, I can certainly add a different path with that for System,... actually that's what I was going to do too.

libs/utils/android/screen.py
joelagnel: Ok.

libs/utils/android/workloads/uibench.py
joelagnel: Re: First comment, there's some weirdness where if I don't do it in unlock() then the screen doesn't unlock. I didn't get too much time last time around to fix that, but I will investigate more today.

Re: Second comment, I normally don't push changes without testing them so I didn't change what I wasn't using ;-) although, now that you mentioned it, I will try to include it in this patch with some amount of testing

joelagnel: Found the reason, and sorry for the sloppy job. Fixing it, cheers

joelagnel: About second comment, I will submit the other workload changes in other patch series please, so that this can be merged first. thanks

joelagnel:
Hope this set is better, thanks so much for your thorough review :-) :+1:

libs/utils/android/workloads/youtube.py
```diff
@@ -66,9 +66,10 @@ def run(self, out_dir, video_url, video_duration_s, collect=''):
         self.collect = collect
 
         # Unlock device screen (assume no password required)
-        System.menu(self._target)
-        # Press Back button to be sure we run the video from the start
-        System.back(self._target)
+        Screen.unlock(self._target)
+
+        # Stop youtube if already running
+        System.force_stop(self._target, self.package)
```
derkling: Should not this better go before the unlock?

However, I would prefer to keep this chunk into a separate patch, which follow the one updating the unlock procedure. At the end that's required just by Youtube and to grant that we start from the beginning.

joelagnel: I don't think it matters whether it is before or after, so I prefer to leave this as it is in the patch.

joelagnel:
Patrick? I pushed an update.

PR #348, opened 2017-03-29
LisaTest: Improve running tests with multiple iterations

bjackman:
Currently when you run a LisaTest with experiments_conf['iterations']!=0, the
test assertions are made for every experiment and the test is failed as soon as
any experiment produces a failure. Only the first failure encountered is
displayed.

With this patch all the results for each wload/conf pair are aggregated.

The problem is only solved across iterations: if multiple wloads or confs fail,
only the first wload/conf pair that was found to have failed will be
displayed.

### DIFF
--- a/libs/utils/test.py
+++ b/libs/utils/test.py
@@ -56,6 +56,9 @@ class LisaTest(unittest.TestCase):
     experiments_conf = None
     """Override this with a dictionary or JSON path to configure the Executor"""
 
+    permitted_fail_pct = 0
+    """The percentage of iterations of each test that may be permitted to fail"""
+
     @classmethod
     def _getTestConf(cls):
         if cls.test_conf is None:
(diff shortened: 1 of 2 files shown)


libs/utils/test.py
```diff
@@ -213,16 +213,25 @@ def experiment_test(wrapped_test, instance, args, kwargs):
...
         except AssertionError as e:
             trace_relpath = os.path.join(experiment.out_dir, "trace.dat")
-            add_msg = "\n\tCheck trace file: " + os.path.abspath(trace_relpath)
-            orig_msg = e.args[0] if len(e.args) else ""
-            e.args = (orig_msg + add_msg,) + e.args[1:]
-            raise
+            add_msg = "Check trace file: " + os.path.abspath(trace_relpath)
+            results[tag].append(str(e) + "\n\t" +  add_msg)
+        else:
+            results[tag].append(None)
```
derkling: I guess this can be required to account for how many tests we executed for a certain wload/configurations... but is not that number exactly the configured number of iterations?

libs/utils/test.py
```diff
@@ -213,16 +213,25 @@ def experiment_test(wrapped_test, instance, args, kwargs):
...
+
+    for tag, tag_results in results.iteritems():
+        fails = [r for r in tag_results if r is not None]
```
derkling: If number of iterations can be used for the formatting of the following `AssertionError` message, can't we just avoid to append `None` at line 228 above and check here just for `len(results[tag]) > 0`?

libs/utils/test.py
```diff
@@ -213,16 +213,25 @@ def experiment_test(wrapped_test, instance, args, kwargs):
...
+        if fails:
```
derkling: Here we should probably add a configuration option to defined an "acceptable number of failures" and raise just in that case, otherwise we can just report a warning message to point the user to failing traces.

libs/utils/test.py
bjackman: Yep, I think you're right, thanks :)

libs/utils/test.py
bjackman: Agree, I have actually implemented that, but I want to add it separately - this one just changes output format. Agree?

derkling: Can we put it in the same PR?

bjackman: Sure

bjackman:
Pushed a new version that uses the `Executor.iterations`. I think it's cleaner, it does mean making that attribute public but I don't think that's a big problem.

ionela-voinescu:
Looks good to me!

PR #349, opened 2017-03-30
Fixes for Capacity Capping EAS test

bjackman:
Two fixes, both issues can be summarised as: the code currently assumes that target operations (starting the workload, capping the frequency) take 0 time. 

@JaviMerino "requesting a review" from you because you wrote this originally, but DW I'm not expecting you to spend your time on this unless you really feel like it :)

### DIFF
--- a/tests/eas/capacity_capping.py
+++ b/tests/eas/capacity_capping.py
@@ -131,11 +131,22 @@ class CapacityCappingTest(unittest.TestCase):
         wload.run(out_dir=cls.env.res_dir, background=True)
         time.sleep(phase_duration)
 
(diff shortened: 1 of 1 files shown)


derkling:
LGTM

bjackman:
BTW, I think a more "proper" solution to the problem currently solved with `write_latency` would be to check the traces to figure out when the capacity was really capped/un-capped, but we don't have time to implement that solution and this one seems to work fine.

bjackman:
Pushed a version, as per @valschneider's out-of-band review, that raises a helpful error if the `write_value` latency is greater than the `phase_duration` (otherwise we call `time.sleep` with a negative number and get a useless "IO Error").

PR #351, opened 2017-03-30
Tweaks for the Generic test suite

bjackman:
- Enable `cpu_frequency` event.
- Disable schedutil ratelimiting.
- Automatically pick sched or schedutil governor if it's available, preferring schedutil.


bjackman:
Removed the one that disables schedutil ratelimits. Instead https://github.com/ARM-software/lisa/pull/352 can highlight if the limits are too long.

PR #354, opened 2017-04-04
Geekbench corporate

valschneider:
I'm not too keen on the naming for the TAB & ENTER input methods. I think press_tab() and press_enter() would've been better, but I also wanted to keep them aligned with the other methods - home(), back()... Feedback on that would be nice.

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -255,6 +255,60 @@ class System(object):
         target.execute('input keyevent KEYCODE_BACK')
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


valschneider:
Set airplane mode during the benchmark

derkling:
Is that really working on more than one single device?
Otherwise, we should report it as "not yet ready to merge"

valschneider:
Well, it does work on both HiKey & Pixel.

bjackman:
This is now supported by wltests

PR #355, opened 2017-04-06
trace: Don't report cpu_frequency if no events from kernel

bjackman:
devlib always introduces fake cpu_frequency events (cpu_frequency_devlib). If
the kernel doesn't produce any 'real' cpu_frequency events the current behaviour
is to use only the devlib ones.

This is misleading as it gives the impression that we have frequency data where
we do not. Instead, if there are no events from the kernel, don't report any
frequency data.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -683,7 +683,8 @@ class Trace(object):
         Verify that all platform reported clusters are frequency coherent (i.e.
         frequency scaling is performed at a cluster level).
         """
-        if not self.hasEvents('cpu_frequency_devlib'):
+        if not self.hasEvents('cpu_frequency_devlib') or \
+           not self.hasEvents('cpu_frequency'):
             return
 
         devlib_freq = self._dfg_trace_event('cpu_frequency_devlib')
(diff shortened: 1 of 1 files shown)


derkling:
Always report cpu_frequency events was an intended behaviour. For example we want to know the frequency a test has been executed when the userspace/performance/powersave governors are in use.

Why should be an issue to have the impression to have frequency data available?
If a "static" governor is in use, the reported frequencies will be always the same and we will have exactly two time as many events as the CPUs count. Can't we use this invariant as a way to identify if we have frequency events generated by the kernel?

bjackman:
Ah yeah, I didn't think about this clearly - we need the devlib events when the trace is enabled but doesn't produce anything. My problem is when the frequency is changing but the events weren't enabled. I think we can detect this properly, will go away and figure it out.

> Why should be an issue to have the impression to have frequency data available?

Because from both an interactive and programmatic point of view it's sensible to see a DataFrame or a plot with 0 or 1 changes in and assume that we can infer the frequency didn't change, which isn't necessarily true.

derkling:
If you do not get frequency switch events it's because: a) you did not enabled them or b) because the OPP has never changed. In the first case it's a bug in your test, in the second it's not an issue at all.

Thus, still do not get why having just fake events (even when OPPs do not change) should be an issue...

bjackman:
> In the first case it's a bug in your test

In an automated test this might be considered a bug but in an interactive session it's a usability issue. Anyway I'll figure this out and post a new PR, this one is the wrong approach.

PR #356, opened 2017-04-07
libs/utils/android: Add systrace configuration

valschneider:
For workloads with collect='systrace', systrace can now be
configured in a similar manner to ftrace. Simply add a systrace
dict in the TestEnv conf like so:

"systrace" : {
    "events" : [
        "gfx",
        "view",
        "sched",
        "freq",
        "idle"
    ],
    "buffsize" : 100 * 1024
}

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -35,26 +35,43 @@ class System(object):
     """
 
     @staticmethod
-    def systrace_start(target, trace_file, time=None,
+    def systrace_start(env, trace_file, time=None,
                        events=['gfx', 'view', 'sched', 'freq', 'idle']):
+        """
+        Start tracing using systrace
+
+        :param env: Target test environment
+        :type env: TestEnv
+
+        :param trace_file: Name of the trace output
+        :type trace_file: str
+
+        :param time: Tracing duration in seconds. If no value is passed,
+                     will keep tracing until tracingStop() is called.
+        :type time: int
+
+        :param events: Events to trace
+        :type events: str list
+        """
 
         log = logging.getLogger('System')
 
         # Check which systrace binary is available under CATAPULT_HOME
         for systrace in ['systrace.py', 'run_systrace.py']:
-                systrace_path = os.path.join(target.CATAPULT_HOME, 'systrace',
+                systrace_path = os.path.join(env.CATAPULT_HOME, 'systrace',
                                              'systrace', systrace)
                 if os.path.isfile(systrace_path):
                         break
         else:
                 log.warning("Systrace binary not available under CATAPULT_HOME: %s!",
-                            target.CATAPULT_HOME)
+                            env.CATAPULT_HOME)
                 return None
 
         #  Format the command according to the specified arguments
-        device = target.conf.get('device', '')
+        device = env.conf.get('device', '')
         if device:
             device = "-e {}".format(device)
+
         systrace_pattern = "{} {} -o {} {}"
         trace_cmd = systrace_pattern.format(systrace_path, device,
                                             trace_file, " ".join(events))
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -24,6 +24,8 @@ import pexpect as pe
 from time import sleep
 
 GET_FRAMESTATS_CMD = 'shell dumpsys gfxinfo {} > {}'
+SYSTRACE_EVENTS_DEFAULT = ['gfx', 'view', 'sched', 'freq', 'idle']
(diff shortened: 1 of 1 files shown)


libs/utils/android/system.py
```diff
@@ -23,17 +23,28 @@
...
-    def systrace_start(target, trace_file, time=None,
-                       events=['gfx', 'view', 'sched', 'freq', 'idle']):
+    def systrace_start(target, trace_file, time=None):
 
         log = logging.getLogger('System')
+        conf = target.conf['systrace']
+
+        events = SYSTRACE_EVENTS_DEFAULT
+        if 'events' in conf:
+            events = conf['events']
```
bjackman: `events = conf.get('events', SYSTRACE_EVENTS_DEFAULT)` and similar below

libs/utils/android/system.py
```diff
@@ -23,17 +23,28 @@
...
     Collection of Android related services
     """
 
     @staticmethod
-    def systrace_start(target, trace_file, time=None,
-                       events=['gfx', 'view', 'sched', 'freq', 'idle']):
+    def systrace_start(target, trace_file, time=None):
 
         log = logging.getLogger('System')
+        conf = target.conf['systrace']
```
bjackman: I've just realised that this `target` is totally misnamed, it's actually a `TestEnv`. Can we refactor that first? In fact we have a lot of asymmetry here between systrace and ftrace :cry: . I think we need to spend some time to move this whole shebang into devlib anyway.

valschneider:
Renamed the TestEnv variable & cleaned up the conf dict access

bjackman:
LGTM, thanks. Since I haven't used this myself I'm not going to hit merge just yet but if nobody else does I'll try to test it.

libs/utils/android/system.py
```diff
@@ -23,15 +23,16 @@
...
 
 class System(object):
     """
     Collection of Android related services
     """
 
     @staticmethod
-    def systrace_start(env, trace_file, time=None,
-                       events=['gfx', 'view', 'sched', 'freq', 'idle']):
+    def systrace_start(env, trace_file, time=None):
```
derkling: Maybe we can keep backward compatible this API by still passing `events=None`?

If `None` events are specified at the call of this method, then we do the dance to load either the configuration provided ones of the default ones.

Thus priorities will be:
1. call parameter
2. configuration file
3. hard-coded defaults

valschneider: ATM `systrace_start()` is only used in workload.py by `tracingStart()`, and the call doesn't specify events - for now it seems all the tracing done with `collect='systrace'` would just use the default events. So unless I'm mistaken, it might be okay to remove `events`, which is redundant anyway with the configuration dict.

valschneider:
As per a discussion with @derkling, having a call parameter for events is useful for tests done without using the workload API. I updated the PR in regards to that.

libs/utils/android/system.py
```diff
@@ -45,12 +46,24 @@ def systrace_start(env, trace_file, time=None,
...
 
         log = logging.getLogger('System')
 
+        # Prioritize custom conf, then environment conf, then default conf
+        if conf is not None:
+            conf = conf
+        elif 'systrace' in env.conf and conf is None:
+            conf = env.conf['systrace']
+        else:
+            conf = {}
```
derkling: Should not this be enough?

```python
if not conf:
    conf = env.conf.get('systrace', {})
```

libs/utils/android/system.py
```diff
@@ -63,9 +76,9 @@ def systrace_start(env, trace_file, time=None,
                 return None
 
         #  Format the command according to the specified arguments
-        systrace_pattern = "{} -e {} -o {} {}"
+        systrace_pattern = "{} -e {} -o {} -b {} {}"
```
derkling: Not sure systrace can really use the buffer parameter, at least when atrace is used on the target side... did you already verified that?

valschneider: Without specifying the buffer size argument, systrace wouldn't record more than ~1 second of intensive workload. It did seem to work by specifying a bigger buffer size, so I suppose it does work ?

libs/utils/android/system.py
valschneider: Yes, that should work too. I liked the if/elif/else though because it's pretty clear what are the priorities (at least to me). On a sidenote, there's a refactoring derp here because `elif 'systrace' in env.conf and conf is None:` should just be `elif 'systrace in env.conf`.

valschneider:
Updated branch following @derkling's comments

PR #357, opened 2017-04-10
wlgen/rta: Fix calibration value string replacement

bjackman:
Currently, under the assumption that the calibration is an int, we don't quote
it. However it may be a string like "CPU0" in which case we _do_ need to quote it.

Also add a test for this case.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -305,6 +305,12 @@ class RTA(Workload):
         self.json = '{0:s}_{1:02d}.json'.format(self.name, self.exc_id)
         ofile = open(self.json, 'w')
         ifile = open(rtapp_conf, 'r')
+
+        # Calibration can either be a string like "CPU1" or an integer, if the
+        # former we need to quote it.
+        if type(calibration) != int:
+            calibration = '"{}"'.format(calibration)
+
         replacements = {
             '__DURATION__' : str(self.duration),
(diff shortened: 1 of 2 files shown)


derkling:
I think that API was meant to always replace `pvalue` with a proper calibration value, otherwise we do calibration at the beginning. If that's the case, why not just hard-coding the CPU in the JSON template?

bjackman:
Good point, we can just do calibration at the beginning and use that. Will update.

bjackman:
Actually on taking another look it turns out it uses "CPU1" for "profile" too, if you don't provide a calibration Running the full calibration is taken care of by the user of the class. Could change that if we wanted to but IMO it's fine as it is now.

libs/wlgen/wlgen/rta.py
```diff
@@ -292,6 +292,12 @@ def _confCustom(self):
         self.json = '{0:s}_{1:02d}.json'.format(self.name, self.exc_id)
         ofile = open(self.json, 'w')
         ifile = open(rtapp_conf, 'r')
+
+        # Calibration can either be a string like "CPU1" or an integer, if the
+        # latter we need to quote it.
```
derkling: you mean "`if the former...`", or... "`if *not* the latter`"...  right?

bjackman: Yep, thanks

bjackman:
Updated with comment fixed.

PR #359, opened 2017-04-10
lisa_shell: fix reported URL

derkling:
Jupyter notebook 4.3.1 introduces protection from cross-site request forgeries,
requiring API requests to either:
- originate from pages served by this server
  (validated with XSRF cookie and toekn)
- authenticate with a token

This patch adds support to generate a random token and report it in the
complete URL thus allowing the user to properly get the complete address
where her server is running.

Signed-off-by: Patrick Bellasi <patrick.bellasi@arm.com>

### DIFF
--- a/src/shell/lisa_shell
+++ b/src/shell/lisa_shell
@@ -185,15 +185,24 @@ fi
 PYDIR="$LISA_HOME/ipynb"
 LOGFILE="$PYDIR/server.log"
 PIDFILE="$PYDIR/server.pid"
+URLFILE="$PYDIR/server.url"
+
+# Generate server URL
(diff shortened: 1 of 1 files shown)


src/shell/lisa_shell
```diff
@@ -185,24 +185,36 @@ fi
...
+    xdg-open $(cat $URLFILE)
 	return 1
 fi
+
 # Start the server bindeed to the specified interface
-echo 'Starting IPython Notebook server...'
-echo "  IP Address :  http://$IPADDR:$PORT/"
+echo
+echo 'Notebook server configuration:'
+echo '  IP Address : ' $URL
```
bjackman: Nit - I guess we should change this to say "URL" instead of "IP address"

derkling:
@bjackman give it a try when you can and then merge... it's not urgent ;-)

PR #360, opened 2017-04-10
gfxinfo: Poll framestats in a thread

mcgeagh:
gfxinfo can be executed with an additional 'framestats' parameter that
gives detailed timestamp information for the last 120 frames.
Added a new parameter to expose this functionality.
As this only captures the last 120 frames, to get all frame data across
the workload we need to poll at least every 2 seconds (assuming 60Hz
vsync) and append the data to the same file. New start and stop
functions for doing this have been added to the workload class, similar
to how tracing is currently handled.

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -268,7 +268,7 @@ class System(object):
         target.execute('dumpsys gfxinfo {} reset'.format(apk_name))
 
     @staticmethod
-    def gfxinfo_get(target, apk_name, out_file):
+    def gfxinfo_get(target, apk_name, out_file, framestats=False):
         """
         Collect frame statistics for the given app.
 
@@ -280,9 +280,14 @@ class System(object):
 
         :param out_file: output file name
         :type out_file: str
+
+        :param framestats: collect framestats and *append* to out_file
+        :type framestats: bool
         """
-        adb_command(target.adb_name,
-                    GET_FRAMESTATS_CMD.format(apk_name, out_file))
+        cmd = GET_FRAMESTATS_CMD.format(apk_name, out_file)
+        if framestats:
+            cmd = cmd.replace('>', 'framestats >>')
+        adb_command(target.adb_name, cmd)
 
     @staticmethod
     def monkey(target, apk_name, event_count=1):
--- a/libs/utils/android/workload.py
+++ b/libs/utils/android/workload.py
@@ -18,6 +18,7 @@
 import logging
 import os
 import re
+import threading
 
 from . import System
 
@@ -42,6 +43,9 @@ class Workload(object):
         # Set of data reported in output of each run
         self.trace_file = None
         self.nrg_report = None
+
+        # Thread used for gfxinfo polling
+        self.gfxinfo_thread = None
 
     def _adb(self, cmd):
         return 'adb -s {} {}'.format(self._target.adb_name, cmd)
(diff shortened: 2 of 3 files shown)


bjackman:
Polling the device over ADB sounds like it would influence the behaviour quite a lot, have you been able to do any trace comparison with and without the polling? Ugly, but it might be less intrusive to spawn a script on the device that does the polling, then at least we avoid the ADB overhead :/

mcgeagh:
@bjackman is there a specific method for comparison you would like me to look into?

I have run a systrace capture for two subtest in UiBench (EditTextTypeActivity and ShadowGridActivity), for two cases - Default gfxinfo method (summary), and my new method (polling).

shadowgridactivity-pollingall0-cluster_freqs
shadowgridactivity-summaryall0-cluster_freqs
edittexttypeactivity-pollingall0-cluster_freqs
edittexttypeactivity-summaryall0-cluster_freqs

They look very similar to me (charts never look identical across multiple iterations with the same setup anyways)

Overview of the systrace files also look very similar.
Polling:
edittexttypeactivity-polling-appsystrace
Summary:
edittexttypeactivity-summary-appsystrace

bjackman:
My concern was that the USB driver + ADB daemon etc will add CPU load and disrupt the workload. It may be that we can't do anything about that.

Also, communicating with the target takes non-negligible time so it's not at all safe to assume that the `dumpsys gfxinfo` command really gets run every 2 seconds. We'll probably need to do it more frequently to avoid missing frames. So (sorry I'm not familiar with these tools) are we at risk of duplicating stats for a frame? I.e. if we aim to collect at t=0, t=2, t=4... but the first collection is delayed due to ADB communication time so that it happens at t=1, when we do a second collection at t=2 can we detect that we already collected stats for ~60 of the previous 120 frames?

mcgeagh:
The default is set to 2s as a compromise of collecting the vast majority of frames (in theory all, in practice potentially fewer) without having the adb overhead as much.
In my own tests, I have been setting it to 1s to have a good overlap to capture every frame when i need that info.

Dumpsys does not avoid duplication and will always return the last 120 frames regardless of the framerate or polling period you give it. However, duplicate frames are easy to remove as a post processing step as all timestamps are identical for a repeated frame. This technique is implemented in WA for example.

libs/utils/android/system.py
```diff
@@ -280,9 +280,14 @@ def gfxinfo_get(target, apk_name, out_file):
...
         :type out_file: str
+
+        :param framestats: collect framestats and *append* to out_file
+        :type framestats: bool
         """
-        adb_command(target.adb_name,
-                    GET_FRAMESTATS_CMD.format(apk_name, out_file))
+        cmd = GET_FRAMESTATS_CMD.format(apk_name, out_file)
+        if framestats:
+            cmd = cmd.replace('>', 'framestats >>')
```
bjackman: I don't like this use of `replace`, there's an invisible dependency where if you change `GET_FRAMESTATS_CMD` you break this code. Maybe instead have a function like `get_framestats_cmd(x)` (with a better name than `x`, not sure what the terminology should be though), that would be used like `get_framestats_cmd(['gfxinfo', 'framstats'])` here?

bjackman:
OK, that sounds good.

derkling:
Isn't that functionality something we can have in devlib?

If we need to periodically run a command on the target I would rather prefer to have a proper pair of shutil functions which allows to start a poller and stop it after a while by collecting and returning all the generated data like if they where collected in one single shot.

bjackman:
Devlib's [GfxInfoFramesInstrument](https://github.com/ARM-software/devlib/blob/master/devlib/instrument/frames.py#L58) now supoprts this via [GfxInfoFramesCollector](https://github.com/ARM-software/devlib/blob/master/devlib/utils/rendering.py#L174)

PR #363, opened 2017-04-11
system.py: gfxinfo_get uses target.execute

mcgeagh:
Redirection to host is unneccessary. Write to file via python instead

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -17,12 +17,9 @@
 
 import logging
 
-from devlib.utils.android import adb_command
 from devlib import TargetError
 import os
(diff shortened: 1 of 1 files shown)


bjackman:
Thanks!

bjackman:
Hey, sorry I never got round to merging this. It seems the gfxinfo support in LISA isn't really needed any more (it's in devlib now, so also works with WA3)  so do you mind if I kill this PR?

valschneider:
For reference: https://github.com/ARM-software/devlib/blob/a3f78cabc17ae2653e6b747207be1f3f27342aa0/devlib/utils/rendering.py#L164

PR #365, opened 2017-04-13
android/benchmark: add iterations command line parameter

jlelli:
It is usually handy to be able to run the same benchmark multiple times
(e.g., to do statistics on results).

Add a --iterations command line parameter to implement the feature.

### DIFF
--- a/libs/utils/android/benchmark.py
+++ b/libs/utils/android/benchmark.py
@@ -59,6 +59,9 @@ class LisaBenchmark(object):
     bm_collect = None
     """Override this with the set of data to collect during test exeution"""
 
+    bm_iterations = 1
+    """Override this with the desired number of iterations of the test"""
+
     def benchmarkInit(self):
         """
         Code executed before running the benchmark
@@ -106,6 +109,9 @@ class LisaBenchmark(object):
                 help='Set of metrics to collect, '
                      'e.g. "energy systrace_30" to sample energy and collect a 30s systrace, '
                      'if specified overrides test defaults')
+        parser.add_argument('--iterations', type=int,
+                default=1,
+                help='Number of iterations the same test has to be repeated for (default 1)')
 
         # Measurements settings
         parser.add_argument('--iio-channel-map', type=str,
@@ -132,6 +138,8 @@ class LisaBenchmark(object):
             self.bm_conf['results_dir'] = self.args.results_dir
         if self.args.collect:
             self.bm_collect = self.args.collect
+        if self.args.iterations:
+            self.bm_iterations = self.args.iterations
 
         # Override energy meter configuration
         if self.args.iio_channel_map:
@@ -203,9 +211,16 @@ class LisaBenchmark(object):
             raise
 
         self._log.info('=== Execution...')
-        self.wl.run(out_dir=self.out_dir,
-                    collect=self._getBmCollect(),
-                    **self.bm_params)
+        for iter_id in range(1, self.bm_iterations+1):
+            self._log.info('=== Iteration {}/{}...'.format(iter_id, self.bm_iterations))
+            out_dir = os.path.join(self.out_dir, "{:03d}".format(iter_id))
+            try:
+                os.makedirs(out_dir)
+            except: pass
+
+            self.wl.run(out_dir=out_dir,
+                        collect=self._getBmCollect(),
+                        **self.bm_params)
 
         self._log.info('=== Finalization...')
         self.benchmarkFinalize()
--- a/libs/utils/android/benchmark.py
+++ b/libs/utils/android/benchmark.py
@@ -62,6 +62,12 @@ class LisaBenchmark(object):
     bm_iterations = 1
     """Override this with the desired number of iterations of the test"""
 
+    bm_iterations_pause = 30
+    """
+    Override this with the desired amount of time (in seconds) to pause
+    for before each iteration
+    """
+
     def benchmarkInit(self):
         """
         Code executed before running the benchmark
@@ -112,6 +118,9 @@ class LisaBenchmark(object):
         parser.add_argument('--iterations', type=int,
                 default=1,
                 help='Number of iterations the same test has to be repeated for (default 1)')
+        parser.add_argument('--iterations-pause', type=int,
+                default=30,
+                help='Amount of time (in seconds) to pause for before each iteration (default 30s)')
 
         # Measurements settings
         parser.add_argument('--iio-channel-map', type=str,
@@ -140,6 +149,8 @@ class LisaBenchmark(object):
             self.bm_collect = self.args.collect
         if self.args.iterations:
             self.bm_iterations = self.args.iterations
+        if self.args.iterations_pause:
+            self.bm_iterations_pause = self.args.iterations_pause
 
         # Override energy meter configuration
         if self.args.iio_channel_map:
@@ -186,6 +197,14 @@ class LisaBenchmark(object):
             return ''
         return self.bm_collect
 
+    def _preRun(self):
+        """
+        Code executed before every iteration of the benchmark
+        """
+        self._log.info('Waiting {}[s] before executing iteration...'\
+                .format(self.bm_iterations_pause))
+        sleep(self.bm_iterations_pause)
+
     def __init__(self):
         """
         Set up logging and trigger running experiments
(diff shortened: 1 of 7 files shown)


libs/utils/android/benchmark.py
```diff
@@ -203,9 +211,15 @@ def __init__(self):
             raise
 
         self._log.info('=== Execution...')
-        self.wl.run(out_dir=self.out_dir,
-                    collect=self._getBmCollect(),
-                    **self.bm_params)
+        for iter in range(self.bm_iterations):
```
ionela-voinescu: It looks good, but I wonder how should we handle it if people would want at some point to reboot between iterations. At the moment we have the possibility to reboot between test run (with different governors and tests within the benchmark) but I wonder if we should use the reboot variable to control a potential reboot between iterations too. What do you think?

jlelli: Right. Actually also waiting some time between iterations might be something people might want to configure. We could set reasonable defaults and add two more command line options?

jlelli: Hummm.. your idea is certainly more elegant.
Couldn't we want however to be able to configure sleep/reboot from Jenkins as well though?
Maybe we can add both ways of configuring sleep/reboot?

ionela-voinescu: This gets complicated by the fact that most often you need a first reboot (to reboot with your kernel) and you can decide after if you need subsequent reboots (between your different tests or iterations of your tests).

I way I think this can work is:
 - The first boot is conditioned by the presence of the boot image as a command line argument.
    That will happen even if a "--reboot" is not passed (see below for meaning of this new argument)
 - Reboots between tests or iterations is controlled by an additional "--reboot" command line argument.
 - A possible delay  is conditioned by "--boot-timeout" (maybe we should change the string) even if "--reboot" is present. I'm wondering here if it's best to add a default value(60s) to this configured argument if --reboot is present. Because if people forget to pass this when they require reboot then we don't wait for the reboot to complete and the call to the benchmark will fail.

This will mean that we remove all handling of a reboot in the test scripts together with the delay after each test run, and we add the following to benchmark.py:
 - reboot_target in __init__ if a boot image is given.
 - a call of this function (maybe called prepare_run) that would include (default):
     - - call to reboot_target if --reboot is present
     - - wait for a period of time if present (boot_timeout)

Both --reboot and --boot_timeout can be configured from Jenkins.

derkling: What I don't like here is that we are not providing a simple support to just wait between two iterations if we do not need to reboot. Why not keeping these two functionalities separated?

For the time being the patch Juri is proposing can be extended with a simple `--interations_pause` parameter to provide a simple yet useful support to inject a small [s] pause between iterations.
For most of the tests we are running now that should be just good enough.

In a second phase we can add a proper rebooting support, with timeout, which is a different concept then a pause between iterations.

derkling: what about:

```python
for iter_id in range(1, self.bm_iterations+1):
```

So that we get more humans iterations ID which starts from 1... and we avoid the small +1 in the following loop?

libs/utils/android/benchmark.py
```diff
@@ -203,9 +211,15 @@ def __init__(self):
...
+            self._log.info('=== Iteration {}/{}...'.format(iter+1, self.bm_iterations))
+            out_dir = self.out_dir + '/{}/'.format(iter)
```
derkling:
```python
out_dir = os.path.join(self.out_dir, "{:03d}".format(iter_id))
```

libs/utils/android/benchmark.py
```diff
@@ -203,9 +211,15 @@ def __init__(self):
...
+            if not os.path.exists(out_dir):
```
derkling: Quite likely it does not exists already:

```python
try:
    os.makedirs(out_dir)
except: pass
```

tests/benchmarks/android_geekbench.py
```diff
@@ -64,19 +64,17 @@ class GeekbenchTest(LisaBenchmark):
     bm_collect = 'ftrace energy'
 
     def benchmarkInit(self):
+        super(GeekbenchTest, self).benchmarkInit()
         self.setupWorkload()
         self.setupGovernor()
-        if self.reboot:
-            self.reboot_target()
```
derkling: With the proposed modifications (e.g. properly placed private methods) we should both fix this horror (:heart:) as well as avoid the need to call the super-class implementation.

tests/benchmarks/android_geekbench.py
```diff
@@ -64,19 +64,17 @@ class GeekbenchTest(LisaBenchmark):
...
 
     def benchmarkFinalize(self):
         if self.delay_after_s:
             self._log.info("Waiting %d[s] before to continue...",
                            self.delay_after_s)
             sleep(self.delay_after_s)
 
-    def __init__(self, governor, test, reboot=False, delay_after_s=0):
-        self.reboot = reboot
```
derkling: In this implementation it was technically possible to reboot only on some instances... with the new API (which uses command line params) I think we miss that.

Well... maybe we can re-add it on-demand the first time we find it useful ;-)

libs/utils/android/benchmark.py
```diff
@@ -201,9 +211,15 @@ def _preRun(self):
         """
         Code executed before every iteration of the benchmark
         """
-        self._log.info('Waiting {}[s] before executing iteration...'\
-                .format(self.bm_iterations_pause))
-        sleep(self.bm_iterations_pause)
+        rebooted = False
+
+        if (self.bm_iterations_reboot):
```
derkling: nit-pick: we don't use parentesisi

libs/utils/android/benchmark.py
```diff
@@ -201,9 +211,15 @@ def _preRun(self):
...
+            rebooted = self.reboot_target()
+
+        if not rebooted:
+            self._log.info('Waiting {}[s] before executing iteration...'\
+                    .format(self.bm_iterations_pause))
```
derkling: nit-pcik: the `.` has to be aligned with the parentesis

libs/utils/android/benchmark.py
```diff
@@ -207,13 +212,22 @@ def _getBmCollect(self):
             return ''
         return self.bm_collect
 
+    def _preInit(self):
+        """
+        Code executed before running the benchmark
+        """
+        # If iterations_reboot is True we are going to reboot before the
+        # first iteration anyway.
+        if (self.bm_reboot and not self.bm_iterations_reboot):
```
derkling: nit-pick: parentesis

libs/utils/android/benchmark.py
```diff
@@ -207,13 +212,22 @@ def _getBmCollect(self):
...
+            self.reboot_target()
+
     def _preRun(self):
         """
         Code executed before every iteration of the benchmark
         """
         rebooted = False
 
-        if (self.bm_iterations_reboot):
+        if (self.bm_reboot and self.bm_iterations_reboot):
```
derkling: same here

tests/benchmarks/android_geekbench.py
```diff
@@ -161,21 +158,18 @@ def setupGovernor(self):
...
-do_reboot = True
 tests_remaining = len(governors) * len(tests)
 tests_completed = 0
 for governor in governors:
     for test in tests:
         tests_remaining -= 1
         delay_after_s = 30 if tests_remaining else 0
         try:
-            GeekbenchTest(governor, test, do_reboot, delay_after_s)
+            GeekbenchTest(governor, test, delay_after_s)
```
derkling: Mmm... with the new "iterations_pause" parameter are not we adding a pause before every iterations, even if it's just one?

If that's the case I think, with the new API you added, here we can get ride of `delay_after_s` as well... and the redefinition of the `benchmarkFinalize()` which are usually used just to add this sleep after a test has completed.

This would make this patch even more nicely red colored... and @JaviMerino (do you remember Javi?) will be one more time your best friend! ;-)

derkling:
Again, that version LGTM... let's wait for @ionela-voinescu to test/merge it.

ionela-voinescu:
It looks good to me. I've already integrated it with CI and I'm now running the final tests, that I expect to run smoothly. Thank you for the changes.

PR #366, opened 2017-04-13
rta: Fix setting up "calibration" field in rt-app global conf

bjackman:
Need to review this more carefully (maybe I can add a unit test for this) but submitting PR now because otherwise I'll definitely forget about this.

The current code breaks calibration for big.LITTLE systems if the 'bl' devlib module isn't loaded, hence marking this as a bug.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -184,104 +184,25 @@ class RTA(Workload):
             for line in self.output['executor'].split('\n'):
                 ofile.write(line+'\n')
 
-    def _getFirstBiggest(self, cpus):
-        # Non big.LITTLE system:
-        if 'bl' not in self.target.modules:
-            # return the first CPU of the last cluster
-            platform = self.target.platform
-            cluster_last = list(set(platform.core_clusters))[-1]
-            cluster_cpus = [cpu_id
-                    for cpu_id, cluster_id in enumerate(platform.core_clusters)
-                                           if cluster_id == cluster_last]
-            # If CPUs have been specified': return the fist in the last cluster
-            if cpus:
-                for cpu_id in cpus:
-                    if cpu_id in cluster_cpus:
-                        return cpu_id
-            # Otherwise just return the first cpu of the last cluster
-            return cluster_cpus[0]
-
-        # big.LITTLE system:
-        for c in cpus:
-             if c not in self.target.bl.bigs:
-                continue
-             return c
-        # Only LITTLE CPUs, thus:
-        #  return the first possible cpu
-        return cpus[0]
-
-    def _getFirstBig(self, cpus=None):
-        # Non big.LITTLE system:
-        if 'bl' not in self.target.modules:
-            return self._getFirstBiggest(cpus)
-        if cpus:
-            for c in cpus:
-                if c not in self.target.bl.bigs:
-                    continue
-                return c
-        # Only LITTLE CPUs, thus:
-        #  return the first big core of the system
-        if self.target.big_core:
-            # Big.LITTLE system
-            return self.target.bl.bigs[0]
-        return 0
-
-    def _getFirstLittle(self, cpus=None):
-        # Non big.LITTLE system:
-        if 'bl' not in self.target.modules:
-            # return the first CPU of the first cluster
-            platform = self.target.platform
-            cluster_first = list(set(platform.core_clusters))[0]
-            cluster_cpus = [cpu_id
-                    for cpu_id, cluster_id in enumerate(platform.core_clusters)
-                                           if cluster_id == cluster_first]
-            # If CPUs have been specified': return the fist in the first cluster
-            if cpus:
-                for cpu_id in cpus:
-                    if cpu_id in cluster_cpus:
-                        return cpu_id
-            # Otherwise just return the first cpu of the first cluster
-            return cluster_cpus[0]
-
-        # Try to return one LITTLE CPUs among the specified ones
-        if cpus:
-            for c in cpus:
-                if c not in self.target.bl.littles:
-                    continue
-                return c
-        # Only big CPUs, thus:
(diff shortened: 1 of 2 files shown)


libs/wlgen/wlgen/rta.py
```diff
@@ -184,47 +184,17 @@ def __postrun(self, params):
...
-        return 0
+        cpus = cpus or range(self.target.number_of_cpus)
+
+        if 'bl'in self.target.modules:
+            cluster = self.target.bl.bigs
+            candidates = sorted(set(self.target.bl.bigs).intersection(cpus))
+            if candidates:
+                return cadidates[0]
+
+        return cpus[0]
```
derkling: Isn't more safe in general to return `cpus[-1]`?
If we did not loaded the `bl` module we still have better chances to get a big CPU that way.

This would require also something like `s/_getFirstBig/_getBig/`

libs/wlgen/wlgen/rta.py
```diff
@@ -184,47 +184,17 @@ def __postrun(self, params):
...
-                return c
-        # Only LITTLE CPUs, thus:
-        #  return the first big core of the system
-        if self.target.big_core:
-            # Big.LITTLE system
-            return self.target.bl.bigs[0]
-        return 0
+        cpus = cpus or range(self.target.number_of_cpus)
+
+        if 'bl'in self.target.modules:
```
derkling: What about doing just:

```python
try:
    return sorted(set(self.target.bl.bigs).intersection(cpus))[-1]
except:
    return cpus[-1]
```

bjackman: I really don't like using exceptions that way unless you can at least catch a very precise exception class; it obscures the intention of the code and it hides errors (so if you misspell `intersection` you just silently get `cpus[-1]` every time) . I know we do this with `TargetError` sometimes but at least in that case we can usually consider that the "except" block is "exceptional" behaviour.

libs/wlgen/wlgen/rta.py
bjackman: Yep, that sounds like a good idea.

bjackman:
Thanks for review. I still need to spend a bit of time thinking about whether this breaks anything and how to test it.

libs/wlgen/wlgen/rta.py
derkling: Well, misspells can still happen in your `if` above :hammer: 

That was perhps just a more pythonic way to get the most common case done without conditional checks... possible exceptions can always be listed.

bjackman: But if you misspell `intersection` with an `if`, you get an `AttributeError`. With a `try` you just get broken behaviour. I think "pythonic" is very debatable here unless you consider "pythonic" to mean "hides common programming errorss", which would definitely be justifiable :tongue:

libs/wlgen/wlgen/rta.py
```diff
@@ -186,72 +186,28 @@ def __postrun(self, params):
...
-                target_cpu = self._getFirstBig()
-                self._log.debug('ref on big cpu: %d', target_cpu)
-            return target_cpu
-
-        # These options are selected only when RTApp has not been
-        # already calibrated
-        if self.cpus is None:
-            target_cpu = self._getFirstBig()
-            self._log.debug('ref on cpu: %d', target_cpu)
+                return min(self.pload.values())
```
ionela-voinescu: I might be missing something, but isn't this suppose to return the key that corresponds to the biggest/smallest pload value? Also, you're returning results in two different formats: one integer here and one string below.

bjackman: It returns the value to set for the 'calibration' key in the rt-app JSON. [This can be either a string like "CPU0" or an integer representing the pLoad value](https://github.com/scheduler-tools/rt-app/blob/master/doc/tutorial.txt#L90).

libs/wlgen/wlgen/rta.py
```diff
@@ -186,72 +186,28 @@ def __postrun(self, params):
 
 
     def _getFirstBig(self, cpus=None):
```
ionela-voinescu: Also, this is not used anymore and can be removed.

bjackman:
I've just updated this with a version that adds some basic tests (which found bugs! :grinning: :grinning: :grinning:) and I'm now fairly confident this is the right approach.

libs/wlgen/wlgen/rta.py
```diff
@@ -184,104 +184,25 @@ def __postrun(self, params):
...
-                target_cpu = self._getFirstBig()
-                self._log.debug('ref on big cpu: %d', target_cpu)
-            return target_cpu
-
-        # These options are selected only when RTApp has not been
-        # already calibrated
-        if self.cpus is None:
-            target_cpu = self._getFirstBig()
-            self._log.debug('ref on cpu: %d', target_cpu)
+                return min(self.pload.values())
```
derkling: Love this... did not noticed it before +1

tests/lisa/test_wlgen_rtapp.py
```diff
@@ -265,3 +265,42 @@ def test_custom_smoke_calib(self):
...
+
+    def test_calibration_conf_pload(self):
+        """Test that the smallest pload value is used, if provided"""
+        cpus = range(self.target.number_of_cpus)
+        conf = self._get_calib_conf(dict(zip(cpus, [c + 100 for c in cpus])))
+        self.assertEqual(conf, 100,
+                         'Calibration not set to minimum pload value')
+
+    def test_calibration_conf_bl(self):
+        """Test that a big CPU is used is big.LITTLE data is available"""
```
valschneider: big CPU is used s/is/if/ big.LITTLE

bjackman: Fixed, thanks

PR #368, opened 2017-04-13
executor: Don't except `init` from freezing on Android

bjackman:
Freezing init doesn't seem to break anything. Also scenarios have been observed
where Android's init eats CPU while the rest of the system is frozen. Maybe
it's spin-waiting on a lock or mailbox for a frozen process.


bjackman:
Oops, re-worded the commit message, double negatives aren't un-confusing!

PR #369, opened 2017-04-19
executor: Fix setting duration for rt-app wloads

bjackman:


bjackman:
I fixed this already :/

PR #370, opened 2017-04-19
AEP simplifications/enhancements

bjackman:
Marking this as RFC until I have tested it a little more thoroughly but I have been using it and it looks good so far. Also, nice red patch.

Basically:

 - Remove some code that does a load of computation we don't seem to use (I need to double check this).
- Add a `data_frame` field to `EnergyReport` that lets you see the samples directly. (CC @joelagnel)

### DIFF
--- a/libs/utils/energy.py
+++ b/libs/utils/energy.py
@@ -258,6 +258,9 @@ class AEP(EnergyMeter):
             # we have already consumed the first line of `f`.
             df = pd.read_csv(f, names=columns)
 
+        if df.empty:
+            raise RuntimeError('No energy data collected')
+
         channels_nrg = {}
         for site, measure in df:
             if measure == 'power':
--- a/libs/utils/energy.py
+++ b/libs/utils/energy.py
@@ -26,6 +26,7 @@ from collections import namedtuple
 from subprocess import Popen, PIPE, STDOUT
 from time import sleep
 
+import pandas as pd
 
 # Default energy measurements for each board
 DEFAULT_ENERGY_METER = {
@@ -61,7 +62,6 @@ DEFAULT_ENERGY_METER = {
 
 }
 
-EnergyCounter = namedtuple('EnergyCounter', ['site', 'pwr_total' , 'pwr_samples', 'pwr_avg', 'time', 'nrg'])
 EnergyReport = namedtuple('EnergyReport', ['channels', 'report_file'])
 
 class EnergyMeter(object):
@@ -219,9 +219,6 @@ class AEP(EnergyMeter):
         # Energy channels
         self.channels = []
 
-        # Time (start and diff) for power measurment
-        self.time = {}
-
         # Configure channels for energy measurements
         self._log.info('AEP configuration')
         self._log.info('    %s', conf)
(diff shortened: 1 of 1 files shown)


bjackman:
Just a thought - probably want to improve the DataFrame layout and document it. E.g. maybe nested columns with top level for measurement site ('BAT') and second level for measurement type ('power'/'current')

bjackman:
OK updated so that we do use nested columns, one level for the measurement site (e.g. 'BAT') and  second level for the measure type (e.g. 'power'/'current'), so a DataFrame looks like:

```
          BAT               
      current  power voltage
0       0.538  2.291   4.258
1       0.543  2.311   4.253
2       0.533  2.266   4.251
3       0.551  2.347   4.257
4       0.551  2.347   4.258
```

libs/utils/energy.py
```diff
@@ -275,22 +242,32 @@ def reset(self):
...
-        self._sample(csv_file)
+        csv_path = os.path.join(out_dir, out_samples)
+        self._aep.get_data(csv_path)
+        with open(csv_path) as f:
+            # Each column in the CSV will be headed with 'SITE_measure'
+            # (e.g. 'BAT_power'). Convert that to a list of ('SITE', 'mesure)
+            # tuples, then pass that as the `names` parameter to read_csv to get
+            # a nested column index. None of devlib's standard measurement types
+            # have '_' in the name so this use of rsplit should be fine.
+            headers = f.readline().strip().split(',')
```
derkling: Should/Can we add a check assert here... dunno if we can get a list of measurements from devlib?

libs/utils/energy.py
```diff
@@ -275,22 +242,32 @@ def reset(self):
...
 
-        # Reformat data for output generation
         channels_nrg = {}
-        for channel in self.channels:
-            self._log.debug('Energy [%16s]: %.6f',
-                            channel.site, channel.nrg)
-            channels_nrg[channel.site] = channel.nrg
+        for site, measure in df:
+            if measure == 'power':
+                channels_nrg[site] = df[site]['power'].sum()
```
derkling: Are you here considering a constant sampling time?

Should we not multiple at least by time reported by the AEP?
Otherwise we will always get "energy correlated" numbers but not a "close to real" energy measure.
... but perhaps it's still good enough for what we wanna do.

libs/utils/energy.py
```diff
@@ -237,36 +234,6 @@ def __init__(self, target, conf, res_dir):
...
 
-    def _get_energy(self, samples, time, idx, site):
-        pwr_total = 0
-        pwr_samples = 0
-        for sample in samples:
-            pwr_total += sample[idx].value
-            pwr_samples += 1
-
-        pwr_avg = pwr_total / pwr_samples
-        nrg = pwr_avg * time
```
derkling: I agree that this `time` can introduce quite a big/variable error... :-(

libs/utils/energy.py
bjackman: Good idea. Devlib gives us the info we need to figure out the expected column headers, I'll try to add a sanity check.

libs/utils/energy.py
bjackman: Yeah this assumes a constant sampling rate, now I think about it I guess I've assumed that the meter automatically scales power according to the sample rate which is probably not true. 

As you said this doesn't matter for our purposes but let's try to get it right anyway. Devlib doesn't actually expose the sample rate right now so I'll try to add it.

derkling: Is not devlib reporting the timestamp for each sample?

In that case you should have it in your dataframe... actually it should be its index.
From that it should be easy to build a delta column which you use to compute a "SW energy counter"... and sum it up that to get the overall energy.

bjackman: Indeed. Devlib doesn't report the timestamp at the moment but I'll try to fix that.

bjackman:
OK, new version which 
- Adds a commit to get rid of the requirement for channel_map in AEP
- Uses https://github.com/ARM-software/devlib/pull/116 to set the DataFrame index and compute energy properly from power.

bjackman:
Removing RFC tag, this is now a Real Pull Request

libs/utils/energy.py
```diff
@@ -237,36 +234,9 @@ def __init__(self, target, conf, res_dir):
...
-
-        ec = EnergyCounter(site, pwr_total, pwr_samples, pwr_avg, time, nrg)
-        return ec
-
-    def _sample(self, csv_file):
         if self._aep is None:
             return
```
mdigiorgio: Why this final check? I think it's unnecessary.

libs/utils/energy.py
```diff
@@ -234,13 +234,7 @@ def __init__(self, target, conf, res_dir):
         self._log.info('   %s', str(self._aep.active_channels))
         self._log.debug('Results dir: %s', self._res_dir)
 
-        if self._aep is None:
-            return
-
     def reset(self):
-        if self._aep is None:
-            return
-
```
mdigiorgio: OK this answers my first question... Why not squashing it into the first patch?

bjackman: Because it isn't related to that patch, except that it edits the same code. The patch to remove `None` checks could be merged independently.

derkling: What about that squashing?

bjackman: I have

derkling:
Tested and it's working... the idea to report a DataFrame as part of the EnergyReport is also a bit :+1: for this series! Let's merge it! :heart_eyes:

PR #372, opened 2017-04-20
Use IPython 5.x to avoid Python 3.3 requirement

bjackman:
This should fix the broken CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@
 sudo: required
 language: python
 install:
-  - pip install --upgrade Cython trappy bart-py devlib psutil wrapt ipython jupyter matplotlib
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -18,7 +18,8 @@
 sudo: required
 language: python
 install:
-  - pip install --upgrade Cython trappy bart-py devlib psutil wrapt ipython jupyter matplotlib
+  # Use IPython 5.x because 6.0+ only supports Python 3.3
```
derkling: Really!?! Wowww... that sounds like a big change... does it means that we will be stuck with an old version of IPython forever?

bjackman: Yeah :( https://github.com/jupyter/roadmap/blob/master/accepted/migration-to-python-3-only.md

PR #376, opened 2017-04-21
Add 'userspace-freezing' support to the capacity_capping test

credp:
Running capacity_capping on a busy android system is an almost guaranteed failure, since it's actually quite reasonable to leave things running on very low OPPs if all the other cpus are already busy. However, if we freeze the userspace then we can be sure that the system is in a state where it should have the expected behaviour.

The test is not based on the executor class, so I took a few liberties rather than rewrite it using the new test classes.
I reused the 'executor' label for the log entries so that the default logging.conf was able to display them.
I took the implementation of freeze/thaw from executor, along with the task list.

Hope it's acceptable.

### DIFF
--- a/tests/eas/capacity_capping.py
+++ b/tests/eas/capacity_capping.py
@@ -83,6 +83,7 @@ class CapacityCappingTest(unittest.TestCase):
 
     """
 
+
     @classmethod
     def setUpClass(cls):
         cls.params = {}
@@ -126,6 +127,7 @@ class CapacityCappingTest(unittest.TestCase):
         wload.conf(kind="profile", params=cls.params)
         phase_duration = WORKLOAD_DURATION_S / 3.
 
+        cls.env.freeze_userspace()
         cls.env.ftrace.start()
 
         wload.run(out_dir=cls.env.res_dir, background=True)
@@ -150,6 +152,7 @@ class CapacityCappingTest(unittest.TestCase):
 
         cls.env.ftrace.stop()
         cls.env.ftrace.get_trace(cls.trace_file)
+        cls.env.thaw_userspace()
 
     def check_residencies(self, cpus, cpus_name, window, phase_description):
         """Helper function to check the residencies of all busy threads on a
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -151,6 +151,22 @@ class TestEnv(ShareState):
     :type force_new: bool
     """
 
+    critical_tasks = {
+        'linux': ['init', 'systemd', 'sh', 'ssh'],
+        'android': [
+            'sh', 'adbd',
+            'usb', 'transport',
+            # We don't actually need this task but on Google Pixel it apparently
+            # cannot be frozen, so the cgroup state gets stuck in FREEZING if we
+            # try to freeze it.
+            'thermal-engine'
+        ]
+    }
+    """
+    Dictionary mapping OS name to list of task names that we can't afford to
+    freeze when using freeeze_userspace.
+    """
+
     _initialized = False
 
     def __init__(self, target_conf=None, test_conf=None, wipe=True,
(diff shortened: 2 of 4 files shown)


tests/eas/capacity_capping.py
```diff
@@ -83,12 +84,33 @@ class CapacityCappingTest(unittest.TestCase):
...
+    Dictionary mapping OS name to list of task names that we can't afford to
+    freeze when using freeeze_userspace.
+    """
+
+
     @classmethod
     def setUpClass(cls):
+        # we don't use the real Executor class here, but we claim to be
+        # it for logging so that messages are visible.
+        cls._log = logging.getLogger('Executor')
```
bjackman: Let's use the 'LisaTest' logger instead.

tests/eas/capacity_capping.py
```diff
@@ -83,12 +84,33 @@ class CapacityCappingTest(unittest.TestCase):
...
+            # We don't actually need this task but on Google Pixel it apparently
+            # cannot be frozen, so the cgroup state gets stuck in FREEZING if we
+            # try to freeze it.
+            'thermal-engine'
+        ]
+    }
+    """
+    Dictionary mapping OS name to list of task names that we can't afford to
+    freeze when using freeeze_userspace.
+    """
```
bjackman: Can we import this instead of duplicating it? Since this is a bit of a hack anyway, we could just import the `Executor` class and then use `Executor.critical_tasks`.

tests/eas/capacity_capping.py
```diff
@@ -109,6 +131,33 @@ def populate_params(cls):
             ).get()
 
     @classmethod
+    def _freeze_userspace(cls):
+        if "freeze_userspace" not in TEST_CONF.get("flags", []):
+            cls._log.info('not asked to freeze userspace');
+            return
```
bjackman: I don't think there's any need to add a flag to the test conf. If someone wants to run without freezing userspace, they're going to have to read the code to figure out how (unless you feel like writing a big README...). By the time they've done that, nobody saves any effort by having a configuration flag, they might as well just comment out the relevant code.

tests/eas/capacity_capping.py
```diff
@@ -109,6 +131,33 @@ def populate_params(cls):
...
+        controllers = [s.name for s in cls.env.target.cgroups.list_subsystems()]
+        if 'freezer' not in controllers:
+            cls._log.warning('No freezer cgroup controller on target. '
+                              'Not freezing userspace')
+            return
+
+        exclude = cls.critical_tasks[cls.env.target.os]
+        cls._log.info('Freezing all tasks except: %s', ','.join(exclude))
+        cls.env.target.cgroups.freeze(exclude)
+        cls.userspace_frozen = True
```
bjackman: I wonder if we can just move the freeze_userspace method into `TestEnv`? Then we won't need to duplicate it here.

(By the way, at some point I want to change the freeze/thaw functionality to use a [context manager](https://docs.python.org/2/library/contextlib.html), so that if we get an exception while userspace is frozen we still unfreeze userspace. That will be much easier if we have this method in one place.)

tests/eas/capacity_capping.py
derkling: Maybe even better, why not doing a refactoring which moves the freezing support into the `TestEnv` class.
At the end this functionality does not really belongs to the `Executor` or, better, having it in the TestEnv it automagically makes it available to many other clients (read: tests, notebooks, etc.)

tests/eas/capacity_capping.py
derkling: Agree, was thinking the same above... and regarding this context manager: nice idea. It can perhaps also be useful for stuff like auto-connecting a device across reboots, etc...

credp:
OK. So... if I have this straight.. Put the freezing support into testenv, make freezing unconditional in capacity_capping. Will do :) :+1:

PR #377, opened 2017-04-24
tests/eas/preliminary: WorkThroughput improvement

valschneider:
_check_work_throughput() had an unused 'margin' variable. It's now used
to check that each CPU does at least 10% more work at its highest OPP
than at its lowest. This also 'fixes' false positives induced by the
fact that every sysbench run is not quite identical, so we could be
comparing 99 events vs 100 events, which should be a failure.

Also added a list of (freq -> work) in the test output should it fail,
which can help figuring out what is causing the failure

### DIFF
--- a/tests/eas/preliminary.py
+++ b/tests/eas/preliminary.py
@@ -131,38 +131,52 @@ class TestWorkThroughput(BasicCheckTest):
         match = re.search(r'(total number of events:\s*)([\d.]*)', bench_out)
         return float(match.group(2))
 
-    def _check_work_throughput(self, cpu):
-        seconds = 1.0
-        margin = 0.2
+    def _check_work_throughput(self, cpu, duration, margin):
         frequencies = self.target.cpufreq.list_frequencies(cpu)
         if len(frequencies) == 1:
             return True
+
         original_governor = self.target.cpufreq.get_governor(cpu)
         original_freq = None
         if original_governor == 'userspace':
             original_freq = self.target.cpufreq.get_frequency(cpu)
-        # set userspace governor
+
+        # Set userspace governor
         self.target.cpufreq.set_governor(cpu, 'userspace')
-        # do each freq in turn
+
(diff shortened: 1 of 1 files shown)


valschneider:
Arguably the (freq -> work) list isn't **that** useful, and can be quite big if there are a lot of OPPs on the target. We could just list the lowest & highest OPPs, but since the test relies on them it would simply be easier to print "work at highest OPP wasn't 10% bigger than work at lowest OPP".

tests/eas/preliminary.py
```diff
@@ -133,37 +133,55 @@ def _run_sysbench_work(self, cpu, duration):
...
+
+        # Format error message
         msg='Work done did not scale with CPU Freq on CPUs: {}'\
-            .format(failed_cpus)
-        self.assertFalse(len(failed_cpus), msg=msg)
+            .format(failed_results.keys())
+        for cpu in failed_results.iterkeys():
+            msg += '\nCPU {}:'.format(cpu)
+            for freq, work in failed_results[cpu].iteritems():
+                msg += '\n\t{}Hz -> {} events'.format(freq, work)
```
bjackman: Maybe say '\n\t{}Hz -> processed {} events' to clarify that the "event" count is a measure of work throughput.

ionela-voinescu:
Looks good to me too. As Valentin I was wondering if there's any point in reporting the full list of performance values for each OPP when it fails, given that it only compares the first and last value for the fail/pass decision. I think it might be worth replacing `Work done did not scale with CPU Freq on CPUs` with `Work at highest OPP wasn't <margin> bigger than work at lowest OPP` just for the user to understand what decides the success/failure of the test.

valschneider:
I removed the (freq -> work) list and simply put an error message, as per @ionela-voinescu's comment. Also, since we only compare lowest & highest OPPs, I removed the runs done at OPPs in-between - their results aren't even used anyway.

PR #379, opened 2017-04-24
libs/utils/energy: Fix error reporting for missing hwmon channel

jlelli:
For some platforms (e.g., JunoR2) channel labels in the energy report
might differ from channel names, and this raises a runtime error (even
if energy is correctly reported).

Fix it by always use channel labels.

Fixes: a41a48dfabb7 ('libs/utils/energy: Slightly improve error for missing hwmon channel')

### DIFF
--- a/libs/utils/energy.py
+++ b/libs/utils/energy.py
@@ -195,11 +195,11 @@ class HWMon(EnergyMeter):
         # Reformat data for output generation
         clusters_nrg = {}
         for channel in self._channels:
-            if channel not in nrg:
+            label = self._channels[channel]
+            if label not in nrg:
                 raise RuntimeError('hwmon channel "{}" not available. '
                                    'Selected channels: {}'.format(
                                        channel, nrg.keys()))
-            label = self._channels[channel]
             nrg_total = nrg[label]['total']
             self._log.debug('Energy [%16s]: %.6f', label, nrg_total)
             clusters_nrg[channel] = nrg_total


bjackman:
So.. I finally took 15 mins to look at this semi-properly and AFAICT the channel_map doesn't do anything except create this bug (it renames the keys in the output dictionary, but I don't see any value in that). @derkling am I missing something?

bjackman:
Untested patch here: https://github.com/bjackman/lisa/commit/893955028488b8f5d57828083f0702319d6a9429

derkling:
Well, the idea was exactly that: rename the default channel names to have something more easy to read in plots. We used the map for example to report if a channel is a big or a LITTLE, on Juno. On same EM then default channel name is really to generic (like Ch0 for the CAPE)... in that case the translation helps to understand what a channel has measured.

derkling:
Note that some of the post-processing scripts:
https://github.com/ARM-software/lisa/blob/master/libs/utils/report.py
are assuming that the channels are named exactly "big" and "LITTLE".

bjackman:
Ah OK, so I guess we should have it be optional, so if a channel doesn't match anything in the channel_map then we just ignore the channel_map. Does that sound sensible?

derkling:
Not really... if we have a channel name specified and it does not match the map, something is wrong... and it would be nice to let the user know.

bjackman:
Well then at least we should allow the `channel_map` to be `None`, in which case you just get all the energy readings by the name the device gives them.

derkling:
That can be an option, agree.

PR #382, opened 2017-05-05
libs/utils/env.py: Add gem5 support

valschneider:
Gem5 is now a board option for TestEnv configuration.
board_settings is used to define the simulation

Results directory was being initialized after reading the
configurations and setting up the targets. This initialization
has been moved up in the code so that is it made before reading
the configurations, the result path is thus now available for
the initialization of certain boards (e.g. gem5)


valschneider:
Naming / structure of the 'board_settings' dict haven't been worked much, and could do with some improvements to be more generic -feedback welcome !

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
...
+            system_info = self.conf['board_settings']['system']
+            simulator_info = self.conf['board_settings']['simulator']
+            simulator_args = ""
+            platform_args = ""
+
+            # Get gem5 binary arguments
+            for args in simulator_info['args']:
+                simulator_args += args + " "
+            # Required for telnet terminal
+            simulator_args += "--listener-mode=on"
```
bjackman: I think you can squash a load of lines into `simulator_args = ' '.join(simulator_info['args'] + ['--listener-mode=on'])` and somether similar for `platform_args`.

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
...
+            self.nrg_model = None
+            self.__modules=['cpufreq']
+
+            system_info = self.conf['board_settings']['system']
+            simulator_info = self.conf['board_settings']['simulator']
+            simulator_args = ""
+            platform_args = ""
+
+            # Get gem5 binary arguments
+            for args in simulator_info['args']:
```
bjackman: Make `args` optional (use `simulator.info.get('args', [])`?

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
...
+                simulator_args += args + " "
+            # Required for telnet terminal
+            simulator_args += "--listener-mode=on"
+
+            # Get platform arguments
+            for args in system_info['platform']['args']:
+                platform_args += args + " "
+            # Mandatory arguments
+            platform_args += " --kernel "     + system_info['kernel'] + \
+                             " --dtb "        + system_info['dtb'] + \
```
bjackman: Nit: use implicit continuation instead of `\` (i.e. wrap expressions in brackets then you don't need to escape the newline)

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
...
+                             " --disk-image " + system_info['disk']
+
+            # Gather all arguments
+            args = simulator_args + " " + \
+                   system_info['platform']['description'] + " " + \
+                   platform_args
```
bjackman: Another `str.join` candidate here

bjackman: Another nit: let's try to keep quotes consistent. LISA code tends to prefer `'` although I've accidentally added plenty of `"` myself.

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
...
+
+            # Setup virtio
+            virtio_args = "--workload-automation-vio={}"
```
bjackman: Is this missing a call to `str.format` or is this the legit arg? If the latter please add a comment.

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
...
+            diod_path = Popen(["which", "diod"], stdout=PIPE).stdout.read().strip()
```
bjackman: `from devlib.utils.misc import which`

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
             self.__modules = ['bl', 'cpufreq']
             platform = Platform(model='pixel')
 
+        # Initialize gem5 platform
+        elif self.conf['board'].upper() == 'GEM5':
+            self.nrg_model = None
+            self.__modules=['cpufreq']
+
+            system_info = self.conf['board_settings']['system']
+            simulator_info = self.conf['board_settings']['simulator']
```
bjackman: Let's call this "gem5_settings" not "board_settings"

libs/utils/env.py
```diff
@@ -439,6 +441,54 @@ def _init_target(self, force = False):
...
+            if not diod_path:
+                raise RuntimeError('Failed to find diod on your host machine, '
+                                   'check your installation or your PATH variable')
+
+            virtio_args += " --which-diod=" + diod_path
+
+            # Merge all arguments
+            platform = devlib.platform.gem5.Gem5SimulationPlatform(
+                name='gem5',
+                gem5_bin=simulator_info['binary'],
```
bjackman: "gem5_bin" is a better name than "binary" IMO.

libs/utils/env.py
valschneider: It's the legit arg, I added a comment

libs/utils/env.py
valschneider: The idea was to try to keep the general structure somewhat generic, e.g. use it for several "boards", but I don't know how realistic that is...

bjackman: Yeah the fact that it already has "system" and "simulator" makes it at least specific to models. *Maybe* we could re-use this structure if we were to integrated something like QEMU or ARM Fast Models but TBH our odds are pretty slim of successfully designing a generic interface without investigating the details of those models in this context.

ipynb/examples/utils/gem5_integration.ipynb
```diff
@@ -0,0 +1,342 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "# Running some binaries on gem5"
```
ionela-voinescu: In my opinion it works better as: "Run workloads on gem5"..

ipynb/examples/utils/gem5_integration.ipynb
```diff
@@ -0,0 +1,342 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "for command, args in commands:\n",
+    "    print \"-----[\" + command + \"]-----\"\n",
+    "    \n",
+    "    path = target.install_if_needed(\"/home/vagrant/lisa/tools/arm64/\" + command)\n",
+    "    print path\n",
+    "    \n",
+    "    target.execute('m5 resetstats')\n",
```
ionela-voinescu: It might be worth adding a link here to http://gem5.org/M5ops to show different arguments of m5.

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
...
+            # Merge all arguments
+            platform = devlib.platform.gem5.Gem5SimulationPlatform(
+                name='gem5',
+                gem5_bin=simulator_info['gem5_bin'],
+                gem5_args=args,
+                gem5_virtio=virtio_args,
+                host_output_dir=self.res_dir
+                #gem5_telnet_port='3456' (not yet supported)
+            )
+
```
ionela-voinescu: In order to make this work properly I've added some more code here to add missing platform information (obtained in my case from a platform json file):
```
            suffix = os.path.splitext(os.path.basename(
                     system_info['platform']['description']))[0]
            board = self._load_board(self.conf['board'], suffix)
            if board:
                platform.core_names = board['cores']
                platform.core_clusters = self._get_clusters(board['cores'])
                platform.big_core = board.get('big_core', None)
```
I've also modified the _load_board function to accept a suffix to the board name. Therefore if the gem5 platform description file is juno.py, it will look for a gem5juno.json platform file to extract this information from.

I believe this to be a bit hacked so I would like some input from you guys on this. It might be better to pass this information to Gem5Simulation platform and let it set, rather than setting it ourselves.

ipynb/examples/utils/gem5_integration.ipynb
```diff
@@ -0,0 +1,342 @@
...
+    "                \"description\" : \"/home/user/platform_juno/juno.py\",\n",
+    "                \"args\" : [\n",
+    "                    # Resume simulation from a previous checkpoint\n",
+    "                    # Checkpoint must be taken before Virtio folders are mounted\n",
+    "                    \"--checkpoint-indir /home/user/gem5/platform_juno/checkpoints\",\n",
+    "                    \"--checkpoint-resume 1\",\n",
+    "                ]\n",
+    "            },\n",
+    "            # Kernel compiled for gem5 with Virtio flags\n",
+    "            \"kernel\"   : \"/home/user/Work/gem5/platform_juno/vmlinux\",\n",
```
bjackman: Do you mean to define  base folder in the notebook or in the actual gem5 API? If the latter then we need to be really careful to ensure we don't place annoying restrictions on the relative locations of the different binaries..

libs/utils/env.py
bjackman: The base `Platform` class has those extra core info attributes as __init__ arguments, maybe we should add them to `Gem5SimulationPlatform` too (or just do the `*args, **kwargs` thing so it inherits all the base class' arguments? Also should try to make it work without needing a <board>.json wherever possible.

We do also definitely need to refactor this whole board thing in LISA but that's a separate issue I suppose.

ipynb/examples/utils/gem5_integration.ipynb
ionela-voinescu: No, I meant the to say that it would be good to define the base in the notebook, not the GEM5 API, just to avoid having all of those long paths, given that most inputs will be present in the same root directory. I don't believe we should have this information in the GEM5 platform class as the restrictions will definitely not apply.

bjackman: OK in that case I agree :+1:

valschneider:
I've updated the branch following @ionela-voinescu's comments - I haven't added the platform information bits as @ionela-voinescu is still working on it and adding bits and pieces.

ionela-voinescu:
@valschneider ,
I've pushed 2 additional commits on top of yours at https://github.com/ionela-voinescu/lisa/commits/gem5_v2. If it's okay with you, you can squash the changes to the notebook in your original commit,  and add the other one separately to your gem5 support. We can then review all three as the final patches.

qperret:
I think it is assumed here that the ${M5_PATH} environment variable is set prior to execution. It is probably OK since most users will have exported it in their .bashrc already but it may be nice to pop up an error message when one tries to use a Gem5 board with ${M5_PATH} unset. Or maybe the M5_PATH value could be passed as an additional field in 'gem5_settings' ?

ionela-voinescu:
Thank you @valschneider for updating the branch. Let me know what you think.

libs/utils/env.py
```diff
@@ -474,19 +485,36 @@ def _init_target(self, force = False):
...
-                gem5_bin=simulator_info['gem5_bin'],
-                gem5_args=args,
-                gem5_virtio=virtio_args,
-                host_output_dir=self.res_dir
+                name = 'gem5',
+                gem5_bin = simulator_info['gem5_bin'],
+                gem5_args = args,
+                gem5_virtio = virtio_args,
+                host_output_dir = self.res_dir,
+                core_names = board['cores'] if board else None,
```
valschneider: Aren't core names / clusters already specified in the python script ?

ionela-voinescu: Usually core names and clusters are detected based on part numbers in cpuinfo, functionality which was broken in GEM5 until a recent release. This configuration would work with the recent release both if you provide a board file or if you don't.

valschneider: Okay, it's all good to me then.

ionela-voinescu:
Anyone in the mood for some gem5 review? It would be good to close this.. Thank you!

libs/utils/env.py
valschneider: Since the core names are optional, how about:
`core_names = board.get('cores', None) if board else None`

ionela-voinescu: When it does not have valid part numbers for CPUs in cpuinfo it uses that information (cores) to differentiate the clusters. I'll rather have it enforce the presence of "cores" in the json file, for this reason.

valschneider: I see. I still find it slightly odd that not passing a json file is accepted, but if you do pass one then you have to provide additional data. You say it's used to differentiate the clusters, is it just for display purposes or something more ? If it's just naming I think we can not enforce it.

ionela-voinescu: So you have two different situations - one in which you don't provide a board file and therefore you accept whatever cpuinfo gives you (you take a risk that that's correct), and another in which you provide it. If you provide the board file I find it important to enforce the presence of information that makes the board file useful. That information it's not only for display purposes, it's also used to differentiate the clusters (establish big/LITTLE clusters of CPUs), information that's later relevant during testing.

valschneider: For a physical platform I would try to reduce as much as possible the amount of required data for the sake of testing simplicity, but since gem5 simulations can be quite lengthy I think it's a good idea to enforce certain parameters before starting a simulation. To me it's good for merging.

valschneider:
I think this is in a relatively good state for merging, any thoughts @bjackman ?

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
...
+
+            # Setup virtio
+            virtio_args = ' '.join(['--which-diod=' + diod_path,
+            # Brackets are there to let the output dir be created automatically
+                                    '--workload-automation-vio={}'])
+
+            # Merge all arguments
+            platform = devlib.platform.gem5.Gem5SimulationPlatform(
+                name='gem5',
+                gem5_bin=simulator_info['gem5_bin'],
```
bjackman: `simulator_info.get('gem5_bin', which('gem5_bin'))`?

ionela-voinescu: The GEM5 binaries used are gem5.fast, gem5.opt, etc., depending on how GEM5 was built and they won't be present in path so it's not useful to have this. It's also not useful to have a default as it depends on how the user built GEM5 and it's up to him to specify it in the configuration.

libs/utils/env.py
```diff
@@ -474,19 +485,36 @@ def _init_target(self, force = False):
...
+            board_file = self.conf['board'].lower() + suffix + '.json';
+            self._board_path = os.path.join(basepath, 'libs/utils/platforms',
+                                      board_file)
+            if not os.path.exists(self._board_path):
+                self._log.warning('Board descriptor file %s was not found.',
+                                  self._board_path)
+                self._board_path = None
+                board = None
+            else:
+                board = self._load_board()
```
derkling: Here we are:
1. refactoring the `_load_board()` call to setup `self._board_path`, which is not required by the existing boards
2. overwriting the anticipated definition of `self._board_path` in the new case of the GEM5 target
3. use the refactored `_load_board()` to actually load a board

I seems to me that all this dance is required just to load a board file which has a suffix.

Cannot we do that by just using the existing `_load_board()` (i.e. no refactoring required) and passing in the exactly `board_file` you build in the first two code line of this chunk?

IOW, cannot this patch be reduced to just 4 lines of code: the three just described here:

```python
suffix = os.path.splitext(os.path.basename(
                system_info['platform']['description']))[0]
                board_file = self.conf['board'].lower() + suffix + '.json';
board = self._load_board(board_file)
```

and the one you add in the following chunk (see my next comment)

valschneider:
Moved some stuff in the first commit as per @derkling's comment. I suppose @ionela-voinescu will reply to the `_load_board()` comment ?

libs/utils/env.py
ionela-voinescu: The `board_file` is required to both load board information and to obtain the `nrg_model`. Previously, both `_load_board` and `_load_em` would obtain the information they needed by constructing the file path from the `board` field of the target configuration. For GEM5 that's not enough and requires part of the platform description as suffix to identify the proper file. So if I only make the change you suggested, I'll have to construct this path (based on a suffix which is convoluted and prone to error - `self.conf['gem5_settings]['system']['platfomr']['description'])` multiple times:
- when `_load_board(board_file)` is called
- when `_load_em(board_file)` is called
- (in future commits) when self.nrg_model is initialised with is based on an energy file (similar to what we have now: juno_energy.py) that will use the same suffix.

In my opinion is more clear to establish this `_board_path` once.
In the current patch (for each of your points 1., 2., above):
- `_load_board()` will use `_board_path` if is set and if the file exists. It will do nothing otherwise (so we maintain the same behaviour as before)
- it will set up a default `_board_path` which is `<board_name>.json` as it was before, and it will overwrite it for GEM5 that needs to add a suffix.

Yes, this dance is not mandatory (I can obtain that suffix multiple times, although it's ugly) but I thought this solution was cleaner for both the existing code and the one added for GEM5.

ionela-voinescu: @derkling What do you think? Does this make sense?

derkling: No, that's not yet convincing to me... looking again at the configurations settings you are adding for gem5 it seems that your concept of "board" cannot be just "gem5". But instead whatever platform configuration you specify under:
`system_info['platform']['description']`

Thus, why not simply overriding the `board` attribute in the gem5 case?

I mean, here you just set:
```python
suffix = os.path.splitext(os.path.basename(system_info['platform']['description']))[0]
# Redefined board to match the specific Gem5 platform we use
self.conf['board'] = self.conf['board'] + suffix
```
and then:
1. the refactoring below is not required
2. we have `self.conf['board']` properly set to represent a specific target

Could that work?

ionela-voinescu: Yes, thank you, that would work nicely. For some reason I was reluctant to modify conf['board'] but looking over the code I don't see a reason not to.
@valschneider: I've updated that patch at: https://github.com/ionela-voinescu/lisa/commit/bde4b3cf0b704c4eaac80b3a04c86a0b0f64ad27.
It's being tested now but if you're okay with this version let me know and I'll merge it once I've verified it.

derkling: AFAIKS by updating `conf['board']` just in the case `board='gem5'` you should eventually break only gem5 related executions... thus you can easily check that everything is working as expected .

At the same time we are also sure to not potentially break other code, since the refactoring is gone. ;-)

valschneider: @ionela-voinescu That looks good to me, I agree it's more sensible to have this done just for gem5. Do you want me to add it to this PR ?

ipynb/examples/utils/gem5_example.ipynb
```diff
@@ -0,0 +1,570 @@
...
+   "source": [
+    "# Initialise command: [binary/script, arguments]\n",
+    "command = 'sysbench'\n",
+    "args = '--test=cpu --max-time=1 run'\n",
+    "\n",
+    "# Install binary if needed\n",
+    "path = target.install_if_needed(\"/home/vagrant/tools/arm64/\" + command)\n",
+    "\n",
+    "# Reset and dump statistics\n",
+    "target.execute('m5 resetstats')\n",
```
derkling: Is there any gem5 support in devlib?

If there is one, these commands (i.e. resetstats/dumpstats) should be better provided via that API.
In case they are not yet available, they can be added there.

In general we don't really want a user has to remember device specific commands...

ipynb/examples/utils/gem5_example.ipynb
```diff
@@ -0,0 +1,570 @@
...
+      "\n",
+      "After sysbench : \n",
+      "-------------------\n",
+      "Seconds in host --> 0.96\n",
+      "Seconds in guest --> 0.021238\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Measure the run times\n",
```
derkling: This is another great example of commands which should be encapsulated into an API, either in devlib or in LISA in this case (since it's about data extraction and presentation).
It would be probably worth to generate and return a DataFrame out of these data.

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
             self.__modules = ['bl', 'cpufreq']
             platform = Platform(model='pixel')
 
+        # Initialize gem5 platform
+        elif self.conf['board'].upper() == 'GEM5':
```
derkling: Since this is also a pretty long setup compare to the other cases above, perhaps we could better move this code into a dedicated function. something like:
```python
elif self.conf['board'].upper() == 'GEM5':
    self.nrg_model = None
    self.__modules=['cpufreq']
    platform = self._init_target_gem5()

#... further down in this file

def _init_target_gem5(self):
    # This branch code
   return platform
```

valschneider:
Updated @ionela-voinescu's board info commit.

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
...
+                                      '--disk-image {}'.format(system_info['disk'])])
+
+            # Gather all arguments
+            args = ' '.join([simulator_args,
+                             system_info['platform']['description'],
+                             platform_args])
+
+            diod_path = which('diod')
+            if diod_path is None:
+                raise RuntimeError('Failed to find diod on your host machine, '
```
derkling: Maybe better (notice the quotes):
`Failed to find "diod" on your host machine`...

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
             self.__modules = ['bl', 'cpufreq']
             platform = Platform(model='pixel')
 
+        # Initialize gem5 platform
+        elif self.conf['board'].upper() == 'GEM5':
+            self.nrg_model = None
+            self.__modules=['cpufreq']
+
+            system_info = self.conf['gem5_settings']['system']
```
derkling: Maybe `"_info"` can be removed, it also returns a variable which has the same name of the configuration attribute... thus aligning code with configuration.

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
...
+            simulator_info = self.conf['gem5_settings']['simulator']
```
derkling: Ditto here for the `simulator` settings

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
...
+            # Get platform arguments
+            platform_args = ' '.join(system_info['platform'].get('args', []) +
+                                     ['--kernel {}'.format(system_info['kernel']),
+                                      '--dtb {}'.format(system_info['dtb']),
+                                      '--disk-image {}'.format(system_info['disk'])])
+
+            # Gather all arguments
+            args = ' '.join([simulator_args,
+                             system_info['platform']['description'],
+                             platform_args])
```
derkling: This all "args" setup code maybe can be defined using a single args variable, which is what you really need for the `Gem5SimulationPlatform` despite being split in different fragments in the properties (all named `args`) in the configuration file:

```python
# gem5 arguments
args  = simulator.get('args', [])
args += ['--listener-mode=on']
# platform definition
args += system['platform']['description']
# platform arguments
args += system['platform'].get('args', [])
args += [
    '--kernel {}'.format(system['kernel']),
    '--dtb {}'.format(system['dsb']),
    '--disk-image {}'.format(system['disk-image'])
]

args = ' '.join(args)
   
```

libs/utils/env.py
```diff
@@ -439,6 +439,51 @@ def _init_target(self, force = False):
...
+
+            diod_path = which('diod')
+            if diod_path is None:
+                raise RuntimeError('Failed to find diod on your host machine, '
+                                   'check your installation or your PATH variable')
+
+            # Setup virtio
+            virtio_args = ' '.join(['--which-diod=' + diod_path,
```
derkling: Is that not just a string you can format straight below, i.e.

```python
platform = devlib.platform.gem5.Gem5SimulationPlatform(
    # other params...
    gem5_virtio = "--which-diod={} --workload-automation-vio={{}}".format(diod_path),
    # other params...
)
```

libs/utils/env.py
derkling: I think this comment does not apply anymore, @bjackman can you check and close it in case?

qperret:
FYI, there is now a devlib module available to manage the stats file of gem5. It would be worth updating the example notebook accordingly.

ionela-voinescu:
@valschneider: The platform support looks good to me. I've given the notebook a run and made some changes. I think it's easier for you to squash this commit (https://github.com/ionela-voinescu/lisa/commit/61cebb25264e438d0633b642b41453477cbee93c) rather than me commenting on each item. Let me know what you think.

ionela-voinescu:
@derkling: May I ping you for another review of the updates? Thank you!

bjackman:
Great, this looks good to me.

libs/utils/env.py
```diff
@@ -554,6 +560,47 @@ def _init_target(self, force = False):
...
+        virtio_args = '--which-diod={} --workload-automation-vio={{}}'.format(diod_path)
+
+        # Merge all arguments
+        platform = devlib.platform.gem5.Gem5SimulationPlatform(
+            name = 'gem5',
+            gem5_bin = simulator['bin'],
+            gem5_args = args,
+            gem5_virtio = virtio_args,
+            host_output_dir = self.res_dir,
+            #gem5_telnet_port = '3456' (not yet supported)
```
derkling: This can be removed... not really useful ;-)

derkling:
Apart for the comment to be removed, the PR now LGTM! ;-)

PR #383, opened 2017-05-07
android: Add gfxinfo module to parse gfxinfo output

joelagnel:
Sample usage:
g = GfxInfo("results/framestats.txt")
print dir(g)
print g.janky_frames
print g.number_missed_vsync

### DIFF
--- a/libs/utils/android/gfxinfo.py
+++ b/libs/utils/android/gfxinfo.py
@@ -67,7 +67,10 @@ class GfxInfo(object):
                 pc = re.sub('[\(\)\%]', '', pc)
                 self.__properties["janky_frames_pc"] = get_value(pc)
                 continue
-
+            # Parse 'nth_percentile: <int>ms' into nth_percentile_ms=<int>
+            if tokens[1].endswith('ms'):
+                tokens[0] = tokens[0] + '_ms'
+                tokens[1] = tokens[1][:-2]
             # Regular parsing
             self.__properties[tokens[0]] = get_value(tokens[1])
 
--- a/libs/utils/android/gfxinfo.py
+++ b/libs/utils/android/gfxinfo.py
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2017, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Parser for dumpsys gfxinfo output
+
+import ast
+
+def get_value(token):
+    try:
+        v = ast.literal_eval(token)
+    except:
+        return token
+    return v
+
+class GfxInfo(object):
+    """
+    Class for parsing and accessing GfxInfo output
+    """
+    __properties = {}
+
+    def __init__(self, path):
+        """
+        Initialize gfxinfo parser object
+
+        :param path: Path to file containing output of gfxinfo
+        """
+        self.path = path
+        self.parse_gfxinfo()
+
+    def parse_gfxinfo(self):
+        """
+        Parser for gfxinfo output, creates a name/value pair
+        """
+        with open(self.path) as f:
+            content = f.readlines()
+
+        # gfxinfo has several statistics of the format
+        # <string>: <value>
+        for line in content:
+            line = line.rstrip()
+            # Ignore any lines that aren't of this format
+            if not ':' in line or line.endswith(':'):
+                continue
+
+            tokens = line.split(':')
+            # convert <string> into a key by replacing spaces with '_'
+            tokens = [t.strip() for t in tokens]
+            tokens[0] = tokens[0].replace(' ', '_').lower()
+
+            self.__properties[tokens[0]] = get_value(tokens[1])
+
+    def __dir__(self):
+        """
+        List all available attributes including ones parsed from
+        gfxinfo output
+        """
+        return self.__properties.keys()
+
+    def __getattr__(self, name):
+        """
+        Get the gfxinfo property using the period operator
+        Ex: obj.number_missed_vsync
+        """
+        return self.__properties[name]
+
+    def __getitem__(self, name):
+        """
+        Get the gfxinfo property using the [] opertator
+        Useful for attributes like "50th_percentile" that can't
+        be fetched with the period operator
+        """
+        return self.__properties[name]
--- a/libs/utils/android/gfxinfo.py
+++ b/libs/utils/android/gfxinfo.py
@@ -16,7 +16,7 @@
 #
 # Parser for dumpsys gfxinfo output
 
-import ast
+import ast, re
 
 def get_value(token):
     try:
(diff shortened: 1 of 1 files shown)


libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,91 @@
...
+    def parse_gfxinfo(self):
+        """
+        Parser for gfxinfo output, creates a name/value pair
+        """
+        with open(self.filename) as f:
+            content = f.readlines()
+
+        for line in content:
+            line = line.rstrip()
+            if not ":" in line:
```
bjackman: Nit: keep quotes consistent (`"` is used here but `'` is used below). Most of the code in LISA prefers `'`.

libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,91 @@
...
+
+            tokens = [t.strip() for t in tokens]
+            tokens[0] = tokens[0].replace(" ", "_").lower()
+
+            if tokens[0] == 'janky_frames':
+                (frames, pc) = tokens[1].split(" ")
+                self.__properties["janky_frames"] = get_value(frames)
+                pc = re.sub("[\(\)\%]", "", pc)
+                self.__properties["janky_frames_pc"] = get_value(pc)
+                continue
```
bjackman: I'd prefer an `else` rather than `continue` here

libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        return self.__properties[name]
+
+    def __getitem__(self, name):
+        """
+        Get the gfxinfo property using the [] opertator
+        Useful for attributes like "50th_percentile" that can't
+        be fetched with the period operator
+        """
+        return self.__properties[name]
```
bjackman: I don't really like having both `__getitem__` and `__getattr__`, I would prefer just `__getitem__` (and also to add the `keys` method). This is personal taste though so let's see how @derkling feels.

libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Class for parsing and accessing GfxInfo output
+    """
+    __properties = {}
+
+    def __init__(self, filename):
+        """
+        Initialize gfxinfo parser object
+
+        filename: File containing output of gfxinfo
```
bjackman: Should be `:param filename: Path to file containing output of gfxinfo` (Also I guess it should technically be "path" not "filename")

libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,91 @@
...
+            content = f.readlines()
+
+        for line in content:
+            line = line.rstrip()
+            if not ":" in line:
+                continue
+
+            tokens = line.split(":")
+            if (len(tokens[1]) < 1):
+                continue
```
bjackman: Could you add comments to these `if`s for the benefit of those who aren't familiar with GfxInfo files?

libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,91 @@
...
+
+            tokens = [t.strip() for t in tokens]
+            tokens[0] = tokens[0].replace(" ", "_").lower()
+
+            if tokens[0] == 'janky_frames':
+                (frames, pc) = tokens[1].split(" ")
+                self.__properties["janky_frames"] = get_value(frames)
```
bjackman: I guess we can just use `int` here and for `janky_frames_pc` instead of `get_value`

libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,91 @@
...
+
+            tokens = line.split(":")
+            if (len(tokens[1]) < 1):
+                continue
+
+            tokens = [t.strip() for t in tokens]
+            tokens[0] = tokens[0].replace(" ", "_").lower()
+
+            if tokens[0] == 'janky_frames':
+                (frames, pc) = tokens[1].split(" ")
```
bjackman: Can you add a comment here too? An example of a "janky frames" line would be good.

libs/utils/android/gfxinfo.py
joelagnel: Ok will fix

libs/utils/android/gfxinfo.py
joelagnel: Ok will fix

libs/utils/android/gfxinfo.py
joelagnel: Will do

libs/utils/android/gfxinfo.py
joelagnel: Ok

libs/utils/android/gfxinfo.py
joelagnel: Frames is actually float. I prefer to keep it consistent with other conversion and just use get_value

joelagnel: I meant frames_pc is float

libs/utils/android/gfxinfo.py
joelagnel: But that would increase indentation? What's wrong with continue

libs/utils/android/gfxinfo.py
joelagnel: Hmm ok. I prefer both but I'm ok with with dropping get_attr  ;-)

libs/utils/android/gfxinfo.py
joelagnel: On second thought, its better with else, so I fixed it up that way, thanks

joelagnel:
I hate github's whole review interface, but anyway just fyi, I updated this PR with new patches.

libs/utils/android/gfxinfo.py
```diff
@@ -0,0 +1,96 @@
...
+        # <string>: <value>
+        for line in content:
+            line = line.rstrip()
+            # Ignore any lines that aren't of this format
+            if not ':' in line:
+                continue
+
+            tokens = line.split(':')
+            # Ignore any lines with ':' but no <value>
+            if (len(tokens[1]) < 1):
```
bjackman: Looks like we could squash this into the if above and have `if ':' not in line or line.endswith(':')`

joelagnel: I love python! That looks much better and will get rid of some lines, fixing

joelagnel:
About dropping `__getattr__` , that also means I have to drop `__dir__` which really is something I don't want to do because its a nice way of listing attributes. I really prefer to keep both. thanks

joelagnel:
@bjackman I had to add back the continue to keep it clean as I added a new special-case. About testing, I have tested it.

libs/utils/android/gfxinfo.py
joelagnel: continue creeped back in..

derkling:
As an overall comment, we have quite a lot of efforts going on the topic of collecting gfxinfo stats.
@credp has also got a complete implementation which takes care to collect data and aggregated/plot them, which is probably going to post part on devlib and part on LISA... and we just discovered that WA (the project where devlib comes from) is also providing a quite complex and complete support for gfxinfo collections:
https://github.com/ARM-software/workload-automation/blob/master/wlauto/instrumentation/fps/__init__.py

Unfortunately the WA instrument has not been (yet) ported to devlib :-/

As a general comment I would probably prefer to wait for @setrofim to make that instrument available in devlib instead of merging a new/different implementation either in LISA or devlib.

bjackman:
So while we have been sitting on this, devlib has gained an instrument for parsing gfxinfo: https://github.com/ARM-software/devlib/blob/master/devlib/instrument/frames.py

It lacks documentation, but nonetheless I think we should use that where possible as it will be used by other tools as well

joelagnel:
Hmm, I missed that. Yes it makes sense to use that. Thanks

joelagnel:
By the way... Connor is pushing or has pushed a lot of code that should also include this so it makes sense to pass these comments along there as well (I'm on paternity leave and writing from phone so I appreciate if someone else could pass along the comments :))

valschneider:
As @bjackman pointed out, gfxinfo lives in devlib now. As such, if any additionnal stuff should be added to gfxinfo stats, it should be submitted there.

PR #395, opened 2017-05-10
energy: Add Monsoon Power Monitor Support

bjackman:
cc: @joelagnel

### DIFF
--- a/libs/utils/energy.py
+++ b/libs/utils/energy.py
@@ -215,34 +215,17 @@ class HWMon(EnergyMeter):
 
         return EnergyReport(clusters_nrg, nrg_file, None)
 
-class AEP(EnergyMeter):
-
-    def __init__(self, target, conf, res_dir):
-        super(AEP, self).__init__(target, res_dir)
-
-        # Configure channels for energy measurements
-        self._log.info('AEP configuration')
-        self._log.info('    %s', conf)
-        self._aep = devlib.EnergyProbeInstrument(
-            self._target, labels=conf.get('channel_map'), **conf['conf'])
-
-        # Configure channels for energy measurements
-        self._log.debug('Enabling channels')
-        self._aep.reset()
-
-        # Logging enabled channels
-        self._log.info('Channels selected for energy sampling:')
-        self._log.info('   %s', str(self._aep.active_channels))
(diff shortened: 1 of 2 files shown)


derkling:
can we split this patch in a refactoring one, just for the AEP with a minimal changelog to explain what's going on, and a second patch which adds the monsoon support?

libs/utils/energy.py
```diff
@@ -280,6 +263,25 @@ def report(self, out_dir, out_energy='energy.json', out_samples='samples.csv'):
...
 
+class AEP(_DevlibContinuousEnergyMeter):
+
+    def __init__(self, target, conf, res_dir):
+        super(AEP, self).__init__(target, res_dir)
+
+        # Configure channels for energy measurements
+        self._log.info('AEP configuration')
+        self._log.info('    %s', conf)
+        self._aep = devlib.EnergyProbeInstrument(
```
derkling: We should now `s/self._aep/self._instrument/`

libs/utils/energy.py
```diff
@@ -280,6 +263,25 @@ def report(self, out_dir, out_energy='energy.json', out_samples='samples.csv'):
...
+            self._target, labels=conf.get('channel_map'), **conf['conf'])
+
+        # Configure channels for energy measurements
+        self._log.debug('Enabling channels')
+        self._aep.reset()
```
derkling: Same here

libs/utils/energy.py
```diff
@@ -280,6 +263,25 @@ def report(self, out_dir, out_energy='energy.json', out_samples='samples.csv'):
...
+
+        # Logging enabled channels
+        self._log.info('Channels selected for energy sampling:')
+        self._log.info('   %s', str(self._aep.active_channels))
```
derkling: And here

libs/utils/energy.py
bjackman: Oops, thanks!

bjackman:
Fixed typo and pushed

derkling:
Can you add on top also the notebook I've posted here:
https://github.com/derkling/lisa/commit/c845055a6588b0a753dce9c3db01cb7f885aa7ed

PR #398, opened 2017-05-17
android: Add support to set volume

joelagnel:

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -275,6 +275,26 @@ class System(object):
         target.execute('input keyevent KEYCODE_SLEEP')
 
     @staticmethod
+    def volume(target, times=1, direction='down'):
+        """
+        Increase or decrease volume
+
+        :param target: instance of devlib Android target
+        :type target: devlib.target.AndroidTarget
+
+        :param times: number of times to perform operation
+        :type times: int
+
+        :param direction: which direction to increase (up/down)
+        :type direction: str
+        """
+        for i in range(times):
+            if direction == 'up':
+                target.execute('input keyevent KEYCODE_VOLUME_UP')
+            elif direction == 'down':
+                target.execute('input keyevent KEYCODE_VOLUME_DOWN')
+
+    @staticmethod
     def gfxinfo_reset(target, apk_name):
         """
         Reset gfxinfo frame statistics for a given app.


libs/utils/android/system.py
```diff
@@ -275,6 +275,26 @@ def sleep(target):
...
+        :type times: int
+
+        :param direction: which direction to increase (up/down)
+        :type direction: str
+        """
+        for i in range(times):
+            if direction == 'up':
+                target.execute('input keyevent KEYCODE_VOLUME_UP')
+            elif direction == 'down':
+                target.execute('input keyevent KEYCODE_VOLUME_DOWN')
```
bjackman: Is there no interface for setting the volume directly?

joelagnel:
There is a way to use 'service call audio', but looks like it requires codes that are android version specific. I am more comfortable with key events, which also seems to be the most widely used method.

bjackman:
Fair enough, It's just a shame to have something unnecessarily stateful. Other than that this looks fine to me.

valschneider:
Is the amount of key presses to go from e.g. lowest volume to highest volume somewhat standard ? It would be nice if we could have a proper mute() function that uses that bounded amount of presses instead of an arbitrarily large value.

joelagnel:
Yes that would be nice but in a separate patch.
This is just for volume change.

On May 17, 2017 9:47 AM, "Valentin Schneider" <notifications@github.com>
wrote:

Is the amount of key presses to go from e.g. lowest volume to highest
volume somewhat standard ? It would be nice if we could have a proper
mute() function that uses that bounded amount of presses instead of an
arbitrarily large value.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ARM-software/lisa/pull/398#issuecomment-302150451>, or mute
the thread

joelagnel:
Also I don't think the number of key presses is standard or we want to rely
on.

On May 17, 2017 4:08 PM, wrote:

Yes that would be nice but in a separate patch.
This is just for volume change.

On May 17, 2017 9:47 AM, "Valentin Schneider" <notifications@github.com>
wrote:

Is the amount of key presses to go from e.g. lowest volume to highest
volume somewhat standard ? It would be nice if we could have a proper
mute() function that uses that bounded amount of presses instead of an
arbitrarily large value.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ARM-software/lisa/pull/398#issuecomment-302150451>, or mute
the thread

valschneider:
Right, it's mostly something that would have been nice to have. The use cases for that will mostly be setting the volume to min or max, but that'll imply arbitrary press counts.

derkling:
If this API is stateless, and we don't know how many key-press you need to get to max/min volume, how do you plan to use it?

joelagnel:
I suppose press enough number of keys to get to the Min or max and take it
from there.

derkling:
That was actually what I was thinking me too.

Initially I was thinking about a couple of methods to set just the max or min volume.
Indeed, I don't think we are interested in not well defined mid volume levels since tests needs to be reproducible... and I guess the volume level can have some noticeable impact on energy consumption.

Than I've noticed a couple of things:
1) If I use `input keyevent KEYCODE_VOLUME_UP` on a just rebooted device... Android update the ring volume, while I guess that in most of the cases we are interested in updating the media volume
2) the `dumpsys audio` command seems to return many interesting info about audio channels, including the a `STREAM_MUSIC` along with `Min` and `Max` supported values.

I think that by reading/parsing that output of we can know exactly which volume we have on the speaker... still we need to find a suitable way to change the Music volume instead of the Ring one.

joelagnel:
Hi Patrick, you are right. But what you said sounds like a more context-aware volume change which can be an additional feature on top of this patch. Do you think we can just include this in, and add the advanced feature as a #issue ? I think its better than having nothing.

joelagnel:
So current versions of android can do this but support was just added to control volume on channels over adb. Which means older android will not be able to. Is that a problem?

derkling:
The main issue now is that, by default it seems we control the ring volume instead of the media volume.
I'm assuming we want to tune the volume for media playback... maybe if we press the volume button while Youtube is in foreground we actually change the media volume?

derkling:
Yes... it seems so... ok for merge and opening an issue to improve that.

joelagnel:
Yes you are right about that. However it is possible to start an app. Change the volume. And then do the activity through intent, like playing a video. But it is an issue as you pointed

PR #399, opened 2017-05-17
tests/eas/preliminary: Make TestSchedutilTunables mainline-friendly

bjackman:
I wrote with a version of schedutil in mind that was
hacked by the EAS project to have two separate rate limiting
tunables, up_rate_limit_us and down_rate_limit_us. I forgot that the
mainline version only has rate_limit_us.

Just check each tunable that `endswith` rate_limit_us to make this
generic across schedutil versions.

### DIFF
--- a/tests/eas/preliminary.py
+++ b/tests/eas/preliminary.py
@@ -190,8 +190,7 @@ class TestEnergyModelPresent(BasicCheckTest):
                 '- No energy model in kernel')
 
 class TestSchedutilTunables(BasicCheckTest):
-    MAX_UP_RATE_LIMIT_US = 20 * 1e3
-    MAX_DOWN_RATE_LIMIT_US = 20 * 1e3
+    MAX_RATE_LIMIT_US = 20 * 1e3
 
     def test_rate_limit_not_too_high(self):
         """Test that the schedutil ratelimiting is not too harsh"""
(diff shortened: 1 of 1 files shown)


tests/eas/preliminary.py
```diff
@@ -190,41 +190,36 @@ def test_energy_model_present(self):
...
-    MAX_DOWN_RATE_LIMIT_US = 20 * 1e3
+    MAX_RATE_LIMIT_US = 20 * 1e3
 
     def test_rate_limit_not_too_high(self):
         """Test that the schedutil ratelimiting is not too harsh"""
         governors = self.target.cpufreq.list_governors(0)
         if 'schedutil' not in governors:
             raise SkipTest('schedutil not present on target')
         self.target.cpufreq.set_all_governors('schedutil')
+        tunables  = self.target.cpufreq.list_governor_tunables(0)
```
valschneider: Why is this being set here when `tunables = self.target.cpufreq.get_governor_tunables(cpu)` is in the loop below ?

bjackman: Oh yeah, this line can be removed, thanks!

valschneider: Feel free to merge once it's removed.

bjackman:
Thanks for review, updated. Also fixed a missing conversion to `int`.

joelagnel:
LGTM

PR #400, opened 2017-05-17
tests/eas/load_tracking: Remove dependency on EnergyModel

bjackman:
This test requires only the max capacities of the CPUs, which is
available in sysfs in mainline (from v4.10). Instead of requiring a
full EnergyModel, just use those files.


valschneider:
Tested on HiKey960 w/ 4.4 kernel, sysfs entry isn't there (no surprises since you mentioned this is a 4.10+ feature). How about digging in /proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/cap_states for versions where /sys/devices/system/cpu/cpu{}/cpu_capacity isn't there ?

bjackman:
Good point - I've updated it to use the energy model where it's available.

tests/eas/load_tracking.py
```diff
@@ -83,14 +95,16 @@ def get_expected_util_avg(self, experiment):
...
         if cpufreq['governor'] == 'userspace':
             freq = cpufreq['freqs'][cpu]
+            max_freq = max(self.te.target.cpufreq.list_frequencies(cpu))
+            cpu_capacity *= float(freq) / max_freq
         else:
             assert cpufreq['governor'] == 'performance'
-            freq = None
-        cpu_capacity = self.te.nrg_model.get_cpu_capacity(cpu, freq)
 
-        scaling_factor = float(cpu_capacity) / self.te.nrg_model.capacity_scale
```
valschneider: Maybe it would be worth it to keep this line ? Admittedly 1024 is supposed to be used everywhere, but there might be some platforms with goofy cap scales ? If that's really unlikely just ignore this comment.

bjackman: Hmm, yes I think we should use `nrg_model.capacity_scale` if it's available.

bjackman:
OK, updated so it should work with wacky capacity scales (haven't tested that but let's cross that bridge if we come to it). Also added a cosmetic commit.

tests/eas/load_tracking.py
```diff
@@ -45,6 +45,15 @@ class _LoadTrackingBase(LisaTest):
...
     }
 
+    @memoized
+    @staticmethod
+    def _get_cpu_capacity(test_env, cpu):
+        if test_env.nrg_model:
+            return test_env.nrg_model.get_cpu_capacity(cpu)
+        else:
+            return test_env.target.read_int(
+                '/sys/devices/system/cpu/cpu{}/cpu_capacity'.format(cpu))
```
bjackman: Hmm - my thinking was that the `EnergyModel` probably read its data from the target anyway so it's a waste of time (and it can be quite a lot of time) to do that again.  If you provided it by hand then you probably had a good reason for doing that so we should respect it.

tests/eas/load_tracking.py
```diff
@@ -45,6 +45,15 @@ class _LoadTrackingBase(LisaTest):
         'modules': ['cpufreq', 'cgroups'],
     }
 
+    @memoized
+    @staticmethod
+    def _get_cpu_capacity(test_env, cpu):
+        if test_env.nrg_model:
+            return test_env.nrg_model.get_cpu_capacity(cpu)
+        else:
```
bjackman: OK

bjackman: Actually... I went to edit this and it's harder to read IMO:

```
        if test_env.nrg_model:
            return test_env.nrg_model.get_cpu_capacity(cpu)
        return test_env.target.read_int(
            '/sys/devices/system/cpu/cpu{}/cpu_capacity'.format(cpu))
```
vs
```
        if test_env.nrg_model:
            return test_env.nrg_model.get_cpu_capacity(cpu)
        else:
            return test_env.target.read_int(
                '/sys/devices/system/cpu/cpu{}/cpu_capacity'.format(cpu))
```
In the second one you can easily see "this either returns X or Y" but in the first one you have to be like "if P then it returns X, then if that hasn't happened it returns Z"... do you know what I mean?

derkling: I would simply add an empty line before the last `return`
```python
        if test_env.nrg_model:
            return test_env.nrg_model.get_cpu_capacity(cpu)

        return test_env.target.read_int(
            '/sys/devices/system/cpu/cpu{}/cpu_capacity'.format(cpu))
```

That reads why better and it also better marks that you are done if you enter the IF branch.

Maybe it's just mine obsession but I cannot really get an `else` after a `return` :fearful:

tests/eas/load_tracking.py
derkling: fair enough...  ok on this.

bjackman:
OK, updated with your comment addressed.

Issue #329, opened 2017-03-11
Add a proper support for commands execution while a test is running


derkling:
While a test is executed, we should ensure the most quite and stable execution environment, such that only test-specific tasks are in execution and exercise the scheduler behaviours.

Unfortunately some tests, like the one merged in #325, can require actions to be performed while the test is in execution. The main issue is that mimicking user actions using the System API send commands from the host to the target while the test is running. This results in a set of new tasks being generated for example at each and every swipe action, tasks which can affects the test itself.
These new tasks are quite evident in the collected trace and they can potentially impact quite a lot in task placement and ultimately in the overall test performance.

There are two possible solutions:

1. find a way to ensure that the ADB server is pinned on CPU0, thus forcing all the commands we send to run on that CPU only. This should mitigate the side-effect of the test instrumentation on the scheduler behaviours.

2. think about a possible API extension for the Benchmark class which allows to generate on-the-fly scripts to be kicked-off on target and executed on background without requiring to generate more task.

The first solution is worth to have in any case and it's more simple to introduce. The second one is much more interesting but can be introduced for completeness as a second step.

derkling:
Fixed by #403

PR #403, opened 2017-05-22
Target-side command execution

valschneider:
This is in response to #329. The example that goes with it is on GMaps, but this uses another PR from devlib (https://github.com/ARM-software/devlib/pull/127), so it cannot be merged yet but is up for comments.

I've looked at the traces on kernelshark, and adbd is running only at the very start & end of the workload, which is a good start.

### DIFF
--- a/libs/utils/android/workloads/gmaps.py
+++ b/libs/utils/android/workloads/gmaps.py
@@ -21,6 +21,7 @@ import logging
 
 from time import sleep
 
+from target_script import TargetScript
 from android import Screen, System
 from android.workload import Workload
 
@@ -90,20 +91,33 @@ class GMaps(Workload):
         # Allow the activity to start
         sleep(1)
 
-        self.tracingStart()
+        script = TargetScript(self._te, "gmaps_swiper.sh")
+        self._log.debug('Accumulating commands')
+
+        for i in range(swipe_count):
+            System.hswipe(script, 20, 80, 100, True)
+            script.append('sleep 1')
+            System.hswipe(script, 20, 80, 100, False)
+            script.append('sleep 1')
+            System.vswipe(script, 40, 60, 100, True)
+            script.append('sleep 1')
+            System.vswipe(script, 40, 60, 100, False)
+            script.append('sleep 1')
+
+        self._log.debug('Accumulation done')
+
+        # Push script to the target
+        script.push()
+
         self._log.info('Opening GMaps to [%s]', loc_url)
         # Let GMaps zoom in on the location
         sleep(2)
 
-        for i in range(swipe_count):
-            System.hswipe(self._target, 20, 80, 100, True)
-            sleep(.5)
-            System.hswipe(self._target, 20, 80, 100, False)
-            sleep(.5)
-            System.vswipe(self._target, 40, 60, 100, True)
-            sleep(.5)
-            System.vswipe(self._target, 40, 60, 100, False)
-            sleep(.5)
+        self.tracingStart()
+
+        self._log.info('Launching target script')
+        script.run()
+        self._log.info('Target script ended')
 
         self.tracingStop()
 
--- a/libs/utils/target_script.py
+++ b/libs/utils/target_script.py
@@ -1,0 +1,116 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2015, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+
+SCRIPT_NAME = 'remote_script.sh'
+
+class TargetScript(object):
+    """
+    This class provides utility to create and run a script
+    directly on a devlib target.
+
+    The execute() method is made to look like Devlib's, so a Target instance can
+    be swapped with an instance of this TargetScript class, and the commands
+    will be accumulated for later use instead of being executed straight away.
+
+    :param env: Reference TestEnv instance. Will be used for some commands
+        that must really be executed instead of accumulated.
+    :type env: TestEnv
+
+    :param script_name: Name of the script that will be pushed on the target,
+        defaults to "remote_script.sh"
+    :type script_name: str
+    """
+
+    _target_attrs = ['screen_resolution', 'android_id', 'abi', 'os_version', 'model']
+
+    def __init__(self, env, script_name=SCRIPT_NAME):
+        self._env = env
+        self._target = env.target
+        self._script_name = script_name
+        self.commands = []
+        
+    # This is made to look like the devlib Target execute()
+    def execute(self, cmd):
+        """
+        Accumulate command for later execution.
+
+        :param cmd: Command that would be run on the target
+        :type cmd: str
+        """
+        self.append(cmd)
+
+    def append(self, cmd):
+        """
+        Append a command to the script.
+
+        :param cmd: Command string to append
+        :type cmd: str
+        """
+        self.commands.append(cmd)
+
+    # Some commands may require some info about the real target.
+    # For instance, System.{h,v}swipe needs to get the value of
+    # screen_resolution to generate a swipe command at a given
+    # screen coordinate percentage.
+    # Thus, if such a property is called on this object,
+    # it will be fetched from the 'real' target object.
+    def __getattr__(self, name):
+        if name in self._target_attrs:
+            return getattr(self._target, name)
+
+        return getattr(super, name)
+            
+    def push(self):
+        """
+        Push a script to the target
+
+        The script is created and stored on the host,
+        and is then sent to the target.
+
+        :param path: Path where the script will be locally created
+        :type path: str
+        :param actions: List of actions(commands) to run
+        :type actions: list(str)
+        """
+
+        actions = ['set -e'] + self.commands + ['set +e']
+        actions = ['#!{} sh'.format(self._target.busybox)] + actions
+        actions = str.join('\n', actions)
+
+        self._remote_path = self._target.path.join(self._target.executables_directory,
+                                                   self._script_name)
+        self._local_path = os.path.join(self._env.res_dir, self._script_name)
+
+        # Create script locally
+        with open(self._local_path, 'w') as script:
+            script.write(actions)
+
+        # Push it on target
+        self._target.push(self._local_path, self._remote_path)
+        self._target.execute('chmod +x {}'.format(self._remote_path))
+
+    def run(self):
+        """
+        Run the previously pushed script
+        """
+
+        if self._target.file_exists(self._remote_path):
+            self._target.execute(self._remote_path)
+        else:
+            raise IOError('Remote script was not found on target device')


libs/utils/android/workloads/gmaps.py
```diff
@@ -90,20 +90,38 @@ def run(self, out_dir, location_search, swipe_count=5, collect=''):
         # Allow the activity to start
         sleep(1)
 
-        self.tracingStart()
-        self._log.info('Opening GMaps to [%s]', loc_url)
-        # Let GMaps zoom in on the location
-        sleep(2)
+        # Hack: screen resolution is memoized, so call it here
```
valschneider: Not too happy with that one here... But I felt the need for a dry-run mode in devlib, as modifying every method in System to add a dry-run seemed absolutely barbaric.

valschneider:
Updated the branch following @bjackman's recommendation. Accumulator/dry-run is now done in LISA instead of devlib.

libs/utils/dummy_target.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @property
+    def accumulated_commands(self):
+        return self._accu_cmds
+
+    def clear_accumulator(self):
+        self._accu_cmds = []
+
+    # This is made to look like the devlib Target execute()
+    # Only the command argument is used, the rest is ignored
```
bjackman: > the rest is ignored

Then we should really throw an error if anything but the default value is provided. Also to be *really* correct we should start the script with `set -e` to get the correct `check_exit_code` behaviour.

libs/utils/android/workload.py
```diff
@@ -88,6 +88,29 @@ def run(self, out_dir, collect='',
...
+        :param actions: List of actions(commands) to run
+        :type actions: list(str)
+        """
+
+        actions = ['#!{} sh'.format(self._target.busybox)] + actions
+        actions = str.join('\n', actions)
+
+        self._scriptpath = '{}/remote_script.sh'.format(self._target.executables_directory)
+
+        self._target.execute('echo \"{}\" > {}'.format(actions, self._scriptpath))
```
bjackman: When hacking similar stuff in the past I have had issues with this approach due to the length of the command (IIRC I found that ADB on Google Pixel has a 4096 byte limit), so it would be more robust to write this to a file and then use `push`. Might be slower, but I think we'll have to live with that.

libs/utils/android/workload.py
```diff
@@ -88,6 +88,29 @@ def run(self, out_dir, collect='',
...
+        self._target.execute('chmod +x {}'.format(self._scriptpath))
+
+    def runTargetScript(self):
+        """
+        Run the previously pushed script
+        """
+        if self._scriptpath:
+            self._target.execute(self._scriptpath)
```
bjackman: Might need to tweak the interface a bit but looks like these methods could go in `DummyTarget`, right?

libs/utils/android/workload.py
derkling: I agree, we should build a script locally and push on target for execution. This allows also to keep a local copy of the executed script in the `Workload::run::out_dir` results folder.

libs/utils/android/workload.py
derkling: Indeed, it makes sense to encapsulate in `DummyTarget` all the functions required to setup and run the script.

libs/utils/android/workloads/gmaps.py
```diff
@@ -90,20 +91,34 @@ def run(self, out_dir, location_search, swipe_count=5, collect=''):
         # Allow the activity to start
         sleep(1)
 
-        self.tracingStart()
+        dummy = DummyTarget(self._target)
```
derkling: Maybe using this kind of naming can make things more easy to read:
```python
script = ScriptTarget(self._target)
#... all actions are accumulated here and will be already part of the ScriptTarget object
script.setup(out_dir) # Here we specify where the local script has to be generated
#... tracing and other stuff
script.run()
#... tracing and other stuff
```

libs/utils/dummy_target.py
valschneider: The reason for adding a "trash collector" argument was to accept calls to `execute()` where `timeout` / `check_exit_code` / `as_root` are set, but all we're really looking for is the command itself.

Arguably handling could be improved by using these arguments to tweak the generated script lines.

libs/utils/android/workloads/gmaps.py
valschneider: If the list of actions is directly grabbed from the accumulation list, maybe there should be a setter property for it ? A user might want to filter out some commands with regexps or something.

libs/utils/dummy_target.py
bjackman: > Arguably handling could be improved by using these arguments to tweak the generated script lines.

Indeed, but until that's done we shouldn't let other code be fooled into thinking they work.

libs/utils/dummy_target.py
```diff
@@ -0,0 +1,48 @@
...
+        self._accu_cmds.append(cmd)
+
+    # Use the DummyTarget methods if they are available
+    # Reach to the real target otherwise
+    def __getattr__(self, name):
+        try:
+            return getattr(super, name)
+        except AttributeError:
+            if self._real_target:
+                return getattr(self._real_target, name)
```
bjackman: Just thinking about this again - for the [reasons I discussed](https://github.com/ARM-software/devlib/pull/127#issuecomment-303364152) on the devlib PR, I still don't think we should pass anything directly through to the real target. Is this needed for the current usage?

valschneider: For the vswipe/hswipe, yes. There's a command sent out to the target to figure out the screen size and generate the input command. We need the real values - so we can either painfully copy every target attribute we may need before accumulating commands, or just reach out to the real target to get this data.

bjackman: Gotcha. Can we whitelist attributes that are OK to use this way?

valschneider: I suppose we could, but we'd have to edit that list every time we stumble upon an API that uses a new attribute. I don't see the problem with letting the target run every command but execute - do you have a specific problematic call in mind ? Maybe it would be easier to blacklist them then.

bjackman: Problematic calls are anything that shares target state with an `execute` call. E.g consider this example (where, unbeknownst to the naive reader, `target` is a dummy):

```
target.execute('cat /sys/devices/system/cpu/online')
target.hotplug.online_all() 
```

Basically any target operation that isn't `memoized` is suspect. We could in theory modify the `memoized` decorator in devlib so that we can detect it...

valschneider: Yeah ok I see what you mean. I'm going to dig around and see if I can find a proper way of handling this, otherwise I'll go for the poor man's whitelist.

valschneider:
I've stumbled on a thing that can extract the decorators of a class' methods: https://stackoverflow.com/a/31197273
The end result works, but it's definitely hacky. Still, `Target`'s memoized attributes should be safe to call - tell me what you think about it.

bjackman:
Phwoar that solution is pretty badass.. but IMO too extreme (IIUC it's actually parsing the Python code...) given that it's possible for us to modify devlib. To be honest the whole dummy `execute` is a bit scary so I prefer a whitelist to avoid spreading usage too far (we can always get rid of the whitelist once we have some confidence in the usage model/API/implementation design).

valschneider:
Yeah, that class has pretty much turned into a playground for me to have (too much) fun with Python. I'll go with a whitelist for now.

libs/utils/target_script.py
```diff
@@ -0,0 +1,84 @@
...
+
+class TargetScript(object):
+    """
+    This class provides utility to create and run a script
+    directly on a devlib target.
+    """
+
+    _target_attrs = ['screen_resolution', 'android_id', 'abi', 'os_version', 'model']
+
+    def __init__(self, env):
```
derkling: Perhaps it could be useful to have:

```python
def __init__(self, env, script_name=SCRIPT_NAME):
    #...
    self.script_name = script_name
```

This would allow eventually to keep a local copy of different scripts generated by different workloads.

libs/utils/target_script.py
```diff
@@ -0,0 +1,84 @@
...
+    # Some commands may require some info about the real target.
+    # For instance, System.{h,v}swipe needs to get the value of
+    # screen_resolution to generate a swipe command at a given
+    # screen coordinate percentage.
+    # Thus, if such a property is called on this object,
+    # it will be fetched from the 'real' target object.
+    def __getattr__(self, name):
+        if name in self._target_attrs:
+            return getattr(self._env.target, name)
+        else:
```
derkling: This else can be removed

libs/utils/target_script.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    This class provides utility to create and run a script
+    directly on a devlib target.
+    """
+
+    _target_attrs = ['screen_resolution', 'android_id', 'abi', 'os_version', 'model']
+
+    def __init__(self, env):
+        self._env = env
+        self.commands = []
```
derkling: Perhaps you can cache `self._target = self._env.target` which is used quite a lot below

libs/utils/target_script.py
```diff
@@ -0,0 +1,84 @@
...
+        :type path: str
+        :param actions: List of actions(commands) to run
+        :type actions: list(str)
+        """
+
+        actions = ['set -e'] + self.commands + ['set +e']
+        actions = ['#!{} sh'.format(self._env.target.busybox)] + actions
+        actions = str.join('\n', actions)
+
+        self._remote_path = '{}/{}'.format(self._env.target.executables_directory, SCRIPT_NAME)
```
derkling: Should be:
```python
self._remote_path = self._env.target.path.join(self._target.executables_directory, self.script_name)
```

libs/utils/target_script.py
```diff
@@ -0,0 +1,84 @@
...
+        self._local_path = '{}/{}'.format(self._env.res_dir, SCRIPT_NAME)
```
derkling: Here instead we can use `os.path.join()`

libs/utils/target_script.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Push it on target
+        self._env.target.push(self._local_path, self._remote_path)
+        self._env.target.execute('chmod +x {}'.format(self._remote_path))
+
+    def run(self):
+        """
+        Run the previously pushed script
+        """
+        if self._remote_path:
```
derkling: Perhaps safer to check with `self._target.file_exists(self._remote_path)` and raise is not?

libs/utils/android/workloads/gmaps.py
```diff
@@ -90,20 +91,33 @@ def run(self, out_dir, location_search, swipe_count=5, collect=''):
         # Allow the activity to start
         sleep(1)
 
-        self.tracingStart()
+        script = TargetScript(self._te)
+        self._log.debug('Accumulating commands')
+
+        for i in range(swipe_count):
+            System.hswipe(script, 20, 80, 100, True)
+            script.commands.append('sleep 1')
```
derkling: Should not be more safe to keep `commands` private and make client code slightly slimmer by exposing an append method, e.g. I would prefer to use:

```python
script.append('sleep 1')
```

libs/utils/target_script.py
derkling: Actually, looking at the example in GMaps, perhaps we can be even more generic by adding an additional

```python
def setup(script_name=SCRIPT_NAME):
    self.script_name = script_name
    self.commands = []
```
In that case, the current `setup()` method should be renamed to `push()`, which is also better describing the fact that we are building and deploying the script on target to be ready for `run()`

The combination of `setup()` and `push()` would allow to create different script, perhaps for different steps of an experiment, push and run them without creating new TargetScript objects.

libs/utils/android/workloads/gmaps.py
valschneider: My reasoning behind exposing commands was to let the user modify it at will. Perhaps a certain call will generate a lot of unneeded commands and the accumulated commands would need to be filtered out. The user could also very well write
`script.execute('sleep 1')`
instead of 
`script.commands.append('sleep 1')`
but I find it makes a lot more sense to use "append" to a command to the script rather than to use "execute"

valschneider:
Added an `append` method to manually add some commands to the script, instead of using the internal list object.

derkling:
can we just add a micro documentation for both `execute` and the new `append` methods?

derkling:
Same for the TargetScript class.. can you add a description of what the ctor parameters define?

valschneider:
Added proper documentation

PR #405, opened 2017-05-24
EAS workshop practical support

cloverpop:
Add new IPython notebooks for EAS workshop in Juno board
1) arch CPUfreq invariance
2) fork migration
3) CGroup and cpuset
4) prefer idle


bjackman:
Hi,

I'm not sure if there's any need to merge these notebooks into LISA - it might be better to just keep them separately and use them for the workshops. The notebooks in LISA are intended to be educational, to be used as starting points for getting to grips with LISA, but all the LISA features covered here are already covered in other notebooks. Do you have a major reason for merging into LISA?

cloverpop:
First of all, I am OK to keep them separately.
But actually what you said "but all the LISA features covered here are already covered in other notebooks",  I did not find straightforward notebooks to demonstrate those EAS features listed in above. If I missed them, please correct me. Even if there is something about freq invariance, IMHO, the original notebook is not useful to show the comparison results to customers directly.

derkling:
In  principle I would not be against the idea to add further tutorial/workshop material to LISA, however this would requires at least a big effort on keep consistency among them and update the with respect to the most recent APIs.

For example, I had a look at the `cgroup_cpuset.ipynb` notbeook you propose and it does not even use the devlib's cgroup API to properly setup CGroups. Here is the example provided by LISA:
   https://github.com/ARM-software/lisa/blob/master/ipynb/examples/devlib/cgroups_example.ipynb
which can probably be improved, but at least it represents the proper way LISA users should configure CGroups. Other notebooks also uses the `target.execute()` API where instead a devlib's one can be used instead.

Finally, merging additional documentation notebook would require an additional maintenance burden since we have to ensure not breaking them every time there is an API update.

Thus, if it's possible to keep them separated, and provided as workshop collateral material, I think it's the better option we have so fare. Otherwise, we should find a way to merge/meld the proposed examples into the notebooks provided by the `ipynb/examples` folder.

PR #406, opened 2017-05-25
Add support for new upstream-friendly load-tracking events

bjackman:
There is a new upstream-friendly version 'sched_load_se' of the
load-tracking trace events added by EAS. Add an abstraction so that
LISA can work with both this new version and the older
'sched_load_avg_task' events.

Trace._sanitize_SchedLoadAvgTask is removed and the logic is moved
into the new function _dfg_task_lt_events in tasks_analysis.

The users of sched_load_avg_cpu in tasks_analysis are updated to use
the new _dfg method (except where their use is specific to the old
sched_load_avg_cpu event).

Also add basic unit tests to exercise the code changed in this
commit.


bjackman:
The CI failure is due to [this changed introduced by Pandas 0.20.1](http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#groupby-describe-formatting) breaking [this `tasks_analysis` code](https://github.com/ARM-software/lisa/blob/master/libs/utils/analysis/tasks_analysis.py#L80), which happens to be exercised by the unit tests added in this PR. So I'll need to

1. Split out a commit that adds that test
2. Fix the issue and submit a PR with that test and fix
3. Rework this PR

But I'm going to leave it open so it doesn't get forgotten.

bjackman:
Phew, finally got the tests passing. Decided I'm just gona leave this as one PR rather than faffing around too much.

bjackman:
Added support for the CPU events (sched_load_cfs_rq) too.

tests/lisa/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import matplotlib
+# Prevent matplotlib from trying to connect to X11 server, for headless testing.
+# Must be done before importing matplotlib.pyplot or pylab
+matplotlib.use('Agg')
```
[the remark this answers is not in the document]
valschneider: Yeah I was about to comment on that. Could this maybe be added to one layer above (`tests/__init__.py`) ? I'm hitting that issue with tests/eas/generics.py so it's a common issue.

valschneider:
Will you drop https://github.com/ARM-software/lisa/pull/518/ since you have the same commits in here ? We want to merge both anyway, it's just a matter or ordering the contents.

bjackman:
I'll drop the commits from here once 518 is merged.

tests/lisa/example_traces/sched_load/platform.json
```diff
@@ -0,0 +1,79 @@
+{
```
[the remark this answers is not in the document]
bjackman: The test_trace automatically tests a) with platform b) with platform with the 'clusters' key removed c) without platform.

So I think it's nice to have this here.

tests/lisa/example_traces/sched_load_avg/platform.json
```diff
@@ -0,0 +1,108 @@
+{
```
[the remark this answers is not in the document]
bjackman: Have added tests for the presence of the extra columns.

bjackman:
Fixed failing test

libs/utils/analysis/tasks_analysis.py
```diff
@@ -186,6 +186,42 @@ def _dfg_rt_tasks(self, min_prio=100):
 
         return rt_tasks
 
+    def _dfg_task_lt_events(self):
```
derkling: Should not `df` be initialised to None?

libs/utils/analysis/tasks_analysis.py
```diff
@@ -384,7 +419,7 @@ def plotBigTasks(self, max_tasks=10, min_samples=100,
             return
 
         # Get the list of events for all big frequent tasks
-        df = self._dfg_trace_event('sched_load_avg_task')
+        df = self._dfg_task_lt_events()
         big_frequent_tasks_events = df[df.pid.isin(big_frequent_task_pids)]
```
derkling: Still missing check for `df not None`?
Like the one you have at line 581

libs/utils/analysis/tasks_analysis.py
```diff
@@ -636,7 +671,7 @@ def _plotTaskSignals(self, axes, tid, signals, is_last=False):
         :type is_last: bool
         """
         # Get dataframe for the required task
-        util_df = self._dfg_trace_event('sched_load_avg_task')
+        util_df = self._dfg_task_lt_events()
```
derkling: Still missing check for `util_df not None`?
Like the one you have at line 581

libs/utils/analysis/tasks_analysis.py
```diff
@@ -699,15 +734,12 @@ def _plotTaskResidencies(self, axes, tid, signals, is_last=False):
...
-
-        if 'cluster' in util_df:
-            data = util_df[util_df.pid == tid][['cluster', 'cpu']]
-            for ccolor, clabel in zip('gr', ['LITTLE', 'big']):
-                cdata = data[data.cluster == clabel]
-                if len(cdata) > 0:
-                    cdata.plot(ax=axes, style=[ccolor+'+'], legend=False)
-
+        util_df = self._dfg_task_lt_events()
+        data = util_df[util_df.pid == tid][['cluster', 'cpu']]
```
derkling: Still missing check for `util_df not None`?
Like the one you have at line 581

libs/utils/analysis/tasks_analysis.py
derkling: Moreover: I think `task_load_events` was a better name... actually, to be more scheduler aligned, what about `sched_task_load`?

libs/utils/analysis/tasks_analysis.py
bjackman: Yup actually it's needed above for big_frequenct_tasks_df

libs/utils/analysis/tasks_analysis.py
bjackman: This one isn't strictly needed as it's checked by the caller. But I'm gona add it in anyway, rather than have invisible dependencies between methods like that. Thanks for pointing it out.

libs/utils/analysis/tasks_analysis.py
bjackman: Ditto

libs/utils/analysis/tasks_analysis.py
```diff
@@ -57,15 +57,15 @@ def _dfg_top_big_tasks(self, min_samples=100, min_utilization=None):
             default: capacity of a little cluster
         :type min_utilization: int
         """
-        if not self._trace.hasEvents('sched_load_avg_task'):
-            self._log.warning('Events [sched_load_avg_task] not found')
+        if self._dfg_task_load_events() is None:
```
derkling: This is kind-of confusing: we check here (L60) and get the df after (L68). That's error prone if we change the code.

I would prefer something like:

```python
df = self._dfg_task_load_events()
if df is None:
    self._log.warning('No trace events for task signals, plot DISABLED')
    return None

# So that from now on we don't use `_dfg_task_load_events()` in this function.
```

libs/utils/analysis/tasks_analysis.py
```diff
@@ -186,6 +186,44 @@ def _dfg_rt_tasks(self, min_prio=100):
...
+            df = df[df.pid != -1]
+
+        if not self._trace.has_big_little:
+            return df
+
+        df['cluster'] = np.select(
+                [df.cpu.isin(self._trace.platform['clusters']['little'])],
+                ['LITTLE'], 'big')
+
+        if 'nrg_model' in self._trace.platform:
```
derkling: Here also I would prefer a:

```python
if `nrg_model` not in self._trace.platform:
    return df
```

PR #408, opened 2017-06-01
android/system/monkey: use 0 physical keys as default.

balsini:
It may happen that, while launching tests on platforms withouth physical
keys (e.g. Hikey), monkey fails returning

> ** SYS_KEYS has no physical keys but with factor 2.0%.

becasue of its default configuration.
This is solved by explicitly defining the --pct_syskeys parameters, 0 as
default value.

### DIFF
--- a/libs/utils/android/system.py
+++ b/libs/utils/android/system.py
@@ -359,7 +359,7 @@ class System(object):
                     GET_FRAMESTATS_CMD.format(apk_name, out_file))
 
     @staticmethod
-    def monkey(target, apk_name, event_count=1):
+    def monkey(target, apk_name, event_count=1, pct_syskeys=0):
         """
         Wrapper for adb monkey tool.
 
@@ -381,8 +381,13 @@ class System(object):
 
         :param event_count: number of events to generate
         :type event_count: int
+        
+        :param pct_syskeys: percentage of physical keys to be used
+        :type pct_syskeys: int
         """
-        target.execute('monkey -p {} {}'.format(apk_name, event_count))
+        target.execute('monkey --pct-syskeys {} -p {} {}'.format(pct_syskeys,
+                                                                 apk_name,
+                                                                 event_count))
 
     @staticmethod
     def list_packages(target, apk_filter=''):


libs/utils/android/system.py
```diff
@@ -359,7 +359,7 @@ def gfxinfo_get(target, apk_name, out_file):
                     GET_FRAMESTATS_CMD.format(apk_name, out_file))
 
     @staticmethod
-    def monkey(target, apk_name, event_count=1):
+    def monkey(target, apk_name, event_count=1, pct_syskeys=0):
```
derkling: There are many different turnings for different events:
   https://developer.android.com/studio/test/monkey.html
It would be nice to make this interface a bit more generic.

Perhaps we can  pass in a dictionary of percentages and map them on supported options.
Something like:

```python
def monkey(target, apk_name, event_count=1, percentages=None):
    PCT_PARAMS = ['touch', 'motion',  ...] # all supported events

    pcts = ''
    for evt, pct in percentages.iteritems():
        if evt not in PCT_PARAMS:
            continue
        if not 0 <= pct <= 100:
            raise ValueError("{} percentage out of [0..100] range")
        pcts += ' --pct-{} {}'.format(evt, value)
    #...
    target.execute('monkey {} -p {} {}'.format(pcts, apk_name, event_count))


# Then it should be possible to:

monkey(target, apk_name, event_count,
       percentages={'syskeys' : 0})


```

balsini: I definitely agree with you.
Another idea could be to make the monkey manager even more independent by defining his own separate class. It could be useful for several automated testing.

What do you think?

derkling: That would make sense, especially if we consider the idea to add more and more functionalities.
In that case however I would suggest you to have a look at [devlib](https://github.com/ARM-software/devlib). 

In the future, we are planning to move most of the LISA's Android API within the devlib library, where it probably makes more sense for them to be. Indeed, devlib is a target abstraction library and thus it would probably be the best place to host a complete new monkey wrapper class.

For me it would be ok to start by merging this simple fix/extension you already got in LISA.
In a second stage we can try to add a more comprehensive class to devlib, where it could be probably hosted in:
https://github.com/ARM-software/devlib/blob/master/devlib/module/android.py

cirosantilli:
I also get the error on hikey.

Does your app open at all with monkey -p without this patch?

Mine doesn't, would that patch fix the not opening as well?

PR #411, opened 2017-06-06
Save & restore cpufreq governor tunables for RTA calibration

bjackman:
Governor tunables are reset to default when switching governors.

This is particularly relevant when using the Executor, which calls
te.calibration() after _setup_cpufreq. If the calibration has not
been done yet, that will result in setting the performance governor
then restoring the previous governor. However if _setup_cpufreq set
up any tunables they are not currently restored. This commit fixes
that.

----

While I'm touching this code, I've also updated it to use the `iter_domains` method I added to devlib.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -102,7 +102,9 @@ class RTA(Workload):
         old_governors = {}
         for domain in target.cpufreq.iter_domains():
             cpu = domain[0]
-            old_governors[cpu] = target.cpufreq.get_governor(cpu)
+            governor = target.cpufreq.get_governor(cpu)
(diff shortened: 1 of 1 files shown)


libs/wlgen/wlgen/rta.py
```diff
@@ -142,8 +144,9 @@ def calibrate(target):
         #   Setting a governor for a cpu will set it for all cpus in the same
         #   clock domain, so only restoring the governor of one cpu per domain
         #   is enough to restore all of the previous governors
-        for cpu, governor in old_governors.iteritems():
```
valschneider: Maybe it is worth mentioning that the tunables are per-domain as well ?

bjackman: Updated the comment

PR #412, opened 2017-06-06
Update rt-app

bjackman:
Sorry this is insanely boring but this basically requires testing by hand. I have tested a selection of notebooks on TC2 and Juno but this is bound to introduce some issues so any testing at all is greatly appreciated.


bjackman:
Er.. temporary NAK, seeing some very strange behaviour with this version of rt-app and the following JSON:

```
{
    "global": {
        "calibration": "CPU1",
        "default_policy": "SCHED_OTHER",
        "duration": -1,
        "logdir": "/home/brendan/devlib-target/run_dir"
    },
    "tasks": {
        "task_my_task": {
            "loop": 1,
            "phases": {
                "p000001": {
                    "loop": 1,
                    "sleep": 1000000
                },
                "p000002": {
                    "loop": 100,
                    "run": 3000,
                    "timer": {
                        "period": 10000,
                        "ref": "task_my_task"
                    }
                }
            },
            "policy": "SCHED_OTHER"
        }
    }
}
```

Need to investigate

bjackman:
Scratch that, "weird behaviour" I was seeing in the trace was rt-app calibrating itself, because I have `"calibration": "CPU1"`. Carry on, this PR is still good.

ionela-voinescu:
I've tested the arm64 binary in gem5. I plan to update the arm64 taskset and trace-cmd binaries to 64bit executables also as the 32bit ones as were causing issues in gem5 so it would be good to merge this soon.

PR #413, opened 2017-06-07
energy_model: Fix read-multiple-files helper

bjackman:
This helper executes a "cat" command to read a large number of files
in one go. In order to avoid this command becomind too long for ADB
to cope with, it was written to use glob patterns to cat the files,
instead of the explicit paths (which it knows).

However a user has reported an issue with this approach, which I
suspect is because the order of the paths returned by `find` differs
from the order of the shell's expansion.

Fix the cat command to explicitly list the paths to be read. In order
to avoid the command-too-long issue, write the potentially-too-long
command to a file, push it, and execute it via sh.

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -38,18 +38,27 @@ def read_multiple_oneline_files(target, glob_patterns):
     target or connection is slow this saves a lot of time when reading a large
     number of files.
 
+    This will only work safely on stationary files, don't try to use it where
+    the glob expansion will change often - for example /proc/**/autogroup would
+    not work because /proc/ entries will likely appear & disappear while we're
+    reading them.
+
     :param target: devlib target object to read from
     :param glob_pattern: Unix glob pattern matching the files to read
     :returns: A dictionary mapping matched paths to the values read. ``{}`` if
               no paths matched the globs.
     """
+    find_cmd = 'find ' + ' '.join(glob_patterns)
     try:
-        paths = target.execute('find ' + ' '.join(glob_patterns)).split()
+        paths = target.execute(find_cmd).split()
     except TargetError:
         return {}
 
-    cmd = 'cat ' + ' '.join(glob_patterns)
+    cmd = '{} | {} xargs cat'.format(find_cmd, target.busybox)
     contents = target.execute(cmd).splitlines()
+
+    if len(contents) != len(paths):
+        raise RuntimeError('File count mismatch while reading multiple files')
 
     return dict(zip(paths, contents))
 


valschneider:
LGTM

bjackman:
Thanks for the review. Before I merge I'm gona come back and test/review this again tomorrow with fresh eyes because I wrote this patch all in one go then found that it worked the first time I ran it... too good to be true.

derkling:
Is not something like this robust enough?
```bash
adb shell 'find /sys/devices/system/cpu/cpu**/topology/*_siblings | while read FILE; do echo -n "$FILE: "; cat $FILE; done'
```
Which returns:
```bash
/sys/devices/system/cpu/cpu0/topology/core_siblings: 0f
/sys/devices/system/cpu/cpu0/topology/thread_siblings: 01
/sys/devices/system/cpu/cpu1/topology/core_siblings: 0f
/sys/devices/system/cpu/cpu1/topology/thread_siblings: 02
/sys/devices/system/cpu/cpu2/topology/core_siblings: 0f
/sys/devices/system/cpu/cpu2/topology/thread_siblings: 04
/sys/devices/system/cpu/cpu3/topology/core_siblings: 0f
/sys/devices/system/cpu/cpu3/topology/thread_siblings: 08
/sys/devices/system/cpu/cpu4/topology/core_siblings: f0
/sys/devices/system/cpu/cpu4/topology/thread_siblings: 10
/sys/devices/system/cpu/cpu5/topology/core_siblings: f0
/sys/devices/system/cpu/cpu5/topology/thread_siblings: 20
/sys/devices/system/cpu/cpu6/topology/core_siblings: f0
/sys/devices/system/cpu/cpu6/topology/thread_siblings: 40
/sys/devices/system/cpu/cpu7/topology/core_siblings: f0
/sys/devices/system/cpu/cpu7/topology/thread_siblings: 80
```
From there you can easily use a regexp to exactly parse IDs and corresponding values?

As an aside comment,it seems that the `read_multiple_oneline_files()` function is used only a couple of time and always to parse em data... if we really wanna got for a script based solution, should not be better to have a new EM related set of functions directly integrated into devlib::shutils?
There are already other  examples of devlib::shutils which parse multiple files are return a single output which is also easy to parse from the python side... for example, mayber we can code a shutil which returns (in the previous example) something like:

```bash
cpu0 core_siblings 0f
cpu0 thread_siblings 01
cpu1 core_siblings 0f
cpu1 thread_siblings 02
cpu2 core_siblings 0f
cpu2 thread_siblings 04
cpu3 core_siblings 0f
cpu3 thread_siblings 08
cpu4 core_siblings f0
cpu4 thread_siblings 10
cpu5 core_siblings f0
cpu5 thread_siblings 20
cpu6 core_siblings f0
cpu6 thread_siblings 40
cpu7 core_siblings f0
cpu7 thread_siblings 80
```

bjackman:
Hmm, yeah I think a generic "shutil" that reads loads of files in one go is a good idea.. I will try to cook something up.

joelagnel:
Aside from the issue of where the script should live, I think its faster if you `cat` a bunch of files at once (in one command), than using `find` to find the files, and then `cat` them individually so I like that in @bjackman 's initial patch. (I could missing context so in that case I am sorry about making a silly comment)

bjackman:
Updated with a version somewhere in the middle - a bit like @derkling solution but using `xargs` instead of `while read` to reflect @joelagnel's point (which is very true - I tested it and a single cat command is indeed massively faster). However I've kept it as two separate commands in an attempt to detect the particular failure mode where the files are not present (rather than a blanket catch of `TargetError`).

PR #415, opened 2017-06-09
Add feature to set test iteration count from command line

bjackman:
This is a kludge, but I think doing this properly basically requires
implementing our own Nose test runner, or even our own test-running
framework entire. In the short term we really need this feature.

nosetests doesn't provide a way for the user to pass command-line
arguments to the test. To get around this, this commit adds a
parameter to lisa-test that sets an environment variable, which
LisaTest then reads to set or override the iterations count.

---

This will go pretty nicely with https://github.com/ARM-software/lisa/pull/348 but they are technically independent.

### DIFF
--- a/libs/utils/test.py
+++ b/libs/utils/test.py
@@ -87,6 +87,14 @@ class LisaTest(unittest.TestCase):
         test_env = TestEnv(test_conf=cls._getTestConf())
 
         experiments_conf = cls._getExperimentsConf(test_env)
+
+        if ITERATIONS_FROM_CMDLINE:
+            if 'iterations' in experiments_conf:
+                cls.logger.warning(
+                    "Command line overrides iteration count in "
+                    "{}'s experiments_conf".format(cls.__name__))
+            experiments_conf['iterations'] = ITERATIONS_FROM_CMDLINE
+
         cls.executor = Executor(test_env, experiments_conf)
 
         # Alias tests and workloads configurations
@@ -226,5 +234,21 @@ def experiment_test(wrapped_test, instance, args, kwargs):
 
 # Prevent nosetests from running experiment_test directly as a test case
 experiment_test.__test__ = False
+
+# Allow the user to override the iterations setting from the command
+# line. Nosetests does not support this kind of thing, so we use an
+# evil hack: the lisa-test shell function takes an --iterations
+# argument and exports an environment variable. If the test itself
+# specifies an iterations count, we'll later print a warning and
+# override it. We do this here in the root scope, rather than in
+# runExperiments, so that if the value is invalid we print the error
+# immediately instead of going ahead with target setup etc.
+try:
+    ITERATIONS_FROM_CMDLINE = int(
+        os.getenv('LISA_TEST_ITERATIONS', '0'))
+    if ITERATIONS_FROM_CMDLINE < 0:
+        raise ValueError('Cannot be negative')
+except ValueError as e:
+    raise ValueError("Couldn't read iterations count: {}".format(e))
 
 # vim :set tabstop=4 shiftwidth=4 expandtab
--- a/src/shell/lisa_shell
+++ b/src/shell/lisa_shell
@@ -284,18 +284,19 @@ Usage: lisa-test [args] FILE[:CLASS]
   to that tool.
 
   Examples:
-    Run all EAS Acceptance tests:
+    Run all EAS Generic tests:
 
-      lisa-test tests/eas/acceptance.py
+      lisa-test tests/eas/generic.py
 
-    Run ForkMigration test from EAS Acceptance suite:
+    Run RampUp test from EAS Generic suite:
 
-      lisa-test tests/eas/acceptance.py:ForkMigration
+      lisa-test tests/eas/generic.py:RampUp
 
-    Run ForkMigration test from EAS Acceptance suite, generating an XML test
+    Run RampUp test from EAS Generic suite, generating an XML test
     report via nose's XUnit plugin (see nosetests documentation):
 
-      lisa-test --with-xunit --xunit-file=report.xml tests/eas/acceptance.py:ForkMigration
+      lisa-test --with-xunit --xunit-file=report.xml tests/eas/generic.py:RampUp
+
 
 EOF
 }
--- a/src/shell/lisa_shell
+++ b/src/shell/lisa_shell
@@ -277,8 +277,12 @@ fi
 
 function _lisa-test-usage {
 cat <<EOF
-Usage: lisa-test [args] FILE[:CLASS]
+Usage: lisa-test [--iterations ITERATIONS] [args] FILE[:CLASS]
   Run automated tests. Tests can be found under the tests/ directory.
+
+  --iterations (or -n) sets the number of iterations to use for
+    each workload/target-conf, for tests that support this usage. 0 means use
+    the test's default iteration count.
 
   This is a wrapper for the 'nosetests' utility, additional arguments are passed
   to that tool.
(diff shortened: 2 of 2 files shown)


bjackman:
> We already changed the source so this is a no-brainer.

Huh? What source?

src/shell/lisa_shell
```diff
@@ -308,19 +319,38 @@ nosetests -v --nocapture --nologcapture \
...
+                        break
+                        ;;
+        esac
+        shift
+done
+
+if [[ $have_more_args == no ]]; then
+        # No arguments provided, default to "help"
+        _lisa-test-usage
+else
```
derkling: return -1 and remove this else?

src/shell/lisa_shell
```diff
@@ -308,19 +319,38 @@ nosetests -v --nocapture --nologcapture \
...
+                help|-h|--help)
+                        _lisa-test-usage
+                        return 0
+                        ;;
+                -n|--wload-iterations)
+                        iterations="$2"
+                        shift
+                        ;;
+                *)
+                        have_more_args=yes
```
derkling: What's that used for?
Can't we go for the usual
```bash
while getopts ...; do
  ...
done
```
and move the following if into the default case?

src/shell/lisa_shell
```diff
@@ -277,9 +277,12 @@ fi
 
 function _lisa-test-usage {
 cat <<EOF
-Usage: lisa-test [args] FILE[:CLASS]
+Usage: lisa-test [--wload-iterations ITERATIONS] [args] FILE[:CLASS]
```
derkling: why not saving @sinkap some typing and name this parameter just `iterations`?

libs/utils/test.py
```diff
@@ -87,6 +87,14 @@ def runExperiments(cls):
         test_env = TestEnv(test_conf=cls._getTestConf())
 
         experiments_conf = cls._getExperimentsConf(test_env)
+
+        if ITERATIONS_FROM_CMDLINE:
```
derkling: Why not naming this just ITERATIONS which matches the name used for the experiments_conf attribute?

libs/utils/test.py
```diff
@@ -227,4 +235,20 @@ def experiment_test(wrapped_test, instance, args, kwargs):
...
+# line. Nosetests does not support this kind of thing, so we use an
+# evil hack: the lisa-test shell function takes an --iterations-count
+# argument and exports an environment variable. If the test itself
+# specifies an iterations count, we'll later print a warning and
+# override it. We do this here in the root scope, rather than in
+# runExperiments, so that if the value is invalid we print the error
+# immediately instead of going ahead with target setup etc.
+try:
+    ITERATIONS_FROM_CMDLINE = int(
+        os.getenv('LISA_TEST_ITERATIONS_COUNT', '0'))
```
derkling: Here instead it should probable be just `LISA_TEST_ITERATIONS`

src/shell/lisa_shell
bjackman: The reason I went for `--wload-iterations` was because the thing that's repeated is each workload and not the test as a whole... but actually I guess that's pretty much the same thing so will change it.

src/shell/lisa_shell
bjackman: Yeah I did have a rationale for doing it this way but I can't remember what it was so it was probably nonsense :D

libs/utils/test.py
bjackman: Because I want it to be as clear as possible that it's a different source for the value.. the mechanism I've implemented here pretty convoluted so I think long descriptive names are gonna be helpful..

src/shell/lisa_shell
bjackman: Although I'd like to keep it as "raw bash" because `getopts` doesn't support longopts like "--iterations". [There's a separate thing called `getopt` but it isn't in OSX etc](https://stackoverflow.com/questions/402377/using-getopts-in-bash-shell-script-to-get-long-and-short-command-line-options) and I guess one day someone might want to get LISA working on Macs. And the raw bash is fine in this case I think.

bjackman:
- Changed the param to "--iterations"
- "LISA_TEST_ITERATIONS_COUNT" -> "LISA_TEST_ITERATIONS"
- Simplified the arg-parsing loop as you suggested

but

- Didn't use `getopts`
- Didn't rename "ITERATIONS_FROM_CMDLINE"

libs/utils/test.py
derkling: Well, we already have a big warning on console to notify the user about the override.

Moreover, I would not be so negative about considering this an hack... nosetest does not support command line parameters and thus, either we can add support for them or use environment variable. An environment variable is still a legal way to defined parameters for an app... ;-)

Moreover, to be honest, I don't think we will even update this code in the near future... mainly because that's doing exactly what we need and it perfectly works for what we need.

However, if you are more comfortable with this name... let's go with it! ;-)

src/shell/lisa_shell
derkling: But does it really make sense to run a test multiple times?
Eventually, and that's what we do, we run a workload multiple time and aggregate results to define if THE test pass or not... thus there should not be ambiguity and I'm pretty sure we will unlikely need in the future to run A test multiple times.

src/shell/lisa_shell
```diff
@@ -308,19 +319,35 @@ nosetests -v --nocapture --nologcapture \
...
-"xHELP")
-	_lisa-test-usage
-	;;
-*)
-	_lisa-test $*
-        local retcode=$?
-esac
-echo
+local iterations=0 # Means use default - see libs/utils/test.py
+local have_more_args=no
```
derkling: Is this still required?

src/shell/lisa_shell
```diff
@@ -308,19 +319,35 @@ nosetests -v --nocapture --nologcapture \
...
+        case "$1" in
+                help|-h|--help)
+                        _lisa-test-usage
+                        return 0
+                        ;;
+                -n|--iterations)
+                        iterations="$2"
+                        shift
+                        ;;
+                *)
```
derkling: Perhaps I'm note really getting what you wanna do here.
Isn't this extremely fragile?

AFAIU we are forced to put `-n/--interations `before the remaining params, otherwise we will pass it to the `__list_test` instead and run just 1 iteration.

Did not tried but I still have the impression that getopts can be useful to make this more robust... maybe I'll give a try later. Can you just post here an example of how you wanna call a test with parameters and multiple iterations?

src/shell/lisa_shell
bjackman: Nope, removing, thanks

src/shell/lisa_shell
bjackman: Hmm yeah that's true this doesn't work if you want to pass extra args to nosetests - that's probably what my rationale for the previous implementation was. Will see what I can come up with...

bjackman:
> But does it really make sense to run a test multiple times?
Eventually, and that's what we do, we run a workload multiple time and aggregate results to define if THE test pass or not... thus there should not be ambiguity and I'm pretty sure we will unlikely need in the future to run A test multiple times.

[Can't reply directly to comment]

Yep, I agree, I meant to change "--wload-iterations" to "--iterations" but missed out the docstring in the commit, will fix.

bjackman:
OK pushed a new version with hopefully a more robust parsing strategy that just picks up "-n" or "--iterations" and passes everything else through to nosetests. Can't figure out how to do this with getopts..

src/shell/lisa_shell
derkling: Is this going to be changed or you still prefer to keep this name?

bjackman: Agh, sorry I thought I changed it.

bjackman:
OK, updated again fixing the doc for "--iterations" parameter.. hopefully I didn't miss anything this time.

ionela-voinescu:
Except for some nitpicks it looks good to me. I'll start some tests of both PRs (https://github.com/ARM-software/lisa/pull/415 and https://github.com/ARM-software/lisa/pull/348) overnight and make changes to CI after these are merged. Good job and thank you for the changes. They are very useful!

PR #417, opened 2017-06-13
Viewer workload

valschneider:
This workload uses the action.VIEW intent. It will use the
most appropriate package for a given URI, e.g. Gmaps app for
a Gmaps location url. If the device doesn't have Google services,
it will open Gmaps in the browser, and so on.

### DIFF
--- a/ipynb/examples/android/workloads/Android_Viewer.ipynb
+++ b/ipynb/examples/android/workloads/Android_Viewer.ipynb
@@ -1,0 +1,321 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "# Generic Android viewer"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
+    "from conf import LisaLogging\n",
+    "LisaLogging.setup()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
+    "%pylab inline\n",
+    "\n",
+    "import json\n",
+    "import os\n",
+    "\n",
+    "# Support to access the remote target\n",
+    "import devlib\n",
+    "from env import TestEnv\n",
+    "\n",
+    "# Import support for Android devices\n",
+    "from android import Screen, Workload, System, ViewerWorkload\n",
+    "from target_script import TargetScript\n",
+    "\n",
+    "# Support for trace events analysis\n",
+    "from trace import Trace\n",
+    "\n",
+    "# Suport for FTrace events parsing and visualization\n",
+    "import trappy\n",
+    "\n",
+    "import pandas as pd\n",
+    "import sqlite3\n",
+    "\n",
+    "from IPython.display import display"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "# Test environment setup\n",
+    "\n",
+    "For more details on this please check out **examples/utils/testenv_example.ipynb**.\n",
+    "\n",
+    "**devlib** requires the ANDROID_HOME environment variable configured to point to your local installation of the Android SDK. If you have not this variable configured in the shell used to start the notebook server, you need to run a cell to define where your Android SDK is installed or specify the ANDROID_HOME in your target configuration.\n",
+    "\n",
+    "In case more than one Android device are connected to the host, you must specify the ID of the device you want to target in **my_target_conf**. Run **adb devices** on your host to get the ID."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true,
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
+    "# Setup target configuration\n",
+    "my_conf = {\n",
+    "\n",
+    "    # Target platform and board\n",
+    "    \"platform\"     : 'android',\n",
+    "    \"board\"        : 'hikey960',\n",
+    "    \n",
+    "    # Device serial ID\n",
+    "    # Not required if there is only one device connected to your computer\n",
+    "    \"device\"       : \"0123456789ABCDEF\",\n",
+    "    \n",
+    "    # Android home\n",
+    "    # Not required if already exported in your .bashrc\n",
+    "    #\"ANDROID_HOME\" : \"/home/vagrant/lisa/tools/\",\n",
+    "\n",
+    "    # Folder where all the results will be collected\n",
+    "    \"results_dir\" : \"Viewer_example\",\n",
+    "\n",
+    "    # Define devlib modules to load\n",
+    "    \"modules\"     : [\n",
+    "        'cpufreq'       # enable CPUFreq support\n",
+    "    ],\n",
+    "\n",
+    "    # FTrace events to collect for all the tests configuration which have\n",
+    "    # the \"ftrace\" flag enabled\n",
+    "    \"ftrace\"  : {\n",
+    "         \"events\" : [\n",
+    "            \"sched_switch\",\n",
+    "            \"sched_wakeup\",\n",
+    "            \"sched_wakeup_new\",\n",
+    "            \"sched_overutilized\",\n",
+    "            \"sched_load_avg_cpu\",\n",
+    "            \"sched_load_avg_task\",\n",
+    "            \"sched_load_waking_task\",\n",
+    "            \"cpu_capacity\",\n",
+    "            \"cpu_frequency\",\n",
+    "            \"cpu_idle\",\n",
+    "            \"sched_tune_config\",\n",
+    "            \"sched_tune_tasks_update\",\n",
+    "            \"sched_tune_boostgroup_update\",\n",
+    "            \"sched_tune_filter\",\n",
+    "            \"sched_boost_cpu\",\n",
+    "            \"sched_boost_task\",\n",
+    "            \"sched_energy_diff\"\n",
+    "         ],\n",
+    "         \"buffsize\" : 100 * 1024,\n",
+    "    },\n",
+    "\n",
+    "    # Tools required by the experiments\n",
+    "    \"tools\"   : [ 'trace-cmd', 'taskset'],\n",
+    "}"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
+    "# Initialize a test environment using:\n",
+    "te = TestEnv(my_conf, wipe=False)\n",
+    "target = te.target"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "# Workload definition\n",
+    "\n",
+    "The Viewer workload will simply read an URI and let Android pick the best application to view the item designated by that URI. That item could be a web page, a photo, a pdf, etc. For instance, if given an URL to a Google Maps location, the Google Maps application will be opened at that location. If the device doesn't have Google Play Services (e.g. HiKey960), it will open Google Maps through the default web browser.\n",
+    "\n",
+    "The Viewer class is intended to be subclassed to customize your workload. There are pre_interact(), interact() and post_interact() methods that are made to be overridden.\n",
+    "\n",
+    "In this case we'll simply execute a script on the target to swipe around a location on Gmaps. This script is generated using the TargetScript class, which is used here on System.{h,v}swipe() calls to accumulate commands instead of executing them directly. Those commands are then outputted to a script on the remote device, and that script is later on executed as the item is being viewed. See **${LISA_HOME}/libs/util/target_script.py**"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true,
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
+    "class GmapsViewer(ViewerWorkload):\n",
+    "    \n",
+    "    def pre_interact(self):\n",
+    "        self.script = TargetScript(te, \"gmaps_swiper.sh\")\n",
+    "\n",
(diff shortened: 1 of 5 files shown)


libs/utils/android/system.py
```diff
@@ -153,6 +153,26 @@ def start_action(target, action, action_args=''):
...
+        :type uri: str
+
+        :param force_new: Force the viewing application to be
+            relaunched if it is already running
+        :type force_new: bool
+        """
+        arguments = '-d {}'.format(uri)
+
+        if force_new:
+            # Flag is ACTIVITY_NEW_TASK & ACTIVITY_CLEAR_TASK
```
bjackman: Link to docs/code?

libs/utils/android/workload.py
```diff
@@ -62,7 +62,7 @@ def _check_availables(cls, test_env):
         _log.debug('Building list of available workloads...')
         for sc in Workload.__subclasses__():
             _log.debug('Checking workload [%s]...', sc.__name__)
-            if sc.package in cls._packages:
+            if sc.package in cls._packages or sc.package == '':
```
bjackman: Can it just be `None`? Empty string seems a bit random

libs/utils/android/workloads/viewer.py
```diff
@@ -0,0 +1,108 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2015, ARM Limited and contributors.
```
bjackman: I don't normally check this but happened to notice this time - year is wrong. Dunno if it makes any difference, probably not TBH...

libs/utils/android/workloads/viewer.py
```diff
@@ -0,0 +1,108 @@
...
+    Android generic Viewer workload
+    """
+    
+    # Let the system pick the best package
+    package = ''
+
+    def __init__(self, test_env):
+        super(Viewer, self).__init__(test_env)
+        self._log = logging.getLogger('Viewer')
+        self._log.debug('Workload created')
```
bjackman: no need for this log

libs/utils/android/workloads/viewer.py
```diff
@@ -0,0 +1,108 @@
...
+        # Set min brightness
+        Screen.set_brightness(self._target, auto=False, percent=0)
+        # Unlock device screen (assume no password required)
+        Screen.unlock(self._target)
+
+        # Force screen in PORTRAIT  mode
+        Screen.set_orientation(self._target, portrait=True)
+
+        System.gfxinfo_reset(self._target, self.package)
+        sleep(1)
```
bjackman: Why sleep? Comment would be good.

ipynb/examples/android/workloads/Android_Viewer.ipynb
```diff
@@ -0,0 +1,312 @@
...
+    "    script = generate_script()\n",
+    "    \n",
+    "    # Get workload\n",
+    "    wload = Workload.getInstance(te, 'Viewer')\n",
+    "        \n",
+    "    # Run workload\n",
+    "    wload.run(out_dir=te.res_dir,\n",
+    "        collect=\"ftrace\",\n",
+    "        uri=\"https://goo.gl/maps/eAbA8ZxPAy32\",\n",
+    "        workload=lambda: script.run())\n",
```
bjackman: `lambda: script.run()` == `script.run`

ipynb/examples/android/workloads/Android_Viewer.ipynb
```diff
@@ -0,0 +1,312 @@
...
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "# Workload definition\n",
+    "\n",
+    "The Viewer workload will simply read an URI and let Android pick the best application to view the item designated by that URI. That item could be a web page, a photo, a pdf, etc. For instance, if given an URL to a Google Maps location, the Google Maps application will be opened at that location. If the device doesn't have Google Play Services (e.g. HiKey960), it will open Google Maps through the default web browser.\n",
+    "\n",
+    "The Viewer workload expects a \"workload\" argument that is a function pointer to some test code to execute while the item is being displayed. In this case we'll simply execute a script on the target, see **${LISA_HOME}/libs/util/target_script.py**"
+   ]
```
bjackman: Can you add a brief explanation of what the `TargetScript` biz is doing? It might be confusing to a new reader.

libs/utils/android/workloads/viewer.py
```diff
@@ -0,0 +1,108 @@
...
+
+        :param out_dir: Path to experiment directory where to store results.
+        :type out_dir: str
+
+        :param uri: The URI of the item to display
+        :type location_search: str
+
+        :param workload: Function pointer to the workload to run once item
+        is opened and being viewed.
+        :type workload: func
```
bjackman: As you suggested offline I think this param is out of place. A "view workload" sounds like it just views something. So better to have a subclass for the gmaps usecase IMO.

libs/utils/android/workload.py
valschneider: I had a talk with @derkling about that. Empty string was suggested so that it is obvious to the developer that his workload requires no package (arguably that could also apply to None)

libs/utils/android/workloads/viewer.py
valschneider: My initial thoughts were to go with the function pointer at first for quick testing, then do a proper subclassing. But now that I've been using that function argument, I must admit I quite like it. It's really easy to implement a new workload straight out of a notebook.

libs/utils/android/workload.py
derkling: We can avoid this check by initialising `cls._packages` to always include the empty string.

libs/utils/android/system.py
```diff
@@ -153,6 +153,26 @@ def start_action(target, action, action_args=''):
         target.execute('am start -a {} {}'.format(action, action_args))
 
     @staticmethod
+    def view_uri(target, uri, force_new=True):
+        """
+        Start a view activity by specifying in URI
```
derkling: "by specifying a URI" ?

libs/utils/android/system.py
```diff
@@ -153,6 +153,26 @@ def start_action(target, action, action_args=''):
...
+
+        :param force_new: Force the viewing application to be
+            relaunched if it is already running
+        :type force_new: bool
+        """
+        arguments = '-d {}'.format(uri)
+
+        if force_new:
+            # Flag is ACTIVITY_NEW_TASK & ACTIVITY_CLEAR_TASK
+            arguments = '{} -f 0x10008000'.format(arguments)
```
derkling: Can we defined this ugly hex as a constant value with a meaningful name at the top of the class?

libs/utils/android/workloads/viewer.py
derkling: I think the rule is to always use the current year a file has been created.

libs/utils/android/workloads/viewer.py
derkling: Is not a function pointer just like a pure virtual method? I would probably better go for a subclass me too.
It could be over-killing for some simple use-cases... but if we need to integrate more complex use-cases in the future we already have the infrastructure in place. Isnt' it?

libs/utils/android/workloads/viewer.py
```diff
@@ -0,0 +1,108 @@
...
+        self.out_dir = out_dir
+        self.collect = collect
+
+        # Set min brightness
+        Screen.set_brightness(self._target, auto=False, percent=0)
+        # Unlock device screen (assume no password required)
+        Screen.unlock(self._target)
+
+        # Force screen in PORTRAIT  mode
+        Screen.set_orientation(self._target, portrait=True)
```
derkling: Are we sure all possible workload will be happy of a portrait mode?
In case, can it make sense to expose this as a parameter?

libs/utils/android/workloads/viewer.py
derkling: If it's related to resetting the GFXInfo, which may well be... we should add a fixup patch where we include this sleep in the above call and remove this ugly sleep from all workloads using gfxinfo.

libs/utils/android/workloads/viewer.py
```diff
@@ -0,0 +1,108 @@
...
+
+        System.gfxinfo_reset(self._target, self.package)
+        sleep(1)
+
+        # Open the requested uri
+        System.view_uri(self._target, uri)
+
+        # Let the activity start
+        sleep(3)
```
derkling: Same here... I would hide the sleep in the above call in such a way to make it kind-of a sync call.
In the future we can make it more robust by checking (maybe via logcat) if the activity has been really started... but that's definitively for another possible future improvement.

libs/utils/android/workload.py
bjackman: > We can avoid this check by initialising cls._packages to always include the empty string.

I much much prefer the explicit check.

libs/utils/android/workloads/viewer.py
bjackman: Yeah I was also thinking about clarity for this workload itself. IMO It's kind of weird to have a "workload" that says "I view something. Also, you have to pass me a 'workload' to run while I'm viewing things". What if you just want to play a video or get a trace for an image decode?

My first thought was it should be possible to set `workload=None`, then I realised setting this parameter is probably a special enough case that it should be its own class.

valschneider: The thing is, viewing an image or playing a video requires some additional logic anyway: you might want to make sure the image decoding is over, same for video playback (and you may even have to start the video itself). So all this workload should be seen as: "I open apps, but you have to tell me what to do once it's loaded". That could be nothing, that could be swiping, clicking, pressing buttons, etc.

The current setup also makes it really easy to quickly sketch up a workload and tweak it on the go. Having a subclass for each and every workload is contrary to that. I agree that some complex/specific uses of the Viewer could be subclasses, but I don't think that should apply to all of them.

derkling: It does not seems so much more code than what you have in your Notebook.

I think we should be able to defined the class in the notebook itself, did not tried in a Notebook but something like that should work:

```python
class GMapsSwipe(Viewer):

    def __init__(self, test_env):
        super(GMapsSwipe, self).__init__(test_env)
        # Create script
        self._script = TargetScript(self._te, "gmaps_swiper.sh")
        # Define commands to execute during experiment
        for i in range(2):
            System.hswipe(self._script, 40, 60, 100, False)
            script.append('sleep 1')
            System.vswipe(self._script, 40, 60, 100, True)
            script.append('sleep 1')
            System.hswipe(self._script, 40, 60, 100, True)
            script.append('sleep 1')
            System.vswipe(self._script, 40, 60, 100, False)
            script.append('sleep 1')
        self._script.push()

    def workload(self):
        self._script.run()

def experiment():
    # Configure governor
    target.cpufreq.set_all_governors('sched')

    # Create the workload and run it
    wload = GMapsSwipe()
    wload.run(out_dir=te.res_dir,
        collect="ftrace",
        uri="https://goo.gl/maps/eAbA8ZxPAy32")
        
    # Dump platform descriptor
    te.platform_dump(te.res_dir)
```

Isn't it?

In that case I would say we have pretty much the same flexibility than using a function pointer... but we also have a structure which allows to easily deploy the new class under workloads and merge it in LISA.

valschneider: Yes that would be fine to me, but this seems to cause problems with the Workload.getInstance() method (Viewer's subclasses are seen as 'Viewer' themselves). I'll try to look into it.

valschneider:
Most comments should be addressed. I've modified the Workload class so that (subclasses of)+ Workload subclasses are also taken into account.

I've also added pre_viewing(), viewing() and post_viewing() methods to the Viewer class for better tweaking.

Once this is merged I will update gmaps/youtube to use it, and see if some simple workloads could be added (e.g. web page browsing).

libs/utils/android/workload.py
```diff
@@ -48,6 +48,14 @@ def _adb(self, cmd):
         return 'adb -s {} {}'.format(self._target.adb_name, cmd)
 
     @classmethod
+    def _subclasses(cls):
```
derkling: That's cool! Thanks!

ipynb/examples/android/workloads/Android_Viewer.ipynb
```diff
@@ -0,0 +1,320 @@
...
+    "import os\n",
+    "\n",
+    "# Support to access the remote target\n",
+    "import devlib\n",
+    "from env import TestEnv\n",
+    "\n",
+    "# Import support for Android devices\n",
+    "from android import Screen, Workload, System\n",
+    "from target_script import TargetScript\n",
+    "from viewer import Viewer\n",
```
derkling: Maybe we can avoid this by moving the definition of the `Viewer` class beside that of `Workload`?
That way it should also be possible to export `Viewer` (which I would also rename into `ViewerWorkload`) from the same `__init__.py` which already exports `Worklaod`, i.e. `libs/utils/android/__init__.py`.

In that case here it should be possible to have just:
```python
from android import Screen, System, ViewerWorkload
```

Not sure will work... but perhaps worth to try ;-)

libs/utils/android/system.py
```diff
@@ -21,9 +21,14 @@
 from devlib import TargetError
 import os
 import pexpect as pe
+from time import sleep
 
 GET_FRAMESTATS_CMD = 'shell dumpsys gfxinfo {} > {}'
 
+# See https://developer.android.com/reference/android/content/Intent.html#setFlags(int)
+FLAG_ACTIVITY_NEW_TASK = 268435456
```
valschneider: I didn't know how python would output hexes so I went with dead stupid integers, I'll fix that

libs/utils/android/workloads/viewer.py
```diff
@@ -0,0 +1,112 @@
...
+    # Let the system pick the best package
+    package = ''
+
+    def __init__(self, test_env):
+        super(Viewer, self).__init__(test_env)
+
+        # Set of output data reported by the viewer
+        self.db_file = None
+
+    def pre_viewing(self):
```
valschneider: Yeah it's mostly to surround the tracing calls. The pre-viewing is there mostly for video playback, as sometimes the VIEW intent is not enough and you need to press play or something before properly "viewing" the video. Since the `viewing()` method is there for the actual workload (swiping, tapping, waiting...), it's then obvious that `pre_viewing()` and `post_tracing()` are executed just before/after the workload itself. `setup/teardown` works fine too.

valschneider: Also for your below comment on the gfxinfo_reset, I can't reply directly to it so I'll do it here: as there are other workloads that use gfxinfo and then call sleep, I was thinking of writing a separate pull request that removes all of those calls and adds it to the System call itself.

valschneider:
Cleaned up things, added docstring, and moved the s/Viewer/ViewerWorkload/ class to **utils/android**. _check_availables() still sees ViewerWorkload as an available workload although it should be a purely abstract workload, so this might be one more thing to clean up.

valschneider:
@derkling Jankbench still works in a Notebook and in the command-line benchmarks, so I *think* I haven't broken anything Workload-related. Do you think you can have another look at this PR ? Thanks.

PR #418, opened 2017-06-14
Add support for custom trace analysis using callbacks

cobrien7:
Some trace analysis tasks are most easily expressed by iterating over
events in chronological order. Add a runCustomAnalysis() method to the
Trace class that will run user-defined functions on trace events in
the order they occurred.

Signed-off-by: Connor O'Brien <connoro@google.com>

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -457,6 +457,58 @@ class Trace(object):
                             tname, self.tasks[tname]['pid'])
         return self.tasks
 
+    def runCustomAnalysis(self, fn_map):
+        """
+        Apply analysis functions to trace events in chronological order.
+
+        This method iterates over a user-specified subset of the available trace
+        event dataframes, calling different user-specified functions for each
+        event type. These functions are passed a dictionary mapping 'Index' and
+        the column names to their values for that row.
+
+        For example, to iterate over trace t, applying your functions analysis_fn1
+        and analysis_fn2 to each sched_switch and sched_wakeup event respectively:
+
+        t.runCustomAnalysis({
+            "sched_switch": analysis_fn1,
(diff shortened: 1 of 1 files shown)


cobrien7:
This change depends on Joel's trappy PR adding a __line column to the trace dataframes (https://github.com/ARM-software/trappy/pull/251)

bjackman:
If this is fast enough then I think it seems like a sensible approach. One request... could you add some basic test coverage? [Here's] a pending commit that adds a unit test for part of the `Trace` class, maybe you could use something like that? (Don't worry about merge conflicts)

Would be great just to have enough coverage that the CI fails on this PR until the "__line" PR is merged in TRAPpy.

derkling:
Cool, @sinkap merged this:
    https://github.com/ARM-software/trappy/pull/251

Thus we can now refresh this PR by adding a commit to progress libs/trappy

You just need to add a patch which "git add libs/trappy" once you pulled master in your local lib/trappy folder. This is an example commit message: https://github.com/ARM-software/lisa/commit/70aa3a3eaf9c3d894642d61c074ad718e7620ed3
Just add a similar patch at the beginning of this PR.

cobrien7:
Just talked to Joel about whether it might make more sense to put this in trappy instead, e.g. as a method for the GenericFTrace class. I'm going to go ahead and do that (and write an accompanying test in trappy/tests) unless there are any objections or problems with that approach

joelagnel:
Is being moved to https://github.com/ARM-software/trappy/issues/254 . Please close.

derkling:
This code is going to be merged in TRAPpy (ARM-software/trappy#254) while in LISA we will expect a new PR to add example analysis using this new API. So, let's close this one.

PR #423, opened 2017-06-18
Benchmark cleanup

joelagnel:
Pushing some benchmark clean ups, please take a look, thanks.

### DIFF
--- a/libs/utils/android/benchmark.py
+++ b/libs/utils/android/benchmark.py
@@ -116,7 +116,7 @@ class LisaBenchmark(object):
 
         # Test customization
         parser.add_argument('--results-dir', type=str,
-                default=None,
+                default=self.__class__.__name__,
                 help='Results folder, '
                      'if specified override test defaults')
         parser.add_argument('--collect', type=str,
--- a/libs/utils/android/benchmark.py
+++ b/libs/utils/android/benchmark.py
@@ -47,7 +47,48 @@ class LisaBenchmark(object):
     directory.
     """
 
-    bm_conf = None
+    bm_conf = {
+
+        # Target platform and board
+        "platform"      : 'android',
+
+        # Define devlib modules to load
+        "modules"     : [
+            'cpufreq',
+        ],
+
+        # FTrace events to collect for all the tests configuration which have
+        # the "ftrace" flag enabled
+        "ftrace"  : {
+            "events" : [
+                "sched_switch",
+                "sched_overutilized",
+                "sched_contrib_scale_f",
+                "sched_load_avg_cpu",
+                "sched_load_avg_task",
+                "sched_tune_tasks_update",
+                "sched_boost_cpu",
+                "sched_boost_task",
+                "sched_energy_diff",
+                "cpu_frequency",
+                "cpu_idle",
+                "cpu_capacity",
+            ],
+            "buffsize" : 10 * 1024,
+        },
+
+        # Default EnergyMeter Configuration
+        "emeter" : {
+            "instrument" : "acme",
+            "channel_map" : {
+                "Device0" : 0,
+            }
+        },
+
+        # Tools required by the experiments
+        "tools"   : [ 'trace-cmd' ],
+
+    }
     """Override this with a dictionary or JSON path to configure the TestEnv"""
 
     bm_name = None
@@ -145,10 +186,6 @@ class LisaBenchmark(object):
 
 
     def _getBmConf(self):
-        if self.bm_conf is None:
-            msg = 'Benchmark subclasses must override the `bm_conf` attribute'
-            raise NotImplementedError(msg)
-
         # Override default configuration with command line parameters
         if self.args.boot_image:
             self.bm_reboot = True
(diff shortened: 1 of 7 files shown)


libs/utils/android/benchmark.py
```diff
@@ -47,7 +47,48 @@ class LisaBenchmark(object):
     directory.
     """
 
-    bm_conf = None
+    bm_conf = {
```
derkling: Yes, I think this kind of make sense given the current code.
Moreover, a sub-class can also easily just re-define a part of the default configuration, for example to change the emeter to use... +1

tests/benchmarks/android_geekbench.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
```
derkling: @ionela-voinescu do you think this can break our internal testing infrastructure?
Would say no, but I would prefer to get a confirmation from you... ;-)

libs/utils/android/benchmark.py
```diff
@@ -116,7 +116,7 @@ def _parseCommandLine(self):
 
         # Test customization
         parser.add_argument('--results-dir', type=str,
-                default=None,
+                default=self.__class__.__name__,
```
bjackman: Worth noting that for synthetic tests we already have a default results dir (https://github.com/ARM-software/lisa/blob/master/libs/utils/env.py#L259), would be good to keep them aligned.

tests/benchmarks/android_geekbench.py
bjackman: IMO if it does then our infra is broken, so I'm find with this :)

libs/utils/android/benchmark.py
derkling: Well... the current implementation uses the path specified by `--results-dir`  that why:
- relative path: append it to the standard `results` dir you references
- absolute path: use whatever you specified

Thus, in the case of the Joel's proposal, the class name would be interpreted as a "relative path" and we will end up with the results under `$LISA_HOME/results/<CLASS_NAME>`... which seems a sensible default to me.

bjackman: It's a sensible default but it's a misalignment - benchmarks use  `$LISA_HOME/results/<CLASS_NAME>` while synthetics etc use `$LISA_HOME/results/<DATE_TIME>`

derkling: The `<DATE_TIME>` is a default if you do not specify a custom path, and the results folder becomes quite confusing and messy once you run it over time. You cannot even know for sure what a `<DATA_TIME>` folder contains exactly.

Should we not fix instead our synthetics to properly generate results using a more clean folder structure, e.g. something like `results/synthetics/<TEST_NAME>`... and maybe we can also append a `<DATE_TIME>` to this base path in case we really want to keep an history of different executions?

bjackman: Yeah, that could be a good idea. I do think having the date+time is a good idea rather than overwriting data when repeating tests. I suppose am entitled millenial but disk space is cheap :stuck_out_tongue:

joelagnel: Correct me if I'm wrong but the existing benchmarks don't use DATE_TIME anyway. Considering that there is some debate here, could we make it a separate issue for a follow-up cleanup?
Also personally I prefer to not append a timestamp. About the consistency argument, if it has to be made consistent, shouldn't we do in in libs/utils/env.py for all result directories? It appears from that code that its done only if results directory isn't specified which is kind of making it consistent anyway (since I'm a LISA n00b, I could be completely off the mark here)

joelagnel:
So based on comments so far, I am guessing this is good to go :+1:

tests/benchmarks/android_geekbench.py
ionela-voinescu: No, it does not break our infrastructure.

libs/utils/android/benchmark.py
bjackman: > Correct me if I'm wrong but the existing benchmarks don't use DATE_TIME anyway.  Considering that there is some debate here, could we make it a separate issue for a follow-up cleanup?

Sure.

> About the consistency argument, if it has to be made consistent, shouldn't we do in in libs/utils/env.py for all result directories? It appears from that code that its done only if results directory isn't specified which is kind of making it consistent anyway (since I'm a LISA n00b, I could be completely off the mark here)

Yes, I think you're right.

libs/utils/android/benchmark.py
```diff
@@ -47,7 +47,48 @@ class LisaBenchmark(object):
...
+                "sched_load_avg_task",
+                "sched_tune_tasks_update",
+                "sched_boost_cpu",
+                "sched_boost_task",
+                "sched_energy_diff",
+                "cpu_frequency",
+                "cpu_idle",
+                "cpu_capacity",
+            ],
+            "buffsize" : 10 * 1024,
```
ionela-voinescu: Might be worth having a higher value here as default. I've chosen 80KB for the CI tests given the  number of trace events.

joelagnel: Can be a follow-up patch? This patch just keeping this in parity with what's already there, appreciate if another patch can be sent for that

joelagnel:
@derkling so good for merge?

joelagnel:
Ping.

PR #424, opened 2017-06-19
command line analysis tools support

joelagnel:
Hi,
Just posting some early patches to get feedback on the command line analysis tools feature. Forgive me for not adding proper docstrings which I'll add when posting follow ups (hence tagged RFC).

The idea is you can  post-process traces or other data in a results directory, and get useful information from them without having to launch a browser or notebook environment. It fits very well with LISA's analysis architecture (libs/utils/analysis/)
Some data is simple enough that it shouldn't need plotting. Also incase of results that need to be made more visual, I am planning to add some code that autogenerates a .html file by running a dynamically generated notebook and automatically launch a browser window to show them) - for example for analysis tools that do need to show something more visual.

Thanks,
Joel

### DIFF
--- a/libs/utils/analysis_tool.py
+++ b/libs/utils/analysis_tool.py
@@ -1,0 +1,102 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2017, ARM Limited, Google and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import argparse
+import logging
+import os
+import select
+from trace import Trace
+from conf import LisaLogging
+from devlib.utils.misc import memoized
+
+class AnalysisTool(object):
+    """
+    A base class for LISA analysis command-line tools.
+
+    This class is intended to be subclassed in order to create a custom
+    analysis tool.
+
+    :param options_parser: A parser object passed by the specific tool
+                           to parse additional options if any (optional).
+    :type options_parser: ArgumentParser
+    """
+
+    def __init__(self, options_parser=None):
+        self.options_parser = options_parser
+
+    def _setup_logging(self):
+        """
+        Setup logging for the analysis tool.
+
+        Analysis tools shouldn't output to standard output as their
+        output can be used by other tools. Configure the logger to
+        output only error messages.
+        """
+        self._log = logging.getLogger('AnalysisTool')
+        level = logging.INFO if self.args.log_stdout else logging.ERROR
+        LisaLogging.setup(level=level)
+
+    def _parse_args(self):
+        """
+        Parse command line arguments given to the tool, also use the parser
+        object if one is passed by the more specific class so that the
+        tool specific options are also parsed.
+        """
+        parser = self.options_parser
+        if not parser:
+            parser = argparse.ArgumentParser(
+                    description='LISA Analysis tool configuration')
+
+        # General options that all tools accept
+        parser.add_argument('--trace', type=str,
+                help='Path of the trace file or directory to be used',
+                required=True)
+
+        parser.add_argument('--log-stdout', action='store_true',
+                help='Output logging to stdout',
+                required=False)
+
+        self.args = parser.parse_args()
+
+    def run_analysis(self):
+        """
+        Run an analysis and store results
+
+        The specific analysis tool should override this and perform
+        analysis and provide results in self.output.
+        """
+        raise NotImplementedError('Analysis tool needs to define run_analysis')
+
+    def run_main(self):
+        """
+        The main function that sets up, runs the analysis and provides
+        output to the user.
+
+        The specific analysis tool should call this function to start the analysis
+        which will perform analysis and report the result. The tool should also
+        override run_analysis to provide custom analysis and optionally override
+        run_output to provide custom output.
+        """
+
+        self._parse_args()
+        self._setup_logging()
+
+        # Create trace object
+        self._log.info('Creating trace object...')
+        self.trace = Trace(None, self.args.trace)
+
+        self.run_analysis()
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -69,7 +69,7 @@ class Trace(object):
     :type plots_prefix: str
     """
 
-    def __init__(self, platform, data_dir, events,
+    def __init__(self, platform, data_dir, events=None,
                  window=(0, None),
                  normalize_time=True,
                  trace_format='FTrace',
@@ -125,7 +125,7 @@ class Trace(object):
             self.plots_dir = self.data_dir
         self.plots_prefix = plots_prefix
 
-        self.__registerTraceEvents(events)
+        self.__registerTraceEvents(events) if events else None
         self.__parseTrace(data_dir, window, normalize_time,
                           trace_format)
         self.__computeTimeSpan()
@@ -219,7 +219,7 @@ class Trace(object):
         :type trace_format: str
         """
         self._log.debug('Loading [sched] events from trace in [%s]...', path)
-        self._log.debug('Parsing events: %s', self.events)
+        self._log.debug('Parsing events: %s', self.events if self.events else 'ALL')
         if trace_format.upper() == 'SYSTRACE' or path.endswith('html'):
             self._log.debug('Parsing SysTrace format...')
             trace_class = trappy.SysTrace
@@ -231,7 +231,8 @@ class Trace(object):
         else:
             raise ValueError("Unknown trace format {}".format(trace_format))
 
-        self.ftrace = trace_class(path, scope="custom", events=self.events,
+        scope = 'custom' if self.events else 'all'
+        self.ftrace = trace_class(path, scope=scope, events=self.events,
                                   window=window, normalize_time=normalize_time)
 
         # Load Functions profiling data
--- a/tools/analysis/android_gfx.py
+++ b/tools/analysis/android_gfx.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+#    Copyright 2017 ARM Limited, Google and contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import pandas as pd
+from time import sleep
+from devlib.exception import TargetError
+from analysis_tool import AnalysisTool
+
+class AndroidGfx(AnalysisTool):
+
+    def get_frame_drops(self, df):
+        if not 'func' in df.columns:
+            return 0
+        return len(df[(df.func == 'FrameMissed') & (df.data == '1')])
+
+    def run_analysis(self):
+        df = self.trace.data_frame.trace_event('tracing_mark_write')
+        print {'frame_drops': self.get_frame_drops(df)}
+
+# Run the analysis and provide results
+if __name__ == "__main__":
+    AndroidGfx().run_main()


joelagnel:
CC @bjackman @derkling

libs/utils/trace.py
```diff
@@ -256,16 +256,17 @@ def __parseTrace(self, path, tasks, window, normalize_time, trace_format):
...
-        self._sanitize_SchedLoadAvgCpu()
-        self._sanitize_SchedLoadAvgTask()
-        self._sanitize_SchedCpuCapacity()
-        self._sanitize_SchedBoostCpu()
-        self._sanitize_SchedBoostTask()
-        self._sanitize_SchedEnergyDiff()
-        self._sanitize_SchedOverutilized()
-        self._sanitize_CpuFrequency()
+        # Santization not possible if platform missing
+        if not self.platform:
```
bjackman: Is this meant to be be `if self.platform`?

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,83 @@
...
+            LisaLogging.setup()
+
+        self._log = logging.getLogger('AnalysisTool')
+
+        # Create trace object
+        self._log.info('Creating trace object...')
+        self.trace = Trace(None, args.trace, [])
+
+	self.output = self.run_analysis()
+	self.run_output()
```
bjackman: This might just be my own weird obsession but I don't like when `__init__` does things other than up data structures - I'd much prefer to have everything ultimately driven by a `analysis_main` function or something, which is called from inside an `if __name__ == "__main__":` - it seems a lot more flexible.

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,83 @@
...
+    def run_analysis(self):
+	raise NotImplementedError('error')
+
+    def run_output(self):
+        if self.gui:
+            raise NotImplementedError('error')
+        print self.output
+
+    def __init__(self, parser=None, gui=False):
+	self.gui = gui
```
derkling: Mind TAB vs spaces... we use always and only spaces...

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self.args = args
+
+        # Analysis tools shouldn't output to standard output as their
+        # output can be used by other tools. Configure the logger to
+        # output only error messages
+        if not args.log_stdout:
+            LisaLogging.setup(level=logging.ERROR)
+        else:
+            LisaLogging.setup()
```
derkling:
```python
level = logging.INFO if args.log_stdout else logging.ERROR
LisaLogging.setup(level=level)
```

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self._log = logging.getLogger('AnalysisTool')
+
+        # Create trace object
+        self._log.info('Creating trace object...')
+        self.trace = Trace(None, args.trace, [])
```
derkling: Perhaps this requires a TRAPpy extension to parse by default all the events, right?
If that's the case I would suggest to add a patch which makes `events` an optional parameter for the `Trace` ctor.

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def run_analysis(self):
+	raise NotImplementedError('error')
+
+    def run_output(self):
+        if self.gui:
+            raise NotImplementedError('error')
+        print self.output
+
+    def __init__(self, parser=None, gui=False):
```
derkling: Maybe better `s/parser/options_parser`?
Moreover, what's an example for the usage of the `gui` option?

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,83 @@
...
+                help='Output logging to stdout',
+                required=False)
+
+        # Parse command line arguments
+        return parser.parse_args()
+
+    def run_analysis(self):
+	raise NotImplementedError('error')
+
+    def run_output(self):
```
derkling: What's the semantic of this method?

libs/utils/analysis_tool.py
derkling: This is ultimately defining an "execution model" for a generic analysis.

We should make it as much generic as possible, maybe something like this:

```python
class BatchAnalysis():
    def __init__():
        # command line parsing
    def parse():
        # Trace parsing and DF generation
    def analyze():
        # Trace post-processing and results generation
    def generate():
        # Output data generation
    def run():
        self.parse()
        self.analyze()
        self.generate()
```

So that a client class usually re_define `parse`, `analyze` and `generate` while the `run` method is the one called by default from the `__main__` of an executable script which is hosted in `tools/analysis`.

libs/utils/trace.py
derkling: I completely agree with the commit message title, but this is like using an axe instead of a bistoury... even when `platform` data are not available we can still (and have to) do many of the trace sanitation updates.
I would better move this check inside each and every `_sanitize_XXX` method to carefully cut out what is not possible.

At the same time, this update requires also to cross check all the analysis and plotting APIs to do the same.
Indeed, right now we assume that we have all the columns that are added by the availability of platform data. This should be done carefully and, since it's a big job by itself, I would rather suggest to these updates on a PR by their own.

Can you remove this patch from this series for the time being while I work on a separate PR to properly remove the platform data dependency? I've just created #427 to track this cleanups.

tools/analysis/android_gfx
```diff
@@ -0,0 +1,39 @@
...
+#
+
+import os
+import pandas as pd
+from time import sleep
+from devlib.exception import TargetError
+from analysis_tool import AnalysisTool
+
+
+class GfxAnalysisTool(AnalysisTool):
```
derkling: I would suggest to use the same name for both the class and the file. The name should also be "kind of" namespace based and details. Thus, I would say that `class AndroidGFX` should be better.

Moreover, I guess this is just an initial simple example and that you wanna add much more GFX related results to be generated by running this script, isn't it?

In that case I would suggest to feed-in some more example to better verify if the API works for even just slightly more complex use-cases.

libs/utils/trace.py
joelagnel: Ok I'll drop it from this PR

libs/utils/analysis_tool.py
valschneider: Since this class is intended to be subclassed, I think it would be nice to add a docstring to all of the "abstract" methods.

tools/analysis/android_gfx
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import pandas as pd
+from time import sleep
+from devlib.exception import TargetError
+from analysis_tool import AnalysisTool
+
```
valschneider: Nitpicking here, there's one more line feed than is needed.

libs/utils/analysis_tool.py
joelagnel: @derkling just wanted some clarification:
*  in your example above, you mentioned 'command line parsing' in `__init__`. But then in the `run` method you have `self.parse` again. Going by @bjackman comment about not doing "actions" in `__init__`, I am wondering if that's what you meant.
* the "executable script" will also host the tool-specific `parse`, `analyze` and `generate` which are overridden in the specific tool. Hope we are on the same page about that?

libs/utils/analysis_tool.py
joelagnel: I'll follow up with better docstrings, sorry. `run_output` is supposed to present output to the user. Every tool will have its own output. Maybe we can provide some generic templates later.

libs/utils/analysis_tool.py
joelagnel: Will add proper docstrings, sorry for making the review hard. GUI is supposed to be for tools that want to show output in the browser instead of text. The idea is for certain tools to create a report automatically (using ipython notebook generation utilities) and automatically open it in the browser.

libs/utils/analysis_tool.py
joelagnel: Sure, I'll make it optional

tools/analysis/android_gfx
joelagnel: Yes this is just an initial example to flesh out the architecture. Will change class naming as you suggest 👍

joelagnel:
Hi @derkling @bjackman 
Please check this and let me know if it looks Ok. Please note:
* I dropped 'make platform optional' patch as @derkling wants to do that so that's needed for these patches to work. But I did add a 'make events optional' patch as @derkling requested.
* I kept the graphics example simple because I don't want to complicate it right now - as I said, this is a simple example.
* I dropped the gui stuff since I didn't want to complicate it and add something not yet in use.

libs/utils/analysis_tool.py
joelagnel: fixed

tools/analysis/android_gfx
joelagnel: fixed

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+
+        self._parse_args()
+        self._setup_logging()
+
+        # Create trace object
+        self._log.info('Creating trace object...')
+        self.trace = Trace(None, self.args.trace)
+
+        self.run_analysis()
+        self.run_output()
```
derkling: Why do we wanna enforce `run_output` and `run_analysis` to be two different functions it then we call them one after the other? I would say that from the standpoint of the "state machine" defined by the case class they are exactly the same state. Thus I would keep just one of the two methods... sub-classes can still split the implementation in two if they want and/or find it useful.

Moreover, I'm not convinced about the naming of these last three methods. I would prefer something more "state/action like"... what about `AnalysisTool::analyse` and `AnalysisTool::run`... where the user is required to put all of its logic and (optional) output production in the `analyse` method.

tools/analysis/android_gfx
```diff
@@ -0,0 +1,39 @@
...
+        super(AndroidGfx, self).__init__()
+
+    def get_frame_drops(self, df):
+        if not 'func' in df.columns:
+            return 0
+        return len(df[(df.func == 'FrameMissed') & (df.data == '1')])
+
+    def run_analysis(self):
+        df = self.trace.data_frame.trace_event('tracing_mark_write')
+
```
derkling: You can skip this empty line... and if you go for the removal of the `run_output` method, here you just go straight for a print statement ;-)

libs/utils/trace.py
```diff
@@ -73,7 +73,7 @@ class Trace(object):
     :type plots_prefix: str
     """
 
-    def __init__(self, platform, data_dir, events,
+    def __init__(self, platform, data_dir, events=None,
```
derkling: :heart_eyes: another step into the direction to make `Trace()` super simple to build, thanks! ;-)

I still have to provide a patch to make `platform` optional as well... will do soon.

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+import os
+import select
+from trace import Trace
+from conf import LisaLogging
+from devlib.utils.misc import memoized
+
+class AnalysisTool(object):
+    """
+    A base class for LISA analysis command-line tools.
+    This class intended to be subclassed in order to create a custom
```
bjackman: Also, empty line after the first line please - https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings (I know we probably don't follow this properly but it's nice to try :P)

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+                    description='LISA Analysis tool configuration')
+
+        # General options that all tools accept
+        parser.add_argument('--trace', type=str,
+                default=None,
+                help='Path of the trace file or directory to be used',
+                required=True)
+
+        parser.add_argument('--log-stdout', action='store_true',
+                default=False,
```
bjackman: Don't think you need the `default=False` when `action='store_true'`

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        parser = self.options_parser
+
+        if not parser:
+            parser = argparse.ArgumentParser(
+                    description='LISA Analysis tool configuration')
+
+        # General options that all tools accept
+        parser.add_argument('--trace', type=str,
+                default=None,
```
bjackman: No need for `default` when `required=True`

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+        self.args = parser.parse_args()
+
+    def run_analysis(self):
+        """
+        Run an analysis and store results
+
+        The specific analysis tool should override this and perform
+        analysis and provide results in self.output
+        """
+        raise NotImplementedError('error')
```
bjackman: Add a slightly more helpful message here?

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+        override run_analysis to provide custom analysis and optionally override
+        run_output to provide custom output.
+        """
+
+        self._parse_args()
+        self._setup_logging()
+
+        # Create trace object
+        self._log.info('Creating trace object...')
+        self.trace = Trace(None, self.args.trace)
```
bjackman: Github makes it look like this commit comes before the one that makes `events` optional. If that's true, please re-order them. [Github might be misleading me though :angry: :sob: :disappointed: :nauseated_face: ](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/)

tools/analysis/android_gfx
```diff
@@ -0,0 +1,39 @@
...
+        self.output = {'frame_drops': self.get_frame_drops(df)}
+
+# Run the analysis and provide results
+AndroidGfx().run_main()
```
bjackman: Should put this inside an `if __name__ == "__main__":`, that way it gets run when you execute this script from the cmdline but not when you import it from elsewhere.

joelagnel: Since I am a python n00b relatively speaking... Does this convention also apply to when you don't expect the scripts to be imported?

libs/utils/analysis_tool.py
joelagnel: Ok

libs/utils/analysis_tool.py
joelagnel: Sure

libs/utils/analysis_tool.py
joelagnel: Ok

libs/utils/analysis_tool.py
joelagnel: Ok got it, thanks

libs/utils/analysis_tool.py
joelagnel: Cool. :-) happy to learn these conventions

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+from devlib.utils.misc import memoized
+
+class AnalysisTool(object):
+    """
+    A base class for LISA analysis command-line tools.
+    This class intended to be subclassed in order to create a custom
+    analysis tool.
+
+    :param options_parser: A parser object passed by the specific tool
+                           to parse specific options if any (optional).
```
joelagnel: Ok

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+        self.output = None
+        self.options_parser = options_parser
+
+    def _setup_logging(self):
+        """
+        Setup logging for the analysis tool.
+
+        Analysis tools shouldn't output to standard output as their
+        output can be used by other tools. Configure the logger to
+        output only error messages
```
joelagnel: Ok will do

libs/utils/analysis_tool.py
joelagnel: I would still want `run_analysis` but I am Ok with dropping output if that's ok :-) deal ?

libs/utils/trace.py
joelagnel: Cool, thanks

joelagnel: Cool, thanks

libs/utils/analysis_tool.py
```diff
@@ -0,0 +1,116 @@
...
+        LisaLogging.setup(level=level)
+
+    def _parse_args(self):
+        """
+        Parse command line arguments given to the tool, also use the parser
+        object if one is passed by the more specific class so that the
+        tool specific options are also parsed
+        """
+        parser = self.options_parser
+
```
joelagnel: Ok

libs/utils/analysis_tool.py
joelagnel: So I think I still want `run_output` now that I think about it more. The reason is in the future if we do any special formatting of results than just print, then it becomes more manageable to do it in one place than modify every script. Second reason is it allows a tool to separate out analysis and reporting logic by just overriding a single function so it's cleaner. I actually don't think it adds much more complexity anyway.

Hope we can agree on something here or maybe @bjackman can make a suggestion and we take the majority vote ;-)

derkling: Now every script produce its output in a single `self.output` variable, which is printed by the default (dummy) implementation of `run_output()`... if in the future you should need any "special formatting" it's quite likely you still need to update all scripts because the simple `self.output` is... too simple.

Do you have any special use-case already in mind? In that case you should provide an example, otherwise your arguments now are not convincing to me.

Regarding `AnalysisTool::run_analysis` vs `AnalysisTool::analyse` I cannot see why you are so much into to that name?
Is there any other reason behind just the name itself?
Otherwise I think we should use @bjackman's vote on this as well ;-)

tools/analysis/android_gfx
derkling: I would say it's more clean (or "pythonic")... even if not strictly required if it will never be included.
But at least it makes clear it is providing a directly executable functionality...

joelagnel:
Got it. Will fix

tools/analysis/android_gfx
joelagnel: Got it.will fix

libs/utils/analysis_tool.py
joelagnel: Ok. Let's drop output call back then, for now and let individual script do printing and formatting.
Regarding the name, it's a personal preference.. I guess to me it sounds more explicit that you're execute something when you say run_. I don't have a strong argument here and I'm ok with @bjackman vote.

bjackman: As for naming I think both `run_analysis` and `analyse` are fine.. probably lean towards `run_analysis` if it gives output too, for the same reason as Joel i.e. `analyse` sort of sounds like it returns some representation of a conclusion whereas `run_analysis` sounds like it has a side effect (i.e. printing output).

derkling: Then let's got for `run_analysis`... ;-)

joelagnel: Cool ;-)

joelagnel:
@derkling fixed all review comments, could you merge?

bjackman:
Just been taking another look at this and doing some thinking - I wonder if instead of just printing the results, the AnalysisTools should

bjackman:
Sorry, I cannot github

bjackman:
What I meant to say was: I wonder if these AnalysisTools should, rather than printing out analysis results directly, return them in a standardized format (maybe just a dict) that can then be formatted in a flexible way (e.g. pretty-print to the console for interactive use but perhaps also save JSON for use in CI)?

The example I'm thinking of is a command-line tool to expose what I've added here: https://github.com/ARM-software/lisa/pull/455. I'd like to return a few different statistics about underutilization, and it would be nice if I know they'll always be formatted in the same way as the other analysis tools.

I might be over-thinking this.

joelagnel:
Hi @bjackman
About your last comment about returning dict than printing, the main usecase I had in mind when I wrote this is PR is that we should print because these tools can be used by other programs to easily parse the machine-readable results. My thought was any analysis results that internal to LISA should be returned by libs/utils/analysis as a DF. The tools/analysis/ stuff is supposed to be just a front end. I'm going to restart this work and post another PR soon,

joelagnel:
Rebased and got rid of init as @bjackman suggested. I think this one has been beaten to death :) lets merge.

bjackman:
> we should print because these tools can be used by other programs to easily parse the machine-readable results.

Yeah, I agree, but I was thinking this would be easier to do if the actual printing is done in a common way. So instead of each analysis tool doing the print, it returns the data in some python representation (dict/DF/something else), then the base class prints it. So then if we wanted to change the format of printing (or have multiple formats, e.g. JSON/CSV) we wouldn't have to change each results processor.

One of the cool ways we could use this is to integrate with [workload-automation](https://github.com/ARM-software/workload-automation) so maybe LISA could be used as a "results processor" for that.

joelagnel:
Hi @bjackman I am ok with changing to making the 'printing' done by the Base class, that way in the future if we have to change it to something else, we only change 1 place. Let me know if this is what you want. I think this is what I was doing in my earlier approaches but I was suggested to do the printing in the specific tool itself.

Speaking in terms of working well with the broader LISA scheme of things, I believe the analysis classes in libs/utils/analysis/ should be the only ones to return DF/dict/whatever strictly, all the others should be the reporters/consumers/presenters. The exception is if the results from those classes need further processing. This is just my thought about delineating things more strictly

bjackman:
OK, that's interesting. Yeah that is what I was suggesting but maybe @derkling disagrees. Perhaps we need to have a cal to discuss this in a little more detail...

PR #426, opened 2017-06-19
energy_model: Don't error if energy model data is dodgy

bjackman:
People have energy models with kinks in the data and still need to
use this module. Just print a warning instead of erroring out.

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -147,6 +147,8 @@ class EnergyModelNode(_CpuTree):
                  cpu=None, children=None, name=None):
         super(EnergyModelNode, self).__init__(cpu, children)
 
+        self._log = logging.getLogger('EnergyModel')
+
         def is_monotonic(l, decreasing=False):
             op = operator.ge if decreasing else operator.le
             return all(op(a, b) for a, b in zip(l, l[1:]))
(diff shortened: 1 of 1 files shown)


valschneider:
Nothing to say about the code itself, but don't we want to let wacky energy values be an error ? Apart from wanting to experiment something particular, I don't see the point of having such an energy model, and thus that's something that should be fixed before testing anything. 

Maybe we should add an energy-model checking test in **tests/eas/preliminary.py** ? That way if someone wants to run some proper EAS tests, the preliminary test fails and asks them to fix their energy model, and otherwise they just get a warning and they can run something non-EAS related. I admit it's mostly nitpicking :)

bjackman:
> Nothing to say about the code itself, but don't we want to let wacky energy values be an error ? Apart from wanting to experiment something particular, I don't see the point of having such an energy model, and thus that's something that should be fixed before testing anything.

The Pixel phone shipped with an energy model that fails these checks..

> Maybe we should add an energy-model checking test in tests/eas/preliminary.py ? That way if someone wants to run some proper EAS tests, the preliminary test fails and asks them to fix their energy model, and otherwise they just get a warning and they can run something non-EAS related. I admit it's mostly nitpicking :)

I like that idea. But in the meantime I think we can go ahead with disabling the errors in `EnergyModel`.

valschneider:
Alright, it's all good to me then.

PR #428, opened 2017-06-19
[RFC] Allow specifying wload objects directly in the executor's experiments_conf

bjackman:
This is basically to allow you to use the Executor without having to use the dictionary grammar to specify the synthetic workloads you want to use.

That basically means instead of:

```
        experiments_conf = {
            'confs': [{
                'tag': "my_conf"
            }],
            "wloads" : {
                "my_wload" : {
                    "type" : "rt-app",
                    "conf" : {
                        "class" : "profile",
                        "params" : {
                            "my_task" : {
                                "kind" : "Periodic",
                                "params" : {
                                    <task params>
                                },
                            },
                        },
                    },
                },
            },
        }
```

You can have something like:

```
        wload = RTA(target, "my_wload")
        wload.conf('profile', {'my_task': Periodic(<task params>).get()})

        experiments_conf = {
            'confs': [{
                'tag': "my_conf",
            }],
            "wloads" : {
                "my_wload" : wload
            }
        }
```

I've marked this "RFC" because a) I haven't done documentation and b) The API is still a bit awkward. For example note that you "name" the workload twice (i.e. the string "my_wload" appears  twice in my example). I think the ideal usage would be like this:

```
        experiments_conf = {
            'confs': [{
                'tag': conf_name
            }],
            "wloads" : [
                RTA(target, wl_name, profile={'mytask': Periodic()})
            ]
        }
```

But there's quite a few more changes needed to achieve that.

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -764,7 +764,7 @@ class TestEnv(ShareState):
                   force=False and we have not installed rt-app.
         """
 
-        if not force:
+        if not force and 'rt-app' not in self.__installed_tools:
             if not self._calib and 'rtapp-calib' in self.conf:
                 self._log.warning('Using configuration provided RTApp calibration')
                 self._calib = {
@@ -772,12 +772,6 @@ class TestEnv(ShareState):
                     for key, value in self.conf['rtapp-calib'].items()
                 }
             return self._calib
-
-        required = force or 'rt-app' in self.__installed_tools
-
-        if not required:
-            self._log.debug('No RT-App workloads, skipping calibration')
-            return
 
         self._log.info('Calibrating RTApp...')
         self._calib = RTA.calibrate(self.target)
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -520,7 +520,8 @@ class TestEnv(ShareState):
             self.target = devlib.LocalLinuxTarget(
                     platform = platform,
                     load_default_modules = False,
-                    modules = self.__modules)
+                    modules = self.__modules,
+                    connection_settings = {'unrooted': True})
         else:
             raise ValueError('Config error: not supported [platform] type {}'\
                     .format(platform_type))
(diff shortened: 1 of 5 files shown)


bjackman:
Another issue came up - need to think about how to make this work nicely with rt-app calibration. Currently `TestEnv` assumes that if you don't have 'rt-app' in the 'tools' field in the test_conf, it doesn't need to run calibration. That's not going to fly from a usability perspective.. need to figure that out before merging.

libs/utils/env.py
```diff
@@ -764,7 +764,13 @@ def calibration(self, force=False):
                   force=False and we have not installed rt-app.
         """
 
-        if not force and self._calib:
+        if not force:
```
valschneider: How about installing rt-app when calibration is requested ? AFAIK `install_tools()` does the difference with already-installed tools, so we can let the `__init__` call to `calibration()` not do anything but the user could call `calibration` with `force=True` to install rt-app ? There would have to be some renaming/extra args so that all of this makes sense to the user, but that's the general idea.

mdigiorgio:
There are some very useful commits in here, any chances to update this PR and try to get it merged?

bjackman:
Hmm so, there is way too much work to do to get this PR ready for merge before I leave. I had forgotten about it and on refreshing my memory I definitely think that it really improves the API. However the fact that I've forgotten about it probably means that improving this API is not necessarily the best investment of time. 

The self-tests for Executor could definitely be useful so I think I will separate that into a different PR.

The changes to triggering rt-app calibration are awkward - I merged something similar recently and then immediately realised it was complete nonsense and had to revert it. So I think that's best left until next time the issue becomes a problem directly. 

As for the rest of it, I'm not sure if it's best to close this PR and raise an issue with a link to my branch (I will leave it around)?

bjackman:
I have split out the new test here: https://github.com/ARM-software/lisa/pull/521

PR #430, opened 2017-06-27
tools/arm64: build 64bit executables for taskset and trace-cmd

ionela-voinescu:
Tested with LISA synthetic tests on GEM5 and Juno.


bjackman:
Gave it a go on Pixel, all looks good.

PR #431, opened 2017-06-27
tests/eas/preliminary: Skip TestSchedDomainFlags when cpu_capacity mi…

bjackman:
…ssing

Some older kernels don't have the cpu_capacity sysfs entry. In that case just
skip the test that needs it.

### DIFF
--- a/tests/eas/preliminary.py
+++ b/tests/eas/preliminary.py
@@ -301,6 +301,9 @@ class TestSchedDomainFlags(BasicCheckTest):
         - Finally, revert to the old cpu_capacity values and check the flag
           returns to its old value.
         """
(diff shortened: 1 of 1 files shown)


tests/eas/preliminary.py
```diff
@@ -301,6 +301,9 @@ def test_asym_cpucapacity(self):
         - Finally, revert to the old cpu_capacity values and check the flag
           returns to its old value.
         """
+        if not self.target.file_exists(self._get_cpu_cap_path(0)):
```
derkling: Can it make sense also to cross-check with the specific kernel version.
I guess that starting from a certain version we expect this sysfs attribute to be exposed... otherwise the test should not be skipped by instead FAILS, isn't it?

Issue #433, opened 2017-06-29
Plotting code misses last sched_overutilized flag


bjackman:
If the last sched_overutilized event in a trace sets overutilized=1 then the plots should show the system as overutilized throughout the remainder of the trace, but instead they ignore that.

valschneider:
Fixed by https://github.com/valschneider/lisa/commit/d7531160ccc2e8151d3a6da618a3118da323dd70

PR #435, opened 2017-07-03
utils/analysis: Fix overutilized plotting

valschneider:
The last event's duration will always be set to NaN, because the time
delta is computed as (next_event time - cur_event time). Since the
plotOverutilized() method uses the time deltas to render the
overutilized flag, this can break things: if the last status event
reports overutilized=1, it won't be rendered (see https://github.com/ARM-software/lisa/issues/433).

This fixes that by 'manually' setting the duration of the last event.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -593,6 +593,10 @@ class Trace(object):
         df['len'] = (df.start - df.start.shift()).fillna(0).shift(-1)
         df.drop('start', axis=1, inplace=True)
 
+        # Fix the last event, which will have a NaN duration
+        # Set duration to trace_end - last_event
+        df.loc[df.index[-1], 'len'] = self.start_time + self.time_range - df.index[-1]
+
         # Build a stat on trace overutilization
         df = self._dfg_trace_event('sched_overutilized')
         self.overutilized_time = df[df.overutilized == 1].len.sum()
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -88,6 +88,9 @@ class Trace(object):
         # The time window used to limit trace parsing to
         self.window = window
 
+        # Whether trace timestamps are normalized or not
+        self.normalize_time = normalize_time
+
         # Dynamically registered TRAPpy events
         self.trappy_cls = {}
 
@@ -126,8 +129,7 @@ class Trace(object):
         self.plots_prefix = plots_prefix
 
         self.__registerTraceEvents(events)
-        self.__parseTrace(data_dir, window, normalize_time,
-                          trace_format)
+        self.__parseTrace(data_dir, window, trace_format)
         self.__computeTimeSpan()
 
         # Minimum and Maximum x_time to use for all plots
@@ -199,7 +201,7 @@ class Trace(object):
         if 'cpu_frequency' in events:
             self.events.append('cpu_frequency_devlib')
 
-    def __parseTrace(self, path, window, normalize_time, trace_format):
+    def __parseTrace(self, path, window, trace_format):
         """
         Internal method in charge of performing the actual parsing of the
         trace.
@@ -209,9 +211,6 @@ class Trace(object):
 
         :param window: time window to consider when parsing the trace
         :type window: tuple(int, int)
-
-        :param normalize_time: normalize trace time stamps
-        :type normalize_time: bool
 
         :param trace_format: format of the trace. Possible values are:
             - FTrace
@@ -232,7 +231,7 @@ class Trace(object):
             raise ValueError("Unknown trace format {}".format(trace_format))
 
         self.ftrace = trace_class(path, scope="custom", events=self.events,
-                                  window=window, normalize_time=normalize_time)
+                                  window=window, normalize_time=self.normalize_time)
 
         # Load Functions profiling data
         has_function_stats = self._loadFunctionsStats(path)
@@ -261,7 +260,7 @@ class Trace(object):
         self._sanitize_CpuFrequency()
 
         # Compute plot window
-        if not normalize_time:
+        if not self.normalize_time:
             start = self.window[0]
             if self.window[1]:
                 duration = min(self.ftrace.get_duration(), self.window[1])
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -130,16 +130,6 @@ class Trace(object):
 
         self.__registerTraceEvents(events)
         self.__parseTrace(data_dir, window, trace_format)
-        self.__computeTimeSpan()
-
-        # Minimum and Maximum x_time to use for all plots
-        self.x_min = 0
-        self.x_max = self.time_range
-
-        # Reset x axis time range to full scale
-        t_min = self.window[0]
-        t_max = self.window[1]
-        self.setXTimeRange(t_min, t_max)
 
         self.data_frame = TraceData()
         self._registerDataFrameGetters(self)
@@ -173,14 +163,9 @@ class Trace(object):
         :param t_max: upper bound
         :type t_max: int or float
         """
-        if t_min is None:
-            self.x_min = 0
-        else:
-            self.x_min = t_min
-        if t_max is None:
-            self.x_max = self.time_range
-        else:
-            self.x_max = t_max
+        self.x_min = t_min if t_min is not None else self.start_time
+        self.x_max = t_max if t_max is not None else self.start_time + self.time_range
+
         self._log.debug('Set plots time range to (%.6f, %.6f)[s]',
                        self.x_min, self.x_max)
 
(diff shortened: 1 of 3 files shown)


libs/utils/analysis/status_analysis.py
```diff
@@ -88,6 +88,11 @@ def plotOverutilized(self, axes=None):
         # Compute intervals in which the system is reported to be overutilized
         bands = [(t, df['len'][t], df['overutilized'][t]) for t in df.index]
 
+        # The last event's duration will always be set to NaN, because the time
+        # delta is computed as (next_event time - cur_event time). This fixes
+        # that by 'manually' setting the duration of the last event.
+        bands[-1] = (bands[-1][0], self._trace.x_max - bands[-1][0], bands[-1][2])
```
derkling: Should not better move this fix into the DataFrame sanitizer, i.e. `Trace::_sanitize_SchedOverutilized`?
This would allow to always get a DF which has already the `len` field properly populated?

valschneider:
Moved the fix in trace.py. This also involves cleaning up some of the time range and window code that was duplicated. I might have missed a few spots here and there so I think the first commit needs to be carefully reviewed as to not break anything.

libs/utils/trace.py
```diff
@@ -242,7 +232,8 @@ def __parseTrace(self, path, tasks, window, normalize_time, trace_format):
             raise ValueError("Unknown trace format {}".format(trace_format))
 
         self.ftrace = trace_class(path, scope="custom", events=self.events,
-                                  window=window, normalize_time=normalize_time)
+                                  window=window if window is not None else (0, None),
```
bjackman: `window=window or (0, None)` should work

libs/utils/trace.py
```diff
@@ -629,14 +629,27 @@ def _sanitize_SchedEnergyDiff(self):
             'Suboptimal Reject')
 
     def _sanitize_SchedOverutilized(self):
-        """ Add a column with overutilized status duration. """
+        """
+        Add a column with overutilized status duration.
+        """
         if not self.hasEvents('sched_overutilized'):
             return
+
```
bjackman: Looks like this and the hunk above are commit noise

libs/utils/trace.py
```diff
@@ -629,14 +629,27 @@ def _sanitize_SchedEnergyDiff(self):
...
 
+        # Fix the last event which will have a NaN duration
+        df.ix[-1, "len"] = self.start_time + self.time_range - df.index[-1]
+
+        # Build a stat on trace overutilization
+        self.overutilized_time = df[df.overutilized == 1].len.sum()
+        self.overutilized_prc = 100. * self.overutilized_time / self.time_range
+
+        self._log.debug('Overutilized time: %.6f [s] (%.3f%% of trace time)',
+                            self.overutilized_time, self.overutilized_prc)
```
bjackman: I think the lines from "Build a stat" onwards are supposed to be added in the previous commit?

libs/utils/trace.py
valschneider: I've moved it back into the previous commit, I can add a "beautifying" commit if you'd like.

derkling:
The first patch is scaring... that convince me that @bjackman is right, we reached the point where automated tests are required. :open_mouth: 

... let see If we can get the first Brendan test seed (which I'm going to add to #390) before adding this one in.

valschneider:
Yeah there's quite a lot of shuffling around, I tried to make sure not to forget anything but I'd be much happier with a proper testing of that. Sadly I needed to use the timespan of the trace and it isn't really straightforward in the current code, so I felt like refactoring.

derkling:
Since now we have got merged an initial set of trace related tests, would you mind to add a test to check we always compute underutilised time.

To make things simple, I would suggest to add "by hand" an _odd number_  of properly crafted "fake" `cpu_overutilized` events in the already existing trace:
   https://github.com/ARM-software/lisa/blob/master/tests/lisa/traces/trace.txt
and do the math to assert that the reported start-end of overutilized times matches with the events and/or the trace beginning/end.

bjackman:
If you're adding manual "synthetic" traces, you might want to use this: https://github.com/bjackman/lisa/commit/f69ce9cacf5affcb75e0b3e9dfebb5e84a32be01

But up to you, personally I'm on the fence about whether it's a worthwhile thing..

valschneider:
@bjackman @derkling Updated with :sparkles:tests:sparkles:. Might want to wait for https://github.com/ARM-software/trappy/pull/268 to be merged to include it in here as well.

tests/lisa/test_trace.py
```diff
@@ -78,3 +78,28 @@ def test_setTaskName(self):
         self.assertEqual(trace.getTaskByName('father'), [1234])
 
         os.remove(self.test_trace)
+
+    def test_time_range(self):
+        """
+        TestTrace: time_range is the duration of the trace
+        """
+        expected_duration = 6.676497
```
valschneider: Not too sure if there's a better way to store this value. We could use ftrace.get_duration() but that's how time_range is set so I don't really see the relevance...

valschneider:
Added trappy update and fixed a bad case of English butchery.

libs/utils/trace.py
```diff
@@ -173,14 +163,9 @@ def setXTimeRange(self, t_min=None, t_max=None):
...
-        if t_min is None:
-            self.x_min = 0
-        else:
-            self.x_min = t_min
-        if t_max is None:
-            self.x_max = self.time_range
-        else:
-            self.x_max = t_max
+        self.x_min = t_min if t_min else self.start_time
+        self.x_max = t_max if t_max else self.start_time + self.time_range
```
bjackman: I think you need `is not None`s in here, otherwise you can't set `t_min=0`, right?

libs/utils/trace.py
```diff
@@ -230,8 +215,13 @@ def __parseTrace(self, path, window, trace_format):
...
             raise ValueError("Unknown trace format {}".format(trace_format))
 
-        self.ftrace = trace_class(path, scope="custom", events=self.events,
-                                  window=window, normalize_time=self.normalize_time)
+        if self.normalize_time:
+            self.ftrace = trace_class(path, scope="custom", events=self.events,
+                                      window=window, normalize_time=self.normalize_time)
+        else:
+            self.ftrace = trace_class(path, scope="custom", events=self.events,
+                                      abs_window=window, normalize_time=self.normalize_time)
```
bjackman: Maybe instead of having two `trace_class` call sites you could have something like
```
# If using normalized time, we need to use TRAPpy's `abs_window` param instead of `window`
if self.normalize_time:
    abs_window = window
    window = (0, None)
else:
    abs_window = (0, None)
    window = window

self.ftrace = trace_class(...)
```
It's actually longer but perhaps clearer? Up to you.

libs/utils/trace.py
valschneider: You're right, good catch.

libs/utils/trace.py
```diff
@@ -590,6 +590,10 @@ def _sanitize_SchedOverutilized(self):
         df['len'] = (df.start - df.start.shift()).fillna(0).shift(-1)
         df.drop('start', axis=1, inplace=True)
 
+        # Fix the last event, which will have a NaN duration
+        # Set duration to trace_end - last_event
+        df.ix[-1, 'len'] = self.start_time + self.time_range - df.index[-1]
```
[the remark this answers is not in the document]
valschneider: Lovely. Thanks for the heads-up.

libs/utils/trace.py
valschneider: While writing this I was thinking we could maybe use a dictionary to describe the args and feed it unpacked to the `trace_class` call - would that be a good practice ?

bjackman: Yeah I thought of that too, but it feels a bit runic, and anyway I think it would end up looking just like the code in my example, right? But yeah it's just style so up to you really. Can keep the current code too if you prefer.

libs/utils/trace.py
```diff
@@ -171,14 +163,9 @@ def setXTimeRange(self, t_min=None, t_max=None):
...
-        if t_min is None:
-            self.x_min = 0
-        else:
-            self.x_min = t_min
-        if t_max is None:
-            self.x_max = self.time_range
-        else:
-            self.x_max = t_max
+        self.x_min = t_min if t_min is not None else self.start_time
+        self.x_max = t_max if t_max is not None else self.start_time + self.time_range
```
valschneider: Added a None check here for the sake of symmetry. A 0 x_max doesn't make a lot of sense, but I think it's less confusing to let that pass than to force the x_max to its maximum possible value if 0 is given.

libs/utils/trace.py
```diff
@@ -231,8 +215,16 @@ def __parseTrace(self, path, window, normalize_time, trace_format):
         else:
             raise ValueError("Unknown trace format {}".format(trace_format))
 
+        # If using normalized time, we should use
+        # TRAPpy's `abs_window` instead of `window`
+        window_kw = {}
```
valschneider: Added a dict for that window argument. I like this version because it only modifies the argument we're interested in. Of course, there's little chance that the default value will be changed from (0, None), but I prefer to be able to leave default values alone.

libs/utils/trace.py
```diff
@@ -628,6 +593,18 @@ def _sanitize_SchedOverutilized(self):
         df['len'] = (df.start - df.start.shift()).fillna(0).shift(-1)
         df.drop('start', axis=1, inplace=True)
 
+        # Fix the last event, which will have a NaN duration
+        # Set duration to trace_end - last_event
+        df.loc[df.index[-1], 'len'] = self.start_time + self.time_range - df.index[-1]
```
valschneider: df.ix shenanigans

valschneider:
@bjackman Marked changes in the code.

libs/utils/trace.py
bjackman: Oh yeah this is nice, I didn't think of using `**` for just the winow arg. :+1:

PR #437, opened 2017-07-03
tests/eas/generic: Tweak criteria for passing tests

bjackman:
- Relax energy threshold for ThreeSmallTasks because we don't really know what
  the ideal task placement is (don't expect energy model to be very accurate
  here), just want to sanity-check the placement (e.g. hopefully fail if tasks
  are all on big CPUs).

- Allow RampDown test a larger threshold to allow for hysteresis in
  down-migration.

- Remove the energy test from the TwoBigThreeSmall, this is similar to the
  ThreeSmall Tasks case but more extreme - let's just check that the throughput
  was provided but ignore the energy-aware placement as our estimation is very
  likely wrong for this workload.

### DIFF
--- a/tests/eas/generic.py
+++ b/tests/eas/generic.py
@@ -329,6 +329,15 @@ class ThreeSmallTasks(_EnergyModelTest):
     """
     Test EAS for 3 20% tasks over 2 seconds
     """
+
+    # The energy estimation for this test is probably not very accurate and this
+    # isn't a very realistic workload. It doesn't really matter if we pick an
+    # "ideal" task placement for this workload, we just want to avoid using big
+    # CPUs in a big.LITTLE system. So use a larger energy threshold that
+    # hopefully prevents too much use of big CPUs but otherwise is flexible in
+    # allocation of LITTLEs.
+    energy_est_threshold_pct = 20
+
     workloads = {
         'three_small' : {
             'type' : 'rt-app',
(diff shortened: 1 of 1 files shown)


tests/eas/generic.py
```diff
@@ -329,6 +329,15 @@ class ThreeSmallTasks(_EnergyModelTest):
...
     Test EAS for 3 20% tasks over 2 seconds
     """
+
+    # The energy estimation for this test is probably not very accurate and this
+    # isn't a very realistic workload. It doesn't really matter if we pick an
+    # "ideal" task placement for this workload, we just want to avoid using big
+    # CPUs in a big.LITTLE system. So use a larger energy threshold that
+    # hopefully prevents too much use of big CPUs but otherwise is flexible in
+    # allocation of LITTLEs.
+    energy_est_threshold_pct = 20
```
derkling: Should we adopt a convention to always use CAPITAL_CASE name for hard-coded magic values?

derkling: Moreover, perhaps this should be better an issue but, this test hardcode 3 small tasks... which will not fit on platforms like pixels for example.

Should not we fix these tests to make them more portable and less TC2 style.... we can still use the LISA's RTApp configuration to ask for a number of small/big tasks which exactly matches the little/big cores number... and update the tests names to be more aligned... for example, this should be s/ThreeSmallTasks/OnlySmallTasks/

What do you think? Should I open an Issue... or should we better (at least try to) fix workload compositions and tests names for this release?

bjackman: > Should we adopt a convention to always use CAPITAL_CASE name for hard-coded magic values?

I prefer lower case here because it's actually a variable (modified by some classes)

> this test hardcode 3 small tasks... which will not fit on platforms like pixels for example.

This will work on any platform. The only case it isn't very useful is if you only have 1 LITTLE and its capacity is less than 30% of big (not likely). Even then, the test will figure out that one or more tasks needs to go on big, and things will work fine.

bjackman:
Merging a version that leaves the task placement test in for the TwoBigThreeSmall. We can come back to it later if necessary

PR #440, opened 2017-07-09
[RFC] CPU residency analysis based on a pivot

joelagnel:
Hi @derkling,
Here's a quick attempt to get meaningful `cgroup_attach_task` events by combining the devlib fake attach events with the real ones. Just posting for any thoughts/ideas/feedback. I've lightly tested it and its working fine for me. Also added in the residency module.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -536,7 +536,7 @@ class Trace(object):
             if not cev in self.available_events:
                 continue
             cdf = self._dfg_trace_event(cev)
-            cdf = cdf[['pid', 'controller', 'cgroup']]
+            cdf = cdf[['__line', 'pid', 'controller', 'cgroup']]
             if not isinstance(df, pd.DataFrame):
                 df = cdf
             else:
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -528,7 +528,8 @@ class Trace(object):
             return df
         return df.loc[df.index.get_level_values(1).isin(listify(functions))]
 
-    def _dfg_cgroup_attach_task(self):
+    # cgroup_attach_task with just merged fake and real events
+    def _cgroup_attach_task(self):
         cgroup_events = ['cgroup_attach_task', 'cgroup_attach_task_devlib']
         df = None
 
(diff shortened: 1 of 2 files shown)


joelagnel:
Here is a notebook to show the use of it:
https://gist.github.com/joelagnel/70959e544c96f54f71d63da011bbbd8a

joelagnel:
Added some more PRs, probably some should be squashed,
the last one builds a complete `sched_switch_cgroup` DF  (`trace.data_frame.sched_switch_cgroup`)

By default `schedtune` and `cpuset` are controllers that are scanned, but a custom controller list can be passed. It results in the following columns in `sched_switch` (only showing the new and a few old columns):
'prev_pid', 'prev_comm', 'prev_schedtune', 'prev_cpuset', 'next_pid', 'next_comm', 'next_schedtune', 'next_cpuset'

Here is a notebook processing trace data from `UiBench`:
https://gist.github.com/joelagnel/0ef27aacae7788b278681a427c4c1647

joelagnel:
So I found a bit of a blocking issue. If a new task is created on fork, it inherits its parent's cgroups but it will not generate an cgroup_attach_task event :(

```
The attachment of each task, automatically inherited at fork by any
children of that task, to a cgroup allows organizing the work load
on a system into related sets of tasks.  A task may be re-attached to
any other cgroup, if allowed by the permissions on the necessary
cgroup file system directories.
```

joelagnel:
@derkling I added in the fork stuff. Seems to work well:
https://github.com/ARM-software/lisa/pull/440/commits/98fa7e33deae5333294419356bc883b26ceaa0ef

joelagnel:
@derkling I added one more commit with the analysis module as well. It probably needs a lot of clean up but I appreciate your testing/review any way and if you see any bugs. I was also thinking of writing test cases for it. I did some manual testing and it is working well.

joelagnel:
CC @sinkap

joelagnel:
Sorry for the accidental push of commits, I will clean this up today. Apologies and thanks.

libs/utils/trace.py
```diff
@@ -536,7 +536,7 @@ def _dfg_cgroup_attach_task(self):
             if not cev in self.available_events:
                 continue
             cdf = self._dfg_trace_event(cev)
-            cdf = cdf[['pid', 'controller', 'cgroup']]
+            cdf = cdf[['__line', 'pid', 'controller', 'cgroup']]
```
mdigiorgio: Squash to previous commit

libs/utils/trace.py
```diff
@@ -542,8 +543,47 @@ def _dfg_cgroup_attach_task(self):
...
+        # Always drop na since this DF is used as secondary
+        fdf.dropna(inplace=True, how='any')
+
+        new_forks_len = len(fdf) / len(controllers)
+
+        fdf = pd.concat([fdf, cdf]).sort(columns='__line')
+
+        if new_forks_len < forks_len:
+            dropped = forks_len - new_forks_len
+            self._log.info("Couldn't attach all forks cgroup with-attach events ({} dropped)".format(dropped))
```
mdigiorgio: self._log.warning

PR #442, opened 2017-07-10
submodules: Update devlib to latest master

ionela-voinescu:
This brings in support for GEM5 power monitoring.


qperret:
I would rather say that this bring support for gem5 "power monitoring" instead of "energy meter" in the commit message. Beside that, it's ok for me :)

ionela-voinescu:
@qperret Fixed commit message.
@all Brought in commit that fixes the syntax error that was causing Travis CI to fail.

Issue #443, opened 2017-07-14
Add Hikey960 board support in LISA


cloverpop:
Hikey960 board CPU topology
4 Cortex-A73 + 4 Cortex-A53

cloverpop:
Patch is submitted.

butterl:
@cloverpop 
> Patch is submitted.
Do you mean Hikey960 already supported by lisa ？ Do I need an extra  compile flag for it?

Thanks,
Butter

bjackman:
@butterl I believe all you need to do is add `"board": "hikey960"`to your `target.config`.

PR #444, opened 2017-07-14
Add Hikey960 board support in LISA #443

cloverpop:
Add Hikey960 board support in LISA


libs/utils/platforms/hikey960.json
```diff
@@ -0,0 +1,37 @@
+{
+	// HiKey960 boards have two clusters
+	// 4 Cortex A73 + 4 Cortex A53 big.LITTLE architecture
+	"board" : {
```
valschneider: Nitpick here, I think spaces would be better than tabs here.

libs/utils/platforms/hikey960.json
```diff
@@ -0,0 +1,37 @@
...
+		"modules"  : [ "bl", "cpufreq", "cpuidle" ]
+	},
+
+    "nrg_model" : {
+        "little" : {
+            "cluster" : {
+                "nrg_max" :   70,
+            },
+	        "cpu" : {
+                "nrg_max" :   1210,
```
valschneider: Where are these numbers coming from ? The "official" energy model in hi3660-sched-energy.dtsi has quite different values.

libs/utils/platforms/hikey960_energy.py
```diff
@@ -0,0 +1,124 @@
...
+    ( 903000,  ActiveState(capacity=428,  power=332)),
+    ( 1421000,  ActiveState(capacity=684,  power=752)),
+    ( 1805000,  ActiveState(capacity=868,  power=1209)),
+    ( 2112000,  ActiveState(capacity=979,  power=1520)),
+    ( 2362000, ActiveState(capacity=1024, power=1715))
+])
+
+CA73_cpu_idle_states = OrderedDict([
+    ('WFI', 10),
+    ('cpu-sleep', 0),
```
valschneider: The CA73 has 4 idle states:
WFI,
cpu-nap
cpu-sleep
cluster-sleep

libs/utils/platforms/hikey960_energy.py
```diff
@@ -0,0 +1,124 @@
...
+])
+
+CA53_cluster_idle_states = OrderedDict([
+    ('WFI', 20),
+    ('cpu-sleep', 20),
+    ('cluster-sleep', 0)
+])
+
+CA53_cpu_active_states = OrderedDict([
+    ( 533000,   ActiveState(capacity=257,   power=160)),
```
valschneider: Same comment here regarding the source of these capacity/power values.

libs/utils/env.py
```diff
@@ -441,6 +448,8 @@ def _init_target(self, force = False):
 
         elif self.conf['board'] != 'UNKNOWN':
             # Initilize from platform descriptor (if available)
+            self._log.error('Warning: UNKNOWN board type detected!!')
```
valschneider: I think this should be moved to a separate commit.

libs/utils/platforms/hikey960.json
cloverpop: I copied it from libs/utils/platforms/hikey.json directly.

libs/utils/platforms/hikey960.json
cloverpop: Don't know why you point to hi3660-sched-energy.dtsi, which is CA72 + CA53 CPUs.
However, Hikey960 is CA73+CA53. 
http://www.96boards.org/product/hikey960/

Even if the energy model can be tuned later, we can check in the support for hikey960 support first of all.

libs/utils/platforms/hikey960_energy.py
cloverpop: I don't know who ever created the EAS energy model in Hikey960 board or not.
If not, all the data we have is the estimation for simulation first. 
Only after we have the hikey960 support in LISA, we can tune the following scripts for Hikey960 board.

libs/utils/env.py
cloverpop: The reason is to give a easier life for the developers who want to add new board support in LISA.
Without the explicit logging information for the "UNKNOWN" board type, it takes longer time for new developers to find out the code here.

This is one reason why customers complain the LISA is not so friendly.

libs/utils/platforms/hikey960.json
valschneider: The hi3660-sched-energy.dtsi file is for the HiKey960 - the CA72 instead of CA73 mistype is a common error with that board (same thing in the Trusted Firmware, it has been fixed actually).

libs/utils/env.py
valschneider: I don't disagree with the code, I just think that moving these few lines in a separate commit would make the history cleaner.

valschneider:
The HiKey960 is in a good state right now, and with hardware video decoding hopefully coming at the end of the month I think we'll be using the platform more and more. We'll need this PR merged sooner than later, but I'd like the following points to be addressed first:

- Merge/Notebook commits should be removed from this PR.
- Move the UNKNOWN board warning in a separate commit
- Energy model values should be aligned with the values used by the kernel, so they should be the same as in hi3660-sched-energy.dtsi.

libs/utils/env.py
ionela-voinescu: Given that these are warnings (and they should be) they should be logged as a warnings with `self._log.warning`. The use of `Warning:` at the beginning of the message strings bring no useful information in this case.

Also, I would propose a rewording of the second message, maybe as "Execution on this target may have unexpected results".

libs/utils/platforms/hikey960_energy.py
```diff
@@ -0,0 +1,124 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from energy_model import (ActiveState, EnergyModelNode, EnergyModelRoot,
+                          PowerDomain, EnergyModel)
+
+from collections import OrderedDict
+
```
ionela-voinescu: Nitpick: It would be good to remove some of the white lines in this file. They have no purpose.

libs/utils/platforms/hikey960_energy.py
ionela-voinescu: > Only after we have the hikey960 support in LISA, we can tune the following scripts for Hikey960 board.

Most of the support for the Hikey960 board is represented by the content of the hikey960.json and hikey96_energy.py files. In my opinion it's important to have correct information here, information matching the one provided in DT.

PR #445, opened 2017-07-14
Bug: Fails to install tool file path without ABI prefix

cloverpop:
Solution:  append the tool file path with ABI prefix to tools_to_install list


bjackman:
The commit "Bug: Fails to install tool file path without ABI prefix" looks good, but this PR includes other stuff that needs to be removed before it can be merged (the merge commits and IPython notebook commits)

cloverpop:
Re-post the patch as https://github.com/ARM-software/lisa/pull/454

bjackman:
Merged via https://github.com/ARM-software/lisa/pull/454 as https://github.com/ARM-software/lisa/commit/74ea6eda578f9b48a32df4c9381b4db8a7f4f001

PR #447, opened 2017-07-20
workloads: add a new camera preview workload

joelagnel:
This workload starts the camera app, waits for a few seconds and then starts data collection (trace, energy etc).

I also have a workload that starts the camera app after tracingStart, I call it 'cameraStartup' however I don't know how useful it is for upstream so I dropped it. cameraStartup was useful for me when I was debugging the time it takes from the start of the camera app to the first preview frame.

### DIFF
--- a/libs/utils/android/workloads/camera_preview.py
+++ b/libs/utils/android/workloads/camera_preview.py
@@ -1,0 +1,105 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2015, ARM Limited, Google and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


joelagnel:
@derkling as you asked me to send this one as it could be useful, here you go :) Connor wrote the sysui workload so I asked him to push that separately, he said we would push it when he gets a chance. Incase he doesn't we can remind him.

bjackman:
Still not totally familiar with the `Workload` API but this looks good to me :)

ionela-voinescu:
It looks good! I've tested it on Pixel with the test at https://github.com/ionela-voinescu/lisa/commit/e2834c7f174c0f992f04d5d49a36f1c7872cccb8. I'll submit this one after yours is merged so we can use the workload more easily in CI.

derkling:
@bjackman I still have some comments to add, but did not yet had time... moreover I've promised to not bother @joelagnel, thus I'll cleanup few things on that patch before merging it ;-)

@ionela-voinescu regarding the tests we have in CI.... perhaps we should remove some governors configurations. Potentially by testing just with the default one (schedutil) and keep it running overnight just to check for regressions along time.

Issue #449, opened 2017-07-21
Bug: RTA property delay_s will cause kernel segmentation fault


cloverpop:
rt-app workload with delay_s property triggers kernel fault immediately.

[28081.878055] delay_starter_1[4059]: unhandled level 0 translation fault (11) at 0x5f746c74
[28081.895414] pgd = ffffffc0b5e9b000
[28081.898809] [5f746c75616665] *pgd=0000000000000000, *pud=0000000000000000
[28081.905976] 
[28081.907552] CPU: 5 PID: 4059 Comm: delay_starter_1 Not tainted 4.4.77-g576a011-dirty #1
[28081.915561] Hardware name: HiKey960 (DT)
[28081.919485] task: ffffffc0787ce080 ti: ffffffc094944000 task.ti: ffffffc094944000
[28081.926973] PC is at 0x43c6c4
[28081.929941] LR is at 0x4028b4
[28081.932904] pc : [<000000000043c6c4>] lr : [<00000000004028b4>] pstate: 00000000
[28081.940298] sp : 0000007fb7ffd610
[28081.943608] x29: 0000007fb7ffd610 x28: 00000000004b16f0 
[28081.948925] x27: 0000007fb7ffe0a0 x26: 00000000004af000 
[28081.954239] x25: 0000000000402ca8 x24: 00000000004a9000 
[28081.959560] x23: 00000000004a8000 x22: 0000007ffffff65f 
[28081.964877] x21: 0000000000000000 x20: 0000000000000000 
[28081.970191] x19: 00000000004a84a0 x18: 0000000000000000 
[28081.975512] x17: 00000000004a9000 x16: 0000000000000000 
[28081.980828] x15: ffffffffffffffff x14: 0000000000000001 
[28081.986141] x13: 0000000000000010 x12: 0101010101010101 
[28081.991464] x11: 7f7f7f7f7f7f7f7f x10: fefefefefeff5c1f 
[28081.996780] x9 : ffffffffffffffff x8 : 705f7413d1617eb5 
[28082.002095] x7 : 0000808080808080 x6 : 0000007fa400190a 
[28082.007417] x5 : ffffffffffffffff x4 : 0000000000005d20 
[28082.012733] x3 : 705f746c75616665 x2 : 000000000000007f 
[28082.018048] x1 : 0000007fa40018d0 x0 : 705f746c75616665

bjackman:
[...]

PR #450, opened 2017-07-21
Bug: RTA property delay_s will case kernel segmentation fault #449

cloverpop:
BUG: RTA workload property delay_s will cause kernel userspace rt-app segmentation fault.
Configure the correct RTA settings to meet the RT-APP requirement ( up to commit=a9a0c6c )

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -411,8 +411,7 @@ class RTA(Workload):
 
             if 'delay' in task.keys():
                 if task['delay'] > 0:
-                    task_conf['phases']['p000000'] = {}
-                    task_conf['phases']['p000000']['delay'] = int(task['delay'] * 1e6)
+                    task_conf['delay'] = int(task['delay'] * 1e6)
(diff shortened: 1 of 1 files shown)


bjackman:
I wasn't able to reproduce the segfault with the version of rt-app in LISA, but it does indeed look like rt-app doesn't recognise the "delay" property for phases so this looks correct to me.

However I think the RTA class is still going to be broken:

```
wl = RTA(te.target, 'mywl', calibration=calib)
wl.conf('profile', 
         params = {'t': Periodic(delay_s=3, duration_s=1) + Periodic(delay_s=2, duration_s=1)).get()})
```

This should result in 

1. 3s delay
2. 1s of periodic behaviour
3. 2s delay
4. 1s of further periodic behaviour.

But part 3 will be ignored. That is, if we compose RTATasks, only the first task's `delay` will be respected.

I think we can probably still merge this PR and then fix that separately though.

bjackman:
OK, submitted a "fix" for my example here: https://github.com/ARM-software/lisa/pull/451

Thanks for the fix, @cloverpop !

PR #451, opened 2017-07-21
rta: Disallow invalid composition of delayed tasks

bjackman:
This example:

    wl = RTA(te.target, 'mywl', calibration=calib)
    wl.conf('profile',
            params = {'t': Periodic(delay_s=3, duration_s=1) +
                    Periodic(delay_s=2, duration_s=1)).get()})

Should semantically result in

1. 3s delay
2. 1s of periodic behaviour
3. 2s delay
4. 1s of periodic behaviour.

But part 3 (the second delay) will be ignored due to the way we use
the "delay" field.

We could probably fix this using a "sleep" event, but to keep things
simple let's just disallow this type of composition, unless we later
find that we need that feature.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -606,6 +606,14 @@ class RTATask(object):
         return self._task
 
     def __add__(self, next_phases):
+        if next_phases._task.get('delay', 0):
+            # This won't work, because rt-app's "delay" field is per-task and
+            # not per-phase. We might be able to implement it by adding a
+            # "sleep" event here, but let's not bother unless such a need
+            # arises.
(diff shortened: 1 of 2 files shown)


derkling:
The API can be considered convoluted... but I think I've added a comment somewhere saying that, if you use composition, the delay_s applies only for the FIRST phase..
That's to retain the semantics of delay_s, which is expected to apply only at the beginning of a task creation.

Does that makes sense?

bjackman:
It makes sense but I still think we should error out instead of silently dropping the "delay_s"

derkling:
Absolutely agree! Well done for the check and test as well! :+1: +1

PR #452, opened 2017-07-21
tests/benchmarks: add android_camera_preview benchmark test

ionela-voinescu:

### DIFF
--- a/tests/benchmarks/android_camera_preview.py
+++ b/tests/benchmarks/android_camera_preview.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python
+
+import os
+from time import sleep
+
+# The workload class MUST be loaded before the LisaBenchmark
+from android import Workload
+from android import LisaBenchmark
+
+from devlib.exception import TargetError
+
+class CameraPreviewTest(LisaBenchmark):
+
+    # Android Workload to run
+    bm_name = 'CameraPreview'
+
+    # Default products to be collected
+    bm_collect = 'ftrace energy'
+
+    def benchmarkInit(self):
+        self.setupWorkload()
+        self.setupGovernor()
+
+    def __init__(self, governor, duration_s):
+        self.governor = governor
+        self.duration_s = duration_s
+        super(CameraPreviewTest, self).__init__()
+
+    def setupWorkload(self):
+        # Create a results folder for each "governor/test"
+        self.out_dir = os.path.join(self.te.res_dir, governor)
+        try:
+                os.stat(self.out_dir)
+        except:
+                os.makedirs(self.out_dir)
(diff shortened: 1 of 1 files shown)


ionela-voinescu:
@derkling in reply to your comment on https://github.com/ARM-software/lisa/pull/447:

The configuration (list of governors, list of tests, duration of benchmark, etc.) of tests in tests/benchmarks is usually there just as an example. Most of the times, in our tests, we change that to a configuration more appropriate to the purpose of our test.

tests/benchmarks/android_camera_preview.py
```diff
@@ -0,0 +1,117 @@
...
+        try:
+            self.target.cpufreq.set_all_governors(self.governor);
+        except TargetError:
+            self._log.warning('Governor [%s] not available on target',
+                             self.governor)
+            raise
+
+        # Setup schedutil parameters
+        if self.governor == 'schedutil':
+            rate_limit_us = 2000
```
qperret: Why not leaving this value as-is ? I'm not saying this is bad, and I can also see that we do the same for other benchmarks but what is the rationale ?

tests/benchmarks/android_camera_preview.py
```diff
@@ -0,0 +1,117 @@
...
+                    tunables = self.target.cpufreq.get_governor_tunables(cpu_id)
+                    self.target.cpufreq.set_governor_tunables(
+                        cpu_id, 'ondemand',
+                        **{'sampling_rate' : tunables['sampling_rate_min']})
+            except TargetError as e:
+                self._log.warning('Failed to set ondemand parameters: {}'\
+                                 .format(e))
+                raise
+            self._log.info('Set ondemand.sampling_rate to minimum supported')
+
```
qperret: What about other governors ? Should we enforce a specific value for `userspace` for ex ?

tests/benchmarks/android_camera_preview.py
```diff
@@ -0,0 +1,117 @@
...
+# Run the benchmark in each of the supported governors
+
+duration_s = 20
+
+governors = [
+    'performance',
+    'ondemand',
+    'interactive',
+    'sched',
```
qperret: Is this still needed ? Maybe for legacy ? schedfreq has been removed from recent android version (>=4.4)

tests/benchmarks/android_camera_preview.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python
+
+import os
+from time import sleep
```
qperret: Not used.

valschneider:
Following a discussion with @ionela-voinescu, sending this to PR hell.

PR #453, opened 2017-07-21
Hikey960 integration

valschneider:
Haven't created a hikey960_energy file yet since it can be read from the device. I still want to do that for post-processing things without the device at hand, but that's less important/likely.

### DIFF
--- a/libs/utils/platforms/hikey960.json
+++ b/libs/utils/platforms/hikey960.json
@@ -1,0 +1,30 @@
+{
+    "board" : {
+	"cores" : [
+	    "A53", "A53", "A53", "A53", 
+	    "A73", "A73", "A73", "A73"
+	],
+	"big_core" : "A73",
+	"modules" : ["bl", "cpufreq", "cpuidle"]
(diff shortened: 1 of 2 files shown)


libs/utils/platforms/hikey960.json
```diff
@@ -0,0 +1,30 @@
+{
+    "board" : {
+	"cores" : [
+	    "A53", "A53", "A53", "A53", 
+	    "A73", "A73", "A73", "A73"
+	],
+	"big_core" : "A73",
+	"modules" : ["bl", "cpufreq", "cpuidle"]
+    },
+    "nrg_model": {
```
derkling: Do these values match the in-kernel EM?

valschneider: Afaik they should: https://android.googlesource.com/kernel/hikey-linaro/+/android-hikey-linaro-4.4/arch/arm64/boot/dts/hisilicon/hi3660-sched-energy.dtsi

PR #454, opened 2017-07-24
Bug: Fails to install tool file path without ABI prefix #445

cloverpop:
Bug: Fails to install tool file path without ABI prefix
Solution: Append the tool file path with ABI prefix to tools_to_install list.


bjackman:
Thanks!

PR #456, opened 2017-07-27
GEM5 Juno support

ionela-voinescu:


libs/utils/env.py
```diff
@@ -600,6 +601,14 @@ def _init_target_gem5(self):
             system['platform']['description']))[0]
         self.conf['board'] = self.conf['board'].lower() + suffix
 
+        energy_file = self.conf['board'] + '_energy';
+        try:
+            self.nrg_model = globals()[energy_file]
```
ionela-voinescu: This patch is a RFC at the moment as I'm not sure if there's a better way to select a specific energy model object, dynamically, based on board and platform information in the target configuration.

libs/utils/platforms/gem5juno.json
```diff
@@ -0,0 +1,30 @@
+{
+    "board" : {
```
valschneider: Other than board naming, is there any reason to have a gem5-specific platform file ? We should be able to use the same values as the "real" thing, right ? Same comment goes to the gem5-specific energy file.

libs/utils/env.py
valschneider: Without digging any deeper I don't see a clean solution for that, however I have a comment: maybe we could change the way platform files are loaded, so that gem5 can use the same files as the "real" platforms ? I understand that in some cases you may still want a gem5-specific file, so maybe we could prioritize them (gem5-specific file first, otherwise real platform file).

ionela-voinescu: 1. It is a good idea to have it default to existing "real" platform and/or energy file, but it won't apply for Juno, unfortunately (please see 2). 
2. I think this is possible for it to use the juno.json file instead of the gem5juno.json file, but not the juno "real" energy file. If you look over the gem5juno_energy.py file you'll see that the order of cores is different and the frequencies are slightly different from the ones of the real platform. This is due to how GEM5 accounts for time (which more knowledgeable people can explain better).
The reason I would choose not to reuse the juno.json file is that by precisely specifying cores and  big_core in gem5juno.json I make sure that this works for older versions of gem5 that had an issue with differentiating cores.
But I'm not sure it's worth the duplication in this case. I'll submit another version with 1.

Thank you!

valschneider: Regarding 2:
I didn't look at the core ordering & frequencies - it makes sense in this case to have a different file. As for specifying cores, that could be done in the "real" platform file (it's the case for the HiKey960 https://github.com/ARM-software/lisa/blob/master/libs/utils/platforms/hikey960.json), but if you need a different file to account for different core ordering anyway it's not really relevant.

derkling:
@ionela-voinescu any further response on @valschneider comments?
Is this PR still valid?

ionela-voinescu:
@derkling  Yes, it's still valid. Let me clean up my local changes and I'll push an update as per @valschneider  comments.

valschneider:
Duuh I seem to have forgotten everything about that one. On top of that the `TestEnv` landscape has changed quite a bit in LISA/next. @ionela-voinescu, what's your take on this?

douglas-raillard-arm:
Looks like providing a platform info file like for any other board should be sufficient. Since PlatformInfo is decoupled from TargetConf, we can use an unlimited number of PlatformInfo descriptions for the same "generic" gem5 target. You can find the doc about available keys here:
https://lisa-linux-integrated-system-analysis.readthedocs.io/en/next/internals.html?highlight=PlatformInfo#lisa.platforms.platinfo.PlatformInfo

And existing (basic) platform info file:
https://github.com/ARM-software/lisa/blob/next/lisa/platforms/juno_r0.yml
(this one is referring to a Python global variable defined in https://github.com/ARM-software/lisa/blob/next/lisa/platforms/juno_r0.py)

PR #459, opened 2017-08-02
ipynb/examples/android/workloads: Fix Gmaps notebook

valschneider:
Some of Trappy's API has been changed, so this brings the notebook
in line with these changes.


bjackman:
This isn't a trappy API, but the change itself is correct AFAICS

valschneider:
@bjackman Right, it's an update of LISA's trace API (which uses Trappy). Relevant PR is https://github.com/ARM-software/lisa/pull/390

PR #462, opened 2017-08-08
Bonus RTA test coverage

bjackman:
Wrote this test while changing the Step class. Turns out I don't need to make that change but might as well merge the test anyway.


mdigiorgio:
LGTM

PR #465, opened 2017-08-10
add kevin and elm chromebook platform files

stephenkyle-ARM:
These are called 'cros-kevin' and 'cros-elm', respectively.


bjackman:
Is 'cros' a standard abbrev for ChromeOS?

stephenkyle-ARM:
Yes. See for instance the argument given for browser under "To run on a Chromebook" here: https://www.chromium.org/developers/telemetry/run_locally

PR #466, opened 2017-08-14
test/eas: add test under CPU hotplug stress

qperret:
The CPU hotplug procedure changes many data structures used by the scheduler and interferes with the task placement, hence resulting in all sorts of bugs. This PR enables testing under CPU hotplug stress as an attemp to trigger those bugs. CPUs are plugged in and off in a random (but repeatable) way while a workload is running. For now, the test is considered passing if the workload simply completes its work and if the system doesn't halt.

One thing worth discussing about the current implementation: the hotplug stress starts at the very beginning and stops at the very end. It does not stop nor restart between iterations of the test, which means that iteration `n` and `n+1` are not assessed over the same hotplug sequence. Not sure if this is a problem or a feature ...

One last thing, I'm not (yet) familiar with the all the test APIs so all feedbacks are more than welcome :)

### DIFF
--- a/libs/utils/target_script.py
+++ b/libs/utils/target_script.py
@@ -101,12 +101,26 @@ class TargetScript(object):
         # Push it on target
         self._remote_path = self._target.install(self._local_path)
 
-    def run(self):
+    def run(self, as_root=False, background=False):
         """
         Run the previously pushed script
         """
-
         if self._target.file_exists(self._remote_path):
-            self._target.execute(self._remote_path)
+            self._run_as_root = as_root
+            self._bg_shell = None
+            if background:
+                self._bg_shell = self._target.background(self._remote_path, 
+                                                    as_root=self._run_as_root)
+            else:
+                self._target.execute(self._remote_path, as_root=self._run_as_root)
         else:
             raise IOError('Remote script was not found on target device')
+
+    def kill(self):
+        """
+        Kill a running script
+        """
+        cmd_pid = '$(pgrep {})'.format(self._script_name)
+        self._target.kill(cmd_pid, as_root=self._run_as_root)
+        if self._bg_shell:
+            self._bg_shell.kill()
--- a/libs/utils/target_script.py
+++ b/libs/utils/target_script.py
@@ -93,17 +93,13 @@ class TargetScript(object):
         actions = ['#!{} sh'.format(self._target.busybox)] + actions
         actions = str.join('\n', actions)
 
-        self._remote_path = self._target.path.join(self._target.executables_directory,
-                                                   self._script_name)
-        self._local_path = os.path.join(self._env.res_dir, self._script_name)
-
         # Create script locally
+        self._local_path = os.path.join(self._env.res_dir, self._script_name)
         with open(self._local_path, 'w') as script:
             script.write(actions)
 
         # Push it on target
-        self._target.push(self._local_path, self._remote_path)
-        self._target.execute('chmod +x {}'.format(self._remote_path))
+        self._remote_path = self._target.install(self._local_path)
 
     def run(self):
         """
--- a/tests/eas/cpuhp.py
+++ b/tests/eas/cpuhp.py
@@ -217,3 +217,28 @@ class _CpuHotplugTest(LisaTest):
         except Exception:
             raise AssertionError("the target is not responsive")
 
+class ThreeSmallTasks(_CpuHotplugTest):
+    """
+    Test EAS for 3 20% tasks over 60 seconds
+    """
+    workloads = {
+        'cpuhp_three_small' : {
+            'type' : 'rt-app',
+            'conf' : {
+                'class' : 'periodic',
+                'params' : {
+                    'duty_cycle_pct': 20,
+                    'duration_s': 60,
+                    'period_ms': 16,
+                },
+                'tasks' : 3,
+                'prefix' : 'many',
+            },
+        },
+    }
+
+    @experiment_test
+    def test_random_hotplugs(self, experiment, tasks):
+        '''Test that the system doesn't crash while under hotplug stress'''
+        self._test_target_is_alive()
+
--- a/tests/eas/cpuhp.py
+++ b/tests/eas/cpuhp.py
@@ -242,3 +242,38 @@ class ThreeSmallTasks(_CpuHotplugTest):
         '''Test that the system doesn't crash while under hotplug stress'''
         self._test_target_is_alive()
 
+class Torture(_CpuHotplugTest):
+    """
+    Torture hotplug with 5 15% tasks over 5 minutes
+    """
+    workloads = {
+        'cpuhp_five_small' : {
+            'type' : 'rt-app',
+            'conf' : {
+                'class' : 'periodic',
+                'params' : {
+                    'duty_cycle_pct': 15,
+                    'duration_s': 300,
+                    'period_ms': 16,
+                },
+                'tasks' : 5,
+                'prefix' : 'many',
+            },
+        },
+    }
+
+    hp_stress = {
+        'seed' : None,                  # Seed of the random number generator
+        'sequence_len' : 100,           # Number of operations in the sequence
+        'sleep' : {
+            'min_ms' : -1,              # Min sleep duration between hotplugs
+            'max_ms' : -1,              # Max sleep duration between hotplugs
+        },                              #   max_ms <= 0 will encode 'no sleep'
+        'max_cpus_off' : sys.maxint,    # Max number of CPUs plugged-off
+    }
+
+    @experiment_test
+    def test_random_hotplugs(self, experiment, tasks):
+        '''Test that the system doesn't crash while under hotplug stress'''
+        self._test_target_is_alive()
+
--- a/tests/eas/cpuhp.py
+++ b/tests/eas/cpuhp.py
@@ -1,0 +1,219 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2017, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os.path
+import logging
+import random
+import json
+import sys
+
+from test import LisaTest, experiment_test
+from target_script import TargetScript
+from devlib.module.hotplug import HotplugModule
+
+class _CpuHotplugTest(LisaTest):
+    """
+    "Abstract" base class for generic EAS tests under CPU hotplug stress
+
+    Subclasses should provide a .workloads member to populate the 'wloads' field
+    of the experiments_conf for the Executor. Furthermore, subclasses may
+    overwrite the .hp_stress member containing the configuration used while
+    generating the random hotplug stress sequence.
+
+    The test will be successful if the system does not halt or kill the
+    workloads because of hotplug stress. It means that the test passes if the
+    workload completes.
+    """
+
+    test_conf = {
+        "ftrace" : {
+            "events" : [
+                "sched_overutilized",
+                "sched_energy_diff",
+                "sched_load_avg_task",
+                "sched_load_avg_cpu",
+                "sched_migrate_task",
+                "sched_switch",
+                "cpu_frequency",
+                "cpu_idle",
+                "cpu_capacity",
+                "cpuhp_enter",
+                "cpuhp_exit",
+                "cpuhp_multi_enter",
+            ],
+        },
+        "modules": ["cgroups", "hotplug"],
+        "tools": ["rt-app", "trace-cmd"],
+    }
+
+    hp_stress = {
+        'seed' : None,                  # Seed of the random number generator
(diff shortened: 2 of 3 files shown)


tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+            cls.hp_stress['seed'] = random.randint(0, sys.maxint)
+        random.seed(cls.hp_stress['seed'])
+        cls.te._log.info('Random sequence of CPU Hotplug generated with: ')
+        cls.te._log.info(cls.hp_stress)
+        with open(os.path.join(cls.te.res_dir, 'hotplug_cfg.json'), 'w') as f:
+            json.dump(cls.hp_stress, f, sort_keys=True, indent=4)
+       
+        # Play with (online) hotpluggable CPUs only
+        cls.hotpluggable_cpus = filter(
+                lambda x: cls.target.file_exists(hp._cpu_path(cls.target, x)),
```
bjackman: Not sure about accessing a "private" member here. Could just duplicate `_cpu_path` here (could still use `HotplugModule.base_path`) as it's pretty trivial.

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+                "sched_switch",
+                "cpu_frequency",
+                "cpu_idle",
+                "cpu_capacity",
+                "cpuhp_enter",
+                "cpuhp_exit",
+                "cpuhp_multi_enter",
+            ],
+        },
+        "modules": ["cgroups", "hotplug"],
```
bjackman: I think you also need to put `'tools': ['ftrace']` here to ensure trace-cmd gets installed on the target. I also think that means you won't need the explicit `te.calibration()` below.

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+    @classmethod
+    def _experimentsInit(cls, *args, **kwargs):
+        # calibration shoud be done before starting spurious hotplugs
+        if not cls.te._calib:
+            cls.te.calibration()
+        
+        # Choose a random seed explicitly if not given
+        if 'seed' not in cls.hp_stress or cls.hp_stress['seed'] is None:
```
bjackman: `if cls.hp_stress.get('seed') is None:`

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+        > while true; do
+        >     echo 0 > /sys/devices/system/cpu/cpu1/online
+        >     echo "Plug off CPU 1" > /sys/kernel/debug/tracing/trace_marker
+        >     sleep 0.4245
+        >     echo 1 > /sys/devices/system/cpu/cpu1/online
+        >     echo "Plug in CPU 1" > /sys/kernel/debug/tracing/trace_marker
+        >     sleep 0.178
+        > done
+        '''
+        script_path = '/tmp/random_cpuhp.sh'
```
bjackman: Better use [`mkstemp` or something](https://docs.python.org/2/library/tempfile.html) rather than a hard-coded path

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+        length of the sequence (that is the number of plug operations) is 
+        defined in the hp_stress member of the class. The actual length of the
+        sequence might differ from the requested one by 1 because it's easier 
+        to implement and it shouldn't be an issue for most test cases.
+        '''
+        cur_on_cpus = cls.hotpluggable_cpus[:]
+        cur_off_cpus = []
+        max_cpus_off = cls.hp_stress['max_cpus_off']
+        i = 0
+        while i < cls.hp_stress['sequence_len'] - len(cur_off_cpus):
```
bjackman: `for i in range(cls.hp_stress['sequence_len'] - len(cur_off_cpus))`

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+            dst = cur_on_cpus if plug_way else cur_off_cpus
+            cpu = random.choice(src)
+            src.remove(cpu)
+            dst.append(cpu)
+            i += 1
+            yield cpu, plug_way
+
+        # Re-plug offline cpus to come back to original state
+        for cpu in cur_off_cpus:
+            yield cpu, 1
```
bjackman: This bit is unnecessary since it's done in `_experimentsFinalize`, no?

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+                cmd = 'echo {} > {}\n'.format(plug_way, sysfs_entry)
+                script.write(shift + cmd)
+                # Log in trace file
+                msg = 'Plug {} CPU {}'.format('in' if plug_way else 'off', cpu)
+                script.write(shift + 'echo "{}" > {}\n'.format(msg, marker))
+                # Sleep if necessary
+                if sleep_max > 0:
+                    sleep_dur_sec = random.randint(sleep_min, sleep_max)/1000.0
+                    script.write(shift + 'sleep {}\n'.format(sleep_dur_sec))
+            script.write('done\n') # end of while
```
bjackman: Outdated comment here

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+# limitations under the License.
+#
+
+import os.path
+import random
+import json
+import sys
+import time
+
+from test import LisaTest, experiment_test
```
bjackman: experiment_test isn't used

tests/eas/cpuhp.py
```diff
@@ -215,3 +215,27 @@ def _random_cpuhp_seq(cls):
...
+                },
+                'tasks' : 3,
+                'prefix' : 'many',
+            },
+        },
+    }
+       
+    @experiment_test
+    def test_random_hotplugs(self, experiment, tasks):
+        return True
```
bjackman: Can you add a docstring like "Test that the system doesn't crash while under hotplug stress"? Also no need to return True, just `pass`

tests/eas/cpuhp.py
qperret: Hmm it's not exactly the same thing ... Since the hotplug sequence runs in an infinite loop in the script, this bit here ensures that the second iteration will start from the same state as the first one. Without that it's likely that you will start the first iteration with all CPUs plugged in, and start the second iteration with some CPUs plugged off. I don't think anything will fail but it just feels cleaner to not re-plug off a CPU that's already offline.
In `_exeperimentsFinalize` we kill the script where it was, possibly in the middle of the hoplug sequence. So we just plug back in all CPUs so that things are clean on the target for future stuff but that's no more part of the test.

tests/eas/cpuhp.py
qperret: Yes you're right. I'll look into that, thanks

tests/eas/cpuhp.py
qperret: Yes that's what I had originally (a duplicate of `_cpu_path`), but it just felt cleaner not to re-do something that's already in devlib. The `HotplugModule.base_path` option is probably better though, I'll do the change

tests/eas/cpuhp.py
qperret: Since `len(cur_off_cpus)` changes as we iterate, I don't think this is equivalent

tests/eas/cpuhp.py
qperret: Of course :)

tests/eas/cpuhp.py
qperret: Ok cool ! Thanks

tests/eas/cpuhp.py
bjackman: Ah yes, sorry.

tests/eas/cpuhp.py
bjackman: Ah yes this is true, I didn't really realise it was running the random sequence in a loop. Why does it do that?

qperret: Just to be sure the hotplug stress never stops if the workload runs for a very long time. I can remove this loop if you prefer but that means the user will have to make sure the generated sequence is long enough to cover his test case

tests/eas/cpuhp.py
qperret: Actually we could remove `- len(cur_off_cpus)` and use a `for` loop instead. It will work just fine but it means that the number of hotplugs in the generated sequence might be less close to `hp_stress['sequence_len']` than it is with the current implementation. It's probably not an issue though.

tests/eas/cpuhp.py
qperret: And actually, the sequence will need to be long enough to cover _all iterations_ of the test, otherwise the last ones might run without hotplug pressure. It's maybe just simpler to leave it in a loop :)

bjackman: Ahh yes, sorry wasn't thinking carefully. Makes sense.

qperret:
Thinking about all this, I start wondering if the current design is the way to go. Maybe what we would like to have, instead of a dedicated test, is a _workload_ that hotplugs CPUs in and out randomly. The good thing is that it could be executed alongside any other workload, and it would be easy to extend all existing tests by simply adding a hotplug workload. Any thoughts about that ?

tests/eas/cpuhp.py
valschneider: Is there a consistent way to detect if the system crashed during the test ? i.e. are we sure all devlib commands being used have a timeout ? I believe there is a default 30s timeout on devlib's `execute()` method, but we should be extra careful here since ADB could just hang and the test would never end.

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+        # Kill hotplug script
+        pid = cls.target.read_int(cls.pid_file)
+        cls.target.kill(pid, as_root=True)
+        cls.hotplug_shell.kill() 
+        # Resume to original state
+        cls.te._log.info('Plugging back in currently offline CPUs')
+        cls.target.hotplug.online(*cls.hotpluggable_cpus)
+
+    @classmethod
+    def _random_cpuhp_script(cls):
```
valschneider: This looks like a job for libs/util/target_script (which is easy for me to say since I wrote it). As per our offline chat, that class will need a few modifications for your test, so do feel free to make it better.

tests/eas/cpuhp.py
valschneider: I think it might be better to keep the script in the results directory (which is what TargetScript does), so that it can be easily re-used/checked in case of failure. It's slightly less relevant here since with a given seed we should always get the same script, but I still think it's a good practice.

tests/eas/cpuhp.py
qperret: Good point. Looks like `target.check_responsive()` in devlib is a good fit for that

qperret:
Thanks guys for the review ! I think this can be useful as is. However, when we'll be able to run several workloads in parallel, we may prefer replacing this test by a hotplug workload that can be used in all existing tests. But I guess running several workloads is a different issue and should be treated separately.

libs/utils/target_script.py
```diff
@@ -89,7 +89,11 @@ def push(self):
         :type actions: list(str)
         """
 
-        actions = ['set -e'] + self.commands + ['set +e']
+        self._pid_file = self._target.path.join(self._target.working_directory,
```
valschneider: We already chatted about it offline, but I'll put it here as well: having to keep track of the process that is executing the script seems strange to me - in a perfect world, issuing a `kill()` via devlib should be enough.

qperret: Yes, I agree a single kill would be far better. Unfortunately, killing `self._bg_shell` stops the shell _on the host_, but the script keeps running on the target. TBH I'm not sure to really understand why, but I couldn't find any other way of solving the issue ...

bjackman: Just jumping in without context here.. can you use `Target.background`?

qperret: That's actually what I do (line 116), but `_bg_shell.kill()` (line 133) does not stop kill the script on the target, just the shell on the host :confused:

qperret:
I came up with something slightly cleaner for the `kill()` call in `TargetScript`. Let me know if it's OK.

qperret:
cc @credp

tests/eas/cpuhp.py
derkling: Still to be removed?

tests/eas/cpuhp.py
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+    @classmethod
+    def _experimentsInit(cls, *args, **kwargs):
+        # Choose a random seed explicitly if not given
+        if cls.hp_stress.get('seed') is None:
+            random.seed() 
+            cls.hp_stress['seed'] = random.randint(0, sys.maxint)
+        random.seed(cls.hp_stress['seed'])
+        cls.te._log.info('Random sequence of CPU Hotplug generated with: ')
```
derkling: Should we better use a local logger, which explicitly reports which class we are.
Something like:
   
```python
# Setup logging
        self._log = logging.getLogger('CpuHotplug')
```

... and just add the new channel `CpuHotplug` to the logger configuration:
https://github.com/ARM-software/lisa/blob/master/logging.conf
by basically copying/updating everything has TestEnv in it ;-)

tests/eas/cpuhp.py
qperret: Actually `experiment_test` is used at the bottom of the file.

tests/eas/cpuhp.py
qperret: This has been addressed with `'tools'; ['ftrace']`

tests/eas/cpuhp.py
qperret: Done

qperret:
I think all comments have been addressed here. I also added a "Torture" test that just plugs CPU in and out as fast as possible during 5 minutes. 
Ready to merge ?

PR #467, opened 2017-08-15
add two tasks to skip-freeze list for linux

stephenkyle-ARM:
Freezing these tasks is causing the kevin chromebook to reboot.


qperret:
I'm OK with this as well. Just by curiosity, what are these two tasks doing ? Are they required for all ChromeOS targets in general or just for the kevin chromebook ?

mdigiorgio:
I guess this can be merged.

PR #468, opened 2017-08-17
libs/utils/energy: add energy meters for gem5

qperret:
The `Gem5PowerInstrument` in devlib exposes continuous power samples during a gem5 simulation by extracting the output of platform-specific power models. This PR introduces an Energy Meter using these power samples to report an estimation of the energy consumption of a gem5 simulation.


qperret:
It looks like github re-ordered my commits without telling me :disappointed:. [I probably played with `rebase` too much](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/). So, obviously, the example notebook on how to use the Energy Meter comes after the commit adding it ...

valschneider:
Afaik if the order of your commits is correct locally, you shouldn't worry about it. The PR tab displays them chronologically to try and keep comments in a sane order, but the underlying git information is correct and they will be merged in the expected order.

qperret:
OK, that's what I was thinking but it's good to have confirmation ! The order is correct on my machine so it shouldn't be an issue.

ionela-voinescu:
It looks good to me! :+1:

libs/utils/energy.py
```diff
@@ -509,4 +511,30 @@ def report(self, out_dir, out_energy='energy.json'):
...
+        # spuriously on OPP changes. Let's use the time channel provided by the 
+        # gem5 power instrument to build the timeline accordingly. 
+        for site, measure in df:
+            if measure == 'time':
+                meas_dur = df[site]['time']
+                break
+        timeline = np.zeros(len(meas_dur))
+        # The time channel gives the elapsed time since previous measurement
+        for i in range(1, len(meas_dur)):
+            timeline[i] = meas_dur[i] + timeline[i-1]
```
valschneider: Tiny nitpick here, dunno if it's a proper Python thing, but I find it better when there are whitespaces around arithmetic operands, i.e. `timeline[i - 1]`

qperret: Done

qperret:
Thanks for the review guys !

> In [6]: Since this is an example notebook, I think it would be good to describe the workload a bit (either in comment or in markdown text), as well as what we're trying to achieve with the workload (here it's just something to run to get energy numbers).

Reorganized this section a little bit and added a description.

> In [8] and beyond: The channel names are explicit enough, but maybe it could be worth writing a bit about what you're doing (summing the channels and such).

Added few comments here and there but not sure if that's enough. Let me know if it's not :smile:.

PR #469, opened 2017-08-17
libs/utils/android/screen: Add screenrecord utility

valschneider:
Implementation isn't the prettiest thing ever. I hit the same problem as @qperret in that I have to send signals to the process that is spawned on the device, and not the ADB command that resides on the host.

I put the code in the Screen class, but with how that class is used we don't have any context, only static classes, so the way to use screenrecord is:

```
data = record_start(target, ...)
<dostuff>
record_stop(target, data)
```

I find that absolutely barbaric, but I also didn't want to put it in the Workload class as it wouldn't be available outside of android workloads. If you guys think it is fine to move it in the Workload class, that'll get us rid of that ugly data passing.

### DIFF
--- a/libs/utils/android/screen.py
+++ b/libs/utils/android/screen.py
@@ -16,8 +16,10 @@
 #
 
 import logging
+
 from system import System
 from time import sleep
+import os
 
 class Screen(object):
     """
@@ -154,5 +156,85 @@ class Screen(object):
        sleep(1)
        System.menu(target)
        System.home(target)
+
+    class Recorder(object):
+        """
+        Android screenrecord utility class
+        """
+
+        def __init__(self, target):
+            self._target = target
+            self._log = logging.getLogger('Recorder')
+
+            self._recording = False
+
+        def start(self, out_file, timestamp=True, max_time=None):
+            """
+            Start recording the screen of the device
+            """
+            if self._recording:
+                raise RuntimeError('Recording already started')
+
+            self._local_file = out_file
+            self._remote_file = self._target.path.join(
+                self._target.working_directory,
+                os.path.basename(out_file)
+            )
+
+            self._max_time = max_time
+
+            options = []
+            if timestamp:
+                options.append('--bugreport')
+            if max_time:
+                options.append('--time-limit {}'.format(max_time))
+
+            record_cmd = 'screenrecord {} {}'.format(
+                ' '.join(options), self._remote_file
+            )
+
+            self._log.debug('screenrecord command = \"{}\"'.format(record_cmd))
+            self._log.info('Starting screen recording')
+
+            self._recording = True
+            self._process = self._target.background(record_cmd, as_root=True)
+
+        def _finalize(self):
+            self._process.wait()
+
+            self._target.pull(self._remote_file,
+                              self._local_file,
+                              as_root=True)
+
+            self._recording = False
+            self._log.info('Recording available at {}'.format(self._local_file))
+
+        def wait(self):
+            """
+            Wait for the recording to end
+            """
+            if not self._recording:
+                raise RuntimeError('Recording not started')
+
+            if not self._max_time:
+                self.stop()
+                raise RuntimeError('max_time not set, stopped Recorder')
+
+            self._log.info('Waiting for recording to end ({}s total)'.format(self._max_time))
+            self._finalize()
+
+        def stop(self):
+            """
+            Force stop recording the screen of the device
+            """
+            if not self._recording:
+                raise RuntimeError('Recording not started')
+
+            # Recording must be ended by Ctrl-C on the device
+            self._log.debug('Sending Ctrl-C to screenrecord')
+            self._target.execute('kill -INT $(pgrep screenrecord)', as_root=True)
+
+            self._finalize()
+
 
 # vim :set tabstop=4 shiftwidth=4 expandtab


libs/utils/android/screen.py
```diff
@@ -155,4 +157,57 @@ def unlock(target):
...
+            options.append('--time-limit {}'.format(max_time))
+
+        remote_file = target.path.join(target.working_directory,
+                                       os.path.basename(out_file))
+
+        record_cmd = 'screenrecord {} {}'.format(
+            ' '.join(options), remote_file
+        )
+
+        target.adb_root(enable=True, force=True)
```
qperret: Do we need that ? What about using `target.background(record_cmd, as_root=True)` on line 189 instead ? If it's functionally equivalent, I think this would be _slighlty_ faster since it won't restart the adb server. Not a big improvement though :)

libs/utils/android/screen.py
```diff
@@ -155,4 +157,57 @@ def unlock(target):
...
+    @staticmethod
+    def record_stop(target, screenrecord_data):
+        """
+        Stop recording the screen of the device
+        """
+        log = logging.getLogger('Screen')
+
+        if not screenrecord_data['max_time']:
+            # Recording must be ended by Ctrl-C on the device
+            target.execute('kill -INT $(pgrep screenrecord)')
```
qperret: `target.kill('$(pgrep screenrecord)')` ?

libs/utils/android/screen.py
valschneider: I actually had to put that there to get it to work, but that might be due to my device that is being very flaky. I'll have a go on another device.

libs/utils/android/screen.py
valschneider: Didn't look into the details, does it send a SIGINT by default ? Afaik screenrecord has a signal handler that intercepts SIGINT to gracefully end the video. Simply killing the process will end up in a botched video file.

valschneider:
@qperret I didn't want to create another class as, after all, this is a screen-related utility. But I think you're right, a non-static class in screen.py is probably the cleanest solution.

qperret:
Maybe a `Recorder` class in `Screen` ? You could the use it like:
```python
rec = Screen.Recorder(target)
rec.start()
# do stuff
rec.stop()
```
I agree putting members to the Screen class is very bad but using it to encapsulate a class like that is OK I think.

libs/utils/android/screen.py
qperret: You can define a `signal` parameter in `target.kill` and I think the default is `SIGTERM`. If you want it more violent you can change it to `SIGKILL` I suppose.

valschneider:
@bjackman || @derkling, as LISA senseis what do you think ?

libs/utils/android/screen.py
valschneider: Mmm `target.kill('$(pgrep screenrecord)')` assumes that the first argument will be expanded in a shell - the only thing specified in Devlib's doc is that the first argument is "the pid to kill", so this may change... And sadly `target.killall()` doesn't seem to work in this situation, so I think I'll stick to `target.execute()`

valschneider:
Moved the code to Screen.Recorder to get rid of that ugly context passing

libs/utils/android/screen.py
qperret: Ah yeah, you're right. If you want to keep the target abstraction you can probably use `target.get_pids_of()` before `target.kill()` though. However, my previous tests with `get_pids_of()` were not very successful (`ps` is not always behaving as expected on Android)  and, if this is really Android-specific, it's probably OK to leave it as it is

libs/utils/android/screen.py
```diff
@@ -155,4 +157,70 @@ def unlock(target):
...
+            if not self._recording:
+                raise RuntimeError('Recording not started')
+
+            if not self._max_time:
+                # Recording must be ended by Ctrl-C on the device
+                self._log.debug('Sending Ctrl-C to screenrecord')
+                self._target.execute('kill -INT $(pgrep screenrecord)', as_root=True)
+
+            self._log.info('Waiting for recording to be completed')
+            self._process.wait()
```
qperret: If I understand correctly we should really wait here only if `max_time` was specified is that correct ? If yes, I think `stop()` should always end the record immediately and the wait semantics should be provided by another method. That way, a user could do `start() ... wait()` to record the screen for certain amount of time or `start() ... stop()` in an event-based fashion with `max_time` just serving as a timeout.

valschneider: Actually the `wait()` is needed in both cases - in one case we're waiting for the timer to end, in the other we're waiting for the SIGINT to be processed (and the recording file to be finalized). As such I think `start()` and `stop()` work well enough.

qperret: Ok, sorry, I was unclear. Let me clarify:

> we're waiting for the SIGINT to be processed

Indeed, but that _should_ be quick, and it's part of a normal `stop()` procedure. When I was saying "we should really wait" I meant that we wait for the recorder to timeout, which can be very long (potentially).

I see two different semantics that should be decorrelated. `Record` should provide 3 methods: `Recorder.start()`, `Recorder.stop()` and `Recorder.wait()`. `Recorder.stop()` should end the recorder immediately, meaning that it will issue a SIGINT unilaterally, without caring about `max_time`,  and use `self._process.wait()` to do that cleanly. On the other hand, `Recorder.wait()` should do nothing else but wait for `max_time` to elapse. That way, we can either:

- use `stop()` to stop the recorder as soon as a workload completes for example (in which case `max_time` will just serve as a timeout ensuring that  we don't record forever if the workload bugs and never ends)
- use `wait()` to record the screen for a specific amount of time (specified by `max_time`).

What do you think ? I guess you already have all the stuff you need to implement that, it's just a bit of refactoring :smiley:

valschneider: > - use stop() to stop the recorder as soon as a workload completes for example (in which case max_time will just serve as a timeout ensuring that we don't record forever if the workload bugs and never ends, causing stop() to never be called)
> - use wait() to record the screen for a specific amount of time (specified by max_time).

I see. I don't really know what use case would mandate blocking on the recording timer, but it doesn't cost much to add. I just don't really like the fact that calling `wait()` with no `max_time` doesn't have any meaning.

valschneider:
Added `wait()` method. As @qperret suggested, `stop()` always stops the recording whereas `wait()` waits for `max_time` to be elapsed.

If no `max_time` is defined, a warning is fired and we'll wait for the end of times, or for the device's disk to get full, whichever comes first. Not the greatest thing, but like I said, I don't know what to do of that `wait()` in this case.

qperret:
We can probably do 2 things about that:

1. use a default `max_time` with a large value. Not sure what large means though (1 hour ?)
2. if we really don't have a `max_time` because it was explicitly set to None, `stop()` then raise an exception with a message such as: "max_time must be set to use wait(). Recorder has stopped()"

qperret:
Thinking about if, just the second option is probably the better.

libs/utils/android/screen.py
qperret: As mentioned below, I agree this is bad, and as, such we should raise an exception with an explanatory message. Moreover, it should be clearly documented that calling `wait()` without a `max_time` value will result in an error.

bjackman:
Quick Q - any reason why we can't put this in devlib as in Instrument? Then we get it for WA3 too...

qperret:
Hmmm very good point. I think it makes sense. And actually it could even make sense to move all the content of `screen.py` into devlib. What do you think @valschneider ?

valschneider:
Yeah about that I had a chat with @derkling - thing is, there's quite a lot of code (like 90%) we have in `system.py` and `screen.py` that could make its way over to Devlib, but I don't really know how much of it all we can move over to WA nor how much we should move over.

bjackman:
Just from skimming the code this bit doesn't seem to have any functional dependencies so we could always start by just doing this in devlib? Should also get @setrofim's view on whether it belongs in devlib/what form it should take.

valschneider:
@bjackman should we close this ? I think it does make more sense to have it in devlib.

bjackman:
Yeah I agree

PR #471, opened 2017-08-24
Add test for NOHZ load updates

bjackman:
This depends on https://github.com/ARM-software/lisa/pull/406 so is not ready for merge yet, but opening it as a placeholder


valschneider:
Will steal this in my own future nohz PR, closing it for now.

PR #473, opened 2017-09-12
Exoplayer

valschneider:


libs/utils/android/system.py
```diff
@@ -464,4 +465,31 @@ def packages_info(target, apk_filter=''):
             return packages
         return None
 
+    @staticmethod
+    def grant_permissions(target, package):
```
valschneider: Nightly thought: this should probably be moved to devlib's android utilities to lessen the work involved in moving all of the System/Screen stuff to devlib (as mentionned in #469

libs/utils/android/system.py
```diff
@@ -464,4 +465,31 @@ def packages_info(target, apk_filter=''):
...
+            'requested permissions:\s*(?P<permissions>(android.permission.+\s*)+)', dumpsys
+        )
+
+        permissions = permissions.group('permissions').replace(" ", "").splitlines()
+        
+        for permission in permissions:
+            try:
+                target.execute('pm grant {} {}'.format(package, permission))
+            except:
+                # Some permissions cannot be granted, ignore them
```
mdigiorgio: Does the app run if you don't grant a permission?

libs/utils/android/workloads/exoplayer.py
```diff
@@ -0,0 +1,184 @@
...
+        self.tracingStart()
+        self._log.info('Playing media file')
+
+        line = monitor.wait_for(REGEXPS['duration'])[0]
+        duration = int(round(float(re.search(REGEXPS['duration'], line)
+                                   .group('duration'))))
+
+        self._log.info('Media duration is {}'.format(duration))
+
+        if self.max_duration and self.max_duration < duration:
```
mdigiorgio: I find `max_duration` and `duration` a bit confusing. I'd suggest a simple renaming, something like:

`duration` -> `media_duration_s` (I'd also put the s at the end to say it's in seconds)
`max_duration` -> `play_duration_s` or `play_period_s`

libs/utils/android/system.py
valschneider: Yes, you just get stuck with a permission request overlayed on top of the app.

mdigiorgio: OK, then if you add this to `devlib` it's probably good to log a debug message saying that a certain permission wasn't granted.

valschneider:
Submitted https://github.com/ARM-software/devlib/pull/169, will update this PR once it's merged.

libs/utils/android/workloads/exoplayer.py
```diff
@@ -0,0 +1,184 @@
...
+
+REGEXPS = {
+    'start'    : '.*Displayed com.google.android.exoplayer2.demo/.PlayerActivity',
+    'duration' : '.*period \[(?P<duration>[0-9]+.*)\]',
+    'end'      : '.*state \[.+, .+, E\]'
+}
+
+class Exoplayer(Workload):
+    """
+    Android Exoplayer workload
```
bjackman: Might be useful to add a note explaining how to get hold of the APK?

libs/utils/android/workloads/exoplayer.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+
+    # Package required by this workload
+    package = 'com.google.android.exoplayer2.demo'
+
+    def __init__(self, test_env):
+        super(Exoplayer, self).__init__(test_env)
+        self._log = logging.getLogger('Exoplayer')
```
bjackman: I think you have to add this logger to logging.conf in the root dir?

libs/utils/android/workloads/exoplayer.py
```diff
@@ -0,0 +1,184 @@
...
+        self._log.debug('Workload created')
```
bjackman: Noise, IMHO

libs/utils/android/workloads/exoplayer.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def _play(self):
+
+        # Grant app all permissions
+        System.grant_permissions(self._target, self.package)
```
bjackman: Maybe we should just have the default install method do this?

libs/utils/android/workloads/exoplayer.py
```diff
@@ -0,0 +1,184 @@
...
+            - 'ftrace'
+            - any combination of the above as a single space-separated string.
+        :type collect: list(str)
+
+        :param media_file: Filepath of the media to play
+            Path on device if 'from_device' is True
+            Path on host   if 'from_device' is False (default)
+        :type media_file: str
+
+        :param from_device: Wether file to play is already on the device
```
bjackman: whether

libs/utils/android/workloads/exoplayer.py
valschneider: Ah yes, that is something I wanted to ask - what's the process for adding an APK to LISA ? It's all built from https://github.com/google/ExoPlayer (after some investigation with Chris we realized we could just use the default 'demo' application).

libs/utils/android/workloads/exoplayer.py
valschneider: Looks like this is something we have in all of our workloads - e.g. Jankbench:
```
self._log = logging.getLogger('Jankbench')
```
Which always requires changing the `logging.conf` file. Maybe we could just use the 'Workload' logger ?

libs/utils/android/workloads/exoplayer.py
valschneider: You mean devlib's install ? Afaik this has to be done every time before launching the app if we clear the app data (`System.force_stop(self._target, self.package, clear=True)`), which we want to do to make sure the app doesn't try to launch a previously-run media file.

bjackman: Ah yep makes sense, OK.

libs/utils/android/workloads/exoplayer.py
```diff
@@ -0,0 +1,183 @@
...
+            Path on device if 'from_device' is True
+            Path on host   if 'from_device' is False (default)
+        :type media_file: str
+
+        :param from_device: Whether file to play is already on the device
+        :type from_device: bool
+
+        :param play_duration_s: If set, maximum duration (seconds) of the media playback
+                                If not set, media will play to completion
+        :type max_duration: int
```
mdigiorgio: s/max_duration/play_duration_s/

mdigiorgio:
Could you please re-update devlib? here is a new commit that adds something to `grant_permissions`.

valschneider:
@mdigiorgio Done + little fixup

bjackman:
Just tested this against a hikey960, looks good.

mdigiorgio:
Looks okay to me as well, go fot it ;)

bjackman:
> Just tested this against a hikey960, looks good.

(The _workload_ looks good, that is.. the Hikey960 overheated :wink: )

PR #474, opened 2017-09-13
Add DataFrame getter for CPU wakeups

bjackman:


libs/utils/analysis/cpus_analysis.py
```diff
@@ -63,6 +63,26 @@ def _dfg_context_switches(self):
         ctx_sw_df.index.name = 'cpu'
         return ctx_sw_df
 
+    def _dfg_cpu_wakeups(self, cpus=None):
+        """"
+        Get irq_handler_entry events that show when a CPU was woken from idle
+
+        :param cpus: List of CPUs to find wakeups for. If None, all CPUs.
```
mdigiorgio: Add `:type cpus: list(int)`

bjackman:
Hmm, on further investigation this is not a good approach to counting CPU wakeups - you don't see softirqs (e.g. the scheduler's nohz balancer kick doesn't show up in this).

Better to just use the `cpu_idle` event (dunno why I didn't think of that, we already have stuff to parse it nicely too). That does mean you don't get the nice info about what caused the wakeup  - the current implementation can give you a clue that lots of the wakeups were caused by a certain interrupt. But that's something to think about another time.

bjackman:
OK uploaded a new version using 'cpu_idle'

libs/utils/analysis/cpus_analysis.py
```diff
@@ -63,6 +63,32 @@ def _dfg_context_switches(self):
         ctx_sw_df.index.name = 'cpu'
         return ctx_sw_df
 
+    def _dfg_cpu_wakeups(self, cpus=None):
+        """"
+        Get irq_handler_entry events that show when a CPU was woken from idle
```
mdigiorgio: `cpu_idle`

bjackman:
Fixed the doc comment and also I realised it was broken because of a missing call to `sort_index` - fixed that and altered the test so it would fail otherwise.

derkling:
Interesting, just one small comment in the patch organization... otherwise this LGTM.

mdigiorgio:
@derkling I think we can merge this. What was your comment? I can't find it. Has it been addressed yet?

tests/lisa/test_trace.py
```diff
@@ -52,6 +52,13 @@ def _get_platform(self):
         with open(os.path.join(self.traces_dir, 'platform.json')) as f:
             return json.load(f)
 
+    def make_trace(self, in_data):
```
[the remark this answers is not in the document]
bjackman: Hmm yeah I just noticed this patch isn't needed at all, it was already merged in a separate PR! Will drop it.

PR #475, opened 2017-09-14
libs/utils/android/workloads: Add PCMark workload

valschneider:
It started off from @ionela-voinescu's #321 but ended up quite different.

So far only the "work" benchmark is supported, as I haven't found a clean way (yet) to start the other benchmarks.

Once the benchmark is done, a .zip archive is pulled from the device, scores are extracted from a .xml file and finally dumped in a .json file.

### DIFF
--- a/libs/utils/android/workloads/pcmark.py
+++ b/libs/utils/android/workloads/pcmark.py
@@ -1,0 +1,181 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2017, Arm Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import re
+import os
+import logging
+import time
+import json
+
+from subprocess import Popen, PIPE
+from zipfile import ZipFile
+
+from android import Screen, System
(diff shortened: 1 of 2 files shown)


libs/utils/android/workloads/pcmark.py
```diff
@@ -0,0 +1,176 @@
...
+        monitor = self._target.get_logcat_monitor(REGEXPS.values())
+        
+        # Start PCMark on the target device
+        System.start_activity(self._target, self.package, self.activity)
+        # Wait a few seconds while application is being loaded
+        time.sleep(5)
+        
+        if test_name.lower() == 'work':
+            # Move to benchmark run page
+            self._target.execute('input keyevent KEYCODE_TAB')
```
valschneider: I just remembered that I added these bits in System back in the day of https://github.com/ARM-software/lisa/pull/354/commits/65402030cffe1dde4513b34ea9adf290c2f047d2, but it might be best not pile more stuff up on that class.

derkling:
LGTM

Didn't tested it but looks interesting. AFAIU you are not using touch/input events to trigger the workload execution, isn't it?

valschneider:
@derkling Just enter/tab key presses, which should hopefully be device agnostic. I can't seem to be able to trigger any other test this way though (PCMark has several workloads), and I don't want to use device-specific touch events.

On a sidenote, WA's revent functionality could make those key presses cleaner - but that's for another time :)

valschneider:
Updated the pydoc to mention that the workload (within the app) has to be installed manually - this is only a one time thing though.

joelagnel:
thanks!

joelagnel:
FYI...

libs/utils/android/workloads/pcmark.py
valschneider: Do you want me to grab that patch and add it to this PR ? It used to live in the Corporate Geekbench PR but it has been closed since there is support in wltest.

valschneider:
* Added the commit that gives tab/enter/dpad wrappers to System.
* Modified workload code to use these wrappers.

PR #479, opened 2017-09-22
platforms: Add target-derived platform info for hikey960

bjackman:
This makes it easier to do platform-specific trace analysis when not connected
to a target.


valschneider:
LGTM

derkling:
Apart for a small line-saving request, that LGTM too and can be merged.

PR #480, opened 2017-09-22
Add thermal plotting in analysis/thermal_analysis

valschneider:
Depends on https://github.com/ARM-software/trappy/pull/269, the trappy submodule is temporarily updated to my local branch for testing purposes.


derkling:
Just added some comments on the referenced trappy PR, have a look there ;-)

libs/utils/plot_utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+        plot_dfr = dfr[dfr[name_field] == name][fields]
+        plot_dfr.columns = cols
+        plot_dfr.plot(ax=ax, drawstyle=drawstyle)
+
+    if setup_plot:
+        ax.set_title(title)
+
+    if xlim is not None:
+        ax.set_xlim(xlim)
```
valschneider: Maybe this could be added to the AnalysisModule itself, so that this can directly use `trace.x_min` & `trace.x_max` ?

derkling:
Did not tested, and I don't think I'll have time to test it, however overall it LGTM.
One main comment is about the possibility to keep the API as it is, but possibly improve the consistency with other analysis APIs. I'm referring for example to the xlim constraints.

Who is the user of the plot_utils.py APIs?

It seems that only the ThermalAnalysis module uses these calls and being a kind-of internal API of TRAPpy, I doubt there are customers which depends on that call. IOW, the plotting functions exposed by the new ThermalAnalysis module should be just good enough to provide a backward compatible API for TRAppy.

Another possibility is perhaps that to move these two functions inside the Analysis base class.
At the end they provide a pretty generic API to plot a DataFrame... which perhaps matches some of the use-cases we have in other analysis modules... thus it would open possibilities for a further refactoring and consolidation of existing analysis modules plot functions.

valschneider:
> Another possibility is perhaps that to move these two functions inside the Analysis base class.
Even though I'm not too hot to do yet another round of refactoring, I do agree with you on that.

valschneider:
> One main comment is about the possibility to keep the API as it is, but possibly improve the consistency with other analysis APIs. I'm referring for example to the xlim constraints.

How so ? If I take for instance `plotCPUFrequencies()`, it's done like this:
```
axes.set_xlim(self._trace.x_min, self._trace.x_max)
```
The difference with the thermal plots is that this is done in `plot_generic()`

derkling:
But plot_generic has xlim as a by default None parameter... thus, if plot_generic is a function of AnalysisModule, then you have a trace on hand to clamp the xlim values the user specify.

valschneider:
Ok yes, I agree with you - it'd make more sense to have that in the AnalysisModule directly.

libs/utils/trace.py
```diff
@@ -629,6 +631,44 @@ def _sanitize_SchedOverutilized(self):
...
 
+    def _sanitize_ThermalTemperature(self):
+        if not self.hasEvents('thermal_temperature'):
+            return
+
+        df = self._dfg_trace_event('thermal_temperature')
+
+        # Temp scale is in milliCelcius
+        temp = df['temp'] / 1000
+        df.loc[df.index, 'temp'] = temp
```
bjackman: - Isn't this `df['temp'] /= 1000`?
- What's wrong with milliCelcius anyway? Changing units just seems like a way to confuse people.

libs/utils/analysis/thermal_analysis.py
```diff
@@ -0,0 +1,243 @@
...
+
+###############################################################################
+# DataFrame Getter Methods
+###############################################################################
+
+    def _dfg_thermal_zone_temperature(self, ids=None):
+        """
+        Get the temperature readings of one or more thermal zone(s).
+
+        :param ids: The thermal zones to consider
```
bjackman: Might want to mention default is all zones

libs/utils/analysis/thermal_analysis.py
```diff
@@ -0,0 +1,243 @@
...
+        :type device: list(str)
+        """
+        df = self._dfg_trace_event('thermal_power_devfreq_limit')
+        df = df[['type', 'freq', 'cdev_state']]
+
+        if devices is not None:
+            df = df[df.type.isin(devices)]
+
+        return df
+
```
bjackman: I don't really see the utility of these `_dfg`s, they just return a pretty trival-to-acquire subset of the full DataFrame, right?

libs/utils/analysis/thermal_analysis.py
```diff
@@ -0,0 +1,243 @@
...
+        plot_df = self._dfg_thermal_zone_temperature(thermal_zones)
+
+        def stringify_tz(id):
+            return plot_df[plot_df.id == id]['thermal_zone'].unique()[0]
+
+        plot_generic(plot_df, 'id', names = thermal_zones,
+                     fields = ['temp'], prettify_name = stringify_tz,
+                     xlim = (self._trace.x_min, self._trace.x_max),
+                     drawstyle = 'steps-post', ax = ax
```
bjackman: Can we make the nomenclature here match trappy's ILinePlot API? Also, no spaces around `=` inside parameter lists please.

libs/utils/analysis/thermal_analysis.py
```diff
@@ -0,0 +1,243 @@
...
+        )
+
+        if thermal_zones is None:
+            suffix = ''
+        else:
+            suffix = '_' + '_'.join(map(str, thermal_zones))
+
+        # Save generated plots into datadir
+        figname = '{}/{}thermal_temperature{}.png'\
```
bjackman: `os.path.join`. Also you can avoid the `\` by putting the whole thing in brackets like

```
   foo = ('bar{}'
          .format('baz'))

File: libs/utils/analysis/thermal_analysis.py [hunk 7]
```diff
@@ -0,0 +1,243 @@
...
+            suffix = ''
+        else:
+            suffix = '_' + '_'.join(map(str, cpus))
+
+        # Save generated plots into datadir
+        figname = '{}/{}thermal_cpufreq_cdev_state{}.png'\
+                  .format(self._trace.plots_dir, self._trace.plots_prefix, suffix)
+        pl.savefig(figname, bbox_inches='tight')
+
+    def plotDevfreqCoolingStates(self, devices = None, ax = None):
```
bjackman: Ditto about spaces

libs/utils/plot_utils.py
```diff
@@ -0,0 +1,138 @@
+# SPDX-License-Identifier: Apache-2.0
```
bjackman: Is this file copied directly from trappy? If so can you say so in the commit message. If not, if possible, could you copy the file exactly from trappy (with a commit message showing exactly where it came from) then make any changes necessary in a separate commit?

valschneider: If it's going into analysis_module, how would I go about that ? Add a commit that adds the file somewhere, then a commit that squashes that in the file + with the modifications ?

libs/utils/analysis/thermal_analysis.py
valschneider: To be honest I think only the cpufreq one is useful since we juggle cpulists & cpumasks around. For the other two, I figured a one-line accessor could always be nice to have, but if I'm wrong I can evince those.

libs/utils/trace.py
valschneider: > Isn't this df['temp'] /= 1000?
Yeah totally. I kept getting a SettingWithCopy warning, and even though that warning was a false positive, I tried to make sure it didn't get fired - that `loc` thing was the cleanest solution I found.

Now, I've re-run the code that directly divides temperature and I don't get that warning anymore...

> What's wrong with milliCelcius anyway?
That's how the data was handled in TRAPpy, I didn't look any further and reproduced it here.

libs/utils/analysis/thermal_analysis.py
valschneider: Do you want me to do a pass around the other plotting methods ? I copied the boilerplate code from frequency_analysis but it looks like it's done this way in the other plotting methods as well.

libs/utils/analysis/thermal_analysis.py
valschneider: > Can we make the nomenclature here match trappy's ILinePlot API?

Do you mean matching the order of the parameters, or going the `**kwargs` route ?

libs/utils/plot_utils.py
bjackman: Just have a commit like "copy file X from trappy" then a commit that makes the changes. Not sure what you mean by "squash" there...

valschneider: Mm probably not the best choice of words - @derkling suggests to move those utility functions in AnalysisModule, that way the `xlim` thing can be handled directly with `axes.set_xlim(self._trace.x_min, self._trace.x_max)`. This means we don't have to pass `xlim` to `plot_generic()`

Correct me if I'm wrong, but in that case I believe the proper course of action would be:
* Have a commit that says "copy file X from TRAPpy"
* Have another commit that says "move utility methods from file X to AnalysisModule"

bjackman: Ahh I see, OK then the two-commit thing probably isn't worth the hassle.

libs/utils/analysis/thermal_analysis.py
bjackman: Oh OK, it doesn't matter too much really.

libs/utils/analysis/thermal_analysis.py
bjackman: I mean the actual names of the parameters - see the penultimate paragraph of [this comment](https://github.com/ARM-software/trappy/pull/269#discussion_r140769563)

libs/utils/analysis/thermal_analysis.py
bjackman: Oh yeah, actually the cpufreq one _is_ useful, and it seems fair to have the devfreq one for symmetry. I don't see the point of the temperature one, but it doesn't really hurt either 🤷‍♂️

libs/utils/analysis/thermal_analysis.py
valschneider: Ah gotcha, skipped that one when I closed the PR

valschneider:
* Updated the `plot_generic()` parameters to match TRAPpy's ILinePlot (only difference is `column**s**` instead of `column` because it really irked me to have a singular variable name that designates a list of things)
* Fixed a bug in `plot_generic`
* Overall cleanup (stray whitespaces & pydoc)

libs/utils/trace.py
derkling: I like better the human readable Celsius... this I'm all for the TRAPpy proposed conversion.

derkling:
Regarding the plot_generic API, since this is a named parameter, if you really wanna be backward compatible, you should probably keep a `column` parameter, which is internally copied into the new `columns` one together with a deprecation warning to inform the user that the misspelled `column` one is going to be removed soon.

Apart from that, although I did not tested it, this LGTM and it can be merged. ;-)

libs/utils/trace.py
valschneider: Argh, I removed it in my last push :(

valschneider:
>  if you really wanna be backward compatible

Well `plot_generic` has only ever lived in pull-requests for now, should I really care about backward compatibility ? Or do you mean compatibility with ILinePlot's API ?

bjackman:
I think Patrick is talking about compatibility with the `ILinePlot` API - bit for me this is not about compatibility, it's just about using the same terminology as other code to improve clarity.

libs/utils/trace.py
bjackman: Hmm, I really don't like the idea of changing units for readability when it affects internal code too - if millicelcius is hard to read then that should be solved when presenting the data (i.e. when you plot it) IMO. But if TRAPpy was already doing it, then let's keep it the same.

valschneider: > But if TRAPpy was already doing it, then let's keep it the same.

Poor choice of words on my part, sorry - the internal dataframe wasn't modified, the milliCelcius -> Celcius conversion was only done for plotting. To be honest, I don't think milliCelcius is that bad - besides, everything temperature-related is in this unit scale (IPA trip points, sensor feedback...)

bjackman: > Celcius conversion was only done for plotting

Oh, if that's true then I didn't read the code properly and I'm happy with the conversion.

valschneider: > Oh, if that's true then I didn't read the code properly and I'm happy with the conversion.

This is a wonderful case of "quiproquo" - I was describing how it was done in TRAPpy, not how I was doing it.
- TRAPpy: only convert to Celcius for plotting
- First version of this PR: Convert internal df to Celcius (affects plotting & analysis)
- Suggested course of action: Don't even convert even for plotting (but maybe add a 'milliCelcius' ylabel).

valschneider:
- Renamed that pesky `columns` parameter to `column`
- Changed `plot_generic` and `plot_setup` to private methods (they were showing up in the autocompletion of trace.analysis.X.*)

derkling:
On 28-Sep 02:14, Brendan Jackman wrote:
> I think Patrick is talking about compatibility with the `ILinePlot`
> API

I think I was a bit confused on this part, I was thinking that the
plot_generic API was pre-existing in Trappy mainline code with a
"column" parameter... while this is not the case but just a new
function added by this PR.

Thus, there are not needs of backward compatibility...

> - bit for me this is not about compatibility, it's just about
> using the same terminology as other code to improve clarity.

... and not even compatibility with the ILinePlot, which is another
API.

Regarding the terminology, I would say that the correct name is what
Valentin is proposing, i.e. columnS. The ILinePlotter has been
extended to accept a list from the original version accepting a single
value. At that time the parameter should have been renamed (I don't
think there was such a massive user-based to justify the backward
compatibility) or otherwise the naming schema I was proposing in the
previous comment should have been applied.

Thus, personally I don't care to keep consistency with something which
is (unfortunately) wrongly named in a different project.

From my side: +1 for the current proposal.

-- 
#include <best/regards.h>

Patrick Bellasi

valschneider:
@bjackman Your thoughts on that ? I can live with either option, but we need a consensus :)

bjackman:
Oh sorry, miscommunication again - I wasn't complaining about 'column' vs 'columns', I agree that 'columns' is better. Just wanted 'filters' and 'pivot' to have the same name. Not worried about identifiers being exactly the same, I just want to avoid having two totally different words for the same concept.

valschneider:
Ah fair enough. I'll re-update the PR with `columns` - we'll get there eventually :)

PR #481, opened 2017-09-27
trace: Drop NaNs from CPU active signal in getClusterActiveSignal

bjackman:
Some CPUs' active signals may begin earlier than others. In that
case, despite the fillna, we may still have NaNs, which break the
reduce() call. Therefore just drop rows with NaNs.

### DIFF
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -847,6 +847,10 @@ class Trace(object):
             )
 
         active.fillna(method='ffill', inplace=True)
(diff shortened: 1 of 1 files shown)


mdigiorgio:
I think it's safe to merge this, but I would suggest to add a short comment in the code to say why you're dropping NaNs after the forward filling.

bjackman:
Oops, forgot about this, have updated now.

PR #484, opened 2017-09-29
Add WA3 results collector and example notebook

bjackman:


bjackman:
So this has somehow ballooned into 800 lines of code, but there _are_ quite a lot of comments... 

I think it can be substantially reduced by using WA's API for doing this, but for now it works and it's probably handy not to have the dependency, so that you can anlyse results without having WA3 installed.

I've added the `get_artifacts` method, which you can use to get hold of a trace for a specific job, it looks like:

```
trace_path = collector.get_artifact('trace-cmd-bin', workload='jankbench', tag='lt=pelt', test='list_view', kernel='foo')
```

I added this because, when playing with this tool, I found that when you find a result like "kernel X is way better on metric Y" your first thought is to look at an example trace for kernel X, so this makes that easier to find.

This can also be used to get things like the Jankbench results DB or the energy samples CSV, in case you want to do analysis that can't be done with the overall results DataFrame.

derkling:
Here are some patches on top of this last your version:
   https://github.com/derkling/lisa/commits/wa3-collector

PR #485, opened 2017-10-02
frequency_analysis: Fix calculating avg_freq

bjackman:
By default, area_under_curve uses method='trapz' which implicitly
does a linear interpolation of the data. This is not desirable for
cpu_frequency, which changes ~instantly and stays constant until the
next data point.


bjackman:
@derkling This popped into my head while working on the WA3 comparison thing.. do you think it's right? We have other uses of area_under_curve which I think are OK using the 'trapz' method:

- One for getting a mean of load tracking signals. I think this an acceptable approximation.
- One for converting series of power data into an energy number. I think this is probably more accurate than 'rect' in that case.

derkling:
Absolutely agree, that was completely bogus! Thanks for the fix!

PR #487, opened 2017-10-09
Analysis/latency: Add task CPU residency utilities

valschneider:
This can be useful to look at the behaviour of a task (or from another point of view, the behaviour of the scheduler regarding this task) and to have an overview of how much time a task has spent on each CPU.

I'm not entirely too sure about with task states should be considered in the runtime sum - from my understanding, anything but sleeping should be accounted, but I could be wrong on that.

### DIFF
--- a/libs/utils/analysis/latency_analysis.py
+++ b/libs/utils/analysis/latency_analysis.py
@@ -286,6 +286,42 @@ class LatencyAnalysis(AnalysisModule):
         run_df = run_df[run_df.next_state.isin(['S', 'x'])][['running_time']]
         return run_df
 
+    @memoized
+    def _dfg_task_residency(self, task):
+        """
+        DataFrame of a task's execution time on each CPU
+
+        The returned DataFrame index is the CPU indexes
+        The DataFrame has just one column:
+        - runtime: the time the task spent being active on a given CPU,
+          in seconds.
+
+        :param task: the task to report runtimes for
+        :type task: int or str
+        """
+        cpus = range(self._platform['cpus_count'])
+        runtimes = {cpu : 0.0 for cpu in cpus}
+
+        df = self._dfg_latency_df(task)
+
+        # Exclude sleep time
+        df = df[df.curr_state != 'S']
+
+        for time, data in df.iterrows():
+            cpu = data['__cpu']
+
+            # When waking up, '__cpu' is NaN but 'target_cpu' is populated instead
+            if np.isnan(cpu):
+                if data['curr_state'] == 'W':
+                    cpu = data['target_cpu']
+                else:
+                    raise RuntimeError('No CPU data for latency_df @{}'.format(time))
+
+            runtimes[cpu] += data['t_delta']
+
+        data = [(cpu, time) for  cpu, time in runtimes.iteritems()]
+        return pd.DataFrame(data, columns=['CPU', 'runtime']).set_index('CPU')
+
 ###############################################################################
 # Plotting Methods
 ###############################################################################
--- a/libs/utils/analysis/latency_analysis.py
+++ b/libs/utils/analysis/latency_analysis.py
@@ -23,6 +23,7 @@ import numpy as np
 import pandas as pd
 import pylab as pl
 import re
+import os
 
 from collections import namedtuple
 from analysis_module import AnalysisModule
@@ -783,6 +784,28 @@ class LatencyAnalysis(AnalysisModule):
         return stats_df.append(pd.DataFrame(
             stats.values(), columns=['running_time'], index=stats.keys()))
 
+    def plotTaskResidency(self, task):
+        """
+        Plot CPU residency of the specified task
+        This will show an overview of how much time that task spent being
+        active on each available CPU, in seconds.
+
+        :param task: the task to report runtimes for
+        :type task: int or str
+        """
+        df = self._dfg_task_residency(task)
+
+        ax = df.plot(kind='bar', figsize=(16, 6))
+        ax.set_title('CPU residency of task {}'.format(task))
+
+        figname = os.path.join(
+            self._trace.plots_dir,
+            '{}task_cpu_residency_{}.png'.format(
+                self._trace.plots_prefix, task
+            )
+        )
+
+        pl.savefig(figname, bbox_inches='tight')
 
 ###############################################################################
 # Utility Methods
--- a/libs/utils/analysis/latency_analysis.py
+++ b/libs/utils/analysis/latency_analysis.py
@@ -146,6 +146,14 @@ class LatencyAnalysis(AnalysisModule):
             - task_latency_df['t_start']
         )
 
+        # Fix the last entry, which will have a NaN state duration
+        # Set duration to trace_end - last_event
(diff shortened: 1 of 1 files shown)


libs/utils/analysis/latency_analysis.py
```diff
@@ -775,6 +776,26 @@ def plotRuntimes(self, task, tag=None, threshold_ms=8, bins=64):
...
+
+        ax = df.plot(kind='bar', figsize=(16, 6))
+        ax.set_title('CPU residency of task {}'.format(task))
+
+        figname = os.path.join(
+            self._trace.plots_dir,
+            'task_cpu_residency_{}.png'.format(
+                self._trace.plots_prefix, task
+            )
+        )
```
mdigiorgio: You could use the same approach as for the other functions, i.e.:

```python
figname = '{}/{}task_cpu_residency_{}.png'.format(self._trace.plots_dir, self._trace.plots_prefix, task)
```

Actually, you were missing the `plots_prefix` in the name of the file :)

valschneider: Woopsy daisy, you're right. Although I think for the path generation, `os.path.join`is preferred.

mdigiorgio:
Sorry, forgot about this. LGTM :+1:

libs/utils/analysis/latency_analysis.py
```diff
@@ -286,6 +286,34 @@ def cr(row):
         run_df = run_df[run_df.next_state.isin(['S', 'x'])][['running_time']]
         return run_df
 
+    @memoized
+    def _dfg_task_residency(self, task):
```
bjackman: Docstring plzzzzzzz

libs/utils/analysis/latency_analysis.py
```diff
@@ -286,6 +286,34 @@ def cr(row):
...
+        cpus = [n for n in range(self._platform['cpus_count'])]
```
bjackman: `cpus = range(self._platform['cpus_count'])`

libs/utils/analysis/latency_analysis.py
```diff
@@ -286,6 +286,34 @@ def cr(row):
...
+        runtimes = {cpu : 0.0 for cpu in cpus}
+
+        df = self._dfg_latency_df(task)
+
+        # Exclude sleep time
+        df = df[df.curr_state != 'S']
+
+        for row in df.iterrows():
+            time = row[0]
+            data = row[1]
```
bjackman: `for time, data in df.iterrows()`

valschneider:
Added docstring & cleaned some stuff up.

libs/utils/analysis/latency_analysis.py
```diff
@@ -286,6 +286,42 @@ def cr(row):
...
+        runtimes = {cpu : 0.0 for cpu in cpus}
+
+        df = self._dfg_latency_df(task)
+
+        # Exclude sleep time
+        df = df[df.curr_state != 'S']
+
+        for time, data in df.iterrows():
+            cpu = data['__cpu']
+            if np.isnan(cpu):
```
derkling: Can this really happen?!?
Since `__cpu` is a common field reporting the CPU in which an event has been generated... I would never expect this to be NaN.

libs/utils/analysis/latency_analysis.py
```diff
@@ -286,6 +286,42 @@ def cr(row):
...
+                if data['curr_state'] == 'W':
+                    cpu = data['target_cpu']
+                else:
```
derkling: Can't we use a continue in the if case and leave this else block not indented?

libs/utils/analysis/latency_analysis.py
```diff
@@ -286,6 +286,42 @@ def cr(row):
...
+                    self._log.warning(
```
derkling: Kind of worrying: what means "results will be affected"?!?

libs/utils/analysis/latency_analysis.py
```diff
@@ -783,6 +784,28 @@ def plotRuntimes(self, task, tag=None, threshold_ms=8, bins=64):
         return stats_df.append(pd.DataFrame(
             stats.values(), columns=['running_time'], index=stats.keys()))
 
+    def plotTaskResidency(self, task):
+        """
+        Plot CPU residency of the specified task
+        This will show an overview of how much time that task spent being
+        active (running, waking up) on each available CPU, in seconds.
```
derkling: Do we highlight running vs waking up time?
Otherwise I would remove "waking up" because its mainly an "implementation/internal detail"

libs/utils/analysis/latency_analysis.py
valschneider: For wakeups, `__cpu` will be unset and `target_cpu` will be set instead:

target_cpu | __cpu | curr_state | next_state | t_start | t_delta
NaN | 6.0 | S | W | 2783.651695 | 0.010125
4.0 | NaN | W | A | 2783.661820 | 0.010525

libs/utils/analysis/latency_analysis.py
valschneider: I think I mostly mean "couldn't get any relevant data for a timestamp". I've never seen that happen so maybe this should be an exception rather than a warning ?

libs/utils/analysis/latency_analysis.py
valschneider: > Do we highlight running vs waking up time?
They're not distinguished in the resulting dataframe. My reasoning is that if a task is waking, it's eating CPU time as well. My plan for this wasn't to create extremely accurate statistics, but rather to get a big picture of the task's lifetime - if you believe something more accurate would actually be useful, I could change that.

libs/utils/analysis/latency_analysis.py
valschneider: I suppose we could - I'm not too hot on using `continue`s when I can avoid them, but that's a personal preference :)

libs/utils/analysis/latency_analysis.py
derkling: Oh right, indeed a wakeup can happen on CPU which is different then the one we wanna run the task on... I forgot there was this asymmetry in the latency dataframe.
Maybe we should also rename that "__cpu" in something like "wakeup_cpu"... just to avoid confusion in the future.

But for the time being, being add a line comment to remember that "__cpu" has a value only for "wakeup" events.

libs/utils/analysis/latency_analysis.py
derkling: I would say that we can just avoid to list running, waking up in the documentation.

```diff
- active (running, waking up) on each available CPU, in seconds.
+ active on each available CPU, in seconds.
```

libs/utils/analysis/latency_analysis.py
derkling: Never mind, in this case with a continue the loop would be broken.

libs/utils/analysis/latency_analysis.py
derkling: yes... if we get not usable data... better to raise.

valschneider:
Addressed @derkling's comments.

PR #490, opened 2017-10-20
Fix TestEnv.calibration()

bjackman:


mdigiorgio:
Why two commits? I think the first one is not necessary.

bjackman:
Well it's doing two different things really:

1. Ensuring the user's provided rt-app configuration can always be accessed
2. Expanding the conditions under which we derive our own calibration.

The code change would look OK as a single commit but I think they're distinct behavioural changes.

mdigiorgio:
But you are removing that line in the following commit...

bjackman:
Yeah, and if it was only one commit then I would be moving and changing it in the same commit, which would be confusing, no?

valschneider:
The 2 commits make sense to me - as @bjackman says, the first is a change and the second is a move. That way the history reads like "I'm changing this. Now I'm moving this" rather than "I'm moving this. By the way, I'm also changing this but you really have to look at what is being changed to notice."

PR #491, opened 2017-10-24
tests/eas/generic: Lower ramp tests duration

valschneider:
Each test in generic.py has a time duration set to 2s. However,
the "time_s" entry for rt-app ramps is the duration of each step,
and not the duration of the whole ramp. As such, each ramp test
lasts 26 seconds instead of 2.

This commit sets the whole duration of each ramp test to 2s.

### DIFF
--- a/tests/eas/generic.py
+++ b/tests/eas/generic.py
@@ -442,7 +442,7 @@ class RampUp(_EnergyModelTest):
                             "start_pct" :  5,
                             "end_pct"   : 70,
                             "delta_pct" :  5,
-                            "time_s"    :  2,
+                            "time_s"    : .5,
                          },
                     },
                 },
@@ -461,7 +461,6 @@ class RampDown(_EnergyModelTest):
     """
     Test EAS for a task ramping from 70% down to 5% over 2 seconds
     """
-
     # The main purpose of this test is to ensure that as it reduces in load, a
     # task is migrated from big to LITTLE CPUs on a big.LITTLE system.
     # This migration naturally happens some time _after_ it could possibly be
@@ -482,7 +481,7 @@ class RampDown(_EnergyModelTest):
                             "start_pct" : 70,
                             "end_pct"   :  5,
                             "delta_pct" :  5,
-                            "time_s"    :  2,
+                            "time_s"    : .5,
                          },
                     },
                 },


valschneider:
I tested this on HiKey960 and it didn't seem to impact the test results nor the shape of the trace (apart from its duration obviously).

tests/eas/generic.py
```diff
@@ -439,10 +444,10 @@ class RampUp(_EnergyModelTest):
...
                         "params" : {
                             'period_ms': WORKLOAD_PERIOD_MS,
-                            "start_pct" :  5,
-                            "end_pct"   : 70,
-                            "delta_pct" :  5,
-                            "time_s"    :  2,
+                            "start_pct" : start_pct,
+                            "end_pct"   : end_pct,
+                            "delta_pct" : delta_pct,
+                            "time_s"    : 2. / ((end_pct - start_pct) / delta_pct),
```
qperret: Maybe factor out this one as well ?

qperret: Argh, sorry, I went too fast. I'm suggesting to factor out  the `2.` in `time_s`. You could give it a nice name like `test_duration` or something like that. That's just a nitpick though.

bjackman: Could always just give rta.Ramp a 'duration_s' argument - if it's specified then it's calculated this way.

derkling:
What about just setting a constant and reasonable step size, e.g. 100ms per step?
Then the test will last for as much required by the number of steps?

Not entirely sure however, was just trying to see which configuration can be easier to validate in post-processing. The advantage of the fixed duration is that, for example, we can know exactly what should be the PELT changes in a predefined time internal.

tests/eas/generic.py
valschneider: @bjackman By "give rta.Ramp a 'duration_s' argument", do you mean modify rta.py do accept a duration_s argument, or just declare it in the workload dict ? I don't see any 'duration_s' parameter in rta so I figured it's the former, but I want to make sure.

bjackman: Yeah modify rta.py

valschneider:
I think @derkling's proposal might be the best way to go in terms of validation. I'll still add a `duration_s` argument to rta.Ramp because it's something I often felt the need for, but it'll be separate from this PR. Any objections ?

bjackman:
Fine by me

qperret:
I'm OK with the idea of a per-step metric but I'm not sure about a fixed duration. What happens if you change the task's period to 200 ms and you keep a 100ms step ? Maybe the per-step metric should be a specific number of task activation ?

valschneider:
@qperret I think the important part of the ramp tests is the steps themselves. I don't know how the workload will look like if the step duration is below the task period, but I guess it will just be rounded up and that's fine to me, we'll still have the utilization steps - my only reason for pushing this PR is that the ramp tests take a lot of time to run when they could be drastically shortened.

derkling:
I agree with the idea to speedup a bit the test, my main concern is that the step time you propose is likely a bit short for PELT to stabilize. If we consider a standard 32ms half-life PELT, a 70% task with 16ms period can take up to 250ms to enter the stable region.
You can easily check these timing by playing with this PELT simulator:
   https://gist.github.com/derkling/6d1d3a3164b0291ef64fa409f61a81cb

For our analysis, I would suggest to use something in the range of 300-500ms to be on the safe side with all our tests.

valschneider:
@derkling Fair point, I think I'll bump up the step time to 500ms - that's still a 4x time improvement for the test, and should remain safe for PELT stability.

valschneider:
Bumped step time to 500ms & update commit message.

PR #492, opened 2017-10-26
libs/utils: add frequency plotting for peripheral clocks

kdub:
Allows the plotting of clocks using the linux common clock tracing
infrastructure tracing. Chart shows the periods the clock is on or
off, as well as frequency information.

Test: run examples/android/workloads/Android_YouTube.ipynb, which
Test: now has a plotPeripheralFrequencies() example

### DIFF
--- a/libs/utils/analysis/frequency_analysis.py
+++ b/libs/utils/analysis/frequency_analysis.py
@@ -27,6 +27,7 @@ import pylab as pl
 from analysis_module import AnalysisModule
 from bart.common.Utils import area_under_curve
 from devlib.utils.misc import memoized
+from matplotlib.ticker import FuncFormatter
 from trappy.utils import listify
 from trace import ResidencyTime, ResidencyData
 
@@ -162,6 +163,85 @@ class FrequencyAnalysis(AnalysisModule):
 ###############################################################################
 # Plotting Methods
 ###############################################################################
+
+    def plotPeripheralClock(self, clk, title='Peripheral Frequency'):
+        """
+        Produce graph plotting the frequency of a particular peripheral clock
+
+        :param title: The title for the chart
+        :type  title: str
+
+        :param clk: The clk name to chart
+        :type  clk: str
+
+        :raises: KeyError
+        """
+        freq = self._trace.getPeripheralClockEffectiveRate(clk)
+        if freq is None or freq.empty:
+            self._log.warning('no peripheral clock events found for clock')
+            return
+
+        fig = plt.figure(figsize=(16,8))
+        gs = gridspec.GridSpec(2, 1, height_ratios=[8, 1])
+        freq_axis = plt.subplot(gs[0])
+        state_axis = plt.subplot(gs[1])
+        plt.suptitle(title, y=.97, fontsize=16, horizontalalignment='center')
+
+
+        # Plot frequency information (set rate)
+        freq_axis.set_title("Clock frequency for " + clk)
+        set_rate = freq['rate'].dropna()
+
+        rate_axis_lib = 0
+        if len(set_rate) > 0:
+            rate_axis_lib = set_rate.max()
+            set_rate.plot(style=['b--'], ax=freq_axis, drawstyle='steps-post', alpha=0.4, label="clock_set_rate value")
+            freq_axis.hlines(set_rate.iloc[-1], set_rate.index[-1], self._trace.x_max, linestyle='--', color='b', alpha=0.4)
+        else:
+            self._log.warning('No clock_set_rate events to plot')
+
+        # Plot frequency information (effective rate)
+        eff_rate = freq['effective_rate'].dropna()
+        if len(eff_rate) > 0 and eff_rate.max() > 0:
+            rate_axis_lib = max(rate_axis_lib, eff_rate.max())
+            eff_rate.plot(style=['b-'], ax=freq_axis, drawstyle='steps-post', alpha=1.0, label="Effective rate (with on/off)")
+            freq_axis.hlines(eff_rate.iloc[-1], eff_rate.index[-1], self._trace.x_max, linestyle='-', color='b', alpha=1.0)
+        else:
+            self._log.warning('No effective frequency events to plot')
+
+        freq_axis.set_ylim(0, rate_axis_lib * 1.1)
+        freq_axis.set_xlim(self._trace.x_min, self._trace.x_max)
+        freq_axis.set_xlabel('')
+        freq_axis.grid(True)
+        freq_axis.legend()
+        def mhz(x, pos):
+            return '%1.2f MHz' % (x*1e-6)
+        freq_axis.get_yaxis().set_major_formatter(FuncFormatter(mhz))
+
+        on = freq[freq.state == 1]
+        state_axis.hlines([0] * len(on),
+                          on['start'], on['start'] + on['len'],
+                          linewidth = 10.0, label='clock on', color='green')
+        off = freq[freq.state == 0]
+        state_axis.hlines([0] * len(off),
+                          off['start'], off['start'] + off['len'],
+                          linewidth = 10.0, label='clock off', color='red')
+
+
+        # Plot time period that the clock state was unknown from the trace
+        indeterminate = pd.concat([on, off]).sort_index()
+        if indeterminate.empty:
+            indet_range_max = self._trace.x_max 
+        else:
+            indet_range_max = indeterminate.index[0] 
+        state_axis.hlines(0, 0, indet_range_max, linewidth = 1.0, label='indeterminate clock state', linestyle='--')
+        state_axis.legend(bbox_to_anchor=(0., 1.02, 1., 0.102), loc=3, ncol=3, mode='expand')
+        state_axis.set_yticks([])
+        state_axis.set_xlabel('seconds')
+        state_axis.set_xlim(self._trace.x_min, self._trace.x_max)
+
+        figname = os.path.join(self._trace.plots_dir, '{}{}.png'.format(self._trace.plots_prefix, clk))
+        pl.savefig(figname, bbox_inches='tight')
 
     def plotClusterFrequencies(self, title='Clusters Frequencies'):
         """
--- a/libs/utils/trace.py
+++ b/libs/utils/trace.py
@@ -874,6 +874,39 @@ class Trace(object):
 
         return cluster_active
 
+    @memoized
+    def getPeripheralClockEffectiveRate(self, clk_name):
+        if clk_name is None: 
+            self._log.warning('no specified clk_name in computing peripheral clock, returning None')
+            return
+        if not self.hasEvents('clock_set_rate'):
+            self._log.warning('Events [clock_set_rate] not found, returning None!')
+            return
+        rate_df = self._dfg_trace_event('clock_set_rate')
+        enable_df = self._dfg_trace_event('clock_enable')
+        disable_df = self._dfg_trace_event('clock_disable')
+        pd.set_option('display.expand_frame_repr', False)
+
+        freq = rate_df[rate_df.clk_name == clk_name]
+        if not enable_df.empty:
+            enables = enable_df[enable_df.clk_name == clk_name]
+        if not disable_df.empty:
+            disables = disable_df[disable_df.clk_name == clk_name]
+
+        freq = pd.concat([freq, enables, disables]).sort_index()
+        if freq.empty:
+            self._log.warning('No events for clock ' + clk_name + ' found in trace')
+            return
+
+        freq['start'] = freq.index
+        freq['len'] = (freq.start - freq.start.shift()).fillna(0).shift(-1)
+        # The last value will be NaN, fix to be appropriate length
+        freq.loc[freq.index[-1], 'len'] = self.start_time + self.time_range - freq.index[-1]
+
+        freq = freq.fillna(method='ffill')
+        freq['effective_rate'] = np.where(freq['state'] == 0, 0,
+                                          np.where(freq['state'] == 1, freq['rate'], float('nan')))
+        return freq
 
 class TraceData:
     """ A DataFrame collector exposed to Trace's clients """
--- a/tests/lisa/test_trace.py
+++ b/tests/lisa/test_trace.py
@@ -17,6 +17,7 @@
 
 import json
 import os
+import numpy as np
 from unittest import TestCase
 
 from trace import Trace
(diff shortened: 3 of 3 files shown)


kdub:
new notebook: (not viewed well in diff)
https://gist.github.com/kdub/6e8f0e064d37ca24c493edc649df5dc8

libs/utils/analysis/frequency_analysis.py
```diff
@@ -115,6 +115,127 @@ def _dfg_cluster_frequency_residency(self, cluster, total=True):
...
+        rate_df['state'] = -1
+
+        pd.set_option('display.expand_frame_repr', False)
+
+        if not enable_df.empty:
+          enable_df = enable_df[enable_df.clk_name == clk]
+          enable_df['clock_setting'] = 1;
+        if not disable_df.empty:
+          disable_df = disable_df[disable_df.clk_name == clk]
+          disable_df['clock_setting'] = 0;
```
mdigiorgio: Different indentation

derkling:
@kdub, any chance you can update this series?

kdub:
Sure, will work on updating now, sorry for delay.

kdub:
Adopted changes, thanks for the suggestion!

libs/utils/analysis/frequency_analysis.py
```diff
@@ -115,6 +115,127 @@ def _dfg_cluster_frequency_residency(self, cluster, total=True):
...
+        """
+        if not self._trace.hasEvents('clock_set_rate'):
+            self._log.warning('Events [clock_set_rate] not found, plot DISABLED!')
+            return
+        rate_df = self._dfg_trace_event('clock_set_rate')
+        enable_df = self._dfg_trace_event('clock_enable')
+        disable_df = self._dfg_trace_event('clock_disable')
+        pd.options.mode.chained_assignment = None
+
+        rate_df = rate_df[rate_df.clk_name == clk]
```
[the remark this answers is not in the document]
kdub: done.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,127 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+
+        :raises: KeyError
+        """
+        if not self._trace.hasEvents('clock_set_rate'):
+            self._log.warning('Events [clock_set_rate] not found, plot DISABLED!')
+            return
+        rate_df = self._dfg_trace_event('clock_set_rate')
+        enable_df = self._dfg_trace_event('clock_enable')
+        disable_df = self._dfg_trace_event('clock_disable')
+        pd.options.mode.chained_assignment = None
```
valschneider: Usually when we face SettingWithCopy warnings we try to quiet them down by using `df.(i)loc ...`, even if the df is being modified properly. 

I see that this is isn't followed in plotClusterFrequencies(), but the warning is eventually restored with `pd.options.mode.chained_assignment = 'warn'`. If we really can't change the code to not have the warning I'd be fine with temporarily disabling it, but then it should be restored eventually.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,127 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+            disable_df['clock_setting'] = 0;
+
+        freq = pd.concat([rate_df, enable_df, disable_df])
+        freq.sort_index(inplace=True)
+
+        if freq.empty:
+            self._log.warning('No events for clock ' + clk + ' found in trace')
+            return
+
+        #this will extend the last frequency value to the end of the timerange.
```
valschneider: This is top-level nitpicking, but... Could you please insert a space between `#` and your comments ? Bonus points if they start with a capital letter too

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,127 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+        freq['timestamp'] = freq.index
+        freq['frequency'] = 0;
+
+        max_freq = 0
+        last_state = 1
+        last_freq = rate_df.iloc[0]['clock_setting']
+        for index, row in freq.iterrows():
+            if row.state == -1 :
+                last_freq = row.clock_setting
+                if max_freq < last_freq: max_freq = last_freq
```
valschneider: Can't we do `max_freq = rate_df.clock_setting.max()` outside of the loop ?
I wonder if this whole loop can't be done with some of pandas' clever functions like `applymap()` to set the frequency to 0 when `clock_setting == 0`, else keep it. The last row logic could then be handled after that.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,127 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+            first_onoff = onoff.iloc[0].timestamp
+
+        #plot time period that the clock state was unknown from the trace
+        state_axis.hlines(0, 0, first_onoff, linewidth = 1.0, label='indeterminate clock state', linestyle='--')
+        state_axis.set_yticks([])
+        state_axis.set_xlabel('seconds')
+        state_axis.set_xlim(self._trace.x_min, self._trace.x_max)
+        state_axis.legend(bbox_to_anchor=(0., 1.02, 1., 0.102), loc=3, ncol=3, mode='expand')
+
+        figname = '{}/{}{}.png'\
```
valschneider: `os.path.join()` would be better.

libs/utils/analysis/frequency_analysis.py
derkling: Likelt plotClusterFrequencies is warning because the pandas library added that check after that function as been added... but, going forward, I would say we should try to avoid warnings instead of muting them.

libs/utils/analysis/frequency_analysis.py
derkling: :+1: for the apply approach

joelagnel:
@kdub any updates on this PR?

libs/utils/analysis/frequency_analysis.py
kdub: Changed to not suppress warning, and corrected offending line to not generate warning.

libs/utils/analysis/frequency_analysis.py
kdub: done

libs/utils/analysis/frequency_analysis.py
kdub: I changed it to use max(), but I didn't find it easy to use the apply() or applymap() approach. I always had to put in more state (namely the frequency value of the  last row) than I could easily capture in the lambda.

libs/utils/analysis/frequency_analysis.py
kdub: done, the other functions could use an update too as a followup.

kdub:
Thanks for taking another look, comments should be fixed.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,124 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+        """
+        if not self._trace.hasEvents('clock_set_rate'):
+            self._log.warning('Events [clock_set_rate] not found, plot DISABLED!')
+            return
+        rate_df = self._dfg_trace_event('clock_set_rate')
+        enable_df = self._dfg_trace_event('clock_enable')
+        disable_df = self._dfg_trace_event('clock_disable')
+
+        rate_df = rate_df[rate_df.clk_name == clk]
+        rate_df['clock_setting'] = rate_df['rate']/1e6
```
valschneider: I get setting with copy warnings here. If this is needed for all clock-related analysis, this could be added to trace sanitization (in **libs/utils/trace.py**) and applied to all the traces (no filtering on clk).

Otherwise, if this is just needed for this analysis, you could do this:
```
rate_df = rate_df[rate_df.clk_name == clk].copy()
```

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,124 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+        rate_df['state'] = -1
+
+        pd.set_option('display.expand_frame_repr', False)
+
+        if not enable_df.empty:
+            enable_df = enable_df[enable_df.clk_name == clk]
+            enable_df['clock_setting'] = 1;
```
valschneider: Ditto for the warning. See above comment or use:
```
enable_df.loc[:, 'clock_setting'] = 1;
```

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,124 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+        last = freq.tail(n=1)
+        last.index = { self._trace.x_max }
+        freq = pd.concat([freq, last])
+        freq['timestamp'] = freq.index
+        freq['frequency'] = 0;
+
+        max_freq = rate_df['clock_setting'].max()
+        last_state = 1
+        last_freq = rate_df.iloc[0]['clock_setting']
+        for index, row in freq.iterrows():
```
valschneider: Had a look at the trace example in trappy, makes more sense now (I didn't know how the dataframes would be built).

It is indeed a tricky one. I couldn't think of a decent `apply()` use, but I think we can make it shorter and easier to read:

```
max_freq = rate_df['clock_setting'].max()
for index, row in freq.iterrows():
    # Get latest frequency update
    curr_freq = rate_df[:index]['clock_setting'].values[-1]
    freq.loc[index, 'frequency'] = 0 if not row.state else curr_freq
```

What do you think ?

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,124 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+        freq = pd.concat([rate_df, enable_df, disable_df])
+        freq.sort_index(inplace=True)
+
+        if freq.empty:
+            self._log.warning('No events for clock ' + clk + ' found in trace')
+            return
+
+        # This will extend the last frequency value to the end of the timerange.
+        last = freq.tail(n=1)
+        last.index = { self._trace.x_max }
```
valschneider: What about directly computing the delta of each event, and "fixing" the last one to extend it to the end of the trace ? This is what we do for overutilized:
```
df = self._dfg_trace_event('sched_overutilized')
df['start'] = df.index

# Fix the last event, which will have a NaN duration
# Set duration to trace_end - last_event
df.loc[df.index[-1], 'len'] = self.start_time + self.time_range - df.index[-1]
```

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,124 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+        rate_df['state'] = -1
+
+        pd.set_option('display.expand_frame_repr', False)
+
+        if not enable_df.empty:
+            enable_df = enable_df[enable_df.clk_name == clk]
+            enable_df['clock_setting'] = 1;
+        if not disable_df.empty:
+            disable_df = disable_df[disable_df.clk_name == clk]
+            disable_df['clock_setting'] = 0;
```
valschneider: Ditto for the warning

libs/utils/analysis/frequency_analysis.py
kdub: Adopted changes, although I couldn't reproduce the warning here with my setup (pandas 0.22)

libs/utils/analysis/frequency_analysis.py
kdub: Adopted changes, although I couldn't reproduce the warning here with my setup (pandas 0.22)

libs/utils/analysis/frequency_analysis.py
kdub: sure, adopted. Thanks!

libs/utils/analysis/frequency_analysis.py
kdub: I don't mind adopting this if you insist, but left as-is, is computing the the delta as fast as just duplicating the last row at a different timestamp to fix up the chart?

libs/utils/analysis/frequency_analysis.py
kdub: Adopted changes, although I couldn't reproduce the warning here with my setup (pandas 0.22).

libs/utils/analysis/frequency_analysis.py
derkling: I like better to have code consistency, just because we can easy recognise similar code patterns and likely, maybe in the future, factorise them into a new common API.

kdub:
@valschneider 
* I refactored the code so the dataframe should be reusable (in the same manner as getCPUActiveSignal). 
* I also reworked the delta computing code to be more like getCPUActiveSignal.
* I added a unit test as well for trace parsing.
* I also regenerated the notebook: https://gist.github.com/kdub/6e8f0e064d37ca24c493edc649df5dc8

kdub:
Anyone have time to review again? Thanks!

joelagnel:
It sad github shows "Changes requested" when changes were already made. Why does github not get it? Atleast they should remove the red from Changes requested since it completely misleads. :-(

Kevin I'll look at this shortly! Thanks.

joelagnel:
LGTM. @derkling  ?

tests/lisa/test_trace.py
```diff
@@ -245,6 +246,37 @@ def test_sched_load_avg_signals(self):
...
+          <idle>-0 [002] 380330000000: clock_enable: bus_clk state=1 cpu_id=2
+          <idle>-0 [002] 380331000000: clock_set_rate: bus_clk state=750000000 cpu_id=2
+          <idle>-0 [000] 380332000000: clock_disable: bus_clk state=0 cpu_id=0
+          <idle>-0 [000] 380333000000: clock_enable: bus_clk state=1 cpu_id=0
+          <idle>-0 [002] 380334000000: clock_set_rate: bus_clk state=100000000 cpu_id=2
+          <idle>-0 [000] 380335000000: clock_disable: bus_clk state=0 cpu_id=0
+          <idle>-0 [004] 380339000000: cpu_idle:             state=1 cpu_id=4
+        """)
+        df = trace.getPeripheralClockEffectiveRate(clk_name='bus_clk')
+        print(df)
```
valschneider: Eh, could have used that trace to test my previous comments ;_;
Anyway, tests are cool, so thanks for adding one. I'd just like the print to go please.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,86 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+            return
+
+        fig = plt.figure(figsize=(16,8))
+        gs = gridspec.GridSpec(2, 1, height_ratios=[8, 1])
+        freq_axis = plt.subplot(gs[0])
+        state_axis = plt.subplot(gs[1])
+        plt.suptitle(title, y=.97, fontsize=16, horizontalalignment='center')
+
+
+        #plot frequency information (set rate)
```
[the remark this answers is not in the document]
kdub: fixed

kdub:
@valschneider Fixed issues found, using ffill was a much simpler solution, thanks!

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +164,85 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+            return
+
+        fig = plt.figure(figsize=(16,8))
+        gs = gridspec.GridSpec(2, 1, height_ratios=[8, 1])
+        freq_axis = plt.subplot(gs[0])
+        state_axis = plt.subplot(gs[1])
+        plt.suptitle(title, y=.97, fontsize=16, horizontalalignment='center')
+
+
+        # plot frequency information (set rate)
```
valschneider: Capitalisation please

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +164,85 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+
+        rate_axis_lib = 0
+        if len(set_rate) > 0:
+            rate_axis_lib = set_rate.max()
+            set_rate.plot(style=['b--'], ax=freq_axis, drawstyle='steps-post', alpha=0.4, label="clock_set_rate value")
+            freq_axis.hlines(set_rate.iloc[-1], set_rate.index[-1], self._trace.x_max, linestyle='--', color='b', alpha=0.4)
+        else:
+            self._log.warning('No clock_set_rate events to plot')
+
+        # plot frequency information (effective rate)
```
valschneider: Ditto capitalisation

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +164,85 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+        state_axis.hlines([0] * len(on),
+                          on['start'], on['start'] + on['len'],
+                          linewidth = 10.0, label='clock on', color='green')
+        off = freq[freq.state == 0]
+        state_axis.hlines([0] * len(off),
+                          off['start'], off['start'] + off['len'],
+                          linewidth = 10.0, label='clock off', color='red')
+
+
+        #plot time period that the clock state was unknown from the trace
```
valschneider: Space + capitalisation

libs/utils/trace.py
```diff
@@ -874,6 +874,39 @@ def getClusterActiveSignal(self, cluster):
...
+            disables = disable_df[disable_df.clk_name == clk_name]
+
+        freq = pd.concat([freq, enables, disables]).sort_index()
+        if freq.empty:
+            self._log.warning('No events for clock ' + clk_name + ' found in trace')
+            return
+
+        freq['start'] = freq.index
+        freq['len'] = (freq.start - freq.start.shift()).fillna(0).shift(-1)
+        #the last value will be NaN, fix to be appropriate length
```
valschneider: Space + capitalisation

kdub:
@valschneider comments and commit ordering fixes done.

PR #493, opened 2017-10-30
submodules: Update TRAPpy to latest master

valschneider:
This change will enable the trace caching by default.


valschneider:
https://github.com/ARM-software/trappy/pull/273 should fix the Travis failure.

valschneider:
The light is now green !

PR #495, opened 2017-10-31
submodules: Update workload-automation to latest next

bjackman:
That means we can delete the exoplayer workload plugin from LISA, as
it is now merged upstream.


derkling:
Now provided by: ARM-software/workload-automation/pull/483

PR #503, opened 2017-11-07
Support abootimg building

Leo-Yan:
This patch set is to enable abootimg building method; if we change to use UEFI bootloader for Hikey960, UEFI using abootimg compatible booting image.

So I think the first patch can be considered merge firstly and the second patch we can postpone to merge until UEFI has been released as official booting image on Hikey960.

### DIFF
--- a/tools/wltests/android/create_aboot_img.sh
+++ b/tools/wltests/android/create_aboot_img.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+
+################################################################################
+# Internal configurations
+################################################################################
+SCRIPT_DIR=$(dirname $(realpath -s $0))
+BASE_DIR="$SCRIPT_DIR/.."
+source "${BASE_DIR}/helpers"
+source "${DEFINITIONS_PATH}"
+
+DEFAULT_KERNEL="${KERNEL_SRC}/arch/${ARCH}/boot/${KERNEL_IMAGE}"
+KERNEL="${KERNEL:-$DEFAULT_KERNEL}"
+
+DEFAULT_DTB="${KERNEL_SRC}/arch/${ARCH}/boot/dts/${KERNEL_DTB}"
+DTB="${DTB:-$DEFAULT_DTB}"
+
+DEFAULT_RAMDISK="${PLATFORM_OVERLAY_PATH}/${RAMDISK_IMAGE}"
+RAMDISK="${RAMDISK:-$DEFAULT_RAMDISK}"
+
+DEFAULT_BOOT_IMAGE="${ARTIFACTS_PATH}/${ANDROID_BOOT_IMAGE}"
+BOOT_IMAGE="${BOOT_IMAGE:-$DEFAULT_BOOT_IMAGE}"
+
+CMDLINE=${CMDLINE:-$KERNEL_CMDLINE}
+
+if [ ! -f ${KERNEL} ] ; then
+	c_error "KERNEL image not found: ${KERNEL}"
+	exit $ENOENT
+fi
+
+if [ ! -f ${DTB} ] ; then
+	c_error "DTB not found: ${DTB}"
+	exit $ENOENT
+fi
+
+if [ ! -f ${RAMDISK} ] ; then
+	c_error "RAMDISK image not found: ${RAMDISK}"
+	c_warning "A valid ramdisk image, which matches the device user-space"
+	c_warning "must be deployed by the user under the required path."
+	c_info "Please refer to the ISTALLATION INSTRUCTIONS"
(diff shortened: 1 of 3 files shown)


tools/wltests/android/create_aboot_img.sh
```diff
@@ -0,0 +1,114 @@
...
+    c_info "- ANDROID_SECOND_SIZE: ${ANDROID_SECOND_SIZE}"
+    ANDROID_SECOND_SIZE="secondaddr = ${ANDROID_SECOND_SIZE}"
+fi
+
+if [ "${ANDROID_RAMDISK_OFFSET}" ]; then
+    c_info "- ANDROID_RAMDISK_OFFSET : ${ANDROID_RAMDISK_OFFSET}"
+    ANDROID_RAMDISK_OFFSET="ramdiskaddr = ${ANDROID_RAMDISK_OFFSET}"
+fi
+
+cat > "${ARTIFACTS_PATH}"/bootimg-960.cfg << EOL
```
derkling: This `bootimg-960.cfg ` seems like an hardcoded filename valid only for hikey960, isnt'it?

We probably need an additional variable in the `definitions` of each platform, to specify the name of the configuration file to generate.

tools/wltests/android/create_aboot_img.sh
```diff
@@ -0,0 +1,114 @@
...
+################################################################################
+
+# Ensure the output folder exists
+mkdir -p $(dirname $BOOT_IMAGE) &>/dev/null
+
+set -x
+c_info "- IMAGE_DTB: $IMAGE_DTB"
+IMAGE_DTB="$(dirname $KERNEL)/Image-dtb"
+cat "${KERNEL}" "${DTB}" > "${IMAGE_DTB}"
+abootimg --create "${BOOT_IMAGE}" -k "${IMAGE_DTB}" -r "${RAMDISK}" -f "${ARTIFACTS_PATH}"/bootimg-960.cfg
```
derkling: This `abootimg` tool should be provided by this patch and be available under `tools/wltests/android`

Have a look at how to use mkbootimg here:
https://github.com/ARM-software/lisa/blob/master/tools/wltests/android/create_boot_img.sh#L75
... on the formatting side as well ;-)

tools/wltests/platforms/hikey960_android-4.4/definitions
```diff
@@ -5,19 +5,21 @@
 
 PLATFORM_NAME="Hikey960_Android"
 
+BOOT_UEFI="${BOOT_UEFI:-YES}"
```
derkling: As per previous comment, this should be better defined as:
```bash
BOOTLOADER=${BOOTLOADER:-uefi}
```
at the end of the following "Boot Image Configuration" section

tools/wltests/android/create_aboot_img.sh
Leo-Yan: Will add variable for the config file.

tools/wltests/android/create_aboot_img.sh
Leo-Yan: abootimg can be installed by Ubuntu apt-get command, mkbooting is not official binary. So I think it's more reasonable to directly use Ubuntu official abootimg, so will add checking and give instructions for abootimg.

tools/wltests/platforms/hikey960_android-4.4/build_images
```diff
@@ -12,9 +12,15 @@ source "${PLATFORM_PATH}/definitions"
 ./build --image --dtbs --silent; ERROR=$?
 [ $ERROR -eq 0 ] || exit $ERROR
 
-./android/create_boot_img.sh; ERROR=$?
-[ $ERROR -eq 0 ] || exit $ERROR
+if [ "x${BOOT_UEFI}"=="xYES" ]; then
```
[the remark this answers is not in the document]
Leo-Yan: Thanks, will do.

tools/wltests/android/create_aboot_img.sh
derkling: Yes, it that case it make sense to just split out a warning if the binary is not available with a note on which packages should be installed. Thanks.

derkling:
Hi @Leo-Yan, any chance to update this according to the above comments?
We are trying to merge as much as possible of the outstanding PR... would be nice if this one also can get it ;-)

douglas-raillard-arm:
Are there still outstanding issues around hikey 960 + UEFI ?

Leo-Yan:
Hi @douglas-raillard-arm ,

I think this patch set is stable and I don't see any requirement so far for UEFI/abootimg on Hikey960.  If you agree with this, let's drop this PR.

PR #504, opened 2017-11-08
tools/wltests: Add support for abootimg

Leo-Yan:
This patch is to support abootimg to generate boot.img, so finally this
can be used by UEFI bootloader; Using abootimg tool it can simplize
the booting images by combining Kernel image and DTB image into single
file.

This patch is to add variable 'BOOTLOADER' to distinguish if the
bootloader is "fastboot" mode or UEFI mode; also add a dedicated script
tools/wltests/android/create_aboot_img.sh for abootimg building.

### DIFF
--- a/tools/wltests/android/create_aboot_img.sh
+++ b/tools/wltests/android/create_aboot_img.sh
@@ -1,0 +1,144 @@
+#!/bin/bash
+
+################################################################################
+# Internal configurations
+################################################################################
+SCRIPT_DIR=$(dirname $(realpath -s $0))
+BASE_DIR="$SCRIPT_DIR/.."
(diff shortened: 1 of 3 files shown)


derkling:
Hi Leo, maybe you don't know but the best way to update PR on github is just to force push your branch with the new version of the patches. If you can do that with your previous branch:
   Leo-Yan:support_abootimg
this sill automatically refresh the previous PR #503 and allows the reviewer to easily check what as been updated.

In other words is not required to open a new PR for each new version of a patch series, just keep force pushing updates on the original branch ;-)

derkling:
To be pushed as a refresh of #503

PR #505, opened 2017-11-08
latency_analysis: Fix task name

Leo-Yan:
Trace::getTaskByPid() returns task name with 'str' type, if use join()
to apply for the string, it inserts comma between every chars and the
task name is not readable.

So fix this by directly using name string from Trace::getTaskByPid().

### DIFF
--- a/libs/utils/analysis/latency_analysis.py
+++ b/libs/utils/analysis/latency_analysis.py
@@ -775,7 +775,7 @@ class LatencyAnalysis(AnalysisModule):
         else:
             raise ValueError("Task must be either an int or str")
 
-        task_label = "{}: {}".format(task_pid, ', '.join(task_names))
+        task_label = "{}: {}".format(task_pid, task_names)
         return TaskData(task_pid, task_names, task_label)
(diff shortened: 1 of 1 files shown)


libs/utils/analysis/latency_analysis.py
```diff
@@ -775,7 +775,7 @@ def _getTaskData(self, task):
         else:
             raise ValueError("Task must be either an int or str")
 
-        task_label = "{}: {}".format(task_pid, ', '.join(task_names))
+        task_label = "{}: {}".format(task_pid, task_names)
```
bjackman: Please can you also rename `task_names` to `task_name`.

Leo-Yan: Sure. I sent another PR for this.

But I want to check if there have updated patch, can I reuse this PR? I tried to "git push" to the same branch, but it failed.

bjackman: You can normally just amend the commit and do a `git push --force`

PR #506, opened 2017-11-08
examples/android/benchmarks: Add Hikey960 UI performance parsing notebook

Leo-Yan:
This script is to add UI performance paring notebook, it analyses the
trace result from CPU level and task level. For CPU level, it checks
task placement and CPU frequency selections, for task level it checks
UI related tasks scheduling latency and runtime period.


valschneider:
Hi Leo, thanks for the submission :)

As this is an example notebook, I think it would be nice to add some explanations as to what the notebook is doing and what information we can draw from the results - you can look at some other notebooks like [Geekbench](https://github.com/ARM-software/lisa/tree/master/ipynb/examples/android/benchmarks/Android_Geekbench.ipynb) or [Gmaps](https://github.com/ARM-software/lisa/tree/master/ipynb/examples/android/workloads/Android_Gmaps.ipynb) for inspiration.

I still haven't found a good way to review notebook commits, so I'll just list my comments based on the notebook cell numbers.

[3] Here it'd be nice to explain how to get an interesting trace. Maybe something even better would be to store this trace in LISA (we already have trace data for LisaInANutshell), load it in the Notebook and add a few lines to explain how this trace was generated.

[5] I think the comments/explanations for the cells should be kept in Markdown cells. Admittedly there is no real added value, the information stays the same, but I think it's a bit more readable. On top of that, it also keeps the same style of the other example notebooks.

[7] I think per-CPU frequency plotting can be dropped, since there's only two frequency domains for the CPUs in the HiKey960.

([14] and forward): This is probably going to be somewhat time-consuming, but it would be really nice to give a brief summary of what is expected for each task (their role/behaviour). Furthermore, maybe an argument could be added to `plot_task_scheduling_latency` to limit the maximum number of plots drawn ? That way it can still be changed when you want to look at a specific issue, but the notebook can be submitted with only one plot per task (IOW only one PID per task name) to make it less "scary".

PR #507, opened 2017-11-08
latency_analysis: Fix task name

Leo-Yan:
Trace::getTaskByPid() returns task name with 'str' type, if use join()
to apply for the string, it inserts comma between every chars and the
task name is not readable.

This patch is to directly use name string from Trace::getTaskByPid();
and rename variable 'task_names' to 'task_name'.

### DIFF
--- a/libs/utils/analysis/latency_analysis.py
+++ b/libs/utils/analysis/latency_analysis.py
@@ -829,21 +829,21 @@ class LatencyAnalysis(AnalysisModule):
                                       ','.join(self._trace.getTaskByPid(pid)))
                 self._log.warning('Returning stats only for PID: %d',
                                   task_pid)
-            task_names = self._trace.getTaskByPid(task_pid)
+            task_name = self._trace.getTaskByPid(task_pid)
 
         # Get task name
         elif isinstance(task, int):
             task_pid = task
-            task_names = self._trace.getTaskByPid(task_pid)
-            if len(task_names) == 0:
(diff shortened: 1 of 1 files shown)


libs/utils/analysis/latency_analysis.py
```diff
@@ -762,21 +762,21 @@ def _getTaskData(self, task):
...
-            task_names = self._trace.getTaskByPid(task_pid)
+            task_name = self._trace.getTaskByPid(task_pid)
 
         # Get task name
         elif isinstance(task, int):
             task_pid = task
-            task_names = self._trace.getTaskByPid(task_pid)
-            if len(task_names) == 0:
+            task_name = self._trace.getTaskByPid(task_pid)
+            if len(task_name) == 0:
```
bjackman: I've just spotted that this is wrong too - `getTaskByPid` returns `None` when it isn't found (it used to return `[]`). Could you change this to `if task_name is None` as well? Sorry I didn't notice this before.

derkling: Yes... this has changed when I've added the patch to always return the lest known name for each PID... before we used to return a list of names in case the task was changing its name after fork. ;-)

Thanks @Leo-Yan for the fix, it was in my todo list since a while +1

derkling:
@Leo-Yan, any chance you can update this patch? It's almost ready to merge ;-)

Leo-Yan:
@derkling, sure, have uploaded new patch for reviewing.

PR #508, opened 2017-11-15
tests/eas/generic: Add plot_expected_util

valschneider:
When running generics.py and trying to troubleshoot task_placement
failures, it can sometime be difficult to understand what was the
expected placement.

This commit adds a `plot_expected_util` method that will plot the
per-CPU utilization that was expected. It is called by default during
the test post-processing, and the plot will be saved as a .png file
in the test results output directory.

An example of how this plot looks like can be found here: https://gist.github.com/valschneider/b8f15ce92c3581a0e7928ae80c8cf306

### DIFF
--- a/tests/eas/generic.py
+++ b/tests/eas/generic.py
@@ -28,6 +28,10 @@ from test import LisaTest, experiment_test
 from trace import Trace
 from unittest import SkipTest
 
+import matplotlib.pyplot as plt
+import numpy as np
+import pylab as pl
+import os
 
 WORKLOAD_PERIOD_MS =  16
 SET_IS_BIG_LITTLE = True
@@ -236,6 +240,10 @@ class _EnergyModelTest(LisaTest):
         :meth:EnergyModel.estimate_from_cpu_util to get a DataFrame showing the
         estimated power usage over time under ideal EAS behaviour.
 
+        :meth:get_optimal_placements returns several optimal placements. They
+        are usually equivalent, but can be drastically different in some cases.
+        Currently only one of those placements is used (the first in the list).
+
         :param experiment: The :class:Experiment to examine
         :returns: A Pandas DataFrame with a column each node in the energy model
                   (keyed with a tuple of the CPUs contained by that node) and a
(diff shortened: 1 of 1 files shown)


tests/eas/generic.py
```diff
@@ -252,9 +256,80 @@ def exp_power(row):
...
+        task_utils_df = self.get_task_utils_df(experiment)
+        nrg_model = self.te.nrg_model
+        trace = self.get_trace(experiment)
+
+        cpus = nrg_model.cpus
+        data = []
+        index = []
+        for time, utils_row in task_utils_df.iterrows():
+            task_utils = utils_row.to_dict()
+            expected_utils = nrg_model.get_optimal_placements(task_utils)[0]
```
bjackman: Is it possible to factor out commonality with the `exp_power` helper above?

tests/eas/generic.py
```diff
@@ -252,9 +256,80 @@ def exp_power(row):
...
+
+            row = []
+            for cpu, util in enumerate(expected_utils):
+                row.append(util)
```
bjackman: Isn't this just `row = expected_utils`?

tests/eas/generic.py
```diff
@@ -252,9 +256,80 @@ def exp_power(row):
...
+            data.append(row)
+            index.append(time)
+
+        df = pd.DataFrame(data, index=index)
+        fig, ax = plt.subplots(len(cpus), 1, figsize=(16, 1.8 * len(cpus)))
+
+        # Check if big.LITTLE data is available for a more detailled plot
+        if trace.has_big_little:
+            little_cap = self.te.platform['nrg_model']['little']['cpu']['cap_max']
+            big_cpus = self.te.platform['clusters']['big']
```
bjackman: `big_cpus` is unused

tests/eas/generic.py
```diff
@@ -252,9 +256,80 @@ def exp_power(row):
...
+            if trace.has_big_little and cpu in little_cpus:
+                ax[cpu].set_ylim((0, little_cap))
+                color = 'blue'
+            else:
+                ax[cpu].set_ylim((0, 1024))
+                color = 'red'
+
+            tdf.plot(ax=ax[cpu], drawstyle='steps-post', title="CPU{}".format(cpu), color=color)
+            ax[cpu].fill_between(tdf.index, tdf, 0, step='post', color=color)
+            ax[cpu].set_ylabel('Utilization')
```
bjackman: "Expected Utilization"? Might be more obvious what the plot really shows.

tests/eas/generic.py
```diff
@@ -252,9 +256,80 @@ def exp_power(row):
...
+                    ax[cpu].axvspan(prev, time, facecolor='gray', alpha=0.1, linewidth=0.0)
+                    ffill = False
+                if util == 0.0:
+                    ffill = True
+
+                prev = time
+
+        figname = os.path.join(experiment.out_dir, 'expected_placement.png')
+        pl.savefig(figname, bbox_inches='tight')
+
```
bjackman: Does this also show the plot inline when you call it from a notebook?

tests/eas/generic.py
valschneider: We could explode the `exp_power` helper into a loop and create the plot + modify the rows in the same place - we're using the same data after all.

tests/eas/generic.py
valschneider: Totally is yeah, must've tried something fancy with cpu indexing and eventually gave up.

tests/eas/generic.py
valschneider: I'm not super knowledgeable about executor/experiments, so I only tested this by running generics.py. I suppose it could be used in a notebook by feeding it experiments from an executor, but I haven't tried this. Is this what you meant ?

bjackman: Yeah. Not important I was just curious.

tests/eas/generic.py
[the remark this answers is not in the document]
valschneider: Originally I wanted to add a sorting method to find all configurations that are *actually* different and plot them, but that has fallen down into my backlog... 

I'll slide in a mention to the fact that we're just picking one for now.

tests/eas/generic.py
```diff
@@ -252,9 +256,80 @@ def exp_power(row):
...
+    def plot_expected_util(self, experiment):
+        """
+        Create a plot of the expected per-CPU utilization for the experiment
+        The plot is then outputted to the test results directory.
+
+        :param experiment: The :class:Experiment to examine
+        :returns: A Pandas DataFrame with a column each node in the energy model
+                  (keyed with a tuple of the CPUs contained by that node) and a
+                  "power" column with the sum of other columns. Shows the
+                  estimated *optimal* power over time.
```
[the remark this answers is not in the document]
valschneider: That's a sad case of a copy/paste gone wrong (and mostly unchecked)

tests/eas/generic.py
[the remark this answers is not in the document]
bjackman: Yeah, I should have just implemented it this way at the beginning. The perfect is the enemy of the good I guess!

valschneider: Still, since `get_optimal_placements` is compute-heavy, it'd be nice if we could run it once per row (instead of twice, which we're doing right now).

bjackman: Oh yeah that's definitely true. You could avoid the double call, or you could memoize it (maybe there was a reason I didn't do that before...), or you could just leave it and tell people to buy fast computers.. all 3 are fine with me :)

valschneider:
- Rewrote the thing to compute utilization only once
- Updated doc of `get_expected_power_df` to mention the `[0]` trickery

valschneider:
Thanks for the review ! I'll wait a bit for people to have a look at it and merge it if it remains ignored :)

PR #512, opened 2017-11-27
ipynb/wltests: sort energy plot by average instead of 99% percentile

qperret:
The energy plots should enable users to quickly see whether their new
feature under test saved energy or not. Although defining "saving
energy" can be tricky, looking at the average value is often a sensible
option to do that, so let's use it by default.

### DIFF
--- a/ipynb/wltests/sched-evaluation-full.ipynb
+++ b/ipynb/wltests/sched-evaluation-full.ipynb
@@ -130,7 +130,7 @@
     "for test in collector.tests(workload='jankbench'):\n",
     "    logging.info(\"Results for: %s\", test)\n",
     "    collector.report(workload='jankbench', metric='device_total_energy',\n",
-    "                     test=\"^{}$\".format(test), sort_on='99%', ascending=True)"
+    "                     test=\"^{}$\".format(test), sort_on='mean', ascending=True)"
    ]
   },
   {
@@ -196,7 +196,7 @@
     "for test in collector.tests(workload='exoplayer'):\n",
     "    logging.info(\"Results for: %s\", test)\n",
     "    collector.report(workload='exoplayer', metric='device_total_energy',\n",
-    "                     test=test, sort_on='99%', ascending=True)"
+    "                     test=test, sort_on='mean', ascending=True)"
    ]
   },
   {
@@ -215,7 +215,7 @@
     "for test in collector.tests(workload='homescreen'):\n",
     "    logging.info(\"Results for: %s\", test)\n",
     "    collector.report(workload='homescreen', metric='device_total_energy',\n",
-    "                     test=test, sort_on='99%', ascending=True)"
+    "                     test=test, sort_on='mean', ascending=True)"
    ]
   },
   {


qperret:
To argue in favour of sorting by `mean` instead of `99%`, here is the results for 10 iterations of homescreen on 10 similar kernel (I touched only documentation files to get different sha1s, but the kernel is the same). 
```
		count		mean		std		min		50%		75%		95%		99%		max
b8c7ccc		10.0		83.865804	2.071522	80.030288	84.077550	85.657704	86.181960	86.201074	86.205853
41047ab		10.0		83.956490	1.778328	80.897373	84.280810	85.217716	86.162340	86.583336	86.688585
f5d089a		10.0		83.765053	1.951244	80.695522	83.592737	85.241977	86.432467	86.728793	86.802875
92c7d9e		10.0		83.309157	1.985682	80.763532	82.777222	84.363756	86.374935	87.292921	87.522418
054ed67		10.0		83.892737	2.296718	80.718918	84.353920	85.025016	86.956492	87.307900	87.395753
0ccaff8		10.0		83.847660	2.365799	81.045893	83.129665	85.082864	87.607949	88.619836	88.872807
base   		10.0		83.760581	2.719289	80.675505	82.750518	85.853952	87.884457	88.780955	89.005080
0c68005		10.0		85.757405	2.860924	80.925355	85.628851	87.321279	89.804905	90.029015	90.085043
43628a2		10.0		84.085558	3.259505	80.648160	83.467422	85.666809	89.251508	90.477546	90.784055
99d8de4		10.0		84.957441	3.687352	81.089235	84.155749	85.561868	90.630351	93.701430	94.469200
```

We can see the mean energy doesn't change much while the 99% percentile changes a lot. This happened for the very same kernels so I would say that the 99% percentile here is essentially measurement noise.

derkling:
As per the previous comments on #510, I'n not entirely convinced about using just the mean, especially when it comes down to so few experiments. I think we should account for stdev somehow, and percentiles does exactly that.

What about switching from using the 99 to using the 75 percentile?
That way we can at least account for 1 stdev on top of the average.

Furthermore, as we already discussed with @credp  in the past, to operate an actual choice we should:
1. define some reasonable thresholds: e.g. a 1-2% difference in energy for a certain benchmark is considered ok because we know what's the average noise error.

Maybe that percentage can be actually computed with a set of experiments like to one you did above (i.e. running multiple times the exact same kernel). 
In your example above, something like:
```python
100 * df['std'].mean() / df['mean'].mean()
2.9691423879778984
```
would suggests that a 3% variation is in the "noise margin" which we can consider acceptable.

2. add a power-vs-performance metric, which allows to define, for example, how much acceptable an energy increase is given certain performance improvement.

qperret:
I agree that selecting the "best" kernel should not be done by looking only at the average, but as I already mentioned on #510, we are sorting the _plots_ here. That means you can see how spread the data is by comparing the size of the boxes. If you look at the plots sorted by average you can see the two dimensions (mean and variance) at once, and it's easy to understand what you see. If you sort the plots by 99% or 75% percentile you're mixing everything together and that makes it confusing IMO.

derkling:
I understand your reasoning when you say that from a graphical perspective we already have both mean and stddev represented. This info is there independently from the ordering, you can always compare two boxplots also in a completely not sorted boxplot.

When it comes to ordering we are kind-of building information on top. We cannot escape from the fact that we are kind-of saying that "something is better" the something else. 
That's why sorting support has been added to start with.

Thus, from a purely graphical reading standpoint, why should we for example represent Kernel A on top of B in cases like this one?

```
                     +----------------------------+
   Kernel A    +-----+      |                     +---------+
                     +----------------------------+

                           +------+
   Kernel B        +-------+   |  +--+
                           +------+
```

qperret:
Because that makes it easy to understand the situation. Kernel A has a lower average value but is more spread. Which one is the best ? I don't know but I have the information I want to know. With a different sorting criteria the information is still there but less readable.

derkling:
@valschneider, @bjackman, @mdigiorgio, @credp help us unlock this please.

Let's got for a vote: should be sort (by default) the energy boxplots by avg or 75 percentile?

valschneider:
[The wheel](http://wheeldecide.com/index.php?c1=avg&c2=percentile&time=5) said "average" for me.

bjackman:
For me average is more intuitive. I would naturally expect the mean to be the first thing you look at, with stddev being something that you visualise too but that is definitely secondary.

PR #513, opened 2017-11-29
test_series: fix WA invocation with multiple devices

qperret:
The ADB device name passed to test_series is currently used to flash and
reboot the board using ADB and Fastboot. However, it is not yet passed
to WA3 which then complains whenever several ADB devices are connected.
This commit fixes this behaviour by making the config fragment currently
used for the ACME more general and appending a "device_config" section
to it.

### DIFF
--- a/tools/wltests/test_series
+++ b/tools/wltests/test_series
@@ -364,14 +364,18 @@ if [ ! -z $WA_AGENDA ]; then
 		list_available_agendas
 		exit $INVAL
 	fi
-	# Generate a WA configuration fragment for the ACME cape
+	# Generate a WA configuration fragment for the ACME cape and ADB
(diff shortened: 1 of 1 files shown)


tools/wltests/test_series
```diff
@@ -380,9 +380,13 @@ energy_measurement:
     # If collecting on multiple channels, or another channel than iio:device0,
     # set them here:
     iio_devices: [$IIO_DEVICES]
+
+# Automatically generated ADB device configuration based on test_series options
```
derkling: nit-pick: usually a `device_config` section comes at the beginning of an agenda file.
Can you move this at the beginning of the `config_local.yaml`?

qperret: Done

PR #521, opened 2017-12-07
Executor self tests

bjackman:
This just adds an extremely simple test, run on a local target, to ensure that the Executor is running to a bare minimum of functionality.

If more in-depth tests need to be added, the `mock` module can be used like in [this example](https://github.com/ARM-software/workload-automation/blob/next/tests/test_runtime_param_utils.py#L11)


libs/utils/executor.py
```diff
@@ -335,26 +336,30 @@ def _setup_rootfs(self, tc):
...
+                self._old_selinux_mode = self.target.execute('getenforce')
+            except TargetError:
+                # Probably the target doesn't have SELinux. No problem.
+                pass
+            else:
+
+                self._log.warning('Setting target SELinux in permissive mode')
+                self.target.execute('setenforce 0', as_root=True)
+        else:
+            self._log.warning('Not mounting tempfs because no root')
```
mdigiorgio: nit: s/tempfs/tmpfs/

bjackman: Fixed

PR #525, opened 2017-12-12
Add more PELT tests

mdigiorgio:
This series adds:

- a couple of improvements to the `_LoadTrackingBase` base class
- a test for PELT tasks' utilization signals while using `schedutil` as governor
- a test for CFS runqueues' utilization

### DIFF
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -24,6 +24,7 @@ from bart.sched import pelt
 from devlib.utils.misc import memoized
 from env import TestEnv
 from executor import Executor
+from math import ceil
 from test import LisaTest, experiment_test
 from time import sleep
 from trace import Trace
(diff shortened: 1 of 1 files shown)


valschneider:
I believe this is redundant with what @Elieva has been working on, am I correct ?

derkling:
@Elieva, can you please have a look at the above patches to check if some of the proposed fixes/updates are still relevant after your #556 patches? Thanks.

Elieva:
I think the third commit(tests: load_tracking: add CFS runqueues' utilization test ) is similar to #551 because they both test the same behavior (task that migrates between cpus).
But the two other commits seem unrelated to the patches #556 and #551

valschneider:
I think this is mostly superseded by the abstraction done in `analysis/load_tracking.py`. Can this be closed?

douglas-raillard-arm:
Closing as the base code has changed too much and the PR would need reimplementation

PR #526, opened 2017-12-13
tests/eas fixes

valschneider:
This PR brings some fixes to the expected_placement plotter in generic.py and also restores the old working directory for AndroidTarget as a temp fix to https://github.com/ARM-software/devlib/issues/225

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -40,7 +40,6 @@ from platforms.pixel_energy import pixel_energy
 
 USERNAME_DEFAULT = 'root'
 PASSWORD_DEFAULT = ''
-WORKING_DIR_DEFAULT = '/data/local/schedtest'
 FTRACE_EVENTS_DEFAULT = ['sched:*']
 FTRACE_BUFSIZE_DEFAULT = 10240
 OUT_PREFIX = 'results'
@@ -164,7 +163,7 @@ class TestEnv(ShareState):
         self.test_conf = {}
         self.target = None
         self.ftrace = None
-        self.workdir = WORKING_DIR_DEFAULT
+        self.workdir = None
         self.__installed_tools = set()
         self.__modules = []
         self.__connection_settings = None
@@ -518,6 +517,7 @@ class TestEnv(ShareState):
             self.target = devlib.LinuxTarget(
                     platform = platform,
                     connection_settings = self.__connection_settings,
+                    working_directory = self.workdir,
                     load_default_modules = False,
                     modules = self.__modules)
         elif platform_type.lower() == 'android':
@@ -525,12 +525,14 @@ class TestEnv(ShareState):
             self.target = devlib.AndroidTarget(
                     platform = platform,
                     connection_settings = self.__connection_settings,
+                    working_directory = self.workdir,
                     load_default_modules = False,
                     modules = self.__modules)
         elif platform_type.lower() == 'host':
             self._log.debug('Setup HOST target...')
             self.target = devlib.LocalLinuxTarget(
                     platform = platform,
+                    working_directory = self.workdir,
                     load_default_modules = False,
                     modules = self.__modules,
                     connection_settings = {'unrooted': True})
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -489,6 +489,11 @@ class TestEnv(ShareState):
         if platform_type.lower() == 'android':
             self.__connection_settings = None
             device = 'DEFAULT'
+
+            # Workaround for ARM-software/devlib#225
+            if not self.workdir:
+                self.workdir = '/data/local/tmp/devlib-target'
+
             if 'device' in self.conf:
                 device = self.conf['device']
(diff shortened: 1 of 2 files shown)


tests/eas/generic.py
```diff
@@ -29,8 +29,6 @@
 from unittest import SkipTest
 
 import matplotlib.pyplot as plt
-import numpy as np
-import pylab as pl
 import os
```
mdigiorgio: Since you are at it you could do a general cleanup of the imports, which means having something like:

```python
import one
import two
# etc.

from foo import bar
# etc.
```

valschneider: Done !

libs/utils/env.py
```diff
@@ -525,6 +525,7 @@ def _init_target(self, force = False):
             self.target = devlib.AndroidTarget(
                     platform = platform,
                     connection_settings = self.__connection_settings,
+                    working_directory = '/data/local/tmp/devlib-target',
```
derkling: While looking at this, I remembered we had already a 'workdir' configuration option, which is configurable via target.conf for example, which AFAIR was there exactly for that scope... however, perhaps due to some refactoring, it's now defined but never used :-1: 

You can see it's definition at line 224 above.

Can you please wire this up to this existing configuration option?

Mind that there is also a WORKING_DIR_DEFAULT, which I don't like so much.... we should probably instead have something like:
```python
elif platform_type.lower() == 'android':
        # Workaround for ARM-software/devlib#225
        self.workdir = conf.get('workdir', '/data/local/tmp/devlib-target')
        self._log.debug('Setup ANDROID target...')
        self.target = devlib.AndroidTarget(...)
#...
else:
             raise ValueError('Config error: not supported [platform] type {}'\
                     .format(platform_type))

self.workdir = self.target.working_directory
```
where this last statement is just to ensure to always have the two variables aligned.

tests/eas/generic.py
```diff
@@ -322,7 +319,7 @@ def plot_expected_util(self, experiment, util_df):
                 prev = time
 
         figname = os.path.join(experiment.out_dir, 'expected_placement.png')
-        pl.savefig(figname, bbox_inches='tight')
+        fig.savefig(figname, bbox_inches='tight')
```
derkling: This should better be on a different patch... either by itself, or eventually squased into "tests/eas/generic: Close exp_placement figure" which seems another fix... in this last case please update the commit message too.

libs/utils/env.py
derkling: In the example above... of course in the Android case you pass in `self.workdir` as `working_directory` in the AndroidTarget ctor ;-)

tests/eas/generic.py
valschneider: AFAIR this change allows us to drop the `pylab` import, hence why it's in this commit.

libs/utils/env.py
valschneider: Devlib already defines default working directories for each of its target types - could I just drop the `WORKING_DIR_DEFAULT` thing ? That way I can just make `self.workdir` default to `None` and let devlib handle it.

tests/eas/generic.py
derkling: Well, in that case, since we already use `pl.savefig` all around in the analysis modules... perhaps removing the pylab module perhaps should be a self contained patch which does the same all around. Do you think that's possible? At least we keep consistency on how we save generated plots...

valschneider: Ok, I'll keep that for another PR.

valschneider:
* Reinstored the use of `pl.savefig` in tests/generic.py
* Reinstored the use of `self.workdir` in env.py

tests/eas/generic.py
```diff
@@ -306,7 +306,6 @@ def plot_expected_util(self, experiment, util_df):
                 color = 'red'
 
             tdf.plot(ax=ax[cpu], drawstyle='steps-post', title="CPU{}".format(cpu), color=color)
-            ax[cpu].fill_between(tdf.index, tdf, 0, step='post', color=color)
```
derkling: Does it make sense to test and use the method if the backend supports it?
I guess this allows to get better pretty locking plots when we run tests from out desktop....

valschneider: I had a quick look for that and it didn't seem very intuitive - that "step" parameter is part of the `fill_between` API so imho it should be supported by all of them. I have a post-it somewhere to remind me to look at that again sometime in the future...

derkling: What about just a simple try: except pass?

valschneider: Not too hot on that but we could do it. 

> I guess this allows to get better pretty locking plots when we run tests from out desktop....

For that you'll also have to change the backend definied in tests/__init__.py

derkling: As per chat we had, seems that fixing this is not enough to allow the removal of the `Agg` backend.
So, let's got with this version.

PR #527, opened 2017-12-15
OPP transitions data frame getters and plots

mdigiorgio:


libs/utils/analysis/frequency_analysis.py
```diff
@@ -110,6 +110,70 @@ def _dfg_cluster_frequency_residency(self, cluster, total=True):
...
+
+        :returns: :mod:`pandas.DataFrame` - number of frequency transitions
+        """
+        if not self._trace.hasEvents('cpu_frequency'):
+            self._log.warn('Events [cpu_frequency] not found, '
+                           'cannot compute frequency transitions!')
+            return None
+
+        if isinstance(entity, str):
+            cpu = self._platform['clusters'][entity.lower()][0]
```
derkling: Have a check, AFAIR now self._platform can be null since we can build a Trace() object without passing in platform data. Thus, eventually, you should have a check here.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -110,6 +110,70 @@ def _dfg_cluster_frequency_residency(self, cluster, total=True):
...
+        else:
+            cpu = entity
+
+        freq_df = self._dfg_trace_event('cpu_frequency')
+        cpu_freqs = freq_df[freq_df.cpu == cpu].frequency
```
derkling: We are trying to maintain a "_df" suffix for all DataFrames.
This is still a DataFrame, thus we should keep the "_df" suffix... or just reuse `freq_df`

libs/utils/analysis/frequency_analysis.py
```diff
@@ -110,6 +110,70 @@ def _dfg_cluster_frequency_residency(self, cluster, total=True):
...
+
+        Requires cpu_frequency events to be available in the trace.
+
+        :param entity: a CPU ID or a cluster name
+        :type entity: int or str
+
+        :returns: :mod:`pandas.DataFrame - number of frequency transitions per
+            second
+        """
+        if not self._trace.hasEvents('cpu_frequency'):
```
derkling: This check is already performed by the following call at line 170.

I would remove this one and make more generic the previous warn at line 127 by just saying something like:
```python
        if not self._trace.hasEvents('cpu_frequency'):
            self._log.warn('Events [cpu_frequency] not found, '
                           'frequency data not available')
```

libs/utils/analysis/frequency_analysis.py
```diff
@@ -511,6 +511,88 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+        :param pct: plot frequency transitions in percentage
+        :type pct: bool
+        """
+        if not self._trace.hasEvents('cpu_frequency'):
+            self._log.warn('Events [cpu_frequency] not found, plot DISABLED!')
+            return
+
+        # Sanitize clusters
+        if clusters is None:
+            _clusters = self._platform['clusters'].keys()
```
derkling: Same note here on `self._platform` being possibly null...

libs/utils/analysis/frequency_analysis.py
```diff
@@ -511,6 +511,88 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+                axes.set_xticklabels([])
+
+        if pct:
+            figtype = "_pct"
+            labeltype = " [%]"
+        else:
+            figtype = ""
+            labeltype = ""
+
+        axes0 = fig.axes[0]
```
derkling: Why not just `axes`, or `ax`, or `_axes`... that 0 really seems odd

libs/utils/analysis/frequency_analysis.py
mdigiorgio: `cpu_freqs` is actually a series that is why I didn't put the `_df` suffix. Should I use `_s`?

mdigiorgio:
Update the PR according to @derkling's comments.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -447,6 +510,96 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
 
+    def plotFrequencyTransitions(self, clusters=None, pct=False):
+        """
+        Plot frequency transitions count of a given cluster.
+
+        Requires cpu_frequency events to be available in the trace.
+
+        :param clusters: name of the clusters to be plotted (all of them by
+            default)
+        :type clusters: str ot list(str)
```
valschneider: s/ot/or/

libs/utils/analysis/frequency_analysis.py
```diff
@@ -110,6 +110,69 @@ def _dfg_cluster_frequency_residency(self, cluster, total=True):
...
+        :returns: :mod:`pandas.DataFrame` - number of frequency transitions
+        """
+        if not self._trace.hasEvents('cpu_frequency'):
+            self._log.warn('Events [cpu_frequency] not found, '
+                           'frequency data not available')
+            return None
+
+        if isinstance(entity, str) or isinstance(entity, unicode):
+            if not self._platform or 'clusters' not in self._platform:
+                self._log.warn('No platform data available. Cannot gather {} '
```
valschneider: Do we really need the cluster topology from the platforms file ? AFAICS we are still able to get the transition data with just a CPU id.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -447,6 +510,96 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+                               'plot DISABLED!'.format(c))
+                return
+
+            if pct:
+                tot = t.transitions.sum()
+                t = t.apply(lambda x: x * 100.0 / tot)
+
+            transitions[c] = t
+            max_cnt = t.transitions.max()
+            if xmax < max_cnt:
```
valschneider: Nitpick: you could squash that into a one-liner: `if xmax < max_cnt: xmax = max_cnt`

libs/utils/analysis/frequency_analysis.py
```diff
@@ -447,6 +510,96 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+                xmax = max_cnt
+
+        yrange = 3 * n_plots
+        for idx, c in enumerate(_clusters):
+            t = transitions[c]
+
+            axes = fig.add_subplot(gs[idx])
+            if pct:
```
valschneider: Could we keep the behaviour of having one bar per frequency with the `pct=True` ? I'm getting staggered bars on a single plot, which makes it hard to figure out the percentage for all frequencies past the first one.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -447,6 +510,96 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+            labeltype = ""
+
+        axes = fig.axes[0]
+        legend_y = axes.get_ylim()[1]
+        axes.annotate('OPP Transitions{}'.format(labeltype),
+                      xy=(0, legend_y), xytext=(-50, 25),
+                      textcoords='offset points', fontsize=18)
+        fig.axes[-1].set_xlabel('Number of transitions{}'.format(labeltype))
+
+        figname = '{}/{}cluster_freq_transitions{}.png'.format(
```
valschneider: Could you use `os.path.join` for this ? i.e.:
```
figname = '{}cluster_freq_transitions{}.png'.format(self._trace.plots_prefix, figtype)
fig.savefig(os.path.join(self._trace.plots_dir, figname), bbox_inches='tight')
```

libs/utils/analysis/frequency_analysis.py
mdigiorgio: Do you mean we should ask the user to always input a CPU ID? That would make the code a lot cleaner actually.

valschneider: Maybe you could go the same way as `plot{Cluster, CPU}Frequencies()`, i.e. have a method that expects a CPU ID (plots all by default) and another one that uses the platform data to do a pretty display with one plot per cluster.

mdigiorgio:
Addressed @valschneider's comments.

libs/utils/analysis/frequency_analysis.py
```diff
@@ -447,6 +495,183 @@ def plotClusterFrequencyResidency(self, clusters=None,
...
+            yrange = 3 * n_plots
+            figtype = ""
+            labeltype = ""
+
+        for idx, c in enumerate(_clusters):
+            t = transitions[c]
+
+            axes = fig.add_subplot(gs[idx])
+            if pct:
+                t.T.plot.barh(ax=axes, figsize=(16, yrange),
```
valschneider: Sooo about that... I was about to say now that we have one bar per frequency, it should also scale to the max percentage encountered, but then we end up with exactly the same plot as with `pct=False`, the only difference is that the ticks represent percentages...

As such I think you may want to forget my previous comment and revert to the old behaviour. Another reason to do that is that **plotClusterFrequencyResidency** behaves this way as well. Sorry about that :l

mdigiorgio: Ah OK, I thought that had been changed.

Will update the PR in a bit.

Issue #529, opened 2017-12-18
nrg_model confusion


cdleonard:
The ["integrating a new board in LISA"](https://github.com/ARM-software/lisa/wiki/Integrating-a-new-board-in-LISA) page claims only a few fields are required but trying to run with such a bare config results in confusing problems (python exceptions).

In particular it seems the TestEnv.platform dict needs an "nrg_model" which seems to partially duplicate EAS energy-costs from DTS? In TestEnv._init_target there is apparently an attempt to read the energy model from target. Unfortunately TestEnv.nrg_model seems to be different from TestEnv.platform['nrg_model'] and the latter can only be loaded from json inside _init_platform.

There are various check for 'nrg_model' in TestEnv.platform in libs/utils/trace.py which makes me think this is optional. But looking at TestEnv._init_platform when init fails it just leaves a None value in the platform dict. I had some luck with the following patch:

```
diff --git libs/utils/env.py libs/utils/env.py
index 013f44f..5ea783b 100644
--- libs/utils/env.py
+++ libs/utils/env.py
@@ -797,6 +797,9 @@ class TestEnv(ShareState):
         else:
             self.platform['nrg_model'] = self._load_em(self.conf['board'])
 
+        if self.platform['nrg_model'] is None:
+            del self.platform['nrg_model']
+
         # Adding topology information
         self.platform['topology'] = self.topology.get_level("cluster")

```

This seems like an ugly hack.

Perhaps the EnergyModel object fetched from the target should be converted to the format that is expected inside platform['nrg_model']? But ideally when no nrg_model is available the whole framework should still run, yes?

derkling:
Hi Leonard, thanks for this bug report!

The energy model story within LISA have evolved along time in two different concepts.

Originally there was only `self.platform['nrg_model']`, which is loaded from JSON by TestEnv._load_em(self.conf['board']). As you correctly point out, that's just a "simplified" EM which reports few key values (e.g. max capacity of LITTLE CPUs) and which is used mainly to embellish some of the plots we can generate with the analysis module.

More recently, we added the support to read an EM from a device's DT, thus generating a complete Python representation of the in-kernel EM data. This representation however does not currently support serialization on disk and it's used mainly to support EAS related tests (i.e. `tests/eas/generic.py`).
When tests are executed we load the EM from the device and use it to validate scheduler decisions based on the optimal expected tasks allocation.
[...]

PR #531, opened 2017-12-21
Fix plotting when no nrg_model is present in platform.json

cdleonard:
Fixes some issues reported in #529


cdleonard:
Looking at Travis CI log it seems the second patch is actually broken on homogeneous systems (without clusters).

cdleonard:
I'll post the patches separately.

PR #532, opened 2017-12-21
TestEnv._init_platform: Remove nrg_model key from platform if None

cdleonard:
There are checks like if "nrg_model" in platform but they return True if
a None value is present. So make such that such a value is never
present.

Signed-off-by: Leonard Crestez <leonard.crestez@nxp.com>

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -795,7 +795,10 @@ class TestEnv(ShareState):
             self.platform['nrg_model'] = self.conf['nrg_model']
         # Try to load the default energy model (if available)
         else:
-            self.platform['nrg_model'] = self._load_em(self.conf['board'])
+            nrg_model = self._load_em(self.conf['board'])
+            # We shouldn't have an 'nrg_model' key if there is no energy model data
+            if nrg_model:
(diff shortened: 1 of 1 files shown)


valschneider:
Hi, sorry for the late reply !

I thought that case had been handled by 5434ab348b3c36dd70b8cc33ae1de13872f544eb but we always set `env.platform['nrg_model']` to the return value of `env._load_em()`, so as you pointed out even if the json entry isn't there the dict key will be added and all of those if conditions will be evaluated to True.

I just have two nitpicks regarding your commit:

1) What about simply turning
```
self.platform['nrg_model'] = self._load_em(self.conf['board'])
```
into
```
nrg_model = self._load_em(self.conf['board'])
# We shouldn't have an 'nrg_model' key if there is no energy model data
if nrg_model: self.platform['nrg_model'] = nrg_model
```

2) We usually prefix our commit headers with the directory where the modifications take place, could you please change yours to "utils/env: Remove nrg_model key from platform if None" ?

Thanks for the submission :)

cdleonard:
Pushed a new version based on your comments.

PR #535, opened 2018-01-03
ipynb/tests: Add notebook for EAS integration boot tests

mdigiorgio:


valschneider:
LGTM overall, I do however have some thoughts/suggestions:

This looks a whole lot like the **tests/eas/preliminary.py** tests, only for the integration branch. Perhaps your tests would be better off in a python script than in a notebook ? I believe it makes more sense for tests, and also it means we could point our CI to those tests for our integration branch jobs. I think for now it would have to live in a different class (maybe in a new **tests/integration** or **tests/mainline** directory in LISA ?), but it could potentially be squashed in **preliminary.py** once mainline/android stuff (hopefully) converge.

One advantage using the LisaTest class for those tests is that it will read from the **target.conf** where the targetted platform should be described, so you can get your board name (tc2, hikey960, etc) from there. As such, the integration test shouldn't require any extra argument, just like the **preliminary.py** or **generic.py** stuff.

Now for the actual code comments:

- In [3]: Can this be dropped ? I don't see it used anywhere
- In [9]: I believe you should OR the flags in variables and pass those variables to the dict, because that's quite a lot of copy/pasting right there.
- In [14]: I see that you only test against CPU0, maybe in some cases we should test against one LITTLE and one big CPU ? Probably not worth it for now though.

mdigiorgio:
Yes, at some point we will want to have the tests as script in LISA rather than a notebook. The reasons why I decided to go for the notebook solution are:

- It's much faster to develop and debug issues
- It's easier to run the same notebook on the different platforms

Regarding your comments:

- In [3] can be dropped yes
- In [9] I don't think it's easy to create such variables, what would a meaningful name be for those? Sometimes same sched domain levels have different flags, how do you distinguish those? I prefer the more verbose way.
- In [14] I agree, it's something we could extend later

valschneider:
> - It's much faster to develop and debug issues
> - It's easier to run the same notebook on the different platforms

Agreed on both points, but from a user perspective I prefer the script way. This does bring the problem of changing **target.conf** to test different boards, but it shouldn't be too hard to import the script in a notebook and change the **target.conf** from there (and since it's JSON it's quite easy).

>In [9] I don't think it's easy to create such variables, what would a meaningful name be for those? Sometimes same sched domain levels have different flags, how do you distinguish those? I prefer the more verbose way.

The juno, tc2 and hikey960 all have the same flags for the DIE/MC domains, so I was thinking of something like `sd_flags_bl = {0: <flag ORing here>, 1: <flag ORing here>}` since we know we have that DIE/MC setup. For other platforms like Hikey we could have an equivalent `sd_flags_smp`, or just keep it as is. It's mostly nitpicking tbh, and if you expect those flags to differ on different BL platforms then you could keep things as they are.

mdigiorgio:
On 04/01/18 11:18, Valentin Schneider wrote:
>       * It's much faster to develop and debug issues
>       * It's easier to run the same notebook on the different platforms
> 
> Agreed on both points, but from a user perspective I prefer the script 
> way. This does bring the problem of changing *target.conf* to test 
> different boards, but it shouldn't be too hard to import the script in a 
> notebook and change the *target.conf* from there (and since it's JSON 
> it's quite easy).
> 
>     In [9] I don't think it's easy to create such variables, what would
>     a meaningful name be for those? Sometimes same sched domain levels
>     have different flags, how do you distinguish those? I prefer the
>     more verbose way.
> 
> The juno, tc2 and hikey960 all have the same flags for the DIE/MC 
> domains, so I was thinking of something like |sd_flags_bl = {0: <flag 
> ORing here>, 1: <flag ORing here>}| since we know we have that DIE/MC 
> setup. For other platforms like Hikey we could have an equivalent 
> |sd_flags_smp|, or just keep it as is. It's mostly nitpicking tbh, and 
> if you expect those flags to differ on different BL platforms then you 
> could keep things as they are.

I don't know, flags depend on the platform. We could use your approach 
for the time being and change it later if needed. Thanks.

mdigiorgio:
Updated with some code to get patch statistics.

derkling:
I would really really really like better to have this as a test script... since it has been already developed and debugged it should now be easy to refactor it into a test... ;-)

What I mainly don't like is the huge section of hard-coded data for each different supported platform.
Since we already have platform specific data under: `libs/utils/platforms`, cannot we simply add the required fields to the JSON defined platform data?
This should at least get rid of all the "Expected Data" section as well as to make available these useful information for other possible use-cases (for example I always wanted to have a full EM available in the platform data).

Regarding the test methods, these are easily convertible into a corresponding set of nosetests tests functions, while `test_eas_integration` will disappear since its functionality are already provided by nosetests.

Finally, regarding the idea to test multiple platforms with a single call, that can also be done by adding a complementary test script, similar to the one we already have under `tools/`, e.g.:
    https://github.com/ARM-software/lisa/blob/master/tools/report.py
which can hard-code the same target.conf you already have in the notebook and start nosetests.

What do you think, can we try to have it converted to a nosetest?

qperret:
Have you had a chance to look at this @douglas-raillard-arm ?

douglas-raillard-arm:
@qperret : I'll convert it to a LISA test at some point, until then I'm using the notebook from this branch. It would be better to keep this PR around so I can look at the comments when I do it though.

derkling:
@douglas-raillard-arm: does it makes sense to create a "feature request" ticket, which ask to translate the notebook in this PR into a suitable test?... so that we keep track of the requirement but at the same time we close this PR which we already decided will never be merged?

douglas-raillard-arm:
@derkling: I'd like to keep the comments on this PR accessible to go through them when I implement the LISA test. Is there a way to put a permalink to this PR in the feature request even if this PR is closed ?

derkling:
Sure, by closing this PR we do not lost the comments... just reference #535 from the new issue ticket ;-)

derkling:
This will be now tracked by #569 and thus can be closed.

PR #536, opened 2018-01-04
utils/analysis_module: Initialize _big/little_cpus even w/o nrg_model

cdleonard:
This makes stuff like plotClusterFrequencies work even if no nrg_model
is available, all that is required here is a list of big and small cpus.

Signed-off-by: Leonard Crestez <leonard.crestez@nxp.com>

### DIFF
--- a/libs/utils/analysis_module.py
+++ b/libs/utils/analysis_module.py
@@ -53,6 +53,10 @@ class AnalysisModule(object):
 
         if self._trace.has_big_little:
             self._little_cap = self._platform['nrg_model']['little']['cpu']['cap_max']
+
+        if ('clusters' in self._platform and
+            'big' in self._platform['clusters'] and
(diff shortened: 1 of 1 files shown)


valschneider:
Hi,

I missed the point of your PR and only realised it after writing a big wall of text. That change means things like plotClusterFrequencies() will still create separate plots for big & LITTLE clusters even when we're missing energy model data (which is fine because we don't need this data for that kind of plot).

I understood it as: make things like plotClusterFrequencies() still create separate plots for !big.LITTLE systems.

Anyway, I think this PR points out that we need separate `has_big_little` and `has_energy_model` flags. Ideally we'd only want to keep the `has_energy_model` one once we truly un-hardcode LISA for big.LITTLE platforms. For now, I think this change is fine - I'll try to spend some time on this separate flag story later.

PR #546, opened 2018-01-22
Wa collector speedup

credp:
Just got a bit frustrated with the speed of loading wa results on my machine, so I made a few edits. For me, this significantly reduces loading time and improves the time spent doing reports.

### DIFF
--- a/libs/utils/wa_results_collector.py
+++ b/libs/utils/wa_results_collector.py
@@ -430,11 +430,11 @@ class WaResultsCollector(object):
         """
         # return
         # value,metric,units
-        metrics_df = pd.DataFrame()
+        extra_metric_list = []
 
         artifacts = self._read_artifacts(job_dir)
         if self.parse_traces and 'trace-cmd-bin' in artifacts:
-            metrics_df = metrics_df.append(
+            extra_metric_list.append(
                 self._get_trace_metrics(artifacts['trace-cmd-bin']))
 
         if 'jankbench_results_csv' in artifacts:
@@ -443,7 +443,7 @@ class WaResultsCollector(object):
             df.loc[:, 'metric'] = 'frame_total_duration'
             df.loc[:, 'units'] = 'ms'
 
-            metrics_df = metrics_df.append(df)
+            extra_metric_list.append(df)
 
         # WA's metrics model just exports overall energy metrics, not individual
         # samples. We're going to extend that with individual samples so if you
@@ -473,7 +473,7 @@ class WaResultsCollector(object):
 
                     df.loc[:, 'units'] = 'watts'
 
-                    metrics_df = metrics_df.append(df)
+                    extra_metric_list.append(df)
                 elif 'output_power' in df.columns and 'USB_power' in df.columns:
                     # Looks like this is from a Monsoon
                     # For monsoon the USB and device power are collected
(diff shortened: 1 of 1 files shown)


derkling:
Testing them right now...

libs/utils/wa_results_collector.py
```diff
@@ -203,6 +203,7 @@ def _read_wa_dir(self, wa_dir):
         # |  |    jankbench job)
         # |- pelt-wk1-jankbench-2/
         #      [etc]
+        self._log.info("reading wa dir %s",wa_dir)
```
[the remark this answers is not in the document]
credp: Done

libs/utils/wa_results_collector.py
```diff
@@ -525,6 +526,10 @@ def _get_metric_df(self, workload, metric, tag, kernel, test):
         """
         Common helper for getting results to plot for a given metric
         """
+        lookup = (workload, metric, tag, kernel, test)
```
derkling: Can we use the same `memoized` attribute defined by devlib, i.e.
```python
from devlib.utils.misc import memoized
```

This will remove a lot of code and improve consistency with other LISA's code paths, e.g.:
https://github.com/ARM-software/lisa/blob/master/libs/utils/analysis/latency_analysis.py#L53

libs/utils/wa_results_collector.py
```diff
@@ -491,9 +492,15 @@ def _wa_get_kernel_sha1(self, wa_dir):
         """
         Find the SHA1 of the kernel that a WA3 run was run against
         """
+        if wa_dir in self._kernel_sha1_cache:
```
derkling: Same here: can we just make `_wa_get_kernel_sha1` a memoized method?

libs/utils/wa_results_collector.py
```diff
@@ -484,9 +484,11 @@ def _get_extra_job_metrics(self, job_dir, workload):
                     df.loc[:, 'metric'] = 'device_power_sample'
                     df.loc[:, 'units'] = 'watts'
 
-                    metrics_df = metrics_df.append(df)
-
-        return metrics_df
+                    extra_metric_list.append(df)
+        if len(extra_metric_list) > 0:
```
derkling: What about keeping the empty DF initialization at the beginning and replace these two lines with just:
```python
if len(extra_metric_list):
    metrics_df = metrics_df.append(extra_metric_list)
return metrics_df
```
nit-pick: I would also add an empty line before the if check.

libs/utils/wa_results_collector.py
```diff
@@ -505,10 +506,15 @@ def _wa_get_kernel_sha1(self, wa_dir):
         return sha1
 
     def _select(self, tag='.*', kernel='.*', test='.*'):
+        key = (tag, kernel, test)
```
derkling: devlib's `memoized`?

libs/utils/wa_results_collector.py
```diff
@@ -530,10 +532,13 @@ def tags(self):
         return self.results_df['tag'].unique()
 
     def tests(self, workload=None):
-        df = self.results_df
         if workload:
-            df = df[df['workload'] == workload]
-        return df['test'].unique()
+            if workload not in self._tests:
```
derkling: devlib's `memoized`?

libs/utils/wa_results_collector.py
```diff
@@ -146,9 +146,11 @@ def __init__(self, base_dir=None, wa_dirs=".*", platform=None,
         self.use_cached_trace_metrics = use_cached_trace_metrics
 
         df = pd.DataFrame()
+        df_list = []
         for wa_dir in wa_dirs:
             self._log.info("Reading wa_dir %s", wa_dir)
-            df = df.append(self._read_wa_dir(wa_dir))
+            df_list.append(self._read_wa_dir(wa_dir))
+        df = df.append(df_list)
```
derkling: If by any chance we end up with an empty df_list... we should add a check here, i.e.:
```python
if len(df_list):
    df = df.append(df_list)
```

PR #548, opened 2018-01-25
skip null size power file

vingu-linaro:
Sometimes, the power measurement file is null. Just skip it in this case
because panda doesn't correctly handle that case


qperret:
Out of curisoty, in which case do you have a null power measurement file ?

vingu-linaro:
From time to time, the energy instrument failed to save the power data in the file but the energy file as already been created. I haven't find the root cause of this situation but the occurrence is really low (may be once every 100 iterations)

PR #549, opened 2018-01-31
Fix Juno R2 energy model

douglas-raillard-arm:
1) Fix Juno R2 energy model.
2) Change the optimal task placement algorithm used by LISA to better match EAS behavior.
EDIT: removed 2)

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -410,50 +410,51 @@ class TestEnv(ShareState):
         if 'board' not in self.conf:
             self.conf['board'] = 'UNKNOWN'
 
+        board_name = self.conf['board'].upper()
+
         # Initialize TC2 board
-        if self.conf['board'].upper() == 'TC2':
+        if board_name == 'TC2':
             platform = devlib.platform.arm.TC2()
             self.__modules = ['bl', 'hwmon', 'cpufreq']
 
         # Initialize JUNO board
-        elif self.conf['board'].upper() in ('JUNO', 'JUNO2'):
+        elif board_name in ('JUNO', 'JUNO2'):
             platform = devlib.platform.arm.Juno()
             self.nrg_model = juno_energy
             self.__modules = ['bl', 'hwmon', 'cpufreq']
 
         # Initialize OAK board
-        elif self.conf['board'].upper() == 'OAK':
+        elif board_name == 'OAK':
             platform = Platform(model='MT8173')
             self.__modules = ['bl', 'cpufreq']
 
         # Initialized HiKey board
-        elif self.conf['board'].upper() == 'HIKEY':
+        elif board_name == 'HIKEY':
             self.nrg_model = hikey_energy
             self.__modules = [ "cpufreq", "cpuidle" ]
             platform = Platform(model='hikey')
 
         # Initialize HiKey960 board
-        elif self.conf['board'].upper() == 'HIKEY960':
+        elif board_name == 'HIKEY960':
             self.__modules = ['bl', 'cpufreq', 'cpuidle']
             platform = Platform(model='hikey960')
 
         # Initialize Pixel phone
-        elif self.conf['board'].upper() == 'PIXEL':
+        elif board_name == 'PIXEL':
             self.nrg_model = pixel_energy
             self.__modules = ['bl', 'cpufreq']
             platform = Platform(model='pixel')
 
         # Initialize gem5 platform
-        elif self.conf['board'].upper() == 'GEM5':
-            self.nrg_model = None
+        elif board_name == 'GEM5':
             self.__modules=['cpufreq']
             platform = self._init_target_gem5()
 
-        elif self.conf['board'] != 'UNKNOWN':
+        elif board_name != 'UNKNOWN':
             # Initilize from platform descriptor (if available)
             board = self._load_board(self.conf['board'])
             if board:
-                core_names=board['cores']
+                core_names = board['cores']
                 platform = Platform(
                     model=self.conf['board'],
                     core_names=core_names,
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -34,7 +34,7 @@ from wlgen import RTA
 from energy import EnergyMeter
 from energy_model import EnergyModel
 from conf import JsonConf
-from platforms.juno_energy import juno_energy
+from platforms.juno_r0_energy import juno_r0_energy
 from platforms.hikey_energy import hikey_energy
 from platforms.pixel_energy import pixel_energy
 
@@ -420,8 +420,10 @@ class TestEnv(ShareState):
         # Initialize JUNO board
         elif board_name in ('JUNO', 'JUNO2'):
             platform = devlib.platform.arm.Juno()
-            self.nrg_model = juno_energy
             self.__modules = ['bl', 'hwmon', 'cpufreq']
+
+            if board_name == 'JUNO':
+                self.nrg_model = juno_r0_energy
 
         # Initialize OAK board
         elif board_name == 'OAK':
--- a/libs/utils/platforms/juno_r0_energy.py
+++ b/libs/utils/platforms/juno_r0_energy.py
@@ -90,7 +90,7 @@ def a57_cpu_node(cpu):
                            active_states=a57_cpu_active_states,
                            idle_states=a57_cpu_idle_states)
 
-juno_energy = EnergyModel(
+juno_r0_energy = EnergyModel(
     root_node=EnergyModelRoot(
         children=[
             EnergyModelNode(
--- a/tests/lisa/test_energy_model.py
+++ b/tests/lisa/test_energy_model.py
@@ -16,7 +16,6 @@
 #
 
 from collections import OrderedDict
-import unittest
 from unittest import TestCase
 import os
 import shutil
@@ -27,9 +26,9 @@ from energy_model import (EnergyModel, ActiveState, EnergyModelCapacityError,
 from trace import Trace
 
 # Import these just to test that they can be constructed
-from libs.utils.platforms.juno_energy import juno_energy
-from libs.utils.platforms.pixel_energy import pixel_energy
-from libs.utils.platforms.hikey_energy import hikey_energy
+import libs.utils.platforms.juno_r0_energy
+import libs.utils.platforms.pixel_energy
+import libs.utils.platforms.hikey_energy
 
 """ A very basic test suite for the EnergyModel class."""
 


qperret:
Since we're doing this it might be worth adding the juno r2 model to `libs/utils/platforms/juno_energy.py`.

src/shell/lisa_shell
```diff
@@ -318,14 +321,22 @@ EOF
...
-nosetests -v --nocapture --nologcapture \
-          --logging-config=logging.conf \
-          $*
+if [ -n "$LISA_RESULTS_DIR" ]; then
+  mkdir -p "$LISA_RESULTS_DIR"
+  # There is no sensible place to store results.xml when the directory is not
+  # set explicitely, as the Python scripts decide where to put the results..
+  # Using results_latest symlink is not an option as it would race if using the
+  # same tree to run multiple tests on different boards at the same time.
+  local xunit_options=(--with-xunit --xunit-file="$LISA_RESULTS_DIR/results.xml")
```
valschneider: Ah, that explains the environment variable. We could read the `results_dir` option from the file conf but that means reading JSON in bash...

src/shell/lisa_shell
```diff
@@ -335,24 +346,37 @@ while [[ $# -gt 0 ]]; do
...
+                        shift
+                        ;;
+                -o|--results-dir)
+                        LISA_RESULTS_DIR="$2"
                         shift
                         ;;
                 *)
                         # Unrecognised args are passed through to nosetests
-                        extra_args="$extra_args $1"
+						extra_args+=("$1")
```
valschneider: That's more than one whitespace too far :p

libs/utils/env.py
```diff
@@ -239,17 +248,27 @@ def __init__(self, target_conf=None, test_conf=None, wipe=True,
...
-                        self.conf.get('results_dir'))
-
-        if self.res_dir and not os.path.isabs(self.res_dir):
-            self.res_dir = os.path.join(basepath, 'results', self.res_dir)
-        else:
+        # Initialize local results folder.
+        # The test configuration overrides the target's one and the environment
+        # variable overrides everything else.
+        self.res_dir = (
+            os.getenv('LISA_RESULTS_DIR') or
```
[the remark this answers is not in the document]
douglas-raillard-arm: Setting this using the JSON file has 2 issues:
1) target.config being about the target, the results dir could very well be different when you test your kernel 1 and kernel 2
2) The test "engine" is split in 2 parts: TestEnv and plumbing classes like that, and nosetests itself. The only way to pass information to both (--xunit-file=<path>) is from the lisa_shell script. It is not straightforward to read target.json from lisa_shel without writing a python script to read the JSON and print the value back.

We are also already applying this pattern for --iteration option of lisa-test.

libs/utils/env.py
```diff
@@ -239,17 +248,27 @@ def __init__(self, target_conf=None, test_conf=None, wipe=True,
...
+            self.conf.get('results_dir') or
+            # THIS MUST NOT BE USED. It is only available for a specific test,
+            # but must not be relied upon as the user is able to override that.
+            self.test_conf.get('results_dir')
```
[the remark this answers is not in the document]
douglas-raillard-arm: Actually, this is used in tests/eas/preliminary.py (TEST_CONF) , so more than one test.
It would be a good idea to update it, but it looks like it is not run on a regular basis (bring up test), so the pain is minimal here.

libs/utils/env.py
```diff
@@ -239,17 +248,27 @@ def __init__(self, target_conf=None, test_conf=None, wipe=True,
...
+        )
+        if not self.res_dir:
+            # If the loop terminates without break, it means no res_dir has been
```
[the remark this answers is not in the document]
douglas-raillard-arm: Outdated comment, I rewrote this part indeed :)

libs/utils/test.py
```diff
@@ -271,8 +271,10 @@ def experiment_test(wrapped_test, instance, args, kwargs):
 # runExperiments, so that if the value is invalid we print the error
 # immediately instead of going ahead with target setup etc.
 try:
-    ITERATIONS_FROM_CMDLINE = int(
-        os.getenv('LISA_TEST_ITERATIONS', '0'))
+    iterations = os.getenv('LISA_TEST_ITERATIONS')
```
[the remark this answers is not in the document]
douglas-raillard-arm: This was already there :) , the only change here is making the interface simpler (allowing empty variable)

src/shell/lisa_shell
```diff
@@ -283,6 +284,8 @@ function _lisa-test-usage {
 cat <<EOF
 Usage: lisa-test [--iterations ITERATIONS] [args] FILE[:CLASS]
   Run automated tests. Tests can be found under the tests/ directory.
+  Only one python file can be specified at a time, otherwise the results might
```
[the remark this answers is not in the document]
douglas-raillard-arm: Yes, it is related to the wrong workload being executed for the 2nd file mentioned on the command line.

src/shell/lisa_shell
douglas-raillard-arm: Ah, accidental conversion of spaces to tabs

valschneider:
I'm ok with this split up as it is now. I would however like a more wordy comment for the 20% capacity thing - and ideally I'd like some other people's opinion on the matter (I have already expressed my thoughts). @derkling, @credp, @qperret  ?

qperret:
Thinking about it I must say that I agree with @valschneider on this one. If the optimal placement is to use little CPUs at 99% of their capacity, then this is an issue in EAS and I'd like the tests to tell us that. If we have issues on Juno right now I vote for changing the size of the tasks in the test, and to remove the accounting of the 20% margin form the optimal task placement.

douglas-raillard-arm:
@qperret : The choice we have to make is:
1) _guess_freqs computes the optimal task placement
2) _guess_freqs simulates the expected behavior of EAS

These are two different things, and the current code is a bit confusing since the functions are named as in 1) but are actually expected to behave like 2) by the rest of the code most of the time, since we compare the output of this function with what the actual EAS code did.

Maybe adding a parameter to these functions to know if we want the optimal placement or EAS expected behavior would allow tests to choose the behavior they need.

qperret:
I vote for 1. The optimal placement should be just that: optimal. It should tell us the behaviour we should aim for, not mimic the behaviour we have. If we're too far off the optimal placement energy wise this is a good indication we have to fix something in EAS :)

douglas-raillard-arm:
Ok, so do try and tweak the tests so that they never fall into the grey area were EAS and optimal placement disagree ? If we don't tweak, they will fail (but if we really want to check that EAS is strictly optimal, that is desired). If we do tweak, they will stay away from EAS limits, which means the test will not show small regressions.

If we mirror EAS behavior (via a parameter), we can check for actual regressions relative to the current EAS code, instead of an absolute "regression" compared to the optimal.

douglas-raillard-arm:
Agreed on using the optimal placement and tweaking tests so they fall in the area where EAS is optimal, and testing the grey area behavior in different tests.

libs/utils/platforms/juno_energy.py
```diff
@@ -90,7 +90,7 @@ def a57_cpu_node(cpu):
                            active_states=a57_cpu_active_states,
                            idle_states=a57_cpu_idle_states)
 
-juno_energy = EnergyModel(
+juno_r0_energy = EnergyModel(
```
derkling: Should we not rename the file too? That's for consistency as well as "to make it obvious" as you say in the commit message..

libs/utils/env.py
```diff
@@ -410,50 +410,56 @@ def _init_target(self, force = False):
         if 'board' not in self.conf:
             self.conf['board'] = 'UNKNOWN'
 
+        board_name = self.conf['board'].upper()
```
derkling: I like this simplification... can we just move them into a dedicated separate patch?
Just to keep it more clean the history and better isolate the patch which change the JUNO energy model... you can keep the patch either before or after... whatever is more convenient.

libs/utils/env.py
```diff
@@ -410,50 +410,56 @@ def _init_target(self, force = False):
...
+
         # Initialize TC2 board
-        if self.conf['board'].upper() == 'TC2':
+        if board_name == 'TC2':
             platform = devlib.platform.arm.TC2()
             self.__modules = ['bl', 'hwmon', 'cpufreq']
 
         # Initialize JUNO board
-        elif self.conf['board'].upper() in ('JUNO', 'JUNO2'):
+        elif board_name == 'JUNO':
```
derkling: Cannot we still try to keep one section per "board":
```python
 elif board_name in in ('JUNO', 'JUNO2'):
    # ...
    if board_name == JUNO:
        self.nrg_model = juno_r0_energy
    elif board_name == JUNO2:
        self.nrg_model = juno_energy
    # ...
```

libs/utils/env.py
douglas-raillard-arm: Done.

libs/utils/env.py
douglas-raillard-arm: Done.

libs/utils/platforms/juno_energy.py
douglas-raillard-arm: Done.

libs/utils/env.py
```diff
@@ -420,9 +420,11 @@ def _init_target(self, force = False):
         # Initialize JUNO board
         elif board_name in ('JUNO', 'JUNO2'):
             platform = devlib.platform.arm.Juno()
-            self.nrg_model = juno_energy
             self.__modules = ['bl', 'hwmon', 'cpufreq']
 
+            if board_name == 'JUNO':
+                self.nrg_model = juno_r0_energy
```
derkling: Aren't we now missing and nrg_model for `JUNO2`?
I.e. look at my previous proposal...

douglas-raillard-arm: It is not missing more than it used to be :-) Since I don't need it at the moment (I always extract it from the target), I'll leave that for another PR and for the person that actually needs the feature.

derkling: Ok, I see what you mean... and at the end that's exactly the same behaviour we had initially... actually, better, since before we was loading a wrong EM for JunoR2.

One last question, what happen on R0, do we still override the above hardcoded EM if we are able to read them from the target?

douglas-raillard-arm: That's right. Later in the same function, there is this bit of code, so the energy model is only read from the target if not already provided:
```python
        if not self.nrg_model:
            try:
                self._log.info('Attempting to read energy model from target')
                self.nrg_model = EnergyModel.from_target(self.target)
```

PR #551, opened 2018-02-02
Cpu migration

Elieva:
This patches add support for new tests where the tasks can migrate from one cpu to another during different phase. The goal is to ensure that the utilization is migrated at the same time as the task.
It adds also two scenarios of cpu migration tests.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -26,7 +26,7 @@ from devlib.utils.misc import ranges_to_list
 
 import logging
 
-_Phase = namedtuple('Phase', 'duration_s, period_ms, duty_cycle_pct')
+_Phase = namedtuple('Phase', 'duration_s, period_ms, duty_cycle_pct, cpus')
 class Phase(_Phase):
     """
     Descriptor for an RT-App load phase
@@ -39,6 +39,10 @@ class Phase(_Phase):
 
     :param duty_cycle_pct: the generated load in [%].
     :type duty_cycle_pct: int
+
+    :param cpus: the list of cpus on which task execution is restricted during
+                 this phase.
+    :type cpus: [int] or int
     """
     pass
 
@@ -338,9 +342,10 @@ class RTA(Workload):
                     task_conf['cpus'] = ranges_to_list(task['cpus'])
                 elif isinstance(task['cpus'], list):
                     task_conf['cpus'] = task['cpus']
+                elif isinstance(task['cpus'], int):
+                    task_conf['cpus'] = [task['cpus']]
                 else:
-                    raise ValueError('cpus must be a list or string')
-
+                    raise ValueError('cpus must be a list or string or int')
 
             # Setup task configuration
             self.rta_profile['tasks'][tid] = task_conf
@@ -401,6 +406,18 @@ class RTA(Workload):
 
                 self.rta_profile['tasks'][tid]['phases']\
                     ['p'+str(pid).zfill(6)] = task_phase
+
+                if phase.cpus:
+                    if isinstance(phase.cpus, str):
+                        task_phase['cpus'] = ranges_to_list(phase.cpus)
+                    elif isinstance(phase.cpus, list):
+                        task_phase['cpus'] = phase.cpus
+                    elif isinstance(phase.cpus, int):
+                        task_phase['cpus'] = [phase.cpus]
+                    else:
+                        raise ValueError('phases cpus must be a list or string \
+                                          or int')
+
 
                 pid+=1
 
@@ -578,9 +595,9 @@ class Ramp(RTATask):
         steps = range(start_pct, end_pct+delta_adj, delta_pct)
         for load in steps:
             if load == 0:
-                phase = Phase(time_s, 0, 0)
+                phase = Phase(time_s, 0, 0, cpus)
             else:
-                phase = Phase(time_s, period_ms, load)
+                phase = Phase(time_s, period_ms, load, cpus)
             phases.append(phase)
 
         self._task['phases'] = phases
@@ -671,7 +688,7 @@ class Pulse(RTATask):
         for load in [start_pct, end_pct]:
             if load == 0:
                 continue
-            phase = Phase(time_s, period_ms, load)
+            phase = Phase(time_s, period_ms, load, cpus)
             phases.append(phase)
 
         self._task['phases'] = phases
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -18,6 +18,8 @@
 import os
 import pandas as pd
 import logging
+import json
+from collections import OrderedDict
 
 from bart.common.Utils import select_window, area_under_curve
 from bart.sched import pelt
@@ -28,6 +30,7 @@ from test import LisaTest, experiment_test
 from time import sleep
 from trace import Trace
 from trappy.stats.grammar import Parser
+from wlgen.rta import Periodic, RTA
 from wlgen.utils import SchedEntity, Task, Taskgroup
 
 UTIL_SCALE = 1024
@@ -536,6 +539,249 @@ class PELTTasksTest(_LoadTrackingBase):
         Test load_avg behaviour for a 50% periodic task
         """
         return self._test_behaviour(experiment, tasks, 'load_avg')
+
+class _CPUMigrationBase(LisaTest):
+    """Base class for shared functionality of tasks migration tests"""
+    test_conf = {
+        'tools'    : [ 'rt-app' ],
+        'ftrace' : {
+            'events' : [
+                'sched_switch',
+                'sched_load_cfs_rq',
+            ],
+        },
+        'modules': ['cpufreq', 'cgroups'],
+    }
+
+    # Allowed error margin
+    allowed_util_margin = UTIL_SCALE * 0.02
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        super(_CPUMigrationBase, cls).runExperiments(*args, **kwargs)
+
+    @classmethod
+    def get_wload(cls, test_env, tasks, workload_name):
+        """
+        Create a specification for a rt-app workload from a list of tasks
+        :param test_env: Environment of the test
+        :type test_env: TestEnv
+
+        :param tasks: lists of tasks
+        :type tasks:[RTATask]
+
+        :param workload_name: name of the workload
+        :type workload_name: str
+
+        :returns: the workload specification
+        """
+        rtapp = RTA(test_env.target, workload_name,
+            calibration=test_env.calibration())
+        rtapp.conf(kind='profile',
+            params={'task{}'.format(i): task.get() for i, task in enumerate(tasks)},
+            run_dir=Executor.get_run_dir(test_env.target))
+        return {
+            'migration': {
+                'type' : 'rt-app',
+                'conf' : {
+                    'class' : 'custom',
+                    'json' : rtapp.json,
+                    'prefix' : 'mig_test',
+                },
+            }
+        }
+
+    def _get_one_task(self, tasks, index):
+        """
+        Get a task from the tasks description
+        :param tasks: Tasks description
+        :type tasks: dict
+
+        :param index: the index of the task
+        :type index: int
+
+        :returns: the task that is in the workload at the given index
+        """
+        task_name = tasks.keys()[index]
+        return tasks[task_name]
+
+    def _get_phases_names(self, task):
+        """
+        Return the phases names of a workload
+        :param task: task from which the phases names are retrived
+        :type task: dict
+
+        :returns: a list of the names of the phases
+        """
+        return task['phases'].keys()
+
+    def _get_duration_s(self, task, phase):
+        """
+        Compute the duration of a given phase
+        :param task: Task from which the phase duration is retrived
+        :type task: dict
+
+        :param phase: phase for which the duration is computed
+        :type phase: str
+
+        :returns: the duration in second of the phase
+        """
+        # Get the loop value for a phase
+        loop = task['phases'][phase]['loop']
+        # Get the period value for a phase in us
+        period_us = task['phases'][phase]['timer']['period']
+        return (loop * period_us) / 1e6
+
+    def _get_duty_cycle(self, task, phase):
+        """
+        Compute the duty clycle of a task during a given phase.
+        :param task: task for which the duty clycle is computed
+        :type task: dict
+
+        :param phase: phase for which the duty cycle is computed
+        :type phase: str
+
+        :returns: the duty cycle in percent of a task for a given phase
+        """
+        # Get the period value for a phase in us
+        period_us = task['phases'][phase]['timer']['period']
+        # Get the run time value for a phase in us
+        run_time_us = task['phases'][phase]['run']
+        return run_time_us * 1e2 / period_us
+
+    def _get_cpus(self, task, phase):
+        """
+        Return the cpu assigned to a taks for a given phase.
+        :param task: The task to get the cpu from
+        :type task: dict
+
+        :param phase: The phase to get the cpu from
+        :type phase: str
+
+        :returns: a list of cpus assigned to a task for a given phase
+        """
+        return task['phases'][phase]['cpus']
+
+    def _get_stable_window(self, df, task):
+        """
+        Return a start and end interval for which the utilization signal
+        should be stabilized for each phase.
+        :param df: data frame of the sched_switch events for the tasks
+        :type df: DataFrame
+
+        :param task: The task to obtain phase information from
+        :type task: dict
+
+        :returns: a dictionary which matches a phase with start time and end
+                  time where the utilization signal should be stabilized.
+        """
+        phase_duration_s = 0
+        window = {}
+        start = df.index[1]
+        phases = self._get_phases_names(task)
+        # Compute the stable start and end time for each phase
+        for phase in phases:
+            start += phase_duration_s
+            phase_duration_s = self._get_duration_s(task, phase)
+            window[phase] = (start + UTIL_AVG_CONVERGENCE_TIME,
+                             start + phase_duration_s)
+        return window
+
+    def _get_util_mean(self, start, end, df, cpus):
+        """
+        Compute the mean utilization per CPU given a time interval
+        :param start: start of the time interval
+        :type start: float
+
+        :param end: end of the time interval
+        :type end: float
+
+        :param df: data frame of the sched_load_cfs_rq event
+        :type df: DataFrame
+
+        :param cpus: list of CPUs for which the mean utilization needs to be
+                     computed
+        :type: cpus: [int]
+
+        :returns: a dictionary which matches a cpu with its mean utilization
+        """
+        util_mean = {}
+        df = df[start:end]
+        for cpu in cpus:
+            util = df[(df.cpu == cpu) & (df.path == '/')].util
+            util_mean[cpu] = area_under_curve(util) / (end - start)
+        return util_mean
+
+    def _get_util_expected(self, tasks, cpus, phase):
+        """
+        Compute the expected utilization per cpu given a phase.
+        :param tasks: json description of the tasks
+        :type tasks: dict
+
+        :param cpus: list of CPUs for which the expected utilization needs to
+                     be computed
+        :type cpus: [int]
+
+        :param phase: name of the phase for which the expected utilization
+                      needs to be computed
+        :type phase: str
+
+        :returns: a dictionary which matches the expected utilization to a cpu
+        """
+        expected = {}
+        for cpu in cpus:
+            expected_val = 0
+            for task in tasks.iteritems():
+                [phase_cpu] = self._get_cpus(task[1], phase)
+                if phase_cpu == cpu:
+                    duty_cycle = self._get_duty_cycle(task[1], phase)
+                    expected_val += 1024 * (duty_cycle / 100.0)
+            expected[cpu] = expected_val
+
+        return expected
+
+    def _test_util_per_cpu(self, experiment, tasks):
+        trace = self.get_trace(experiment)
+        if not trace.hasEvents('sched_switch'):
+            raise ValueError('No sched_switch events. '
+                             'Does the kernel support them?')
+        if not trace.hasEvents('sched_load_cfs_rq'):
+            raise ValueError('No sched_load_cfs_rq events. '
+                             'Does the kernel support them?')
+        cpus = set()
+        # Load the JSON tasks description
+        tasks_desc = json.load(open(experiment.wload.params['custom']),
+                               object_pairs_hook=OrderedDict)['tasks']
+        sw_df = trace.data_frame.trace_event('sched_switch')
+        # Filter the event related to the tasks
+        sw_df = sw_df[sw_df.next_comm.isin(tasks_desc.keys())]
+
+        util_df = trace.data_frame.trace_event('sched_load_cfs_rq')
+        phases = self._get_phases_names(self._get_one_task(tasks_desc, 0))
+
+        # Compute the interval where the signal is stable for the phases
+        window = self._get_stable_window(sw_df,
+                                         self._get_one_task(tasks_desc, 0))
+
+        msg = 'Saw util {} on cpu {}, expected {} during phase {}'
+        for phase in phases:
+            # Get all the cpus where tasks are running during this phase
+            for task in tasks_desc.iteritems():
+                cpus.update(self._get_cpus(task[1], phase))
+
+            # Get the mean utilization per CPU
+            util_mean = self._get_util_mean(window[phase][0], window[phase][1],
+                                            util_df, cpus)
+            # Get the expected utilization per CPU
+            expected = self._get_util_expected(tasks_desc, cpus, phase)
+
+            # Check that the expected utilization value and the measured one
+            # match for each CPU.
+            for cpu in cpus:
+                self.assertAlmostEqual(util_mean[cpu], expected[cpu],
+                                       delta=self.allowed_util_margin,
+                                       msg=msg.format(util_mean[cpu], cpu,
+                                                      expected[cpu], phase))
 
 class _PELTTaskGroupsTest(LisaTest):
     """
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -850,6 +850,73 @@ class OneTaskCPUMigrationTest(_CPUMigrationBase):
         """Test util stable range for a migrated 25% task"""
         return self._test_util_per_cpu(experiment, tasks)
 
+class TwoTasksCPUMigrationTest(_CPUMigrationBase):
+    """
+    Goal
+    ====
+    Checks that the utilization is correctly updated for CPU when two tasks are
+    migrated from a CPU to another.
+
+    Detailed Description
+    ====================
+    Two tasks are created. Each task defined two phases as follow (the CPU
+    is the one corresponding for the hikey960 it may be another CPU on
+    another platform):
+
+              Phase 0         Phase 1
+            CPU | Util      CPU | Util
+    Task 0:   4 | 20%         5 | 20%
+    Task 1:   5 | 50%         4 | 50%
+
+    Expected Behaviour
+    ==================
+    The utilization of a CPU is sampled during both phases and should be
+    observed as follow:
+
+            Phase 0         Phase 1
+            Util            Util
+    CPU 4:   20%             50%
+    CPU 5:   50%             20%
+    """
+    @classmethod
+    def _getExperimentsConf(cls, test_env):
+        tasks = []
+        # Get the 2 CPUs with the highest capacities
+        cpus = sorted(test_env.calibration(),
+                      key=test_env.calibration().get)[:2]
+        # tasks creation
+        task0_phase0 = Periodic(duty_cycle_pct=20, duration_s=1, period_ms=16,
+                                cpus=cpus[0])
+        task0_phase1 = Periodic(duty_cycle_pct=20, duration_s=1, period_ms=16,
+                                cpus=cpus[1])
+        tasks.append(task0_phase0 + task0_phase1)
+
+        task1_phase0 = Periodic(duty_cycle_pct=50, duration_s=1, period_ms=16,
+                               cpus=cpus[1])
+        task1_phase1 = Periodic(duty_cycle_pct=50, duration_s=1, period_ms=16,
+                               cpus=cpus[0])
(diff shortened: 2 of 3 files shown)


tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        :param phase: name of the phase for which the expected utilization
+                      needs to be computed
+        :type phase: str
+        :return a list containing the expected utilization for each CPU
+        """
+        expected = []
+        for cpu in cpus:
+            expected_val = 0
+            for task in tasks.iteritems():
+                period_us = task[1]['phases'][phase]['timer']['period']
```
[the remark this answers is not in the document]
derkling: If useful,  you can also add wrapper methods which returns directly a list with values for each phase.

Something like (not sure it works... but gives the idea):
```python
def _get_periods(self, task):
   return [period for 
           phase  in task['phases'] and
           period in task['phases'][phase]['timer']['period']
    ]
```

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+                run_time_us = task[1]['phases'][phase]['run']
+                duty_cycle = run_time_us * 100 / period_us
+                [phase_cpu] = task[1]['phases'][phase]['cpus']
+                expected_val += 1024 * (duty_cycle / 100.0) \
+                                if phase_cpu == cpu else 0
+            expected.append(expected_val)
+
+        return expected
+
+    def _get_cpus(self, tasks):
```
derkling: +1, that's exactly what I'm proposing also above for the other params

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+        :param phase_id: index (starting from 0) of the phase for which the
+                         stable range is computed
+        :type phase: int
+        :return a start time and a end time where the utilization signal should
+                be stabilized.
+        """
+        task_names = tasks.keys()
+        phase_duration_s = 0
+        phase_names = tasks[task_names[0]]['phases'].keys()
```
derkling: Utility get method for this too?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        df = df[df.next_comm.isin(task_names)]
+        start = df.index[1]
+        # Compute the total duration of the previous phases to get the start
+        # time of the given phase
+        for phase in (phase_names[: phase_id+1]):
+            start += phase_duration_s
+            loop = tasks[task_names[0]]['phases'][phase]['loop']
```
derkling: Utility get method for this too?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+            period_us = tasks[task_names[0]]['phases'][phase]['timer']['period']
```
derkling: according to above comments... here you should already have a get method.

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+            phase_duration_s = (loop * period_us) / 1e6
```
derkling: Actually, all this for can be a `_get_durations` utility get method which returns the duration of each phase.

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+        end = start + phase_duration_s - UTIL_AVG_CONVERGENCE_TIME
```
derkling: Why do we trim at the end?... the signal should be stable at that time... and by double trimming we need not necessary longer phases, isn't it?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        sw_df = trace.data_frame.trace_event('sched_switch')
+        util_df = trace.data_frame.trace_event('sched_load_cfs_rq')
+
+        phases = tasks_desc[tasks_desc.keys()[0]]["phases"].keys()
+        cpus = self._get_cpus(tasks_desc)
+
+        msg = 'Saw util {} on cpu {}, expected {} during phase {}'
+        for j in range(len(phases)):
+            # Compute the interval where the signal is stable for a phase
+            stable_start, stable_end = self._get_stable_range(sw_df,
```
derkling: Would use just `start` and `end`... we know the context or you can make it explicit with just a comment.

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+                                                              tasks_desc, j)
+            # Get the mean utilization per CPU
+            util_mean = self._get_util_mean(stable_start, stable_end,
+                                            util_df, cpus)
+            # Get the expected utilization per CPU
+            expected = self._get_util_expected(tasks_desc, cpus, phases[j])
+            for i in range(len(cpus)):
```
derkling: Add empty line and comment before this loop.

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+                error_margin = expected[i] * 0.04
```
derkling: This `0.04` looks a bit arbitrary.. anyway shouldn't it be a test const defined somewhere at the top?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+                self.assertAlmostEqual(util_mean[i], expected[i],
+                                       delta=error_margin,
```
derkling: mmm.... with the above definition of error_margin, are you kind-of defining a margin which is proportional to the actual absolute value?

This means that for smaller utilizations we are more restrictive... however, not sure we can expect a better precision from PELT... maybe a const value is just good enough?

tests/eas/load_tracking.py
```diff
@@ -740,6 +740,77 @@ def _test_util_per_cpu(self, experiment, tasks):
...
+
+            Phase 0         Phase 1
+            Util            Util
+    CPU 4:   35%             10%
+    CPU 5:   5%              30%
+    """
+    @classmethod
+    def _getExperimentsConf(cls, test_env):
+        # Get the 2 CPUs with the highest capacities
+        cap_cpu = sorted(test_env.calibration(),
```
derkling: Why that name? What about just `cpus`... the meaning is already in the comment.

tests/eas/load_tracking.py
```diff
@@ -740,6 +740,77 @@ def _test_util_per_cpu(self, experiment, tasks):
...
+                         key=test_env.calibration().get)[:2]
+        tasks = []
+        # tasks creation
+        task0_phase0 = Periodic(duty_cycle_pct=25, duration_s=1, period_ms=16,
+                                cpus=[cap_cpu[0]])
```
derkling: Would really like to be able to pass in just an `int` when a single CPU is required... can we make wlgen smart enough, if not already, to generated that list when you pass in an `int`?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        Generate a json file which describes the different tasks.
+        :param test_env: Environment of the test
+        :type test_env: TestEnv
+
+        :param tasks: lists of task
+        :type tasks:[RTATask]
+
+        :param json_file_name: name of the json file
+        :type json_file_name: str
+        :return the generate file name without extensin
```
derkling: That's a bit confusing... looking at rtapp.conf() code it just returns a "test label"... and that's not the name the "generated json" file name..

derkling: If that happens to be the same name of the generated json file (to be checked) then why not adding the ".json" extension here... or even better... return the full path of the generated json file.
This would match better with the method name... and simplify callsites code.

tests/eas/load_tracking.py
```diff
@@ -811,6 +811,74 @@ def test_util_task_migration(self, experiment, tasks):
...
+
+            Phase 0         Phase 1
+            Util            Util
+    CPU 4:   20%             50%
+    CPU 5:   50%             20%
+    """
+    @classmethod
+    def _getExperimentsConf(cls, test_env):
+        # Get the 2 CPUs with the highest capacities
+        cap_cpu = sorted(test_env.calibration(),
```
derkling: same comment here as per the other test

tests/eas/load_tracking.py
```diff
@@ -740,6 +740,77 @@ def _test_util_per_cpu(self, experiment, tasks):
...
+
+        task2_phase = Periodic(duty_cycle_pct=5, duration_s=1, period_ms=16,
+                               cpus=[cap_cpu[1]])
+        tasks.append(task2_phase + task2_phase)
+
+        json_file = cls.generate_json(test_env, tasks, 'CPUMigration0')
+        wload = cls.get_wload(json_file+'.json', len(tasks))
+
+        conf = {
+            'tag' : '1task_cpu_migrate',
```
derkling: Can we use a more "namespace like" naming... since these are "cpu_migration" tests... with wither 1 or 2 tasks, what about `cpu_migration_one_task` and `cpu_migrate_two_tasks` (for the other test)?

tests/eas/load_tracking.py
```diff
@@ -811,6 +811,74 @@ def test_util_task_migration(self, experiment, tasks):
...
+        task1_phase0 = Periodic(duty_cycle_pct=50, duration_s=1, period_ms=16,
+                               cpus=[cap_cpu[1]])
+        task1_phase1 = Periodic(duty_cycle_pct=50, duration_s=1, period_ms=16,
+                               cpus=[cap_cpu[0]])
+        tasks.append(task1_phase0 + task1_phase1)
+
+        json_file = cls.generate_json(test_env, tasks, 'CPUMigration1')
+        wload = cls.get_wload(json_file+'.json', len(tasks))
+        conf = {
+            'tag' : '2tasks_cpu_migrate',
```
derkling: `cpu_migrate_two_tasks` ?

libs/wlgen/wlgen/rta.py
```diff
@@ -39,6 +39,9 @@ class Phase(_Phase):
 
     :param duty_cycle_pct: the generated load in [%].
     :type duty_cycle_pct: int
+
+    :param cpus: cpus on which the phase can be executed
```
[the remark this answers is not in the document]
Elieva: Done.
Yes indeed, it is a list of cpus that is expected here or a int if there is only one cpu (see comments for load_tracking.py line 777 to 780)

libs/wlgen/wlgen/rta.py
```diff
@@ -398,6 +403,8 @@ def _confProfile(self):
                         'run': running_time,
                         'timer': {'ref': tid, 'period': period},
                     }
+                    if phase.cpus:
+                        task_phase['cpus'] = phase.cpus
```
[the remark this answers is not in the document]
Elieva: Yes better like that
Done

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+        :param json_file_name: name of the json file
+        :type json_file_name: str
+        :return the generate file name without extensin
+        """
+        rtapp = RTA(test_env.target, json_file_name,
+            calibration=test_env.calibration())
+        return rtapp.conf(kind='profile',
+            params={'task'+str(i): task.get() for i, task in enumerate(tasks)},
+            run_dir=test_env.target.working_directory+'/run_dir/')
```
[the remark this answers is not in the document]
Elieva: Done: use directly test_env.target.working_directory

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        },
+        'modules': ['cpufreq', 'cgroups'],
+    }
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        super(_CPUMigrationBase, cls).runExperiments(*args, **kwargs)
+
+    @classmethod
+    def generate_json(cls, test_env, tasks, json_file_name):
```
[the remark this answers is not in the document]
Elieva: Yes much better :)

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+            run_dir=test_env.target.working_directory+'/run_dir/')
+
+    @classmethod
+    def get_wload(cls, json_file, tasks_nb):
+        """
+        Get a specification for a rt-app workload from a json file
+        :param json_file: path to the file that contains the tasks description
+        :type json_file: str
+
+        :param tasks_nb: number of tasks contained by the file
```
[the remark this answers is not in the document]
Elieva: Done: I removed it. Actually it is used by the test Executor but not in the case of a 'custom' configuration.

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+        :param cpus: list of CPUs for which the mean utilization needs to be
+                     computed
+        :type: cpus: [int]
+        :return a list containing the mean utilization for each CPU
+        """
+        util_mean = []
+        df = df[start:end]
+        for cpu in cpus:
+            util_per_cpu = df[(df.cpu == cpu) & (df.path == '/')].util
```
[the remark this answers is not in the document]
Elieva: Done: Rename util_per_cpu to util and return a dictionary instead of a list

tests/eas/load_tracking.py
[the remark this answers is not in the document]
Elieva: I added an example of task description in the get_wload function
For the getter methods, I have directly added _get_duration and _get_duty_cycle

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        :return a list containing the expected utilization for each CPU
+        """
+        expected = []
+        for cpu in cpus:
+            expected_val = 0
+            for task in tasks.iteritems():
+                period_us = task[1]['phases'][phase]['timer']['period']
+                run_time_us = task[1]['phases'][phase]['run']
+                duty_cycle = run_time_us * 100 / period_us
+                [phase_cpu] = task[1]['phases'][phase]['cpus']
```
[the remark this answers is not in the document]
Elieva: The tests are expected that tasks are running on only one cpu per phase. If tasks can run on several cpus  it is harder to determine what is the expected utilization per cpu as there is uncertainty of where the tasks has been running and for how long. 

Use _get_cpus -> done

tests/eas/load_tracking.py
Elieva: Ok

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: I have directly done the _get_durations method (see comment below)

tests/eas/load_tracking.py
Elieva: I have directly done the _get_duration method (see comment below)

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Agree. Done, I removed the trimming for the end value.

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done, I defined it for the class as other tests seem to have different margin values (some have 5%, some have 2%...)

tests/eas/load_tracking.py
Elieva: Done, I changed it for  a constant value which is 2% of 1024 (the utilization scale). The idea is to not be sensitive to parasitic task that may execute as well on these cpus. It is still a bit arbitrary. May you have a better solution?

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done: it can now be a list, a int or a string (like for the task cpus)

tests/eas/load_tracking.py
Elieva: Done: change to return the json file name from rtapp (rtapp.json).

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done (cpu_migration_two tasks for consistency with previous comment)

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        Generate a json file which describes the different tasks.
+        :param test_env: Environment of the test
+        :type test_env: TestEnv
+
+        :param tasks: lists of task
+        :type tasks:[RTATask]
+
+        :param workload_name: name of the workload
+        :type workload_name: str
+        :return the json file name
```
derkling: Should we add an empty line before this return?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+    allowed_util_margin = UTIL_SCALE * 0.02
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        super(_CPUMigrationBase, cls).runExperiments(*args, **kwargs)
+
+    @classmethod
+    def generate_json(cls, test_env, tasks, workload_name):
+        """
+        Generate a json file which describes the different tasks.
```
derkling: I think we usually (dunno why) keep the first line right after the three `"""`... I think it's a documentation processor "requirement"

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        """
+        rtapp = RTA(test_env.target, workload_name,
+            calibration=test_env.calibration())
+        rtapp.conf(kind='profile',
+            params={'task{}'.format(i): task.get() for i, task in enumerate(tasks)},
+            run_dir=test_env.target.working_directory)
+        return rtapp.json
+
+    @classmethod
+    def get_wload(cls, json_file):
```
derkling: What about squashing the `generate_json` method into this one?
We do not really use the other method without calling this one right after.

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        """
+        Get a specification for a rt-app workload from a json file
+        :param json_file: path to the file that contains the tasks description
+        :type json_file: str
+
+        Example of expected task description:
+        "task0": {
```
derkling: The above comment will allow to remove this documentation section... since we know that we will always get a correct json file... isn't it?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+    def _get_duration_s(self, tasks, phase):
+        """
+        Compute the duration of a given phase
+        :param task: Tasks description
+        :type task: dict
+
+        :param phase: phase for which the duration is computed
+        :type phase: str
+        :return the duration in second of the phase
```
derkling: Space before return?... same for following documentation

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+        :param phase: phase for which the duty cycle is computed
+        :type phase: str
+        :return the duty cycle in percent of a task for a given phase
+        """
+        # Get the period value for a phase in us
+        period_us = task['phases'][phase]['timer']['period']
+        # Get the run time value for a phase in us
+        run_time_us = task['phases'][phase]['run']
+        return run_time_us * 100 / period_us
```
derkling: I assume we are ok with an integer duty-cycle... maybe let's specify it in the above return docstring?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+    def _get_cpu(self, task, phase):
```
derkling: `_get_cpus`?... we can get a list back, isn't it?

... and the next methods should be named maybe `_get_tasks_cpus`?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        :type tasks: dict
+
+        :param phase: phase to get the cpu allocation
+        :return a list of CPUs on which tasks are running for the given phase
+        """
+        cpus = []
+        for task in tasks.iteritems():
+            [cpu] = self._get_cpu(task[1], phase)
+            if cpu not in cpus:
+                cpus.append(cpu)
```
derkling: What about using a `set` in the list? It should take care of duplicates... and then we can convert to a list at return time...

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        :type phase: str
+        :return a dictionary which matches the expected utilization to a cpu
+        """
+        expected = {}
+        for cpu in cpus:
+            expected_val = 0
+            for task in tasks.iteritems():
+                duty_cycle = self._get_duty_cycle(task[1], phase)
+                [phase_cpu] = self._get_cpu(task[1], phase)
+                expected_val += 1024 * (duty_cycle / 100.0) \
```
derkling: What about just:
```python
if hase_cpu == cpu:
   expected_val += 1024 * (duty_cycle / 100.0)
```

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        for phase in phase_names:
+            start += phase_duration_s
+            phase_duration_s = self._get_duration_s(tasks, phase)
+            window[phase] = (start + UTIL_AVG_CONVERGENCE_TIME,
+                             start + phase_duration_s)
+        return window
+
+    def _test_util_per_cpu(self, experiment, tasks):
+        trace = self.get_trace(experiment)
+        if 'sched_switch' not in trace.available_events:
```
derkling: We have an API for that... what about:
```python
if not trace.hasEvents('sched_switch'):
```

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+            raise ValueError('No sched_switch events. '
+                             'Does the kernel support them?')
+        if 'sched_load_cfs_rq' not in trace.available_events:
```
derkling: same here

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,406 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        task1_phase = Periodic(duty_cycle_pct=10, duration_s=1, period_ms=16,
+                               cpus=cpus[0])
+        tasks.append(task1_phase + task1_phase)
+
+        task2_phase = Periodic(duty_cycle_pct=5, duration_s=1, period_ms=16,
+                               cpus=cpus[1])
+        tasks.append(task2_phase + task2_phase)
+
+        json_file = cls.generate_json(test_env, tasks, 'cpu_migration0')
+        wload = cls.get_wload(json_file)
```
derkling: So, with the above comment... this will be a single call to `get_wload`

tests/eas/load_tracking.py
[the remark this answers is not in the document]
Elieva: This was not working because the test executor uses another path for the log directory (os.path.join(test_env.target.working_directory, '/run_dir/'). So to manage to get the value given by the executor a static function has been added to the executor to get this run_dir. It can be now used here to get the good run_dir

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: I have changed duty_cycle to be a float to avoid to increase the error when computing the utilization per cpu by the truncation to integer.

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Yes it is better with a set. It is not needed to be converted back to a list.
Done.

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,367 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+                'type' : 'rt-app',
+                'conf' : {
+                    'class' : 'custom',
+                    'json' : rtapp.json,
+                    'prefix' : 'mig_test',
+                },
+            }
+        }
+
+    def _get_one_task(self, tasks):
```
derkling: Mmm... would say either `_get_first_task` or `_get_one_task(self, tasks, id)`... and I would slightly prefer the second.

Elieva: Done: _get_one_task(self, tasks, id)

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,371 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        :param test_env: Environment of the test
+        :type test_env: TestEnv
+
+        :param tasks: lists of tasks
+        :type tasks:[RTATask]
+
+        :param workload_name: name of the workload
+        :type workload_name: str
+
+        :return the workload specification
```
valschneider: AFAICT the syntax for the "return" statement should be:
`:returns:  the workload specification`
(https://stackoverflow.com/a/24385103)

Not sure if this would create an actual error when generating the doc. I noticed that our sphinx setup doesn't build the doc for the tests, I need to fix that.

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,371 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+    # Allowed error margin
+    allowed_util_margin = UTIL_SCALE * 0.02
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        super(_CPUMigrationBase, cls).runExperiments(*args, **kwargs)
+
+    @classmethod
+    def get_wload(cls, test_env, tasks, workload_name):
+        """Create a specification for a rt-app workload from a list of tasks
```
valschneider: So this is super nitpicky of me and I feel ashamed of asking you to change this but at the same time I feel like I need to say it...

The 2 styles of docstring I've seen in LISA are
```
"""hello world"""
```
and
```
"""
hello world
"""
```
So I think it would be nice if your docstrings fit in either of those 2 categories. Sorry for having an easily-triggered OCD :(

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,371 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+
+class TwoTasksCPUMigrationTest(_CPUMigrationBase):
+    """Goal
+    =======
+    Checks that the utilization is correctly updated for CPU when two tasks are
+    migrated from a CPU to another.
+
+    Detailed Description
+    ====================
+    Three tasks are created. Each task defined two phases as follow (the CPU
```
valschneider: Two tasks ?

tests/eas/load_tracking.py
```diff
@@ -537,6 +540,206 @@ def test_load_avg_behaviour(self, experiment, tasks):
...
+        :type tasks:[RTATask]
+
+        :param json_file_name: name of the json file
+        :type json_file_name: str
+        :return the generate file name without extensin
+        """
+        rtapp = RTA(test_env.target, json_file_name,
+            calibration=test_env.calibration())
+        return rtapp.conf(kind='profile',
+            params={'task'+str(i): task.get() for i, task in enumerate(tasks)},
```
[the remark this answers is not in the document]
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

valschneider:
LGTM, @derkling can we merge this before hitting 100 comments ?

Issue #544, opened 2018-01-18
Reduce the size of results folders with wltest


qperret:
When using sched-evaluation-full-traced.yaml, the size of the result folder for 1 kernel can easily reach 25GB. With just 4 kernels, 100GB are needed. This is because the trace files are large, and contain a lot of  trace events that are rarely necessary. It might be a good idea to collect less trace events in order to reduce the size of the results, and save some significant space when many kernels are tested.

qperret:
Hopefully this was solved by #552

PR #552, opened 2018-02-05
wa/config: specify the list of default trace events to collect

qperret:
By default, WA's list of events to collect is the following:
        ['sched*', 'irq*', 'power*', 'thermal*']
As a result, the trace files are generally massive. This becomes an
issue when sched-evaluation-full-traced is in use as the result folder
can be > 25GB per kernel, mostly because of the trace files.

To avoid this, let's select by default a list of trace events we care
and drop all the others. This global parameter can be overridden in each
agenda if needed by setting 'trace_events' in the global section.

### DIFF
--- a/tools/wa_user_directory/config.yaml
+++ b/tools/wa_user_directory/config.yaml
@@ -4,6 +4,28 @@ device: generic_android
 trace-cmd:
   buffer_size: 102400
   report: false
+  events: [
+    "sched_switch",
+    "sched_wakeup",
+    "sched_wakeup_new",
+    "sched_overutilized",
+    "sched_load_avg_cpu",
+    "sched_load_avg_task",
+    "sched_pelt_se",
+    "sched_load_se",
+    "sched_load_cfs_rq",
+    "sched_load_waking_task",
+    "cpu_capacity",
+    "cpu_frequency",
+    "cpu_idle",
+    "sched_tune_config",
+    "sched_tune_tasks_update",
+    "sched_tune_boostgroup_update",
+    "sched_tune_filter",
+    "sched_boost_cpu",
+    "sched_boost_task",
+    "sched_energy_diff"
+  ]
 
 # Disable re-trying things that go wrong
 max_retries: 0


qperret:
This should address #544

tools/wa_user_directory/config.yaml
```diff
@@ -4,6 +4,25 @@ device: generic_android
 trace-cmd:
   buffer_size: 102400
   report: false
+  events: [
+    "sched_switch",
+    "sched_wakeup",
+    "sched_wakeup_new",
+    "sched_overutilized",
+    "sched_load_avg_cpu",
```
derkling: These are old version format events... can we run on kernels where we use the new set of events for PELT signals?

tools/wa_user_directory/config.yaml
```diff
@@ -4,6 +4,25 @@ device: generic_android
...
+    "sched_load_avg_task",
+    "sched_load_waking_task",
+    "cpu_capacity",
+    "cpu_frequency",
+    "cpu_idle",
+    "sched_tune_config",
+    "sched_tune_tasks_update",
+    "sched_tune_boostgroup_update",
+    "sched_tune_filter",
```
derkling: What in case certain events are not available? Does WA still configure the remaining ones?

tools/wa_user_directory/config.yaml
qperret: Yeah we can add those, I'll align this with what is done in the PELT tests.

tools/wa_user_directory/config.yaml
qperret: You just get a warning like "the event xxx is not available" and all the other ones are collected properly.

tools/wa_user_directory/config.yaml
valschneider: IMO we could just list the old & new ones.

qperret:
I just added the new PELT events to the list. I'll re-run my tests to make sure everything is OK before we merge :)

qperret:
I just ran some test, everything looks fine with the new list of events :+1:

PR #553, opened 2018-02-06
Misc lisa-test improvements

douglas-raillard-arm:
Fix some docstrings and comments
Add more control over lisa-test command behavior.
Make the indentation style of lisa_shell coherent.
Remove ASCII escape codes from lisa_shell output when stdout is not a terminal.


src/shell/lisa_shell
```diff
@@ -318,14 +327,22 @@ EOF
 }
 
 function _lisa-test {
-nosetests -v --nocapture --nologcapture \
-          --logging-config=logging.conf \
-          $*
+if [ -n "$LISA_RESULTS_DIR" ]; then
```
derkling: Can we add at the top a default initialization for the results dir?
Something like:
```bash
LISA_RESULTS_DIR=${LISA_RESULTS_DIR:-'lisa_tests_results'}
```
not sure if the default makes sense... to be verified.

This will allow to skip this `if` and use instead

```bash
[ -d "$LISA_RESULTS_DIR" ] || mkdir -p "$LISA_RESULTS_DIR"
local xunit_options=(--with-xunit --xunit-file="$LISA_RESULTS_DIR/results.xml")
```

src/shell/lisa_shell
```diff
@@ -335,24 +352,37 @@ while [[ $# -gt 0 ]]; do
...
+                        LISA_TEST_ITERATIONS="$2"
+                        shift
+                        ;;
+
+                -t|--target-conf)
+                        LISA_TARGET_CONF="$2"
+                        shift
+                        ;;
+                -o|--results-dir)
+                        LISA_RESULTS_DIR="$2"
```
derkling: As suggested above... wouldn't be nice to have default values for all the above variables?

src/shell/lisa_shell
```diff
@@ -74,12 +74,12 @@ echo -ne "$LISASHELL_DEFAULT"
 ################################################################################
 
 function _lisa-update-usage {
-    echo "Usage: lisa-update (CMD)"
+	echo "Usage: lisa-update (CMD)"
```
derkling: This means you are not using VIM? :)

... since this conflicts with the modeline at the end :/

src/shell/lisa_shell
```diff
@@ -281,13 +282,21 @@ fi
 
 function _lisa-test-usage {
 cat <<EOF
-Usage: lisa-test [--iterations ITERATIONS] [args] FILE[:CLASS]
+Usage: lisa-test [--iterations ITERATIONS] [--target-conf CONF_FILE]
```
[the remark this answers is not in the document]
douglas-raillard-arm: I don't have strong feelings, ssh man page does it one way, grep man page another way. Let's go for grep's style since it is easier to maintain :-)

src/shell/lisa_shell
```diff
@@ -281,13 +282,21 @@ fi
...
+	  [--results-dir RESULTS_DIR] [args] FILE[:CLASS]
+
   Run automated tests. Tests can be found under the tests/ directory.
+  Only one python file can be specified at a time, otherwise the results might
+  not be valid.
```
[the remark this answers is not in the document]
douglas-raillard-arm: Done

src/shell/lisa_shell
douglas-raillard-arm: That is unfortunately not possible. This directory is by default based on the date, so no fixed default is acceptable. If there is no fixed default, there is no way for env.py to know if this variable is set by the user of set as a default value, unless you add an extra variable LISA_RESULTS_DIR_IS_DEFAULT=0/1  to signal that. If you don't have this information, the Python code can't decide whether to use LISA_RESULTS_DIR or target.config settings.

src/shell/lisa_shell
douglas-raillard-arm: Since they are read using os.getenv in the Python scripts, you can assume they default to an empty string.

* LISA_TEST_ITERATIONS: the behavior has not changed
* LISA_TARGET_CONF: The default value is set by Python code. Maybe we should push these defaults to lisa_shell when possible and assume the environment variables are set in the Python code. 
* LISA_RESULTS_DIR: discussed above

src/shell/lisa_shell
douglas-raillard-arm: The modeline at the end is:
`# vim: set tabstop=4:`
From Vim doc: "Number of spaces that a Tab in the file counts for."

This does not tell vim that you need to use tabs or spaces, and since you introduced this mixture in 33108ec6, you probably want to review these bits of configuration in your setup :-)
Currently, there is no way to avoid review comments: either you use spaces and somebody will notice a nearby line that uses tabs, or you use tabs and somebody will notice a nearby line uses spaces, so either we don't care at all or we choose either tabs or spaces.

libs/utils/env.py
```diff
@@ -64,16 +64,20 @@ class TestEnv(ShareState):
       want to use to run the experiments
     - a test configuration (test_conf) defining which SW setups we need on
       that HW target
-    - a folder to collect the experiments results, which can be specified
-      using the test_conf::results_dir option and is by default wiped from
-      all the previous contents (if wipe=True)
+    - a folder to collect the experiments results, which can be specified using
```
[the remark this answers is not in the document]
derkling: I guess the configuration depends on the editor you use... the modeline at the end of the file should works out-of-the-box if you edit the file in vim...

libs/utils/env.py
```diff
@@ -203,8 +207,13 @@ def __init__(self, target_conf=None, test_conf=None, wipe=True,
             self._log.info('Loading custom (file) target configuration')
             self.conf = self.loadTargetConfig(target_conf)
         elif target_conf is None:
-            self._log.info('Loading default (file) target configuration')
-            self.conf = self.loadTargetConfig()
+            env_target_conf = os.environ.get('LISA_TARGET_CONF')
```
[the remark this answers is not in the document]
derkling: Not sure to get your point here... what I'm proposing above is just to use the ENV variable to override the target.conf file to use.

libs/utils/env.py
```diff
@@ -239,17 +248,26 @@ def __init__(self, target_conf=None, test_conf=None, wipe=True,
...
-        if self.res_dir and not os.path.isabs(self.res_dir):
-            self.res_dir = os.path.join(basepath, 'results', self.res_dir)
-        else:
+        # Initialize local results folder.
+        # The test configuration overrides the target's one and the environment
+        # variable overrides everything else.
+        self.res_dir = (
+            os.getenv('LISA_RESULTS_DIR') or
+            self.conf.get('results_dir') or
+            # THIS MUST NOT BE USED. It is only available for a specific test,
```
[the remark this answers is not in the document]
derkling: Since we have control on all tests... I'm more of the idea to have a single API, which is an improvement of the current one, and update the tests to use the new one.

src/shell/lisa_shell
derkling: Right... I think that's my local setup where I also set expand tab... but... anyway... if that's a fix across all the file I'm still ok with it... just keep it on a dedicated patch (which you already do) and check that what you get is still a reasonable indentation... for example lines 253-255 in fb164a7 seems to be a bit ugly

src/shell/lisa_shell
derkling: cannot we have the current data expanded as a default in the above initialisation?

I mean... something like:
```bash
LISA_RESULTS_DIR=${LISA_RESULTS_DIR:-$(date +%Y%m%d)'}
```
where you use the format string used somewhere else to initialize that folder?

libs/utils/env.py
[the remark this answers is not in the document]
douglas-raillard-arm: My point was that the code in my original commit was already wrapped at 80 columns. The modeline does not include textwidth, so it does not set any wrapping. We can add it though.

libs/utils/env.py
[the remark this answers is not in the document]
douglas-raillard-arm: I'll change self.loadTargetConfig so that these 2 are equivalents:
```self.loadTargetConfig()
self.loadTargetConfig("")
```
This way, this code will work as you expect it:
```
env_target_conf = os.environ.get('LISA_TARGET_CONF', '')
self.conf = self.loadTargetConfig(env_target_conf)
```

libs/utils/env.py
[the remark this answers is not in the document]
douglas-raillard-arm: Ok, I'll update these tests so they don't rely on this key in test conf.

src/shell/lisa_shell
douglas-raillard-arm: If you do that, what behavior do you want when you set "results-dir" in target.config ? The only solution would be to have target.config winning over the environment variable. This is broken since the XML file would end up in the wrong directory in that case. Going that way means also removing the option from target.config ...

src/shell/lisa_shell
douglas-raillard-arm: Right, I missed that one and line 340

douglas-raillard-arm:
@qperret : it does not fix this issue, however I quite likely have the fix but I need to test it better before issuing the PR.

src/shell/lisa_shell
derkling: +1

PR #554, opened 2018-02-07
Add peripheral clock name info to the platform descriptor.

kdub:
This allows for users to designate clocks-of-interest for their run,
and also allows tools like tools/plot.py to produce charts for
peripheral clock info.

Change-Id: Iafc6515ecebdb17bb3bf519ea924aa294f8f1da2

### DIFF
--- a/libs/utils/analysis/frequency_analysis.py
+++ b/libs/utils/analysis/frequency_analysis.py
@@ -163,6 +163,127 @@ class FrequencyAnalysis(AnalysisModule):
 # Plotting Methods
 ###############################################################################
 
+    def plotPeripheralClock(self, title='Peripheral Frequency', clk='unknown'):
+        """
+        Produce graph plotting the frequency of a particular peripheral clock
+
+        :param title: The title for the chart
+        :type  title: str
+
+        :param clk: The clk name to chart
+        :type  clk: str
+
+        :raises: KeyError
+        """
+        if not self._trace.hasEvents('clock_set_rate'):
+            self._log.warning('Events [clock_set_rate] not found, plot DISABLED!')
+            return
+        rate_df = self._dfg_trace_event('clock_set_rate')
+        enable_df = self._dfg_trace_event('clock_enable')
+        disable_df = self._dfg_trace_event('clock_disable')
+        pd.options.mode.chained_assignment = None
+
+        rate_df = rate_df[rate_df.clk_name == clk]
+        rate_df['clock_setting'] = rate_df['rate']/1e6
+        rate_df['state'] = -1
+
+        pd.set_option('display.expand_frame_repr', False)
+
+        if not enable_df.empty:
+          enable_df = enable_df[enable_df.clk_name == clk]
+          enable_df['clock_setting'] = 1;
+        if not disable_df.empty:
+          disable_df = disable_df[disable_df.clk_name == clk]
+          disable_df['clock_setting'] = 0;
+
+        freq = pd.concat([rate_df, enable_df, disable_df])
+        freq.sort_index(inplace=True)
+
+        if freq.empty:
+            self._log.warning('No events for clock ' + clk + ' found in trace')
+            return
+
+        #this will extend the last frequency value to the end of the timerange.
+        last = freq.tail(n=1)
+        last.index = { self._trace.x_max }
+        freq = pd.concat([freq, last])
+        freq['timestamp'] = freq.index
+        freq['frequency'] = 0;
+
+        max_freq = 0
+        last_state = 1
+        last_freq = rate_df.iloc[0]['clock_setting']
+        for index, row in freq.iterrows():
+            if row.state == -1 :
+                last_freq = row.clock_setting
+                if max_freq < last_freq: max_freq = last_freq
+                if last_state == 0:
+                    freq.loc[index, 'frequency'] = 0
+                else:
+                    freq.loc[index, 'frequency'] = last_freq
+
+            if row.state == 1:
+                freq.loc[index, 'frequency'] = last_freq
+                last_state = row.state
+
+            if row.state == 0:
+                freq.loc[index, 'frequency'] = 0
+                last_state = row.state
+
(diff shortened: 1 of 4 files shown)


tools/plots.py
```diff
@@ -142,7 +155,7 @@ def plotdir(run_dir, platform):
         logging.info('No performance data found')
 
     # Load Trace Analysis modules
-    trace = Trace(platform, run_dir)
+    trace = Trace(platform, run_dir, [], trace_format="systrace")
```
joelagnel: Just checking why `trace_format` should be systrace. tools/plot.py should should be generic, not android-specific right?

tools/plots.py
```diff
@@ -68,17 +68,30 @@ def main():
...
     if args.plots == 'all':
-        args.plots = 'tasks clusters cpus stune ediff edspace'
+        args.plots = 'tasks clusters cpus stune ediff edspace peripherals'
 
     # For each rtapp and each run
     if args.outdir is not None:
 
         # Load platform descriptior
         platform = None
-        plt_file = os.path.join(args.outdir, 'platform.json')
```
joelagnel: This part should be written properly. I think the old code is also wrong. `plt_file` can never be a directory, so I don't get the whole `isfile` stuff. Seems the below code can be simplified. The `else` path shouldn't be possible

libs/utils/env.py
```diff
@@ -761,6 +761,18 @@ def _load_em(self, board):
             return None
         return board.json['nrg_model']
 
+    def _load_peripherals(self, board):
+        peripherals_path = os.path.join(basepath,
```
joelagnel: This json loading code is pretty redundant. Can we abstract it better? It seems `_load_board` and `_load_em` both have similar code.

tools/plots.py
kdub: Fixed. libs.utils.analysis.Trace loads systrace and ftrace a bit inconsistently (the distinction between file and directory isn't clear). New code will detect systrace and call Trace appropriately.

kdub:
Fixed, thanks for taking a look!

libs/utils/env.py
kdub: done.

tools/plots.py
kdub: I reworked this so that its clearer what's going on. The input is still a directory, but now it loads the traces and platforms a bit more intelligently.

kdub:
This branch is now stacked on PR 492. (I had thought that one landed here in github, the plotting tool needs a function introduced in that PR)

libs/utils/analysis/frequency_analysis.py
```diff
@@ -163,6 +163,127 @@ def _dfg_cpu_frequency_transition_rate(self, cpu):
...
+
+        :raises: KeyError
+        """
+        if not self._trace.hasEvents('clock_set_rate'):
+            self._log.warning('Events [clock_set_rate] not found, plot DISABLED!')
+            return
+        rate_df = self._dfg_trace_event('clock_set_rate')
+        enable_df = self._dfg_trace_event('clock_enable')
+        disable_df = self._dfg_trace_event('clock_disable')
+        pd.options.mode.chained_assignment = None
```
valschneider: Usually when we face SettingWithCopy warnings we try to quiet them down by using `df.(i)loc ...`, even if the df is being modified properly.

tools/plots.py
```diff
@@ -59,31 +59,44 @@
 parser.add_argument('--plots', type=str,
         default='all',
         help='List of plots to produce (all,')
+parser.add_argument('--platform_file', type=str,
```
valschneider: I didn't even know this file existed, so... @derkling, it's in your hands :p

joelagnel: Huh, this is just the platform file. I though that's very basic thing for LISA (EAS needs platform information, etc).

tools/plots.py
```diff
@@ -176,6 +173,7 @@ def plotdir(run_dir, platform):
         trace.analysis.frequency.plotCPUFrequencies()
         trace.analysis.frequency.plotCPUFrequencyResidency()
 
+    print platform
```
joelagnel: Is this needed? If so, can we add it to logging?

libs/utils/env.py
```diff
@@ -803,15 +781,15 @@ def _init_platform(self):
...
             self.platform['nrg_model'] = self.conf['nrg_model']
         # Try to load the default energy model (if available)
-        else:
-            nrg_model = self._load_em(self.conf['board'])
-            # We shouldn't have an 'nrg_model' key if there is no energy model data
-            if nrg_model:
-                self.platform['nrg_model'] = nrg_model
+        # We shouldn't have an 'nrg_model' key if there is no energy model data
+        elif 'nrg_model' in board.json:
+            self.platform['nrg_model'] = board.json['nrg_model']
```
joelagnel: thumbs up for cleanliness

libs/utils/env.py
```diff
@@ -747,42 +747,20 @@ def _init_platform_smp(self):
 
         self.platform['cpus_count'] = len(self.target.core_clusters)
 
-    def _load_em(self, board):
-        em_path = os.path.join(basepath,
-                'libs/utils/platforms', board.lower() + '.json')
-        self._log.debug('Trying to load default EM from %s', em_path)
```
joelagnel: These log messages will be missing in the new code. Is that really Ok?

tools/plots.py
valschneider: I meant the **tools/plots.py** file. I've never used it - @derkling gave me a quick intro to what it is.

joelagnel: Ok sounds good, thanks

tools/plots.py
joelagnel: Hmm, so it looks like the new commit isn't updated in this PR.  As per "https://github.com/ARM-software/lisa/pull/554/commits", this commit (f8082d6702d) is 21 days old. Could you update it?

PR #555, opened 2018-02-07
add a manual test experiment.

kdub:
Add a manual test experiment to facilitate running experimental
tests that are either tricky to script, or used as an initial
data collection to test out a theory.

### DIFF
--- a/libs/utils/android/workloads/manual.py
+++ b/libs/utils/android/workloads/manual.py
@@ -1,0 +1,93 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2018, ARM Limited, Google and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import re
+import os
+import logging
+
+from time import sleep
+
+from target_script import TargetScript
+from android import Screen, System
+from android.workload import Workload
+
+
+class Manual(Workload):
+    """
+    Android Manual workload
+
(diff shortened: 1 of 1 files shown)


libs/utils/android/workloads/manual.py
```diff
@@ -0,0 +1,93 @@
...
+        self.collect = collect
+
+        # Set min brightness
+        Screen.set_brightness(self._target, auto=False, percent=0)
+
+        # Set timeout to min value
+        Screen.set_timeout(self._target, seconds=0)
+
+        # Prevent screen from dozing
+        Screen.set_doze_always_on(self._target, on=False)
```
valschneider: `Screen` doesn't have that method... Yet :)

libs/utils/android/workloads/manual.py
```diff
@@ -0,0 +1,93 @@
...
+        # Turn on airplane mode
+        System.set_airplane_mode(self._target, on=True)
+
+        # Unlock device screen (assume no password required)
+        Screen.unlock(self._target)
+
+        self.tracingStart(screen_always_on=False)
+        raw_input("Tracing and energy collection started.\n\
+                   Press ENTER when the manual experiment is concluded.")
+        print("collecting logs...")
```
valschneider: `self._log.info()` would be better

libs/utils/android/workloads/manual.py
```diff
@@ -0,0 +1,93 @@
...
+from android import Screen, System
+from android.workload import Workload
+
+
+class Manual(Workload):
+    """
+    Android Manual workload
+
+    This workload will set up a device, start tracing, start collection,
+    disconnect USB, and pause. User can then conduct a manual experiment,
```
valschneider: I don't see code related to handling the USB.

derkling:
@kdub do you have any chance to work on the feedback comments from @valschneider?

We are in the middle of a "merge window"... meaning we allocated a bit more time this week to get as much as possible PR merged into LISA. Thus, would be great if you can update this one, and other from yours, to have a good chance to merge it soon.

derkling:
Closed by error, let's reopen it.

valschneider:
I think it's still a good wload to have. It'd probably be nice to have as a script, so that you can do something like:

```
./manual_experiment --kind android --device DEADBEEF
<play with the device>
ENTER
```

I'll assign that to myself to get it in LISA/next.

PR #556, opened 2018-02-13
cgroups migration test

Elieva:
These patches add tests and support for tests where tasks are migrated from one cgroups to another during their execution and control that the utilization of each cgroups is migrated with them.

### DIFF
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -984,7 +984,11 @@ class _PELTTaskGroupsTest(LisaTest):
                     # Run tasks
                     se.wload.run(out_dir=test_dir, cpus=se.cpus,
                                  cgroup=se.parent.name, background=True)
-            sleep(max_duration)
+
+            sleep(max_duration / 2.0)
+            # Wake up to migrate tasks between cgroups
+            cls._migrate_task(te)
+            sleep(max_duration / 2.0)
 
             te.ftrace.stop()
 
@@ -994,6 +998,10 @@ class _PELTTaskGroupsTest(LisaTest):
 
         # Extract trace
         cls.trace = Trace(None, test_dir, te.ftrace.events)
+
+    @classmethod
+    def _migrate_task(cls, test_env):
+        return
 
     def _test_group_util(self, group):
         if 'sched_load_se' not in self.trace.available_events:
@@ -1006,6 +1014,7 @@ class _PELTTaskGroupsTest(LisaTest):
             raise ValueError('No sched_switch events. '
                              'Does the kernel support them?')
 
+        max_duration = 0
         task_util_df = self.trace.data_frame.trace_event('sched_load_se')
         tg_util_df = self.trace.data_frame.trace_event('sched_load_cfs_rq')
         sw_df = self.trace.data_frame.trace_event('sched_switch')
@@ -1014,15 +1023,22 @@ class _PELTTaskGroupsTest(LisaTest):
         for se in self.root_group.iter_nodes():
             if se.name == group:
                 tg = se
+            if se.is_task:
+                max_duration = max(max_duration, se.duration_s)
 
         if tg is None:
             raise ValueError('{} taskgroup does not exist.'.format(group))
 
         # Only consider the time interval where the signal should be stable
+        # after the migration phase
         tasks_names = [se.name for se in tg.iter_nodes() if se.is_task]
         tasks_sw_df = sw_df[sw_df.next_comm.isin(tasks_names)]
-        start = tasks_sw_df.index[0] + UTIL_AVG_CONVERGENCE_TIME
-        end = tasks_sw_df.index[-1] - UTIL_AVG_CONVERGENCE_TIME
+
+        start = tasks_sw_df.index[0] + \
+                UTIL_AVG_CONVERGENCE_TIME + \
+                max_duration / 2.0
+        end = tasks_sw_df.index[-1]
+
         task_util_df = task_util_df[start:end]
         tg_util_df = tg_util_df[start:end]
 
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -18,6 +18,7 @@
 import os
 import pandas as pd
 import logging
+import re
 import json
 from collections import OrderedDict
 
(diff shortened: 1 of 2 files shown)


tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+
+        # Get the task and taskgroup for the migration
+        for se in cls.root_group.iter_nodes():
+            if se.name == 't1_2':
+                task = se
+            if se.name == '/tg2':
+                new_taskgroup = se
+
+        # Get the PID of the task that is not migrated
+        tg1_task = cgroups.controllers['cpu'].tasks('/tg1')
```
derkling: s/tg1_task/tg1_tasks/?

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+    def test_util_root_group(self):
+        """Test utilization propagation to cgroup root"""
+        return self._test_group_util('/')
+
+    def test_util_tg1_group(self):
+        """Test utilization propagation to cgroup /tg1"""
+        return self._test_group_util('/tg1')
+
+    def test_util_tg1_1_group(self):
+        """Test utilization propagation to cgroup /tg2"""
```
derkling: This comment suggests a different test function name, is that intentional?

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+        # Get the task and taskgroup for the migration
+        for se in cls.root_group.iter_nodes():
+            if se.name == 't1_2':
+                task = se
+            if se.name == '/tg2':
+                new_taskgroup = se
+
+        # Get the PID of the task that is not migrated
+        tg1_task = cgroups.controllers['cpu'].tasks('/tg1')
+        for pid, data in tg1_task.iteritems():
```
derkling: this `iteritems` return `pid, (task_name, cmdline)`... let's `s/data/(task_name, cmdline)/` to make more clear in the code below what we are testing...

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+            if data[0] == 'rt-app' and re.search(r't1_1', data[1]):
```
derkling: This will be:
```python
if task_name == 'rt-app' and re.search(r't1_1', cmdline):
```

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+                exclude = '{}'.format(pid)
```
derkling: Aren't we looking up for a single task here?
In that case we can also break and skip all other tasks, isn't it?
If you do that, please however add a comment to make it clear.

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+
+        # Migrate t1_2 to the group /tg2 without migrating t1_1
+        task.change_taskgroup(new_taskgroup)
+        srcg = cgroups.controllers['cpu'].cgroup('/tg1').directory
+        dstg = cgroups.controllers['cpu'].cgroup('/tg2').directory
+        cgroups.cgroups_tasks_move(srcg, dstg, exclude=exclude)
```
derkling: Actually, here you can already pass in a regexp of tasks to not move... do we really need the code in the above lines 1389-1393?

IOW, isn't this:
```python
cgroups.cgroups_tasks_move(srcg, dstg, exclude='t1_1')
```
just enough?

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+
+    @classmethod
+    def _migrate_task(cls, test_env):
+        task = None
+        new_taskgroup = None
+        exclude = ''
+        cgroups = test_env.target.cgroups
+
+        # Get the task and taskgroup for the migration
+        for se in cls.root_group.iter_nodes():
```
derkling: This seems more like a `class Task(SchedEntity)` internals... cannot we fold these bits into the implementation of the `change_taskgroup` method, in such a way that we have a more easy API for test writers where you just specify that you wanna move a given task `t1_2` into a given group `/tg2`?

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+
+    @classmethod
+    def _getExperimentsConf(cls, test_env):
+        # Run all workloads on a CPU with highest capacity
+        cls.target_cpu = min(test_env.calibration(),
+                             key=test_env.calibration().get)
+
+        # Create taskgroups
+        cpus = test_env.target.list_online_cpus()
+        mems = 0
```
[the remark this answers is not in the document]
Elieva: Done: The value has been hard-coded in the function call.

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+        cls.root_group = Taskgroup("/", cpus, mems, test_env)
+        tg1 = Taskgroup("/tg1", cpus, mems, test_env)
+        tg2 = Taskgroup("/tg2", cpus, mems, test_env)
+
+        # Create tasks
+        period_ms = 16
+        duration_s = 4
+        cpus = [cls.target_cpu]
```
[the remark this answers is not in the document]
Elieva: Done: I changed to use directly cls.target_cpu (which is equal to the biggest cpu)

tests/eas/load_tracking.py
```diff
@@ -1293,3 +1294,119 @@ def test_util_tg4_group(self):
...
+
+        # Create taskgroups
+        cpus = test_env.target.list_online_cpus()
+        mems = 0
+        cls.root_group = Taskgroup("/", cpus, mems, test_env)
+        tg1 = Taskgroup("/tg1", cpus, mems, test_env)
+        tg2 = Taskgroup("/tg2", cpus, mems, test_env)
+
+        # Create tasks
+        period_ms = 16
```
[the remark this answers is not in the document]
Elieva: Done : I moved it up

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Oops done :)

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: I think "exclude='t1_1'  is not enough. The way i think it works is the following: the cgroups_tasks_move will filter the task from the following information:
PID, COMMAND  and as the COMMAND is "rt-app" and is common to all the created tasks it cannot be used to filter tasks. The only way, I can see, is to find the PID of the task which is not migrated and use the PID to exclude the task.

tests/eas/load_tracking.py
Elieva: Done: I added a get_children function in Task(SchedEntity) which take a name in argument and return the task object associate to this name such that the loop to find the object is coded into the Task class

tests/eas/load_tracking.py
derkling: This seems an interesting new use-case... cannot we try to extend the devlib API to filter on the command line parameters?
That patch will allow to simplify your code... as well as to improve devlib ;-)

tests/eas/load_tracking.py
```diff
@@ -1273,3 +1290,224 @@ def test_util_tg4_group(self):
...
+    target_cpu = 0
+    root_group = None
+    trace = None
+
+    @classmethod
+    def _getExperimentsConf(cls, test_env):
+        # Run all workloads on a CPU with highest capacity
+        cls.target_cpu = [min(test_env.calibration(),
+                              key=test_env.calibration().get)]
+        period_ms = 16
```
derkling: Cannot these two be class attributes?

tests/eas/load_tracking.py
```diff
@@ -1273,3 +1290,224 @@ def test_util_tg4_group(self):
...
+        duration_s = 4
+        # Create taskgroups
```
derkling: nit-pick: empty line before comment?

tests/eas/load_tracking.py
```diff
@@ -1273,3 +1290,224 @@ def test_util_tg4_group(self):
...
+        }
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        super(NestedCgroupsMigrationTest, cls).runExperiments(*args, **kwargs)
+
+    @classmethod
+    def _migrate_task(cls, test_env):
+        exclude = ''
+        cgroups = test_env.target.cgroups
```
derkling: that's just because of my OCD: can we swap these two lines... it seems to read better to me :)
... sorry :(

tests/eas/load_tracking.py
```diff
@@ -1273,3 +1290,224 @@ def test_util_tg4_group(self):
...
+
+class NestedCgroupsMigrationTest(_PELTTaskGroupsTest):
+    """
+    Test PELT utilization for task groups migration.
+    Initial group hierarchy:
+                            +-----+
+                            | "/" |
+                            +-----+
+                             /
+                            /
```
derkling: can we use just one `/`  for each edge... we should be also to save quit some lines :)

tests/eas/load_tracking.py
```diff
@@ -1273,3 +1290,224 @@ def test_util_tg4_group(self):
...
+        task.change_taskgroup(new_taskgroup)
+        srcg = cgroups.controllers['cpu'].cgroup('/tg1').directory
+        dstg = cgroups.controllers['cpu'].cgroup('/tg2').directory
+        cgroups.cgroups_tasks_move(srcg, dstg, exclude=exclude)
+
+    def test_util_root_group(self):
+        """Test utilization propagation to cgroup root"""
+        return self._test_group_util('/')
+
+    def test_util_tg1_group(self):
```
derkling: Thinking about the real meaning of these tests.... maybe if we find a more descriptive name it can be easier to understand what is failing when something is failing...

Thus, what about a renaming similar to:
- `s/test_util_root_group/test_util_aggregation/`
- `s/test_util_tg1_group/test_util_move_out/`
- `s/test_util_tg2_group/test_util_move_in/`

derkling: Better proposals are more then welcome... and the test docstring should also be update to better describe the tested condition. For example, for the last test, maybe something like:

```python
def test_util_move_in(self):
"""
Test utilization when a task enter a group.
"""
```

tests/eas/load_tracking.py
```diff
@@ -1273,3 +1290,224 @@ def test_util_tg4_group(self):
...
+    target_cpu = 0
+    root_group = None
+    trace = None
+
+    @classmethod
+    def _getExperimentsConf(cls, test_env):
+        # Run all workloads on a CPU with highest capacity
+        cls.target_cpu = [min(test_env.calibration(),
+                              key=test_env.calibration().get)]
+        period_ms = 16
```
derkling: Same comment as below (sorry, I've started the review from the bottom)... cannot these be class attributes?

tests/eas/load_tracking.py
```diff
@@ -1273,3 +1290,224 @@ def test_util_tg4_group(self):
...
+        duration_s = 4
+        # Create taskgroups
```
derkling: And the empty line here as well :(

tests/eas/load_tracking.py
Elieva: Done in devlib

tests/eas/load_tracking.py
Elieva: Done for period_ms and duration_s (but cls.target_cpu need to be in the function because test_env is only available there)

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Kind of done: exclude = '' has been removed

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

tests/eas/load_tracking.py
Elieva: Done

derkling:
LGTM... almost ready to merge... you just need to add at the base a patch to progress the devlib submodule to: ARM-software/devlib@8f3dc05f97889bdd607688fed3167cdbd4bcf937

PR #558, opened 2018-02-13
Misfit tests

valschneider:
All is in the title...


tests/eas/misfit.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def get_idle_time(self, trace, start, end, cpus):
+        """
+        Return for how long 'cpus' were idle in [start, end]
+        """
+        idle_time = 0
+
+        for cpu in cpus:
+            busy_df = trace._compute_steps_duration(
+                pd.DataFrame(trace.getCPUActiveSignal(cpu), columns=['state'])
```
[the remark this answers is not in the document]
credp: I'm not so sure. I think a Series is the proper data format for a single time-series signal and it should be cheap to convert to a DataFrame when we need it. AFAIK there are some Series APIs which can be useful for manipulating time-based signals like this (logical operations, testing values at arbitrary timestamps, etc.).

valschneider:
Note for self: A "preliminary" test that checks if any misfit migration happened should be added. It'll allow us to more easily see if "misfit migration happened but wasn't optimal" vs "misfit is completely borked"

libs/utils/trace.py
```diff
@@ -886,6 +886,69 @@ def _compute_steps_duration(self, df):
...
+            return df
+
+        res = df[start:end].copy()
+
+        # There's a 99% probability that events won't fit snuggly within within
+        # the interval. Find the one preceding the interval, and deduce the rest
+        # mathemagically.
+        left = df[:start].tail(1)
+        right = res.tail(1)
+
```
derkling: We can easily return here, since we need to generate just one row DF:

```python
if res.empty:
    new_delta = df[start:].head(1).index - start
    res = pd.DataFrame(df[:start].tail(1).values,
                       columns=df.columns, index=[start])
    res[col_name] = new_delta
    return res
```

Moreover, I've come up with this other implementation for the whole function:

```python
def slice(df, start, end, col_name):
    # Sanity check boundaries
    start = max(start, df.index[0])
    end = min(end, df.index[-1])
    
    # Let's start by building the first row
    values = df[:start].tail(1).values
    delta_first = df[start:].head(1).index - start
    begin_df = pd.DataFrame(values, columns=df.columns, index=[start])
    begin_df[col_name] = delta_first

    # Get a "middle" slice from the original data frame
    middle_df = df[start:end]

    # If the middle slice is empty, then we just need to first row previously computed
    if middle_df.empty:
        return begin_df

    # Otherwise, the middle slice can be appended to the first row
    final_df = begin_df.append(middle_df)

    # And finally fix the delta for the last row
    delta = df[:end].tail(1)[col_name].values[0]
    delta_last = delta + (end - df[end:].head(1).index).values[0]
    final_df.iloc[-1, final_df.columns.get_loc(col_name)] = delta_last

    return final_df.reset_index().drop_duplicates(keep='last', subset='index').set_index('index') 
```
I've tested with some corner cases... but it certainly deserves some more third party testing.
To me, it seems more easy to read... what do you think.

libs/utils/trace.py
```diff
@@ -886,6 +886,69 @@ def _compute_steps_duration(self, df):
 
         return df
 
+    def _windowify_steps_df(self, df, (start, end), len_column='len'):
+        """
+        Slice a dataframe to [start:end], and work on the time data so that it
+        makes sense within the interval.
```
derkling: This functionality is so tight related to the new `_compute_steps_duration` which I'm wondering if it does not make more sense to fold this code into that call. We can have a default `(start, end) = (0, None)` which does not do any trimming at all... otherwise, you just add this functionality at then 

Based on previous patch comment, the new prototype would become:

```python
def addEventsDeltas(self, df, col_name='delta', trim=(0, None))
```

"Slice a dataframe using the specified [start:end] timestamps, and fix col_name values to match the new time boundaries"?

libs/utils/trace.py
```diff
@@ -886,6 +886,69 @@ def _compute_steps_duration(self, df):
...
+
+        _windowify_steps_df(df, (16.5, 17.5)) =>
+
+        Time len state
+        16.5  .5   0
+        17    .5   1
+
+        _windowify_steps_df(df, (16.2, 16.8)) =>
+
+        Time len state
```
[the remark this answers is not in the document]
derkling: Just notice you added one in the next patch :+1:

libs/utils/trace.py
derkling: Actually, your next patch test can give it a first check... did not run it although. ;-)

libs/utils/trace.py
```diff
@@ -874,6 +867,25 @@ def getClusterActiveSignal(self, cluster):
 
         return cluster_active
 
+    def _compute_steps_duration(self, df):
```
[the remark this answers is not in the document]
valschneider: Comment #1: I like your name much more. Also good point on the exception.
Comment #2: Agreed, although I think the casing would be
```python
def add_events_deltas(self, df, col_name='delta')
```
right ?

libs/utils/trace.py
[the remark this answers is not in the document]
valschneider: Yeah I initially wrote this comment without an actual test and then heard the ghost of @bjackman saying "tests plz", so added one and found a bug \o/

libs/utils/trace.py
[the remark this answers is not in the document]
derkling: There are other utility functions in trace.py which uses the camel case syntax I've proposed... but have a better check ;-)

libs/utils/trace.py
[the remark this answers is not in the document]
derkling: Do you remember @bjackman ?!?!

libs/utils/trace.py
```diff
@@ -886,6 +886,69 @@ def _compute_steps_duration(self, df):
...
+
+        # There's a 99% probability that events won't fit snuggly within within
+        # the interval. Find the one preceding the interval, and deduce the rest
+        # mathemagically.
+        left = df[:start].tail(1)
+        right = res.tail(1)
+
+        # "void" to fill between 'start' and the first event in the interval
+        if not left.empty:
+            left_duration = max(0, left.index[0] + left[len_column].values[0] - start)
```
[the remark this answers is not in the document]
valschneider: Well for normally shaped dfs I would say it could work.The example dataframe has continuous events... But what if it doesn't ? That's totally the case for my use of it: I take the latency_df and filter events that I don't care about (I only keep the active state). So I end up with a dataframe that has a 'len' column but those are not continuous.

This was all very clear to me when I wrote it but now it's really confusing tbh.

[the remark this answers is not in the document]
derkling: Ok, I was missing this use-case... then probably also the solution below does not fits that use-case.

However, I don't see why you need to filter (i.e. keep only active states) and then trim after...

libs/utils/trace.py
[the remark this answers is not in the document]
valschneider: No you're right, I was looking at the wrong place.

libs/utils/trace.py
```diff
@@ -874,6 +867,27 @@ def getClusterActiveSignal(self, cluster):
...
+            raise RuntimeError("Column {} is already present in the dataframe".
+                               format(col_name))
+
+        df['start'] = df.index
+        df[col_name] = (df.start - df.start.shift()).fillna(0).shift(-1)
+        df.drop('start', axis=1, inplace=True)
+
+        # Fix the last event, which will have a NaN duration
+        # Set duration to trace_end - last_event
+        df.loc[df.index[-1], 'len'] = self.start_time + self.time_range - df.index[-1]
```
derkling: :thinking: ... should not `s/'len'/col_name/` ?

libs/utils/trace.py
```diff
@@ -874,6 +867,27 @@ def getClusterActiveSignal(self, cluster):
 
         return cluster_active
 
+    def addEventsDeltas(self, df, col_name='len'):
```
derkling: It's now a more generic API... let's match the default with the method name, i.e. `s/len/delta`

libs/utils/trace.py
```diff
@@ -888,6 +888,95 @@ def addEventsDeltas(self, df, col_name='len'):
         # Set duration to trace_end - last_event
         df.loc[df.index[-1], 'len'] = self.start_time + self.time_range - df.index[-1]
 
+    def slice_df(self, df, start, end, len_col='len'):
```
derkling: I would say the API looks more generic if we:
1. rename the method to `trim_df`... since that's doing much more then a "standard slice"
2. replace `len_col` with `column`, which more specifically defines that the values that will be trimmed according to the new start,end boundaries.

The documentation below should better state the expected DF input format, i.e. the column should report deltas among consecutive DFs.

libs/utils/trace.py
```diff
@@ -888,6 +888,95 @@ def addEventsDeltas(self, df, col_name='len'):
...
+        """
+        Slice a dataframe to [start:end], and work on the time data so that it
```
derkling: What about the following text:

```python
Slice a dataframe of deltas in [start:end] and ensure we have an event at exactly those boundaries.

The input dataframe is expected to have a "column" which reports the time
delta between consecutive rows, as for example dataframes generated by
addEventsDeltas().

The returned dataframe is granted to have an initial and final event at the
specified "start" ("end") index values, which values are the same of the last
event before (first event after) the specified "start" ("end") time.

Examples:
```

.. here your examples below.

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+    @staticmethod
+    def _get_cpu_capacity(test_env, cpu):
+        """
+        :returns: The capacity value of a given CPU on the target
+        """
+        if test_env.nrg_model:
+            return test_env.nrg_model.get_cpu_capacity(cpu)
+
+        return test_env.target.read_int(
+            '/sys/devices/system/cpu/cpu{}/cpu_capacity'.format(cpu))
```
derkling: This sound like an API suitable for devlib... isn't it?

```python
test_env.target.cpu_capacity(cpu)
```

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+
+        :returns: a dict of dicts. The first level is the integer representing
+        the sched_domain index (sched_domain<key>/* entries), the second level
+        is the sched domain field (sched_domainX/<key> entry)
+        """
+        data = test_env.target.read_tree_values(
+            '/proc/sys/kernel/sched_domain/cpu{}/'.format(cpu), depth=2
+        )
+
+        return {int(key.lstrip('domain')) : values for key, values in data.iteritems()}
```
derkling: Another cool devlib API... ;-)

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+    def _has_asym_cpucapacity(cls, test_env):
+        """
+        :returns: Whether SD_ASYM_CPUCAPACITY is set on any sd in the system
+        """
+        for cpu in xrange(test_env.platform['cpus_count']):
+            for domain, data in cls._get_sd_info(test_env, cpu).iteritems():
+                if int(data['flags']) & SD_ASYM_CPUCAPACITY == SD_ASYM_CPUCAPACITY:
+                    return True
+
+        return False
```
derkling: Don't what to repeat myself... but... would not be nice to have something like:

```python
test_env.target.cpus_have_asym_capacity()
```

:sunglasses:

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+        cpus = {}
+        for cpu in xrange(test_env.platform['cpus_count']):
+            cap = cls._get_cpu_capacity(test_env, cpu)
+            if cap not in cpus:
+                cpus[cap] = []
+
+            cpus[cap].append(cpu)
+
+        capacities = sorted(cpus.keys())
+        return [cpus[capacity] for capacity in capacities]
```
derkling:
```python
test_env.target.cpus_by_capacity()
```
?

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+
+    @memoized
+    @classmethod
+    def _get_max_lb_interval(cls, test_env):
+        """
+        :returns: The absolute maximum load-balance interval in seconds
+        """
+        HZ = cls._get_hz(test_env)
+        return ((HZ * test_env.platform['cpus_count']) / 10) * (1. / HZ)
```
derkling: Is that hardcoded `/10` intentional?... I guess yess... can we add a bit of doc explaining why 10?

And.. does it makes sense for this also to be a devlib API?

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+    Test Misfit task migration happens at idle balance (staggered test case)
+
+    This test spawns nr_cpus 100% tasks. The tasks running on bigger-capacity
+    CPUs will finish first, and it is expected of them to instantly pull the
+    tasks running on smaller-capacity CPUs via idle-balance.
+    """
+
+    # How long the tasks will be pinned to their "starting" CPU. Doesn't have
+    # to be long (we just have to ensure they spawn there), so arbitrary value
+    pin_delay = 0.001
```
derkling: Just to make is clear, what about `s/pin_delay/pin_delay_s/`

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+
+    # Somewhat arbitrary delay - long enough to ensure
+    # rq->avg_idle > sysctl_sched_migration_cost
+    idling_delay = 1
```
derkling: same here?

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+            "delay": int(cls.idling_delay * 1000000),
+            "loop": 1,
+            "phases": OrderedDict([("pinned", {}), ("free", {})]),
+            "policy": "SCHED_OTHER"
+        }
+
+        for key in task_template["phases"].iterkeys():
+            task_template["phases"][key]["loop"] = 1
+
+        task_template["phases"]["pinned"]["run"] = int(1e6 * cls.pin_delay)
```
derkling: What about adding above a nice `_to_ns(seconds)` converter?

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+        # We're pinning stuff in the first phase, so give it ample time to
+        # clean the pinned logic out of balance_interval
+        task_template["phases"]["free"]["run"] = int(
+            1e6 * 1.1 * cls._get_max_lb_interval(test_env))
```
derkling: Why 1.1?

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+            wload["tasks"]["misfit_{}".format(cpu)] = task
+
+        wload_path = os.path.join(test_env.res_dir, 'staggered_template.json')
+        with open(wload_path, 'w') as fh:
+            res = json.dumps(wload)
+            # Tokens that represent integers musn't be escaped by quotes.
+            # The reason strings are used is to dump the json more easily.
+            for key in ['__PVALUE__']:
+                res = res.replace('"{}"'.format(key), key)
+            fh.write(res)
```
derkling: I'm not really a big fan of this way to generate rt-app JSONs... especially now that we support cpumasks per phases since 0e5d5853d

Not testes, but I think something like this should do its job...

```python
pinned_time = # whatever you like
free_time = # whatever you like
all_cpus = range(test_env.platform['cpus_count']);
tasks = {}
for cpu in all_cpus:
    task_pinned = Periodic(duty_cycle_pct=100, duration_s=pinned_time, period_ms=16, cpu=cpu)
    task_free = Periodic(duty_cycle_pct=100, duration_s=free_time, period_ms=16, cpu=all_cpus)
    task = task_pinned + task_free
    tasks["task_{}".format(cpu)] = task.get()

rtapp.conf(
    kind='profile',
    params=tasks
)
```

Is the new API not good enough for your use-case?

tests/eas/misfit.py
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+    def get_idle_time(self, trace, start, end, cpus):
+        """
+        :returns: How long 'cpus' were idle in the [start, end] interval
+        """
+        idle_time = 0
+
+        for cpu in cpus:
+            busy_df = pd.DataFrame(trace.getCPUActiveSignal(cpu), columns=['state'])
```
derkling: So, trace.getCPUActiveSignal(cpu) is memoized... but here we keep creating DFs...
Maybe you can add this line in an above memoized wrapper to avoid generating again ad again the same DF... but I'm not sure how many times you call this on the same CPU...

tests/eas/misfit.py
valschneider: Yeah it's defined like that in the kernel:
```c
max_load_balance_interval = HZ/10;
```

tests/eas/misfit.py
valschneider: Aye aye cap'n

tests/eas/misfit.py
valschneider: It's a very scientifically proven number to have something > 1 to ensure that even with the pinned logic we did get one balance attempt.

libs/utils/trace.py
valschneider: :sweat:

libs/utils/trace.py
valschneider: Speaking of names, I was wondering if `squash_df` wouldn't be actually clearer. It's not a "clean cut" as would imply slice or trim - you actually squash stuff within an interval.

libs/utils/trace.py
valschneider: I think the main reason I did that is because the overutilized dataframe used to be built with that 'len' column, and I didn't want to break anything. But that's not really an issue, I can just specify 'len' for overutilized instead of making it the norm for everyone else.

tests/eas/misfit.py
derkling: Then let's define it as a constant here too... just keep it float to avoid rounding errors...

valschneider: Do you mean just the `10` ? The thing is it's then scaled by the number of CPUs, and the point of this method is to abstract it all away. For all I know this could change at some point in the kernel, so I just hide it all under a single wrapper.

What I linked is the initial value but for SMP we actually do that:
```c
max_load_balance_interval = HZ*num_online_cpus()/10;
```
Which is what the whole method does.

tests/eas/misfit.py
valschneider: AFAIR the sysfs entry isn't present for older kernels, so kind of rely on the energy model values, and those are known to LISA only. It could be possible to squash that into devlib but it'd be tedious.

tests/eas/misfit.py
valschneider: I think those would make sense to have in a new module e.g. `asym`.

We already have `bl` but I've always had problems with it, I think mostly because you have to set up a platform and give it core names and then it will try to match core names, etc etc.

We could go with a new module that handles those things as done here, i.e. it reads stuff from the target. We could also pave the way for DynamIQ stuff, and remove all of our weird code where we try to do what devlib.bl does without using it because of the issues mentioned above.

tests/eas/misfit.py
derkling: Right... maybe just add the formula above (i.e. `max_load_balance_interval = HZ/10;`) in the docstring... maybe with an lxr reference to where in the kernel code this constant is defined.

tests/eas/misfit.py
valschneider: That new API wasn't there when I wrote this :p

derkling:
On 02-Mar 12:08, Valentin Schneider wrote:
> valschneider commented on this pull request.
> 
> > +
> +    @classmethod
> +    def setUpClass(cls, *args, **kwargs):
> +        super(_MisfitMigrationBase, cls).runExperiments(*args, **kwargs)
> +
> +    @memoized
> +    @staticmethod
> +    def _get_cpu_capacity(test_env, cpu):
> +        """
> +        :returns: The capacity value of a given CPU on the target
> +        """
> +        if test_env.nrg_model:
> +            return test_env.nrg_model.get_cpu_capacity(cpu)
> +

This only is the domain of the devlib API, read also the next
comment...

---8<---
> +        return test_env.target.read_int(
> +            '/sys/devices/system/cpu/cpu{}/cpu_capacity'.format(cpu))
---8<---


> 
> AFAIR the sysfs entry isn't present for older kernels, so kind of

Right...

> rely on the energy model values, and those are known to LISA only.

The devlib API can still be implemented to return 1024 if the value is
not available.

> It could be possible to squash that into devlib but it'd be tedious.

The above code gives priority to the LISA energy model, and that's
fine. If we don't have an EM, we can try to use the devlib API to read
the capacity... and get 1024 it the kernel do not report it.

Does that makes sense?

-- 
#include <best/regards.h>

Patrick Bellasi

derkling:
On 02-Mar 12:13, Valentin Schneider wrote:
> valschneider commented on this pull request.
> 
> 
> 
> > +        task_template["phases"]["free"]["run"] = int(
> +            1e6 * 1.1 * cls._get_max_lb_interval(test_env))
> +
> +        for cpu in xrange(test_env.platform['cpus_count']):
> +            task = deepcopy(task_template)
> +            task["phases"]["pinned"]["cpus"] = [cpu]
> +            wload["tasks"]["misfit_{}".format(cpu)] = task
> +
> +        wload_path = os.path.join(test_env.res_dir, 'staggered_template.json')
> +        with open(wload_path, 'w') as fh:
> +            res = json.dumps(wload)
> +            # Tokens that represent integers musn't be escaped by quotes.
> +            # The reason strings are used is to dump the json more easily.
> +            for key in ['__PVALUE__']:
> +                res = res.replace('"{}"'.format(key), key)
> +            fh.write(res)
> 
> That new API wasn't there when I wrote this :p

Yes I know... it's not your fault... however we should still try now
to use what we have.

Hope the example code I provided is good enough..

-- 
#include <best/regards.h>

Patrick Bellasi

derkling:
On 02-Mar 12:12, Valentin Schneider wrote:
> valschneider commented on this pull request.
> 
> 
> 
> > +        :returns: The value of CONFIG_HZ on the target
> +        """
> +        return int(test_env.target.config.get('CONFIG_HZ'))
> +
> +    @memoized
> +    @classmethod
> +    def _has_asym_cpucapacity(cls, test_env):
> +        """
> +        :returns: Whether SD_ASYM_CPUCAPACITY is set on any sd in the system
> +        """
> +        for cpu in xrange(test_env.platform['cpus_count']):
> +            for domain, data in cls._get_sd_info(test_env, cpu).iteritems():
> +                if int(data['flags']) & SD_ASYM_CPUCAPACITY == SD_ASYM_CPUCAPACITY:
> +                    return True
> +
> +        return False
> 
> I think those would make sense to have in a new module e.g. `asym`.

Why not just add them to target?... with a proper naming...
We already have generic platform APIs in that module... and this is
not bl specific... it still makes sense to read the capacity to
understand that a platform is SMP

> 
> We already have `bl` but I've always had problems with it, I think
> mostly because you have to set up a platform and give it core names
> and then it will try to match core names, etc etc.
> 
> We could go with a new module that handles those things as done
> here, i.e. it reads stuff from the target. We could also pave the
> way for DynamIQ stuff, and remove all of our weird code where we try
> to do what devlib.bl does without using it because of the issues
> mentioned above.

Could be useful only if we know that we really have tons of new APIs
to add... and we need to find a suitable yet generic name.
Not sure however it's not enough to go just with keep adding them to
target.

> 
> -- 
> You are receiving this because your review was requested.
> Reply to this email directly or view it on GitHub:
> https://github.com/ARM-software/lisa/pull/558#discussion_r171832404

-- 
#include <best/regards.h>

Patrick Bellasi

valschneider:
@derkling I think I took care of all your comments except the devlib ones. Will look into that next week.

valschneider:
Now uses the devlib Sched module, which just got merged :D

tests/eas/misfit.py
credp: I vote for a different asymmetric cpu module in devlib too

derkling:
This patch:
   analysis/latency_analysis: Guard against None platform
should be dropped.

There is already a (hopefully better) fix in this series https://github.com/ARM-software/lisa/pull/602, specifically: https://github.com/ARM-software/lisa/pull/602/commits/ecc493c5f548df013b1f178ac79b323a0c31afa8

valschneider:
* Fixed some false negatives in the preemption test

valschneider:
* Added some delay between each task completion, which should prevent the issue we've seen where one task is about to end, some sshd task briefly takes its place and another ending task tries to idle_balance pull the sshd task (but can't)

valschneider:
Fat Finger Syndrome...

PR #560, opened 2018-02-15
Using branch names in wltest series

balsini:
Extend wltest with the capability of using, in addition to the SHA1,
also branch names to checkout to a given kernel commit.
This allows to keep the tests updated with the HEAD commit of the given
branch name, without the need to update the series file every time the
HEAD of the branch changes.


tools/wltests/test_series
```diff
@@ -943,8 +943,14 @@ COMMIT_ID=1
...
-	# Extract SHA1 from commit description
-	COMMIT_SHA1=${COMMITS%% *}
+	# Extract commit identifier from commit description
+	COMMIT_NAME=${COMMITS%% *}
+
+	# Translate possible branch name to SHA1
+	AVAIL_COMMITS=`echo $(git -C $KERNEL_SRC show-ref --heads -s $COMMIT_NAME; \
+			      git -C $KERNEL_SRC show-ref -s $COMMIT_NAME) \
+			      $COMMIT_NAME`
+	COMMIT_SHA1=${AVAIL_COMMITS%% *}
```
derkling: We should fix it to fail in case there are local and remote branches with the same name.

For example, this is what I get on my machine:
```bash
# We have a local and remote branch with the same name:
$> git -C . show-ref users/patbel01/lkml/utilest_v5
68568b57832476e20fdb020266329a824f061d96 refs/heads/users/patbel01/lkml/utilest_v5
0b05d1859714e22013c5c882c3d3adfaa0de9121 refs/remotes/arm-pdsw-drk/users/patbel01/lkml/utilest_v5

# Here is the results by specifying the remote (arm-pdsw-drk):
$> git -C . show-ref arm-pdsw-drk/users/patbel01/lkml/utilest_v5
0b05d1859714e22013c5c882c3d3adfaa0de9121 refs/remotes/arm-pdsw-drk/users/patbel01/lkml/utilest_v5
```

Something like this should work for all the cases...

```bash
COMMIT=${COMMITS%% *}
COMMITS=$(git -C $KERNEL_SRC show-ref -s $COMMIT)
[ $? -eq 0 ] || COMMIT_SHA1 = $COMMIT
if [ echo $COMMIT_SHA1 | wc -l) -gt 1 ]; then
    report error
fi
```

... just add a better error message explaining the user she has to use either a SHA1 or a unique branch name (i.e. including the remote repo name if required)

balsini:
Code updated :) @derkling

PR #563, opened 2018-02-23
Futuristic userspace freezer

valschneider:
Lazarus version of https://github.com/ARM-software/lisa/pull/376

### DIFF
--- a/tests/eas/capacity_capping.py
+++ b/tests/eas/capacity_capping.py
@@ -83,6 +83,7 @@ class CapacityCappingTest(unittest.TestCase):
 
     """
 
+
     @classmethod
     def setUpClass(cls):
         cls.params = {}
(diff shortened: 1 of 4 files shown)


tests/eas/capacity_capping.py
```diff
@@ -126,30 +127,31 @@ def run_workload(cls):
...
-                    "Increase WORKLOAD_DURATION_S or speed up target connection")
-            time.sleep(phase_duration - write_latency)
-
-        write_and_sleep(min_frequency)
-        write_and_sleep(max_frequency)
-
-        cls.env.ftrace.stop()
-        cls.env.ftrace.get_trace(cls.trace_file)
+        with cls.env.freeze_userspace():
+            cls.env.ftrace.start()
```
derkling: Since we are already creating such a big diff to basically add just an indentation...
can we go further and make it even more clean by moving the bulk of this code under a private function?
```python
with cls.env.freese_userspace():
    cls._run_workload()
```

valschneider:
@derkling Done. Didn't create a private function  but encapsulated the call to `run_workload()` with the userspace freeze - I tested and it still works fine.

PR #568, opened 2018-02-26
Loadtracking.py: Cgroups test userspace freezing update

Elieva:
The class _PELTTaskGroupsTest(LisaTest) has been updated to
freeze the userspace as expected by the new syntax.
Namely, the userspace is freezed in a "with" statement to
catch error and unfreeze correctly in this case.


valschneider:
Missed that one during the rewrite, thanks !

PR #570, opened 2018-02-27
Check for kernel SHA1 to match also signed tags'

balsini:
Since a signed tag has a SHA1 that differs from the one associated with
the commit to which it is referring, this was causing false negatives
when verifying the correspondence between the current kernel version of
the device and of the local repository.

This patch, when there is no correspondence between the current SHA1 of
the repository and the kernel of the device, verifies if there is a tag
pointing to the same SHA1 and performs the comparison with it.

It also inverts the substirng checking: it checks if the SHA1 of the
kernel is contained in the SHA1 of the repo, since this last should be
longer.
In order to do this, the following steps are followed:
- take the kernel release (uname -a), e.g.
  "4.4.88-g0e6b355-00003-g36e8d7f";
- take the last element, considering '-' as separator, e.g. "g36e8d7f";
- remove the first character, e.g. "36e8d7f";
- check if this SHA1 is contained in the SHA1 of the current repository.

### DIFF
--- a/tools/wltests/test_series
+++ b/tools/wltests/test_series
@@ -791,7 +791,15 @@ match_sha1() {
 	CURRENT=$($ADB shell 'uname -r' | awk -F "-g" '{print $NF}')
 	[[ $COMMIT_SHA1 =~ $CURRENT* ]] && return $OK
 
-	return $OK
+	# It may happen that a tag pointing to a given commit has
+	# a different SHA1 (signed tag), so check if this this is
+        # the case.
+	TAG_NAME=$(git -C $KERNEL_SRC describe --exact-match $COMMIT_SHA1 2>/dev/null) || return $EAGAIN
+	TAG_SHA1=$(git -C $KERNEL_SRC show-ref --hash $TAG_NAME)
+	c_info "   CommitID of local repo corresponds to tag: $TAG_NAME"
+	[[ $TAG_SHA1 =~ $CURRENT* ]] && return $OK
+
+	return $EAGAIN
 }
 
 build_sha1() {
--- a/tools/wltests/test_series
+++ b/tools/wltests/test_series
@@ -786,10 +786,10 @@ match_sha1() {
 	COMMIT_SHA1=$1
 
 	c_info "Current kernel: "
-	CURRENT=$($ADB shell 'uname -a')
-	c_info "   $CURRENT"
+	c_info "   $($ADB shell 'uname -a')"
 
-	[[ $CURRENT = *$COMMIT_SHA1* ]] || return $EAGAIN
+	CURRENT=$($ADB shell 'uname -r' | awk -F "-g" '{print $NF}')
+	[[ $COMMIT_SHA1 =~ $CURRENT* ]] && return $OK
 
 	return $OK
 }


tools/wltests/test_series
```diff
@@ -786,12 +786,25 @@ match_sha1() {
 	COMMIT_SHA1=$1
 
 	c_info "Current kernel: "
-	CURRENT=$($ADB shell 'uname -a')
-	c_info "   $CURRENT"
+	c_info "   $($ADB shell 'uname -a')"
 
-	[[ $CURRENT = *$COMMIT_SHA1* ]] || return $EAGAIN
+	CURRENT=$($ADB shell 'uname -r | awk -F "-" "{print \$NF}"')
```
derkling: Let's do the awk on the host side, where is more likely to have the command available.

tools/wltests/test_series
```diff
@@ -786,12 +786,25 @@ match_sha1() {
...
+	CURRENT=${CURRENT:1}
```
derkling: This also can be folded in the previous command, thus:
```python
CURRENT=$($ADB shell 'uname -r' | awk -F "-g" "{print \$NF}"')
```

tools/wltests/test_series
```diff
@@ -786,12 +786,25 @@ match_sha1() {
...
 
-	return $OK
+	[[ $COMMIT_SHA1 =~ $CURRENT* ]] && return $OK
```
derkling: The above chunks can be moved in a separate patch, which corresponds to the change you already describe in the commit message when you say "It also inverts the substirng..." 

There are also some typos in the commit message, e.g: s/substirng/substring/

tools/wltests/test_series
```diff
@@ -786,12 +786,25 @@ match_sha1() {
...
+
+	# It may happen that a tag pointing to a given commit has
+	# a different SHA1 (signed tag), so check if this this is
+        # the case.
+	TAG_NAME=$(git -C $KERNEL_SRC describe --exact-match $COMMIT_SHA1 2> /dev/null) || return $EAGAIN
```
derkling: nit-pick `s|2> /dev/null|2>/dev/null` (i.e. remove the space)

tools/wltests/test_series
```diff
@@ -786,12 +786,25 @@ match_sha1() {
...
+	TAG_SHA1=$(git -C $KERNEL_SRC show-ref --hash $TAG_NAME)
+
+	if [[ $TAG_SHA1 =~ $CURRENT* ]]; then
```
derkling: Not entirely sure, but could it happen that we find a TAG which still do not match the CURRENT?
In that case, is not more generic to always report the tag found?

```python
TAG_SHA1=$(git -C $KERNEL_SRC show-ref --hash $TAG_NAME)
c_info "   Corresponding tag: $TAG_NAME"
[[ $TAG_SHA1 =~ $CURRENT* ]] && return $OK

return $EAGAIN
```

As per above comment, I would put this last chunk into a dedicated patch, since this is really adding a new functionality while the previous chunk is improving an existing one.... as a bonus you also get two more patches in LISA :-P

tools/wltests/test_series
balsini: Good.

tools/wltests/test_series
balsini: Good.

tools/wltests/test_series
balsini: Good.

tools/wltests/test_series
balsini: OMG. :)
Good.

tools/wltests/test_series
balsini: It does not change too much, but OK, your solution would be more verbose.
BUT since the last log message shows the kernel version of the device, this would be misleading.

I'll try to change the message text accordingly.

balsini:
@derkling , the new version is online :+1:

PR #572, opened 2018-03-02
RFC: static sanity checks for energy models

Elieva:
@derkling, @credp, @qperret, @valschneider 
Please, can you have a look to check if it covers all the tests that you had in mind for the sanity check of the energy model? and if the global form of the tests is fine for you (that the tests are executed in preliminary.py and test_energy_model.py etc...) or if you have other suggestions?

### DIFF
--- a/libs/utils/platforms/hikey960_energy.py
+++ b/libs/utils/platforms/hikey960_energy.py
@@ -1,0 +1,115 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2016, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from collections import OrderedDict
+
+from energy_model import (ActiveState, EnergyModelNode, EnergyModelRoot,
+                          PowerDomain, EnergyModel)
+
+a53_cluster_active_states = OrderedDict([
+    (533000, ActiveState(power=12)),
+    (999000, ActiveState(power=22)),
+    (1402000, ActiveState(power=36)),
+    (1709000, ActiveState(power=67)),
+    (1844000, ActiveState(power=144)),
+])
+
+# TODO warn if any of the idle states aren't represented by power domains
+a53_cluster_idle_states = OrderedDict([
+    ("WFI",               12),
+    ("cpu-sleep-0",       12),
+    ("cluster-sleep-0",   0),
+])
+
+a53_cpu_active_states = OrderedDict([
+    (533000, ActiveState(capacity=133, power=87)),
+    (999000, ActiveState(capacity=250, power=164)),
+    (1402000, ActiveState(capacity=351, power=265)),
+    (1709000, ActiveState(capacity=429, power=388)),
+    (1844000, ActiveState(capacity=462, power=502)),
+])
+
+a53_cpu_idle_states = OrderedDict([
+    ("WFI",               5),
+    ("cpu-sleep-0",       0),
+    ("cluster-sleep-0",   0),
+])
+
+a53s = [0, 1, 2, 3]
+
+def a53_cpu_node(cpu):
+    return EnergyModelNode(cpu=cpu,
+                           active_states=a53_cpu_active_states,
+                           idle_states=a53_cpu_idle_states)
+
+a72_cluster_active_states = OrderedDict([
+    (903000, ActiveState(power=102)),
+    (1421000, ActiveState(power=124)),
+    (1805000, ActiveState(power=221)),
+    (2112000, ActiveState(power=330)),
+    (2362000, ActiveState(power=433)),
+])
+
+a72_cluster_idle_states = OrderedDict([
+    ("WFI",               102),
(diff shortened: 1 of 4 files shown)


libs/utils/platforms/hikey960_energy.py
```diff
@@ -0,0 +1,115 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2016, ARM Limited and contributors.
```
valschneider: 2018, Arm ?

libs/utils/platforms/hikey960_energy.py
```diff
@@ -0,0 +1,115 @@
...
+])
+
+a53s = [0, 1, 2, 3]
+
+def a53_cpu_node(cpu):
+    return EnergyModelNode(cpu=cpu,
+                           active_states=a53_cpu_active_states,
+                           idle_states=a53_cpu_idle_states)
+
+a72_cluster_active_states = OrderedDict([
```
valschneider: Nit: the dt reports A72 but it's actually A73...

tests/utils/em.py
```diff
@@ -0,0 +1,313 @@
...
+    # requiered opp
+    for cap in capacities:
+        if cap >= util:
+            power = powers[capacities.index(cap)] * (float(util) / cap)
+            return float(cap) / power
+
+    raise ValueError('The utilization {} is larger than the possible '
+                     'opp for {}'.format(util, cpu.name))
+
+def check_overutilized_area(energy_model):
```
valschneider: All of the other tests look fine to me, but this one got me slightly confused.
IIUC, in short you want to assert here that the biggest OPPs on the LITTLEs are still more energy efficient than the smallest OPPs on the bigs, and that we don't skip those because of the 80% margin.

Maybe that should actually be split into two tests, i.e:
1) Is there any OPP on the bigs that is more power-efficient than on the LITTLEs ?
2) Is there any OPP on the LITTLEs that is more power-efficient that on the bigs but we'll skip it because of that 80% margin ?

What do you think ?

Elieva: Thanks for your comment
Yes it seems better that way :) I will work on it

tests/eas/preliminary.py
```diff
@@ -188,6 +190,66 @@ def test_energy_model_present(self):
                 '- Kernel built without (CONFIG_SCHED_DEBUG && CONFIG_SYSCTL)\n'
                 '- No energy model in kernel')
 
+class TestEnergyModelSanity(BasicCheckTest):
```
qperret: I don't think this test should live in `tests/eas/preliminary.py`. Some platforms will fail those tests but that doesn't mean EAS won't work on them. The semantic for `prelminary.py` is: if one of the tests fail, go fix your platform before running anything else because something is broken. Your test is somewhat different, so I'd prefer to to have it in a new file called `tests/eas/energy_model.py` or something like that. That said, others might disagree. @credp @derkling : any thoughts ?

tests/utils/em.py
```diff
@@ -0,0 +1,313 @@
...
+    succeed = True
+    for cpu in energy_model.cpu_nodes:
+        powers = _get_powers_list(cpu)
+        cap = _get_capacities_list(cpu)
+        energy = [float(powers[i])/cap[i] for i in range(len(powers))]
+        # Check that power / cap is strickly increasing which is equivalent
+        # to check that the cap / power (energy efficiency) is decreasing
+        if not _is_list_strickly_increasing(energy):
+            failed_nodes.append(cpu.name)
+            succeed = False
```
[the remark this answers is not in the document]
Elieva: The goal here was to get the list of cpus for which the energy model does not respect the property. If i add the break there it would only return the first faulty cpu.

tests/lisa/test_energy_model.py
```diff
@@ -423,3 +424,74 @@ def test_estimate_from_trace(self):
             row = df.iloc[i]
             self.assertAlmostEqual(row.name, exp_index, places=4)
             self.assertDictEqual(row.to_dict(), exp_values)
+
+class TestEnergyModelSanityCheck(TestCase):
```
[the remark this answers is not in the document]
Elieva: Done --> As discussed offline, I added tests here for the sanity check functions using an hardcoded model. And I moved the test for the platform energy models given in Lisa in a separate test file in tests/eas/energy_model.py

tests/utils/em.py
```diff
@@ -0,0 +1,313 @@
...
+
+    :returns: The list of capacities associates with the given node.
+    """
+    cap = []
+    for freq in node.active_states.keys():
+        cap.append(node.active_states[freq].capacity)
+    return cap
+
+def _is_list_strickly_increasing(l):
+    return all(x<y for x, y in zip(l, l[1:]))
```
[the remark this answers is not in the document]
Elieva: Done --> I changed it for a for loop

tests/eas/preliminary.py
Elieva: Done --> the tests has been moved into tests/eas/energy_model.py

tests/utils/em.py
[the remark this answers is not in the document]
douglas-raillard-arm: all() solution is slowed down by having to call next() on the iterator for every new item (and restoring its state). Using a list comprehension solves that.
The for loop is fast because it does not need to have this iterator state to be restored every time and is similar to all([...]).
The fastest solution of all (on a sorted list) is a plain "l == sorted(l)". This is because it just involves 2 calls into C code (sorted builtin and comparison), instead of executing Python. CPython is so slow that anything that reduces the number of opcodes executed will reduce the runtime, regardless of the algorithmic complexity of what is actually done:
              0 LOAD_FAST                   0 (l)
              2 LOAD_GLOBAL             0 (sorted)
              4 LOAD_FAST                   0 (l)
              6 CALL_FUNCTION         1
              8 COMPARE_OP              2 (==)
             10 RETURN_VALUE

qperret:
See  #577

PR #574, opened 2018-03-09
load_tracking.py: Fix root group util test

douglas-raillard-arm:
The root group contains more tasks than what we put in it, and we cannot
completely control when they execute. This leads to test failure because
the measured utilization in this group is higher than expected. Since
the kind of children (tasks and groups) the root group has are also
found in other groups we fully control, we can simply drop these tests
without reducing coverage.


tests/eas/load_tracking.py
```diff
@@ -1180,9 +1171,9 @@ class UnbalancedHierarchy(_PELTTaskGroupsTest):
                                        +-----+
                                        | "/" |
                                        +-----+
-                                       /     \
-                                      /       \
-                                  +------+   t0_1
+                                       /
```
derkling: Since you are at modifying this comment... can you also remove all the double `/` that we use to represent connections?
... it would make the file shorted and more consistent with the tests @Elieva recently added.

tests/eas/load_tracking.py
```diff
@@ -943,7 +943,7 @@ class _PELTTaskGroupsTest(LisaTest):
         'modules': ['cpufreq', 'cgroups'],
     }
     # Allowed error margin
-    allowed_util_margin = 0.02
+    allowed_util_margin = 0.04
```
derkling: Why do we need to increase this... that's a 4% margin, right?
What's the failure rate if we keep it to 2%

derkling: I've also just notice that `allowed_util_margin` this seems to be something similar to `ERROR_MARGIN_PCT` defined at the top of the file... where it's defined to be 15%

Despite I reviewed @Elieva code, I did not notice this difference before... which could explain why the original tests from @mdigiorgio was passing and these new are not.

Should not we use the same margin across all these tests?... and in that case I would really like the smallest number (2% or 4%)... I'm just concerned that above tests could fails... but maybe it's for the same reason this patch is trying to fix... and then perhaps we can find a solution suitable for both casese.

tests/eas/load_tracking.py
```diff
@@ -1155,12 +1152,6 @@ def _getExperimentsConf(cls, test_env):
     def setUpClass(cls, *args, **kwargs):
         super(TwoGroupsCascade, cls).runExperiments(*args, **kwargs)
 
-    def test_util_root_group(self):
-        """
-        Test utilization propagation to cgroup root
-        """
-        return self._test_group_util('/')
```
derkling: Since the root group can be kind-of different then child groups... I was wondering if we can easily go for a slightly different fix where, in `_test_group_util` we use `assertGreaterEqual` instead of `assertAlmostEqual` at L1075 above.

Actually, since in `_test_group_util` there are two different checks, one for childs and one for the group itself, maybe we can have a more generic solution by passing in the assert functions are parameters, e.g.

```python
def _test_group_util(self, group, group_assert=self.assertAlmostEqual, child_assert=self.assertAlmostEqual):
```

And the test the root groups with:
```python
return self._test_group_util('/', group_assert=self.assertGreaterEqual)
```

tests/eas/load_tracking.py
douglas-raillard-arm: I don't remember the exact figure, but with 4%, we don't have any failure at all on 800 iterations. I haven't reviewed all the tests to see how they use the margin. It would make little sense to me to force using the same margin for unrelated quantities. The 15% may benefit from more testing to see if it actually needs to be that large.

tests/eas/load_tracking.py
douglas-raillard-arm: @Elieva  and I tested with assertGreaterEqual, and we concluded that changing the assert based on the group was just more code bloat without any real added value (unless the root group is actually  special so we want to test it just because it is the root).
Having different asserts also requires recoding a one-sided assertAlmostEqual since assertGreaterEqual fails sometimes when it is slightly smaller (which is expected).

tests/eas/load_tracking.py
derkling: Well, it's seems to be used also for PELT realted tests, in `PELTTasksTest`... but, in any case, maybe we can keep this bit for another time.

tests/eas/load_tracking.py
derkling: Mmm... have you tryied by replacing both the `assertAlmostEqual` into `_test_group_util`?

Because I think it's needed only for the top level group... and again, with the above solution we can actually use it only when we test the root group.

tests/eas/load_tracking.py
douglas-raillard-arm: Done

tests/eas/load_tracking.py
douglas-raillard-arm: What we need is to check that X is within these ranges:
* for the root group: [X - delta1; X + delta2], with delta2 > delta1
* for the other groups: [X - delta1; X + delta1]

delta2 needs to be large enough to account for tasks "randomly" executing, so it needs to be significantly larger than delta1. delta1 is still needed as the average we compute is can sometimes be slightly smaller than the measured value. assertAlmostEqual cannot be used for root group, neither can assertGreaterEqual unless you relax the test even more. All in all, I don't really see the value of writing complicated code if the root group does not have any special property. We can just test other groups with similar set of children and we will get the same coverage of utilization computation and propagation to parents.

derkling: In your example above, why can't use assertGreaterEqual(X - delta1) for the group we call `_test_group_util()` on?

douglas-raillard-arm: assertGreaterEqual(X - delta1) check that [X - delta1; +infinite] instead of that: [X - delta1; X + delta2]
That is going to only catch half of the issues, that would also be caught when testing the other groups, at the cost of extra bloat to support it.

douglas-raillard-arm: Outcome: we settled on the fact that the root group is somehow special in the kernel (different code paths), so testing non-root groups does not provide full coverage. As a result, the root group will be checked with a "greater or equal" test to avoid issues when unexpected tasks are executing in it.

PR #576, opened 2018-03-13
Wltest monsoon composite energy

credp:
Hi guys, this is what I'm using right now for Monsoon energy meter support. It's slightly ahead of the one we merged already. I think I need to rebase and push again before we merge this.

### DIFF
--- a/libs/utils/wa_results_collector.py
+++ b/libs/utils/wa_results_collector.py
@@ -213,6 +213,37 @@ class WaResultsCollector(object):
 
         # results.csv contains all the metrics reported by WA for all jobs.
         df = pd.read_csv(os.path.join(wa_dir, 'results.csv'))
+        # When using Monsoon, the device is a single channel which reports
+        # two metrics. This means that devlib's DerivedEnergymeasurements class
+        # cannot see the output. Due to the way that the monsoon.py script
+        # works, it looks difficult to change Monsoon over to the Acme way of
+        # operating. As a workaround, let's mangle the results here instead.
+        unique_metrics = df['metric'].unique()
+        if 'device_total_energy' not in unique_metrics:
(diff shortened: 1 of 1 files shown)


derkling:
I've already merged a patch (c51081abf093b) which fixes the same issue of the first patch in this series, which can then be dropped.

For the second patch instead I need some more time to review/test it.

The last patch should not be required if we rebase the second one on the master branch.

libs/utils/wa_results_collector.py
```diff
@@ -206,6 +206,37 @@ def _read_wa_dir(self, wa_dir):
 
         # results.csv contains all the metrics reported by WA for all jobs.
         df = pd.read_csv(os.path.join(wa_dir, 'results.csv'))
+        # When using Monsoon, the device is a single channel which reports
+        # two metrics. This means that devlib's DerivedEnergymeasurements class
+        # cannot see the output. Due to the way that the monsoon.py script
+        # works, it looks difficult to change Monsoon over to the Acme way of
+        # operating. As a workaround, let's mangle the results here instead.
+        __metrics = df['metric'].unique()
```
douglas-raillard-arm: __metrics will be mangled into _WaResultsCollector__metrics by the Python parser since it appears inside a class definition block. That is not really harmful as this name is not really exposed, but that can still be confusing when looking at backtraces. The leading double underscore convention in the kernel does not mix well with Python unfortunately:
```
>>> class A:
...   def f(self):
...    print(__metric)
...
>>> a=A()
>>> a.f()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in f
NameError: name '_A__metric' is not defined
```

credp: I changed the name to unique_metrics instead in this new version.

PR #577, opened 2018-03-15
Sanity check for energy model

Elieva:
Close #572 
This patches add sanity tests for the energy model available in Lisa and allow to test also the energy model available on the target.


qperret:
See #612

PR #580, opened 2018-03-21
Avoid file sharing between platforms

Elieva:
When run in parallel on several platforms, some files with the workload description were shared between the different platforms which was resulting in calibration failures.
This patch modifies the creation of the workload file to give them an unique name to avoid the sharing 
and changed the control sequence in the calibration to avoid using the 'bl' module of the target.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -20,6 +20,7 @@ import json
 import os
 import re
 
+from tempfile import NamedTemporaryFile
 from collections import OrderedDict
 from wlgen import Workload
 from devlib.utils.misc import ranges_to_list
@@ -218,14 +219,11 @@ class RTA(Workload):
 
             return 'CPU{0:d}'.format(target_cpu)
 
-    def _confCustom(self):
+    def _confCustom(self, tf):
 
         rtapp_conf = self.params['custom']
 
-        self._log.info('Loading custom configuration:')
-        self._log.info('   %s', rtapp_conf)
         self.json = '{0:s}_{1:02d}.json'.format(self.name, self.exc_id)
-        ofile = open(self.json, 'w')
 
         calibration = self.getCalibrationConf()
         # Calibration can either be a string like "CPU1" or an integer, if the
@@ -252,6 +250,8 @@ class RTA(Workload):
                                  "a filename or an embedded dictionary")
         else:
             # We assume we are being passed a filename instead of a dict
+            self._log.info('Loading custom configuration:')
+            self._log.info('   %s', rtapp_conf)
             ifile = open(rtapp_conf, 'r')
 
         for line in ifile:
@@ -259,20 +259,20 @@ class RTA(Workload):
                 raise ValueError('Workload duration not specified')
             for src, target in replacements.iteritems():
                 line = line.replace(src, target)
-            ofile.write(line)
+            tf.write(line)
 
         if isinstance(ifile, file):
             ifile.close()
-        ofile.close()
+        tf.close()
 
-        with open(self.json) as f:
-            conf = json.load(f)
-        for tid in conf['tasks']:
+        with open(tf.name) as f:
+            self.rta_profile = json.load(f)
+        for tid in self.rta_profile['tasks']:
             self.tasks[tid] = {'pid': -1}
 
         return self.json
 
-    def _confProfile(self):
+    def _confProfile(self, tf):
 
         # Sanity check for task names
         for task in self.params['profile'].keys():
@@ -433,9 +433,9 @@ class RTA(Workload):
 
         # Generate JSON configuration on local file
         self.json = '{0:s}_{1:02d}.json'.format(self.name, self.exc_id)
-        with open(self.json, 'w') as outfile:
-            json.dump(self.rta_profile, outfile,
-                      indent=4, separators=(',', ': '))
+        json.dump(self.rta_profile, tf,
+                  indent=4, separators=(',', ': '))
+        tf.close()
 
         return self.json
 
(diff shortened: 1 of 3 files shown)


libs/wlgen/wlgen/rta.py
```diff
@@ -505,14 +505,19 @@ def conf(self,
...
-        # Setup class-specific configuration
-        if kind == 'custom':
-            self._confCustom()
-        elif kind == 'profile':
-            self._confProfile()
-
-        # Move configuration file to target
-        self.target.push(self.json, self.run_dir)
+        try:
+            tf = NamedTemporaryFile(delete=False)
```
douglas-raillard-arm: Would benefit from:
with NamedTemporaryFile() as tf:
 ...

libs/wlgen/wlgen/rta.py
```diff
@@ -505,14 +505,19 @@ def conf(self,
...
+            # Setup class-specific configuration
+            if kind == 'custom':
+                self._confCustom(tf)
+            elif kind == 'profile':
+                self._confProfile(tf)
+
+            # Move configuration file to target
+            self.target.push(tf.name, os.path.join(self.run_dir,self.json))
+        finally:
```
douglas-raillard-arm: the "finally" clause becomes unnecessary as the as the context manager will close and delete the file. That also avoids the issue of tf not being defined if NamedTemporaryFile raises an exception while creating the file.

libs/wlgen/wlgen/rta.py
```diff
@@ -252,27 +250,29 @@ def _confCustom(self):
...
                 raise ValueError('Workload duration not specified')
             for src, target in replacements.iteritems():
                 line = line.replace(src, target)
-            ofile.write(line)
+            tf.write(line)
 
         if isinstance(ifile, file):
             ifile.close()
-        ofile.close()
+        tf.close()
```
douglas-raillard-arm: I don't think tf should be closed since the context manager takes care of that.

libs/wlgen/wlgen/rta.py
```diff
@@ -433,9 +433,9 @@ def _confProfile(self):
 
         # Generate JSON configuration on local file
         self.json = '{0:s}_{1:02d}.json'.format(self.name, self.exc_id)
-        with open(self.json, 'w') as outfile:
-            json.dump(self.rta_profile, outfile,
-                      indent=4, separators=(',', ': '))
+        json.dump(self.rta_profile, tf,
+                  indent=4, separators=(',', ': '))
+        tf.close()
```
douglas-raillard-arm: same comment as in _confCustom: tf should probably not be closed here

libs/wlgen/wlgen/rta.py
```diff
@@ -168,14 +168,18 @@ def calibrate(target):
...
-            if pload[bcpu] > pload[lcpu]:
-                log.warning('Calibration values reports big cores less '
-                            'capable than LITTLE cores')
-                raise RuntimeError('Calibration failed: try again or file a bug')
-            bigs_speedup = ((float(pload[lcpu]) / pload[bcpu]) - 1) * 100
+        if target.big_core:
+            bcpu = [i for i, t in enumerate(target.core_names) if t == target.big_core]
+            lcpu = [i for i, t in enumerate(target.core_names) if t == target.little_core]
+            for l in lcpu:
+                for b in bcpu:
```
douglas-raillard-arm: nitpicking: you can use itertools.product for these kind of loops :)
https://docs.python.org/2/library/itertools.html#itertools.product

libs/wlgen/wlgen/rta.py
```diff
@@ -252,27 +250,29 @@ def _confCustom(self):
...
 
-        with open(self.json) as f:
-            conf = json.load(f)
-        for tid in conf['tasks']:
+        with open(tf.name) as f:
+            self.rta_profile = json.load(f)
+        for tid in self.rta_profile['tasks']:
             self.tasks[tid] = {'pid': -1}
 
         return self.json
```
douglas-raillard-arm: should probably be updated to tf.name

libs/wlgen/wlgen/rta.py
```diff
@@ -433,9 +433,9 @@ def _confProfile(self):
...
         # Generate JSON configuration on local file
         self.json = '{0:s}_{1:02d}.json'.format(self.name, self.exc_id)
-        with open(self.json, 'w') as outfile:
-            json.dump(self.rta_profile, outfile,
-                      indent=4, separators=(',', ': '))
+        json.dump(self.rta_profile, tf,
+                  indent=4, separators=(',', ': '))
+        tf.close()
 
         return self.json
```
douglas-raillard-arm: should probably be updated to tf.name

libs/wlgen/wlgen/rta.py
```diff
@@ -168,14 +168,18 @@ def calibrate(target):
...
-            bigs_speedup = ((float(pload[lcpu]) / pload[bcpu]) - 1) * 100
+        if target.big_core:
+            bcpu = [i for i, t in enumerate(target.core_names) if t == target.big_core]
+            lcpu = [i for i, t in enumerate(target.core_names) if t == target.little_core]
+            for l in lcpu:
+                for b in bcpu:
+                    if pload[b] > pload[l]:
+                        log.warning('Calibration values reports big cores less'
+                                    ' capable than LITTLE cores')
+                        raise RuntimeError('Calibration failed: try again or '
```
douglas-raillard-arm: Could we just throw the exception with the message of log.warning ? that should be enough for debugging since we will get the exception message in the xUnit file generated by nosetest.

douglas-raillard-arm:
Close as an equivalent reworked PR will be pushed soon

PR #581, opened 2018-03-28
events: target config can specified additional events

Elieva:
The target config file can specified additionnal events that will be
registered during the tests with the label 'debug_events'.
These events will be added to the list of events needed by the test
and can be used afterward to help debugging if needed.

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -720,6 +720,10 @@ class TestEnv(ShareState):
         events = FTRACE_EVENTS_DEFAULT
         if 'events' in ftrace:
             events = ftrace['events']
+        if 'debug_events' in self.conf:
+            # Need to convert from unicode object to str
+            dbg_events = [s.encode('utf-8') for s in self.conf['debug_events']]
+            events = list(set(events).union(dbg_events))
 
         functions = None
         if 'functions' in ftrace:


libs/utils/env.py
```diff
@@ -720,6 +720,10 @@ def _init_ftrace(self, force=False, conf=None):
         events = FTRACE_EVENTS_DEFAULT
         if 'events' in ftrace:
             events = ftrace['events']
+        if 'debug_events' in self.conf:
+            # Need to convert from unicode object to str
+            dbg_events = [s.encode('utf-8') for s in self.conf['debug_events']]
+            events = list(set(events).union(dbg_events))
```
douglas-raillard-arm: would change to:
        mandatory_events = set(ftrace.get('events', []))
        dbg_events = set(self.conf.get('debug_events', []))
        events = sorted(e.encode('ascii') for e in (mandatory_events | dbg_events))

Doing it this way avoids a NameError when debug_events is present but not 'events'
'ascii' codec is preferable because if the rest of the code is expecting an str instead of unicode, it probably won't be able to deal with multibyte sequences in any case. That means it's better to catch such non-ascii characters early by only allowing ascii characters.

douglas-raillard-arm:
Close as a reworked PR will be pushed soon

PR #583, opened 2018-04-06
generic.py: fix EnergyModelWakeMigration

douglas-raillard-arm:
Change the task utilization so that the optimal placement agrees with
EAS placement. 50% was high enough for EAS to move the task to a big
CPU, but the optimal placement was expecting a little CPU. With 70%,
they both agree that it should be on a big CPU.

### DIFF
--- a/tests/eas/generic.py
+++ b/tests/eas/generic.py
@@ -567,7 +567,7 @@ class RampDown(_EnergyModelTest):
 
 class EnergyModelWakeMigration(_EnergyModelTest):
     """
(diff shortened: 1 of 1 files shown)


valschneider:
LGTM, I'd just like a little more detail in the commit message in regards to what the problem was (something that explains that this happened on a platform where the LITTLE CPUs had unexpectedly high capacities)

derkling:
Let's merge as it is... although in the future I would really suggest to generate load for each synthetic based on energy model reported capacities.

PR #586, opened 2018-04-06
Some changes proposal and fixes

vingu-linaro:
Hi these, patches are few changes that helps to use the .ipnb like sched-full-evaluation.ipnb as a python script for post processing.

### DIFF
--- a/libs/utils/wa_results_collector.py
+++ b/libs/utils/wa_results_collector.py
@@ -1006,7 +1006,7 @@ class WaResultsCollector(object):
                         # Find a p-value which hopefully represents the
                         # (complement of the) certainty that any difference in
                         # the mean represents something real.
-                        pvalue =  ttest_ind(group_results, base_results, equal_var=False).pvalue
+                        _, pvalue = ttest_ind(group_results, base_results, equal_var=False)
 
                     comparisons.append(Comparison(
                         metric, test, inv_id,
--- a/libs/utils/wa_results_collector.py
+++ b/libs/utils/wa_results_collector.py
@@ -111,12 +111,19 @@ class WaResultsCollector(object):
                      this can take some time, so the extracted metrics are
                      cached in the provided output directories. Set this param
                      to False to disable this caching.
+
+    :param display_charts: This class uses IPython.display module to render some
+                           charts of workloads' results. But we also want to use
+                           this class without rendering any charts when we are
+                           only interested in table of figures. Set this param
+                           to False if you only want table of results but not
+                           display them.
     """
     RE_WLTEST_DIR = re.compile(r"wa\.(?P<sha1>\w+)_(?P<name>.+)")
 
     def __init__(self, base_dir=None, wa_dirs=".*", platform=None,
                  kernel_repo_path=None, parse_traces=True,
-                 use_cached_trace_metrics=True):
+                 use_cached_trace_metrics=True, display_charts=True):
 
         self._log = logging.getLogger('WaResultsCollector')
 
@@ -143,6 +150,7 @@ class WaResultsCollector(object):
         if not self.parse_traces:
             self._log.warning("Trace parsing disabled")
         self.use_cached_trace_metrics = use_cached_trace_metrics
+        self.display_charts = display_charts
 
         df = pd.DataFrame()
         df_list = []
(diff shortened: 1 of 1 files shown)


vingu-linaro:
Hi,

Any comments on the patches ?

libs/utils/wa_results_collector.py
```diff
@@ -113,10 +113,12 @@ class WaResultsCollector(object):
     """
     def __init__(self, base_dir=None, wa_dirs=".*", platform=None,
                  kernel_repo_path=None, parse_traces=True,
-                 use_cached_trace_metrics=True):
+                 use_cached_trace_metrics=True, charts=True):
```
valschneider: `display_charts` would be a more fitting name

libs/utils/wa_results_collector.py
```diff
@@ -458,11 +463,17 @@ def _get_extra_job_metrics(self, job_dir, workload):
         # think we can simplify this.
         for artifact_name, path in artifacts.iteritems():
             if os.stat(path).st_size == 0:
-                self._log.info(" no data for %s",  path)
+                #self._log.info(" no data for %s",  path)
```
valschneider: Don't we want to keep that log here ? If we don't, the commented line can be removed.

libs/utils/wa_results_collector.py
```diff
@@ -458,11 +463,17 @@ def _get_extra_job_metrics(self, job_dir, workload):
...
                 continue
 
             if artifact_name.startswith('energy_instrument_output'):
-                df = pd.read_csv(path)
+                #self._log.info(" power file %s",  path)
```
valschneider: Stray comment ?

libs/utils/wa_results_collector.py
```diff
@@ -972,7 +972,7 @@ def find_comparisons(self, base_id=None, by='kernel'):
                         # Find a p-value which hopefully represents the
                         # (complement of the) certainty that any difference in
                         # the mean represents something real.
-                        pvalue =  ttest_ind(group_results, base_results, equal_var=False).pvalue
+                        tmpstat, pvalue = ttest_ind(group_results, base_results, equal_var=False)
```
valschneider: I think this might be due to discrepancies between the scipy version we use and the one used in the VM. Are you using an older distro ?

Otherwise, the change is fine with me, with a tiny nitpick: since `tmpstat` is not used, you can just write that line as:
``` python
_, pvalue = ttest_ind(...)
```

libs/utils/wa_results_collector.py
vingu-linaro: ok

libs/utils/wa_results_collector.py
vingu-linaro: yes. a debug trace that should have been removed

libs/utils/wa_results_collector.py
vingu-linaro: yes probably.
I have commented for testing because it was generating a lot of log when you parse a wa directory but you aborted WA after few tests

libs/utils/wa_results_collector.py
vingu-linaro: I'm using the vagrant image:
Linux vagrant-ubuntu-trusty-64 3.13.0-135-generic #184-Ubuntu SMP Wed Oct 18 11:55:51 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

libs/utils/wa_results_collector.py
```diff
@@ -113,10 +113,12 @@ class WaResultsCollector(object):
     """
     def __init__(self, base_dir=None, wa_dirs=".*", platform=None,
                  kernel_repo_path=None, parse_traces=True,
-                 use_cached_trace_metrics=True):
+                 use_cached_trace_metrics=True, charts=True):
 
         self._log = logging.getLogger('WaResultsCollector')
 
+        self.charts = charts
```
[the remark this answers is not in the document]
vingu-linaro: ok.
I have also added the following description for the parameter:
+
+    :param display_charts: This class uses IPython.display module to render some
+                           charts of workloads' results. But we also want to use
+                           this class without rendering any charts when we are
+                           only interested in table of figures. Set this param
+                           to False if you only want table of results but not
+                           display them.
     """

libs/utils/wa_results_collector.py
valschneider: I think some people have been running Trusty without issues, but then maybe they bypassed the package manager to get a newer version of scipy (seems that the return fields were renamed between scipy 0.15.1 and scipy 0.16.0).

In any case, the change still works on newer versions so we can go with it

credp: @valschneider - I think this means that the review is OK now?

PR #588, opened 2018-04-10
cgroups migration test: catch error when migrating task

douglas-raillard-arm:
The PID of the task that is not migrated need to be found
by looking into the cgroups file and filtering these information.
If something goes wrong it may happen that the task PID is not
found. In this case an error is raised.

Signed-off-by: Douglas Raillard <douglas.raillard@arm.com>

### DIFF
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -1370,9 +1370,12 @@ class CgroupsMigrationTest(_PELTTaskGroupsTest):
         task = cls.root_group.get_child('t1_2')
         new_taskgroup = cls.root_group.get_child('/tg2')
         task.change_taskgroup(new_taskgroup)
-        tg1_task = cgroups.tasks('/tg1', filter_tname='rt-app',
-                                 filter_tcmdline='t1_1')
-        exclude = next(iter(tg1_task))
+        try:
+            [exclude] = cgroups.tasks('/tg1', filter_tname='rt-app',
+                                      filter_tcmdline='t1_1').keys()
+        except ValueError:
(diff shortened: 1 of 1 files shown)


tests/eas/load_tracking.py
```diff
@@ -1370,9 +1370,12 @@ def _migrate_task(cls, test_env):
         task = cls.root_group.get_child('t1_2')
         new_taskgroup = cls.root_group.get_child('/tg2')
         task.change_taskgroup(new_taskgroup)
-        tg1_task = cgroups.tasks('/tg1', filter_tname='rt-app',
-                                 filter_tcmdline='t1_1')
-        exclude = next(iter(tg1_task))
+        try:
+            [exclude] = cgroups.tasks('/tg1', filter_tname='rt-app',
```
valschneider: The previous version used a ```next(iter(x))``` construct, whereas here we expect the ```keys()``` call to return a list with a single item. Is it a safe assumption ?

douglas-raillard-arm: It is safe since we only have one task in that group. next(iter()) is not a great pattern since you take the risk of having a wild StopIteration exception bubbling. Since that exception is swallowed silently when it is raised by an iterator, it can lead to some debugging issues (a loop stopping for seemingly no reason).

douglas-raillard-arm: exclude is not a list, it is the 1st item of the iterable returned by cgroups.task. We can make that more explicit with:
excludes = cgroups.task(...)
```
if len(excludes) != 1:
    raise RuntimeError(...)
exclude  = excludes[1]
```

PR #589, opened 2018-04-13
Add trace analysis scripts

cobrien7:
runnable: measure task runnable times
runtime: calculate runtime of tasks as percent of CPU.

### DIFF
--- a/tools/analysis/runnable.py
+++ b/tools/analysis/runnable.py
@@ -1,0 +1,165 @@
+#!/usr/bin/env python
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2017, ARM Limited, Google, and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from collections import namedtuple
+import trappy
+import argparse
+
+EventData = namedtuple('EventData', ['time', 'event', 'data'])
+LatData = namedtuple('LatData', ['pid', 'switch_data', 'wake_data', 'last_wake_data', 'running', 'wake_pend', 'latency', 'lat_total'])
+
+parser = argparse.ArgumentParser(description='Analyze runnable times')
+
+parser.add_argument('--trace', dest='trace_file', action='store', required=True,
+                    help='trace file')
+
+parser.add_argument('--rt', dest='rt', action='store_true', default=False,
+                    help='only consider RT tasks')
+
+parser.add_argument('--normalize', dest='normalize', action='store_true', default=False,
+                    help='normalize time')
+
+parser.add_argument('--rows', dest='nrows', action='store', default=20, type=int,
+                    help='normalize time')
+
+parser.add_argument('--total', dest='lat_total', action='store_true', default=False,
+                    help='sort by total runnable time')
+
+parser.add_argument('--start-time', dest='start_time', action='store', default=0, type=float,
+                    help='trace window start time')
+
+parser.add_argument('--end-time', dest='end_time', action='store', default=None, type=float,
+                    help='trace window end time')
+
+args = parser.parse_args()
+
+path_to_html = args.trace_file
+rtonly = args.rt
+nrows = args.nrows
+
+# Hash table of pid -> LatData named tuple
+latpids = {}
+
+def switch_cb(data):
+    event = "switch"
+    prevpid = data['prev_pid']
+    nextpid = data['next_pid']
+    time = data['Index']
+
+    e = EventData(time, event, data)
+
+    # prev pid processing (switch out)
+    if latpids.has_key(prevpid):
+        if latpids[prevpid].running == 1:
+            latpids[prevpid] = latpids[prevpid]._replace(running=0)
+        if latpids[prevpid].wake_pend == 1:
+            print "Impossible: switch out during wake_pend " + str(e)
+            latpids[prevpid] = latpids[prevpid]._replace(wake_pend=0)
+            return
+
+    if not latpids.has_key(nextpid):
+        return
+
+    # nextpid processing  (switch in)
+    pid = nextpid
+    if latpids[pid].running == 1:
+        print "INFO: previous pid switch-out not seen for an event, ignoring\n" + str(e)
+        return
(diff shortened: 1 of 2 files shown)


valschneider:
Thanks for the submission ! I definitely want to test/review this but it'll have to wait a bit longer, sorry... Stay tuned !

tools/analysis/runnable.py
```diff
@@ -0,0 +1,165 @@
...
+def switch_cb(data):
+    event = "switch"
+    prevpid = data['prev_pid']
+    nextpid = data['next_pid']
+    time = data['Index']
+
+    e = EventData(time, event, data)
+
+    # prev pid processing (switch out)
+    if latpids.has_key(prevpid):
```
douglas-raillard-arm: Please use "if prevpid in latpids:" as has_key is deprecated

tools/analysis/runnable.py
```diff
@@ -0,0 +1,165 @@
...
+        if latpids[prevpid].running == 1:
```
douglas-raillard-arm: As "running" is a boolean, can be simplified into:
        if latpids[prevpid].running:

This also applies to testing for 0:
        if not latpids[prevpid].running:

tools/analysis/runnable.py
```diff
@@ -0,0 +1,165 @@
...
+args = parser.parse_args()
+
+path_to_html = args.trace_file
+rtonly = args.rt
+nrows = args.nrows
+
+# Hash table of pid -> LatData named tuple
+latpids = {}
+
+def switch_cb(data):
```
douglas-raillard-arm: *_cb functions should take latpids as a parameter to avoid wild accesses to the latpids global.
functools.partial can be used to tie latpids and the function when setting up the callbacks:
https://docs.python.org/2/library/functools.html#functools.partial

tools/analysis/runnable.py
```diff
@@ -0,0 +1,165 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from collections import namedtuple
+import trappy
+import argparse
+
+EventData = namedtuple('EventData', ['time', 'event', 'data'])
+LatData = namedtuple('LatData', ['pid', 'switch_data', 'wake_data', 'last_wake_data', 'running', 'wake_pend', 'latency', 'lat_total'])
```
douglas-raillard-arm: Maybe LatData could benefit from being renamed to LatencyData to make it more obvious

tools/analysis/runnable.py
```diff
@@ -0,0 +1,165 @@
...
+
+    latpids[pid] = latpids[pid]._replace(last_wake_data = e, wake_pend=1)
+
+
+if args.normalize:
+    kwargs = { 'window': (args.start_time, args.end_time) }
+else:
+    kwargs = { 'abs_window': (args.start_time, args.end_time) }
+
+systrace_obj = trappy.SysTrace(name="systrace", path=path_to_html, \
```
douglas-raillard-arm: Unnecessary backslash at the end of the line

tools/analysis/runnable.py
```diff
@@ -0,0 +1,165 @@
...
+systrace_obj.apply_callbacks({ "sched_switch": switch_cb, "sched_wakeup": wake_cb, \
+                               "sched_waking": wake_cb })
+
+# Print the results: PID, latency, start, end, sort
+if args.lat_total:
+    result = sorted(latpids.items(), key=lambda x: x[1].lat_total, reverse=True)
+else:
+    result = sorted(latpids.items(), key=lambda x: x[1].latency, reverse=True)
+
+print "\t".join([
```
douglas-raillard-arm: nitpicking: Could use str.format for better readability

tools/analysis/runtime.py
```diff
@@ -0,0 +1,159 @@
...
+parser = argparse.ArgumentParser(description='Analyze runtimes')
+
+parser.add_argument('--trace', dest='trace_file', action='store', required=True,
+                    help='trace file')
+
+parser.add_argument('--normalize', dest='normalize', action='store_true', default=False,
+                    help='normalize time')
+
+parser.add_argument('--rows', dest='nrows', action='store', default=20, type=int,
+                    help='normalize time')
```
douglas-raillard-arm: copy/paste issue in help message

tools/analysis/runtime.py
```diff
@@ -0,0 +1,159 @@
...
+        self.last_start_time = time
+        self.total_time = np.float64(0.0)
+        self.start_time = -1
+        self.end_time = -1
+        self.running = 1
+        self.maxrun = -1
+
+parser = argparse.ArgumentParser(description='Analyze runtimes')
+
+parser.add_argument('--trace', dest='trace_file', action='store', required=True,
```
douglas-raillard-arm: action='store' is not needed since this is the default (and is expected)

tools/analysis/runtime.py
```diff
@@ -0,0 +1,159 @@
...
+                    help='trace file')
+
+parser.add_argument('--normalize', dest='normalize', action='store_true', default=False,
```
douglas-raillard-arm: "default=False" unneeded since it is implied by action='store_true'

tools/analysis/runtime.py
douglas-raillard-arm: Are "dest=" arguments needed since they are bound to a usual variable before being used by the script ? Renaming already happens there. For the ones that are not renamed, a default dest is computed using the long option name, replacing dashes by underscores

tools/analysis/runtime.py
```diff
@@ -0,0 +1,159 @@
...
+nrows = args.nrows
+
+# Hash table for runtimes
+runpids = {}
+
+starttime = None
+endtime = None
+
+def switch_cb(data):
+    global starttime, endtime
```
douglas-raillard-arm: If some global state is needed (more than just the latpid), it may be worth putting these callbacks into a class and keep track of the state in the instance.

tools/analysis/runnable.py
```diff
@@ -0,0 +1,165 @@
...
+    time = data['Index']
+
+    e = EventData(time, event, data)
+
+    if rtonly and data["prio"] > 99:
+        return
+
+    if  not latpids.has_key(pid):
+        latpids[pid] = LatData(pid, switch_data=None, wake_data=None,
+                last_wake_data = e, running=0, latency=-1, lat_total=0, wake_pend=1)
```
douglas-raillard-arm: Boolean attributes should be booleans True/False instead of int 0/1

douglas-raillard-arm:
Seems related to https://github.com/ARM-software/lisa/pull/424 in essence. 

Analysis infrastructure has had a steady stream of improvements since then, and `lisa-plot` command line tool is now a thing. I guess it would make sense to generalize that into `lisa-trace plot` for plots and `lisa-trace report` for textual metrics report. It would display the results of analysis methods returning something similar to lisa.tests.base.TestMetric . The tool would take care of formatting all that in nice tables.

PR #593, opened 2018-04-18
Fix calibration

douglas-raillard-arm:
RTA calibration process is relying on pushing an rt-app configuration file to the target. Since that file used to always have the same name, it created a race condition if more than one LISA instance was running on the same LISA source tree.

Fix that by using a proper temporary file instead of relying on a deterministically-created one.

Thanks to @Elieva for discovering that and creating the first draft of this PR.

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -20,6 +20,7 @@ import json
 import os
 import re
 
+from tempfile import NamedTemporaryFile
 from collections import OrderedDict
 from wlgen import Workload
 from devlib.utils.misc import ranges_to_list
@@ -218,12 +219,11 @@ class RTA(Workload):
 
             return 'CPU{0:d}'.format(target_cpu)
 
-    def _confCustom(self):
+    def _confCustom(self, tf):
 
         rtapp_conf = self.params['custom']
 
         self.json = '{0:s}_{1:02d}.json'.format(self.name, self.exc_id)
-        ofile = open(self.json, 'w')
 
         calibration = self.getCalibrationConf()
         # Calibration can either be a string like "CPU1" or an integer, if the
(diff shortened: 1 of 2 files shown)


libs/wlgen/wlgen/rta.py
```diff
@@ -505,18 +506,21 @@ def conf(self,
...
-        # Move configuration file to target
-        self.target.push(self.json, self.run_dir)
-
-        self.rta_cmd  = self.target.executables_directory + '/rt-app'
-        self.rta_conf = self.run_dir + '/' + self.json
-        self.command = '{0:s} {1:s} 2>&1'.format(self.rta_cmd, self.rta_conf)
+        with NamedTemporaryFile() as tf:
+            # Setup class-specific configuration
+            if kind == 'custom':
+                self._confCustom(tf)
```
derkling: The two conf methods already return the JSON corresponding to the file generated.

Cannot we just remove the JSON file generated from within the two `_conf*` methods and have a single write to the temp file here? This would requires also some additional cleanup on locals definition and meaning...

I've come up with something like this:
   https://github.com/derkling/lisa/commit/19a49b2a5a87348e56079eb92ec1c108b9dc06cd

Not yet tested however.

libs/wlgen/wlgen/rta.py
```diff
@@ -254,25 +254,27 @@ def _confCustom(self):
...
         for line in ifile:
             if '__DURATION__' in line and self.duration is None:
                 raise ValueError('Workload duration not specified')
             for src, target in replacements.iteritems():
                 line = line.replace(src, target)
-            ofile.write(line)
+            temp_json.append(line)
+
+        temp_json = '\n'.join(temp_json)
+        self.rta_profile = json.loads(temp_json)
```
derkling: Don't like the usage of this attribute... which was introduced explicitly for the PROFILE case...
We should have instead a generic configuration container.... as in the example below...

valschneider:
@douglas-raillard-arm Can we close this?

douglas-raillard-arm:
Fixed by design in LISA/next, the rtapp description is now created in a dedicated subfolder of the TestEnv, which is not shared with any other code.

PR #594, opened 2018-04-18
Rework results directory

### LINKED ISSUE #316: lisa-test broken when running multiple test suites with a single command
E.g. `lisa-test tests/eas/heavy_load.py tests/eas/load_tracking.py` results in the `heavy_load` test failing due to "cpu_idle" events being missing.  Probably `TestEnv`'s test_conf is being overwritten at the wrong time.

douglas-raillard-arm:
The current behavior relies on TestEnv being initialized once per lisa-test command. This will create a results_dir directory that will be used by all tests. The default argument wipe=True does not wipe this folder since it is ignored after the first TestEnv is created. Unfortunately, this has a number of downsides: TestEnv are sharing a results_dir that is supposed to be private. It makes it harder to map files to the test they belong to. The shared state in TestEnv also means the trace events are initialized once per lisa-test command, so the first test in alphabetical order will get its required events but it will be a gamble for the following tests.

The fix consists in limiting the shared state to the target attribute and closely related attributes. The rest of TestEnv is now initialized properly for every test, including a new results directory base on the name of the test. The base folder containing all the tests' results directory can be specified with -o lisa-test option.

close #316

### DIFF
--- a/ipynb/energy/EnergyModel_ClusterEnergy.ipynb
+++ b/ipynb/energy/EnergyModel_ClusterEnergy.ipynb
@@ -178,7 +178,7 @@
    "source": [
     "# Initialize a test environment using:\n",
     "# the provided target configuration (my_conf)\n",
-    "te = TestEnv(target_conf=my_conf, force_new=True)\n",
+    "te = TestEnv(target_conf=my_conf)\n",
     "target = te.target"
    ]
   },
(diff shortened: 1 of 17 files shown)


valschneider:
@douglas-raillard-arm I think we can close this now, right?

douglas-raillard-arm:
Fixed by design in LISA/next, which we will soon be using for all tasks.

PR #598, opened 2018-05-16
Fix jankbench wa support

vingu-linaro:
Hi,

These 2 patches fix the support of the computation of statistic for total_frame_duration

Regards,


vingu-linaro:
Hi,

Any comments on the patches ?

PR #600, opened 2018-06-06
utils/wa_results_collector: Change sha1 acquisition method

valschneider:
To map a WA run with a kernel commit sha1, we currently read whatever
__meta/target_info.json reports for "kernel_release" (which should be
the equivalent of uname -r).

However, there are cases where this doesn't seem to be reported. For
instance, a HiKey960 built with hikey960_defconfig will not report the
commit sha1 in its version data, which will cause an exception in the
results collector code.

Since the sha1 is included in the results directory name anyway, use
that instead as a more reliable source.

### DIFF
--- a/libs/utils/wa_results_collector.py
+++ b/libs/utils/wa_results_collector.py
@@ -112,6 +112,8 @@ class WaResultsCollector(object):
                      cached in the provided output directories. Set this param
                      to False to disable this caching.
     """
+    RE_WLTEST_DIR = re.compile(r"wa\.(?P<sha1>\w+)_(?P<name>.+)")
+
     def __init__(self, base_dir=None, wa_dirs=".*", platform=None,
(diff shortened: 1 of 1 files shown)


derkling:
LGTM, however I was wondering if it could make sense to check before the WA output... and fallback to the dir parsing only in case the first one should fails.

valschneider:
@derkling Makes sense to me. I was hoping to find a "centralized" source of information for that sha1, but as you point out the data could be generated without using **test_series**. I'll fold the old method back in.

vingu-linaro:
Hi, I confirm that we are not encoding the sha1 in the WA output folder name when using wa_results_collector in our ci loop

valschneider:
@vingu-linaro Thanks for the confirmation, I'll only add reading the dir name as a fallback.

PR #604, opened 2018-06-13
Make cpufreq module optional for rt-app calibration

jlelli:
If cpufreq module is not used (e.g., KVM target) rt-app calibration
should still work as usual, no need to set frequency to max in this case
though.

Remove the hard dependecy towards cpufreq module.

Signed-off-by: Juri Lelli <juri.lelli@redhat.com>

### DIFF
--- a/libs/wlgen/wlgen/rta.py
+++ b/libs/wlgen/wlgen/rta.py
@@ -102,6 +102,39 @@ class RTA(Workload):
         self.setCallback('postrun', self.__postrun)
 
     @staticmethod
+    def __setGovernorsPerformance(target):
+        if 'cpufreq' not in target.modules:
+            log = logging.getLogger('RTApp')
+            log.warning('cpufreq module not loaded,'
+                        ' skipping setting frequency to max')
+            return {}
+
+        # Save previous governors
+        old_governors = {}
+        for domain in target.cpufreq.iter_domains():
+            cpu = domain[0]
+            governor = target.cpufreq.get_governor(cpu)
+            tunables = target.cpufreq.get_governor_tunables(cpu)
+            old_governors[cpu] = governor, tunables
+
+        target.cpufreq.set_all_governors('performance')
+
+        return old_governors
+
+    @staticmethod
+    def __restoreGovernors(target, old_governors):
+        if 'cpufreq' not in target.modules:
+            return
+
+        # Restore previous governors
+        #   Setting a governor & tunables for a cpu will set them for all cpus
+        #   in the same clock domain, so only restoring them for one cpu
+        #   per domain is enough to restore them all.
+        for cpu, (governor, tunables) in old_governors.iteritems():
+            target.cpufreq.set_governor(cpu, governor)
+            target.cpufreq.set_governor_tunables(cpu, **tunables)
+
+    @staticmethod
     def calibrate(target):
         """
         Calibrate RT-App on each CPU in the system
@@ -115,15 +148,7 @@ class RTA(Workload):
         # Setup logging
         log = logging.getLogger('RTApp')
 
-        # Save previous governors
-        old_governors = {}
-        for domain in target.cpufreq.iter_domains():
-            cpu = domain[0]
-            governor = target.cpufreq.get_governor(cpu)
-            tunables = target.cpufreq.get_governor_tunables(cpu)
-            old_governors[cpu] = governor, tunables
-
-        target.cpufreq.set_all_governors('performance')
+        old_governors = RTA.__setGovernorsPerformance(target)
 
         # Create calibration task
         max_rtprio = int(target.execute('ulimit -Hr').split('\r')[0])
@@ -154,13 +179,7 @@ class RTA(Workload):
                 pload[cpu] = int(pload_match.group(1))
                 log.debug('>>> cpu%d: %d', cpu, pload[cpu])
 
-        # Restore previous governors
-        #   Setting a governor & tunables for a cpu will set them for all cpus
-        #   in the same clock domain, so only restoring them for one cpu
-        #   per domain is enough to restore them all.
-        for cpu, (governor, tunables) in old_governors.iteritems():
-            target.cpufreq.set_governor(cpu, governor)
-            target.cpufreq.set_governor_tunables(cpu, **tunables)
+        RTA.__restoreGovernors(target, old_governors)
 
         log.info('Target RT-App calibration:')
         log.info("{" + ", ".join('"%r": %r' % (key, pload[key])


jlelli:
@derkling suggested changes on IRC

libs/wlgen/wlgen/rta.py
```diff
@@ -101,6 +101,39 @@ def __init__(self,
...
         self.setCallback('postrun', self.__postrun)
 
+    @staticmethod
+    def __setGovernorsPerformance(target):
+        if 'cpufreq' not in target.modules:
+            log = logging.getLogger('RTApp')
+            log.warning('cpufreq module not loaded,'
+                        ' skipping setting frequency to max')
+            return {}
+
```
derkling: Since we are at that, there can be another condition to test: the performance governor should be available.
```python
if 'performance' not in target.cpufreq.list_governors(0):
   log.warning('performance cpufreq governor not available, '
               'skipping setting frequency to max')
   return {}

```

libs/wlgen/wlgen/rta.py
```diff
@@ -101,6 +101,39 @@ def __init__(self,
         # Setup RTA callbacks
         self.setCallback('postrun', self.__postrun)
 
+    @staticmethod
+    def __setGovernorsPerformance(target):
+        if 'cpufreq' not in target.modules:
+            log = logging.getLogger('RTApp')
+            log.warning('cpufreq module not loaded,'
+                        ' skipping setting frequency to max')
```
derkling: can we keep this space at the end of the previous sentence?

libs/wlgen/wlgen/rta.py
```diff
@@ -101,6 +101,39 @@ def __init__(self,
...
+        # Save previous governors
+        old_governors = {}
+        for domain in target.cpufreq.iter_domains():
+            cpu = domain[0]
+            governor = target.cpufreq.get_governor(cpu)
+            tunables = target.cpufreq.get_governor_tunables(cpu)
+            old_governors[cpu] = governor, tunables
+
+        target.cpufreq.set_all_governors('performance')
+
```
derkling: Let's add a log here to:
```python
log.warning('Using governors: %s', target.cpufreq.get_all_governors())
```

derkling: This means that we can move:

```python
log = logging.getLogger('RTApp')
```

right at the beginning of the function

libs/wlgen/wlgen/rta.py
```diff
@@ -101,6 +101,39 @@ def __init__(self,
...
+            return
+
+        # Restore previous governors
+        #   Setting a governor & tunables for a cpu will set them for all cpus
+        #   in the same clock domain, so only restoring them for one cpu
+        #   per domain is enough to restore them all.
+        for cpu, (governor, tunables) in old_governors.iteritems():
+            target.cpufreq.set_governor(cpu, governor)
+            target.cpufreq.set_governor_tunables(cpu, **tunables)
+
```
derkling: And here too:

```python
log.warning('Governors restored to: %s', target.cpufreq.get_all_governors())
```

valschneider:
So I was thinking this could benefit from being packaged in a context manager, like was done for freezing the userspace: https://github.com/ARM-software/lisa/commit/cdf25f2961e22d5cc8b97c68c5a121a54b09adc8

I don't know what would be the best place for it to live - maybe inside devlib's cpufreq module ?
I'd say it would look something like this:

``` python
@contextlib.contextmanager
def use_governor(governor, strict=False):
    old_governors = {}
    
    for domain in target.cpufreq.iter_domains():
        cpu = domain[0]
        governor = target.cpufreq.get_governor(cpu)
        tunables = target.cpufreq.get_governor_tunables(cpu)
        old_governors[cpu] = governor, tunables
        
    target.cpufreq.set_all_governors(governor)

    yield

    # Restore previous governors
    #   Setting a governor & tunables for a cpu will set them for all cpus
    #   in the same clock domain, so only restoring them for one cpu
    #   per domain is enough to restore them all.
    for cpu, (governor, tunables) in old_governors.iteritems():
        target.cpufreq.set_governor(cpu, governor)
        target.cpufreq.set_governor_tunables(cpu, **tunables)


with use_governor("performance"):
    # Do whatever you want !
```

And then you can fold in checks for "is the governor I'm asking supported ?".

derkling:
I think the main issue for having this into devlib is that, if we don't have the cpufreq module (which is the case Juri is trying to fix) we cannot access the method you propose... unless... @valschneider: can we have a context manager defined for a static method?

If the above is not possible, it's fine for me to have this API living in rta.py.

valschneider:
Hmm yeah so that's actually a downside. One way to still use it would be:
``` python
if "cpufreq" in target.modules:
    with use_governor("performance"):
        run_calibration()
else:
    run_calibration()
```

Which isn't great... But we're already facing that problem with userspace freezing: https://github.com/ARM-software/lisa/blob/master/libs/utils/executor.py#L266
Another way would be to define it as a static method in devlib.cpufreq as you propose, and then within the method you can skip everything if the passed target doesn't have the cpufreq module loaded.

libs/wlgen/wlgen/rta.py
```diff
@@ -101,6 +101,39 @@ def __init__(self,
         # Setup RTA callbacks
         self.setCallback('postrun', self.__postrun)
 
+    @staticmethod
+    def __setGovernorsPerformance(target):
```
douglas-raillard-arm: Double leading underscore has a special meaning in Python and must not be used unless there is a good reason. This definition will define: RTA._RTA__setGovernorsPerformance, which is unlikely to be what you expect, and will prevent subclasses from overriding it in a sane way. Internal methods can use a single underscore to mark them as not exposed as a user API.

Details on name mangling here: https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

libs/wlgen/wlgen/rta.py
douglas-raillard-arm: I'd prefer no line break for user-exposed message for the same reason as in the kernel coding style: if you want to know where a message comes from, you grep for it, and splitting the message will defeat simple grep commands.

douglas-raillard-arm:
I don't have the full background of the discussion but the context manager needs to take "target" as a parameter anyway, so there is no reason for this check not to be included inside the context manager:
```
if "cpufreq" in target.modules:
```
However, having a such an API that silently bails out is a recipe for lot of troubles down the line, so this silent behavior should be enabled by a parameter that defaults to raising an exception, even if we end up specifying that parameter all the time.

valschneider:
Hi,

I implemented that governor context manager in here: https://github.com/ARM-software/devlib/pull/316
I'm also cleaning up lots of stuff around LISA, so when it comes to rtapp I'll make it work without the cpufreq module.

valschneider:
It's all implemented in here for now: https://github.com/valschneider/lisa/blob/lisa-next/lisa/wlgen/rta.py#L394
Closing this for now, although you should wait for a proper release ;)

PR #605, opened 2018-06-15
RFC: Load from simplified em

credp:
Since we have the new EM in the android-mainline-tracking branch, I figured we should be able to read it from the target. Here's a go at adding it to LISA.

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -24,7 +24,7 @@ import re
 import pandas as pd
 import numpy as np
 
-from devlib.utils.misc import memoized, mask_to_list
+from devlib.utils.misc import memoized, mask_to_list, ranges_to_list
 from devlib import TargetError
 from trappy.stats.grammar import Parser
 
@@ -774,7 +774,105 @@ class EnergyModel(object):
         return sorted(ret, key=lambda x: x[0])
 
     @classmethod
-    def from_target(cls, target):
+    def from_simplifiedEM_target(cls, target,
+            directory='/sys/devices/system/cpu/energy_model'):
+        """
+        Create an EnergyModel by reading a target filesystem on a device with
+        the new Simplified Energy Model present.
+
+        This uses the energy_model sysctl added by EAS patches to exposes
+        the frequency domains, together with a tuple of capacity, frequency
+        and active power for each CPU. This feature is not upstream in mainline
+        Linux (as of v4.17), and only exists in Android kernels later than
+        android-4.14.
+
+        Wrt. idle states - the EnergyModel constructed won't be aware of
+        any power data or topological dependencies for entering "cluster"
+        idle states since the simplified model has no such concept.
+
+        Initialises only Active States for CPUs and clears all other levels.
+
+        :param target: Devlib target object to read filesystem from. Must have
+                       cpufreq and cpuidle modules enabled.
+        :returns: Constructed EnergyModel object based on the parameters
+                  reported by the target.
+        """
+        if 'cpuidle' not in target.modules:
+            raise TargetError('Requires cpuidle devlib module. Please ensure '
+                               '"cpuidle" is listed in your target/test modules')
+
+        # Simplified EM on-disk format (for each frequency domain):
+        #    /sys/devices/system/cpu/energy_model/<frequency_domain>/..
+        #        ../capacity
+        #           contains a space-separated list of capacities in increasing order
+        #        ../cpus
+        #           cpulist-formatted representation of the cpus in the frequency domain
+        #        ../frequency
+        #           space-separated list of frequencies in corresponding order to capacities
+        #        ../power
+        #           space-separated list of power consumption in corresponding order to capacities
+        # taken together, the contents of capacity, frequency and power give you the required
+        # tuple for ActiveStates.
+        # hence, domain should be supplied as a glob, and fields should be
+        #   capacity, cpus, frequency, power
+
+        sysfs_em = target.read_tree_values(directory, depth=3)
+
+        if not sysfs_em:
+            raise TargetError('Simplified Energy Model not exposed '
+                              'at {} in sysfs.'.format(directory))
+
+        cpu_to_fdom = {}
+        for fd, fields in sysfs_em.iteritems():
+            cpus = ranges_to_list(fields["cpus"])
+            for cpu in cpus:
+                cpu_to_fdom[cpu] = fd
+            sysfs_em[fd]['cpus'] = cpus
+            sysfs_em[fd]['frequency'] = map(int, sysfs_em[fd]['frequency'].split(' '))
+            sysfs_em[fd]['power'] =  map(int, sysfs_em[fd]['power'].split(' '))
+            sysfs_em[fd]['capacity'] = map(int, sysfs_em[fd]['capacity'].split(' '))
+
+        def read_active_states(cpu):
+            fd = sysfs_em[cpu_to_fdom[cpu]]
+            cstates = zip(fd['capacity'], fd['power'])
+            active_states = [ActiveState(c, p) for c, p in cstates]
+            return OrderedDict(zip(fd['frequency'], active_states))
+
+        def read_idle_states(cpu):
+            # idle states are not supported in the new model
+            # record 0 power for them all, but name them according to target
+            names = [s.name for s in target.cpuidle.get_states(cpu)]
+            return OrderedDict((name, 0) for name in names)
+
+        # Read the CPU-level data
+        cpus = range(target.number_of_cpus)
+        cpu_nodes = []
+        for cpu in cpus:
+            node = EnergyModelNode(
+                cpu=cpu,
+                active_states=read_active_states(cpu),
+                idle_states=read_idle_states(cpu))
+            cpu_nodes.append(node)
+
+        root = EnergyModelRoot(children=cpu_nodes)
+        freq_domains = [sysfs_em[fdom]['cpus'] for fdom in sysfs_em]
+
+        # We don't have a way to read the idle power domains from sysfs (the kernel
+        # isn't even aware of them) so we'll just have to assume each CPU is its
+        # own power domain and all idle states are independent of each other.
+        cpu_pds = []
+        for cpu in cpus:
+            names = [s.name for s in target.cpuidle.get_states(cpu)]
+            cpu_pds.append(PowerDomain(cpu=cpu, idle_states=names))
+
+        root_pd = PowerDomain(children=cpu_pds, idle_states=[])
+        return cls(root_node=root,
+                   root_power_domain=root_pd,
+                   freq_domains=freq_domains)
+
+    @classmethod
+    def from_sd_target(cls, target, filename=
+            '/proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/{}'):
         """
         Create an EnergyModel by reading a target filesystem
 
@@ -804,8 +902,7 @@ class EnergyModel(object):
                                '"cpuidle" is listed in your target/test modules')
 
         def sge_path(cpu, domain, group, field):
-            f = '/proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/{}'
-            return f.format(cpu, domain, group, field)
+            return filename.format(cpu, domain, group, field)
 
         # Read all the files we might need in one go, otherwise this will take
         # ages.
@@ -896,6 +993,49 @@ class EnergyModel(object):
         return cls(root_node=root,
                    root_power_domain=root_pd,
                    freq_domains=freq_domains)
+
+    @classmethod
+    def from_target(cls, target):
+        """
+        Create an EnergyModel by reading a target filesystem
+
+        If present, load an EM provided via dt using from_sd_target, since these
+        devices make the full EM available via the sched domain in sysfs. If
+        there is no EM at this location, attempt to load the simplified EM
+        made available via dedicated sysfs files.
+
+        :param target: Devlib target object to read filesystem from. Must have
+                       cpufreq and cpuidle modules enabled.
+        :returns: Constructed EnergyModel object based on the parameters
+                  reported by the target.
+        """
+        _log = logging.getLogger('EMReader')
+
+        def em_present_in_sd(target):
+            try:
+                cpu = target.list_online_cpus()[0]
+                d = '/proc/sys/kernel/sched_domain/cpu{}/domain0/group0/energy'.format(cpu)
+                _log.info('Checking if directory {} exists'.format(d))
+                return target.directory_exists(d)
+            except Exception:
+                return False
+
+        def simplified_em_present_in_cpusysfs(target):
+            try:
+                d = '/sys/devices/system/cpu/energy_model'
+                _log.info('Checking if directory {} exists'.format(d))
+                return target.directory_exists(d)
+            except Exception:
+                return False
+
+        if em_present_in_sd(target):
+            _log.info('Attempting to load EM from sched domains.')
+            return cls.from_sd_target(target)
+        elif simplified_em_present_in_cpusysfs(target):
+            _log.info('Attempting to load simplified EM.')
+            return cls.from_simplifiedEM_target(target)
+        else:
+            raise TargetError('Unable to probe for energy model on target.')
 
     def estimate_from_trace(self, trace):
         """
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -1011,31 +1011,51 @@ class EnergyModel(object):
         """
         _log = logging.getLogger('EMReader')
 
-        def em_present_in_sd(target):
-            try:
-                cpu = target.list_online_cpus()[0]
-                d = '/proc/sys/kernel/sched_domain/cpu{}/domain0/group0/energy'.format(cpu)
-                _log.info('Checking if directory {} exists'.format(d))
-                return target.directory_exists(d)
-            except Exception:
-                return False
+        # To add a new EM reader type, the following is required:
+        # 1. Create an inline function to test for EM presence which takes a
+        #    target as the first parameter. Any exceptions raised here will
+        #    be caught in the loader loop.
+        # 2. Create a function which returns an EnergyModel as a member of this
+        #    class, also with a target as the first parameter.
+        # 3. Add an entry to the em_loaders dict where 'check' contains the
+        #    inline function and 'load' contains the class member function
+        # 4. If you need any additional data, add it to the em_loaders dict - any
+        #    additional keys will be passed to both 'check' and 'load' functions
+        #    as named parameters.
 
-        def simplified_em_present_in_cpusysfs(target):
-            try:
-                d = '/sys/devices/system/cpu/energy_model'
-                _log.info('Checking if directory {} exists'.format(d))
-                return target.directory_exists(d)
-            except Exception:
-                return False
+        # Utility functions to determine if we should try to use a particular
+        # EM loader function.
+        def em_present_in_sd(target, filename=None):
+            cpu = target.list_online_cpus()[0]
+            f = filename.format(cpu, 0, 0, 'cap_states')
+            return target.file_exists(f)
+        def simplified_em_present_in_cpusysfs(target, directory=None):
+            return target.directory_exists(directory)
 
-        if em_present_in_sd(target):
-            _log.info('Attempting to load EM from sched domains.')
-            return cls.from_sd_target(target)
-        elif simplified_em_present_in_cpusysfs(target):
-            _log.info('Attempting to load simplified EM.')
-            return cls.from_simplifiedEM_target(target)
-        else:
-            raise TargetError('Unable to probe for energy model on target.')
+        # em_loaders dictionary joins EM loaders and the identifying functions
+        # with any associated metadata
+        em_loaders = {
+            'sd'    : { 'check': em_present_in_sd,
+                        'load': cls.from_sd_target,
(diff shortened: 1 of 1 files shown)


libs/utils/energy_model.py
```diff
@@ -814,8 +1024,7 @@ def sge_path(cpu, domain, group, field):
         sge_file_values = read_multiple_oneline_files(target, sge_globs)
 
         if not sge_file_values:
-            raise TargetError('Energy Model not exposed in sysfs. '
-                              'Check CONFIG_SCHED_DEBUG is enabled.')
+            return cls.from_simplifiedEM_target(target)
```
valschneider: Hmmm what about splitting this more cleanly ? I'm thinking about something like this:

``` python
os = target.os
version = target.kernel_version.parts

if os == 'android' and parts >= (4, 14):
    return from_target_v2(...)
else:
    return from_target_v1(...)
```

libs/utils/energy_model.py
```diff
@@ -773,6 +773,216 @@ def _find_core_groups(cls, target):
...
+                    if domain not in domains:
+                        domains.append(domain)
+            available_sge_freq_domains=domains
+        except:
+            pass
+        #_log.info("Available domains: {}".format(available_sge_freq_domains))
+
+        # convert a comma-separated list with ranges to a plain list
+        # e.g. '0,3-5' -> [ '0', '3', '4', '5' ]
+        def cpu_csl_to_list(cpu_csl):
```
valschneider: AFAICT that's already implemented by **devlib.utils.misc.ranges_to_list** (there's a lot of useful "str thing to cpu list" functions in there, when in doubt it's good to have a look)

libs/utils/energy_model.py
```diff
@@ -773,6 +773,216 @@ def _find_core_groups(cls, target):
...
+                                           cpu_or_list, cpu_csl))
+                else:
+                    list_cpus = cpu_or_list.split('-')
+                    if len(list_cpus) == 2:
+                        result.extend([str(c) for c in range(int(list_cpus[0]), int(list_cpus[1])+1)])
+            #_log.info('list {} converted to {}'.format(cpu_csl, result))
+            return result
+
+        # work out which frequency domain contains a particular CPU
+        def freq_domain_for_cpu(cpu):
```
valschneider: What about building a dict for that ? I don't know how to use that sge thing, so I'll use **read_tree_values** (it creates a dict) as an exemple:

``` python
sysfs_em = target.read_tree_values("/sys/devices/system/cpu/energy_model/", depth=3)

cpu_to_fdom = {}

for freq_domain, fields in sysfs_em.iteritems():
    cpus = ranges_to_list(fields["cpus"])
    for cpu in cpus:
        cpu_to_fdom[cpu] = freq_domain
    
```

qperret: +1, using `read_tree_values()` sounds like the right method to parse the EM. We could actually use it to do more that just this CPU to fd mapping. We can probably convert the strings of "sysfs_em" into values and get a nice representation of the EM which can then be used directly to build the ActiveState objects. Based on your code, something as simple as:
```python
from devlib.utils.misc import *
from devlib.utils.types import *

sysfs_em = target.read_tree_values("/sys/devices/system/cpu/energy_model/", depth=3)
cpu_to_fdom = {}

for fd, fields in sysfs_em.iteritems():
    cpus = ranges_to_list(fields["cpus"])
    for cpu in cpus:
        cpu_to_fdom[cpu] = fd
    sysfs_em[fd]['cpus'] = cpus
    sysfs_em[fd]['frequency'] = map(integer, sysfs_em[fd]['frequency'].split(' '))
    sysfs_em[fd]['power'] =  map(integer, sysfs_em[fd]['power'].split(' '))
    sysfs_em[fd]['capacity'] = map(integer, sysfs_em[fd]['capacity'].split(' '))
``` 
gives the following `sysfs_em` struct (which has all we need, I think):

```
{'fd0': {'capacity': [133, 250, 351, 428, 462],
  'cpus': [0, 1, 2, 3],
  'frequency': [533000, 999000, 1402000, 1709000, 1844000],
  'power': [28, 70, 124, 187, 245]},
 'fd4': {'capacity': [391, 616, 782, 915, 1024],
  'cpus': [4, 5, 6, 7],
  'frequency': [903000, 1421000, 1805000, 2112000, 2362000],
  'power': [243, 500, 804, 1161, 1571]}}
```
And then,with this structure, `read_active_states()` can be simplified to something as simple as:
```python
def read_active_states(cpu, level):
    fd = sysfs_em[cpu_to_fdom[cpu]]
    cstates = zip(fd['capacity'], fd['power'])
    active_states = [ActiveState(c, p if not level else 0) for c, p in cstates]
    return OrderedDict(zip(sorted(fd['frequency']), active_states))
```

libs/utils/energy_model.py
```diff
@@ -773,6 +773,216 @@ def _find_core_groups(cls, target):
...
+                      freq_domain = cpu
+                      break
+                  cpus_path = sge_path(fdom, 'cpus')
+                  cpus_list = cpu_csl_to_list(read_sge_file(cpus_path))
+                  #_log.info('looking for cpu {} in list {} for fdom {}'.format(cpu, cpus_list, fdom))
+                  if cpu in cpus_list:
+                      #_log.info('matched cpu {} to fdom {}'.format(cpu, fdom))
+                      freq_domain = fdom
+                      break
+            if not freq_domain:
```
douglas-raillard-arm: nit: Such pattern can be done directly with the for loop:
```
for x in y:
  if condition:
    break
# else clause executes when the loop terminates without a break. You can think of it as matching the "if condition: break" inside the loop
else:
 XXX
```

libs/utils/energy_model.py
credp: If we do this, we can't test the integration branch with it - there isn't a link between version or OS which we can rely on. Maybe we should refactor to provide 'is_this_model_present' methods..

libs/utils/energy_model.py
credp: I spent at least half an hour searching for various combinations of list and cpu but missed it. I'll switch over.

libs/utils/energy_model.py
valschneider: Right, then maybe we can try to determine which EM is there by looking at the sysfs entries instead.

douglas-raillard-arm: What we want to avoid is silent errors when there is actually nothing valid to read there. As long as it can fail properly, it should be fine.
Another solution could be to add a target config entry but it's easier if we can just autodetect it from the target.

libs/utils/energy_model.py
```diff
@@ -897,6 +1000,39 @@ def read_idle_states(cpu, domain_level):
...
+            # ages.
+            sge_globs = ['/proc/sys/kernel/sched_domain/cpu**/domain**/group**/energy/cap_states',]
+            sge_file_values = read_multiple_oneline_files(target, sge_globs)
+
+            if not sge_file_values:
+                return False
+            return True
+
+        if em_present_in_sd(target):
+            return cls.from_sd_target(target)
```
valschneider: Could you add some logging in here (or at the head of either EM loading methods) specifying which EM is being loaded ? So far we only have this log in **env.py**:
``` python
self._log.info('Attempting to read energy model from target')
```

Being able to tell from the log which one was actually loaded would be great :)

credp: If you use the simplified loader, you get a message but I agree I should probably let you know which it thinks you have.

credp:
@valschneider I tried to address your logging comment and pushed another new version :)

libs/utils/energy_model.py
```diff
@@ -774,7 +774,105 @@ def _find_core_groups(cls, target):
...
+        # taken together, the contents of capacity, frequency and power give you the required
+        # tuple for ActiveStates.
+        # hence, domain should be supplied as a glob, and fields should be
+        #   capacity, cpus, frequency, power
+
+        try:
+            sysfs_em = target.read_tree_values("/sys/devices/system/cpu/energy_model/", depth=3)
+        except:
+            raise TargetError('Energy Model not exposed in sysfs. '
+                              'Check CONFIG_SCHED_DEBUG is enabled.')
```
qperret: `CONFIG_SCHED_DEBUG` isn't required for the simplified EM and will never be (as long as the EM framework is decoupled from the scheduler at least ...)

libs/utils/energy_model.py
```diff
@@ -897,6 +995,43 @@ def read_idle_states(cpu, domain_level):
...
+        _log = logging.getLogger('EMReader')
+
+        def em_present_in_sd(target):
+            def sge_path(cpu, domain, group, field):
+                f = '/proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/{}'
+                return f.format(cpu, domain, group, field)
+            # Read all the files we might need in one go, otherwise this will take
+            # ages.
+            sge_globs = ['/proc/sys/kernel/sched_domain/cpu**/domain**/group**/energy/cap_states',]
+            sge_file_values = read_multiple_oneline_files(target, sge_globs)
```
qperret: Is it possible to replace that with a simple `target.file_exists()`

libs/utils/energy_model.py
```diff
@@ -897,6 +995,43 @@ def read_idle_states(cpu, domain_level):
...
+
+            if not sge_file_values:
+                return False
+            return True
+
+        if em_present_in_sd(target):
+            _log.info('Attempting to load EM from sched domains.')
+            return cls.from_sd_target(target)
+        else:
```
qperret: `elif target.file_exists('/sys/devices/system/cpu/energy_model/fd0'):` ?

libs/utils/energy_model.py
```diff
@@ -897,6 +995,43 @@ def read_idle_states(cpu, domain_level):
...
+            _log.info('Attempting to load simplified EM.')
+            return cls.from_simplifiedEM_target(target)
+
```
qperret: else complain something ?

libs/utils/energy_model.py
```diff
@@ -774,7 +774,105 @@ def _find_core_groups(cls, target):
...
+        idle states since the simplified model has no such concept.
+
+        Initialises only Active States for CPUs and clears all other levels.
+
+        :param target: Devlib target object to read filesystem from. Must have
+                       cpufreq and cpuidle modules enabled.
+        :returns: Constructed EnergyModel object based on the parameters
+                  reported by the target.
+        """
+        if 'cpufreq' not in target.modules:
```
qperret: Why do we need the `cpufreq` module ? The frequency exposed directly in the EM should be enough no ?

libs/utils/energy_model.py
```diff
@@ -774,7 +774,105 @@ def _find_core_groups(cls, target):
...
+        for fd, fields in sysfs_em.iteritems():
+            cpus = ranges_to_list(fields["cpus"])
+            for cpu in cpus:
+                cpu_to_fdom[cpu] = fd
+            sysfs_em[fd]['cpus'] = cpus
+            sysfs_em[fd]['frequency'] = map(int, sysfs_em[fd]['frequency'].split(' '))
+            sysfs_em[fd]['power'] =  map(int, sysfs_em[fd]['power'].split(' '))
+            sysfs_em[fd]['capacity'] = map(int, sysfs_em[fd]['capacity'].split(' '))
+
+        def read_active_states(cpu, level):
```
qperret: Is `level` required in the end ?

libs/utils/energy_model.py
```diff
@@ -774,7 +774,105 @@ def _find_core_groups(cls, target):
...
+        except:
+            raise TargetError('Energy Model not exposed in sysfs. '
+                              'Check CONFIG_SCHED_DEBUG is enabled.')
+        cpu_to_fdom = {}
+        for fd, fields in sysfs_em.iteritems():
+            cpus = ranges_to_list(fields["cpus"])
+            for cpu in cpus:
+                cpu_to_fdom[cpu] = fd
+            sysfs_em[fd]['cpus'] = cpus
+            sysfs_em[fd]['frequency'] = map(int, sysfs_em[fd]['frequency'].split(' '))
```
qperret: Nit: devlib has an `integer()` utility function in `devlib.utils.types`, but not sure if needed here

libs/utils/energy_model.py
```diff
@@ -774,7 +774,105 @@ def _find_core_groups(cls, target):
...
+            sysfs_em[fd]['power'] =  map(int, sysfs_em[fd]['power'].split(' '))
+            sysfs_em[fd]['capacity'] = map(int, sysfs_em[fd]['capacity'].split(' '))
+
+        def read_active_states(cpu, level):
+            fd = sysfs_em[cpu_to_fdom[cpu]]
+            cstates = zip(fd['capacity'], fd['power'])
+            active_states = [ActiveState(c, p if not level else 0) for c, p in cstates]
+            return OrderedDict(zip(sorted(fd['frequency']), active_states))
+
+        def read_idle_states(cpu, domain_level):
```
qperret: Same question, is `domain_level` needed ?

libs/utils/energy_model.py
credp: Same as above - I think we only know that the fd index will be the cpuid of the first cpu in a group. If the energy_model directory is present, maybe that's enough to indicate the feature should be there and we can fail if the content is not present or not structured as expected.

libs/utils/energy_model.py
credp: Yes, actually we don't need cpufreq for the simplified EM loading, only for the traditional model.

libs/utils/energy_model.py
credp: Not any more - I was keeping it in in case we needed a zero-energy cluster model for the task placement calculations, but it works fine without. Will remove.

libs/utils/energy_model.py
credp: that's probably why it failed when I had integer there, is there a reason we should use devlib types instead of python types?

libs/utils/energy_model.py
credp: Done

libs/utils/energy_model.py
credp: Done

libs/utils/energy_model.py
credp: Done

libs/utils/energy_model.py
credp: OK, I'm trying a directory check approach in the new version. Done.

libs/utils/energy_model.py
credp: Nope!

credp:
I tested this loading new and old style EM on my Juno R2 and it works fine.

libs/utils/energy_model.py
valschneider: Looks like this type is only really there for conversion from binary/octal/hex/whatevs (it will read the string and appropriatly handle "0x", "0b" & co prefixes). I think here we do really expect base 10 frequencies, right ?

credp: Yes, it's a plain base 10 number. If it isn't I think we want an error.

qperret: Yeah, that should be fine

derkling:
I guess we miss a +1 from @douglas-raillard-arm and then this can go in ;)

libs/utils/energy_model.py
```diff
@@ -774,7 +774,102 @@ def _find_core_groups(cls, target):
...
+            sysfs_em[fd]['cpus'] = cpus
+            sysfs_em[fd]['frequency'] = map(int, sysfs_em[fd]['frequency'].split(' '))
+            sysfs_em[fd]['power'] =  map(int, sysfs_em[fd]['power'].split(' '))
+            sysfs_em[fd]['capacity'] = map(int, sysfs_em[fd]['capacity'].split(' '))
+
+        def read_active_states(cpu):
+            fd = sysfs_em[cpu_to_fdom[cpu]]
+            cstates = zip(fd['capacity'], fd['power'])
+            active_states = [ActiveState(c, p) for c, p in cstates]
+            return OrderedDict(zip(sorted(fd['frequency']), active_states))
```
douglas-raillard-arm: Shouldn't it be zip and then sorted ?
```
OrderedDict(sorted(zip(fd['frequency'], active_states)))
```
Unless it is assumed that fd['capacity'] is already sorted and fd['frequency'] is in random order.

libs/utils/energy_model.py
```diff
@@ -774,7 +774,102 @@ def _find_core_groups(cls, target):
...
+
+        # We don't have a way to read the idle power domains from sysfs (the kernel
+        # isn't even aware of them) so we'll just have to assume each CPU is its
+        # own power domain and all idle states are independent of each other.
+        cpu_pds = []
+        for cpu in cpus:
+            names = [s.name for s in target.cpuidle.get_states(cpu)]
+            cpu_pds.append(PowerDomain(cpu=cpu, idle_states=names))
+
+        root_pd=PowerDomain(children=cpu_pds, idle_states=[])
```
douglas-raillard-arm: nit: spaces around "=" in assignments (not when passing parameters)

libs/utils/energy_model.py
```diff
@@ -897,6 +992,49 @@ def read_idle_states(cpu, domain_level):
...
+        """
+        _log = logging.getLogger('EMReader')
+
+        def em_present_in_sd(target):
+            try:
+                cpu = target.list_online_cpus()[0]
+                d = '/proc/sys/kernel/sched_domain/cpu{}/domain0/group0/energy'.format(cpu)
+                _log.info('Checking if directory {} exists'.format(d))
+                return target.directory_exists(d)
+            except:
```
douglas-raillard-arm: except should match a specific exception and be restricted to the least amount of code like:
```
+            try:
+                cpu = target.list_online_cpus()[0]
+            except IndexError:  
+                return False
+            else:
+                d = '/proc/sys/kernel/sched_domain/cpu{}/domain0/group0/energy'.format(cpu)
+                _log.info('Checking if directory {} exists'.format(d))
+                return target.directory_exists(d)
```

Or even:

```
+            online_cpus = target.list_online_cpus()
+            if online_cpus:
+                cpu = online_cpus[0]
+                d = '/proc/sys/kernel/sched_domain/cpu{}/domain0/group0/energy'.format(cpu)
+                _log.info('Checking if directory {} exists'.format(d))
+                return target.directory_exists(d)
+            else:  
+                return False
```
```

File: libs/utils/energy_model.py [hunk 17]
```diff
@@ -897,6 +992,49 @@ def read_idle_states(cpu, domain_level):
...
+                return False
+
+        def simplified_em_present_in_cpusysfs(target):
+            try:
+                d = '/sys/devices/system/cpu/energy_model'
+                _log.info('Checking if directory {} exists'.format(d))
+                return target.directory_exists(d)
+            except:
```
douglas-raillard-arm: Same as in em_present_in_sd

libs/utils/energy_model.py
```diff
@@ -774,7 +774,102 @@ def _find_core_groups(cls, target):
...
+        #        ../power
+        #           space-separated list of power consumption in corresponding order to capacities
+        # taken together, the contents of capacity, frequency and power give you the required
+        # tuple for ActiveStates.
+        # hence, domain should be supplied as a glob, and fields should be
+        #   capacity, cpus, frequency, power
+
+        try:
+            sysfs_em = target.read_tree_values("/sys/devices/system/cpu/energy_model/", depth=3)
+        except:
```
douglas-raillard-arm: except must match a single exception, or a tuple if needed, to avoid catching things like KeyboardInterrupt

libs/utils/energy_model.py
qperret: No, both should be sorted in the same order, so there is probably no need to call `sorted`

credp: Agreed, we should not sort - if the EM returns incorrect values, let's be able to see them.

libs/utils/energy_model.py
credp: I want to catch any exception - I don't want anything which might cause an exception in one check to prevent us from doing other checks to see which EM should be loaded,

libs/utils/energy_model.py
qperret: I think CPU0 is guaranteed to be the first CPU of at least one freq domain, so fd0 should always exist. But yeah, checking the directory should be just fine anyway ;-)

libs/utils/energy_model.py
```diff
@@ -1011,31 +1011,48 @@ def from_target(cls, target):
...
+                        'load': cls.from_simplifiedEM_target,
+                        'directory': '/sys/devices/system/cpu/energy_model' }
+                     }
+
+        for loader_type in em_loaders:
+            args = em_loaders[loader_type]
+            check = args['check']
+            load = args['load']
+            args.pop('check')
+            args.pop('load')
```
credp: I can combine these like check = args.pop('check') instead of this.

libs/utils/energy_model.py
```diff
@@ -1011,31 +1011,48 @@ def from_target(cls, target):
...
+            'sd'    : { 'check': em_present_in_sd,
+                        'load': cls.from_sd_target,
+                        'filename': '/proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/{}' },
+            'sysfs' : { 'check': simplified_em_present_in_cpusysfs,
+                        'load': cls.from_simplifiedEM_target,
+                        'directory': '/sys/devices/system/cpu/energy_model' }
+                     }
+
+        for loader_type in em_loaders:
+            args = em_loaders[loader_type]
```
credp: I probably should modify a copy rather than the original dictionary.

derkling:
@douglas-raillard-arm seems to be happy with this which has already been tested with our last integration branch. Let's merge it them ;)

PR #607, opened 2018-06-25
XZ2 support for wltests

valschneider:
Tested locally, had to add some intrusive changes to have something working and not spend a whole week on it.

DTBO generation is currently broken on the kernel (https://github.com/sonyxperiadev/kernel-copyleft/tree/51.1.A.2.xxx), and even after getting the DTBO files flashing them seems to soft-brick the phone. This only flashes boot.img for now.


tools/wltests/build
```diff
@@ -287,41 +288,40 @@ fi
...
 if [ ${MAKE_MENUCONFIG} -eq $YES ]; then
 	c_info "Making [menuconfig]..."
-	(set -x; cd "${KERNEL_SRC}"; make menuconfig)
+	(set -x; cd "${KERNEL_SRC}"; make ${BUILD_FLAGS} menuconfig)
 fi
 
 if [ ${MAKE_SAVEDEFCONFIG} -eq $YES ]; then
 	c_info "Making [savedefconfig]..."
-	(set -x; cd "${KERNEL_SRC}"; make savedefconfig)
+	(set -x; cd "${KERNEL_SRC}"; make ${BUILD_FLAGS} savedefconfig)
```
qperret: Dumb question: is this needed for defconfig-related stuff ?

tools/wltests/platforms/xz2_android/build_images
```diff
@@ -0,0 +1,75 @@
...
+source "${BASE_DIR}/helpers"
+source "${PLATFORM_PATH}/definitions"
+
+################################################################################
+# Build all images required to test a new kernel
+################################################################################
+
+warn="\
+Update the 'definitions' file if needed, but be aware that the values must \
+be in sync with the userspace currently flashed on the device or it won't boot\
```
qperret: Nit: maybe we'd like to be able set those from command line as options of the lisa-wltest-series command ? Updating the file is a bit of a pain since it's git-versioned. But I guess we can also export the env variable directly so ...

tools/wltests/platforms/xz2_android/definitions
```diff
@@ -0,0 +1,52 @@
+################################################################################
+# Build definitions for Hikey960
```
qperret: Not Hikey960 AFAICT :wink:

tools/wltests/platforms/xz2_android/definitions
```diff
@@ -0,0 +1,52 @@
...
+################################################################################
+
+PLATFORM_NAME="xz2"
```
qperret: What about akari ? And also, I think the folder should be called akari_v4.9, to be consistent with other platforms.

valschneider: Actually I willingly went against the naming convention of the other platforms because specifying the kernel version doesn't bring a lot of value TBH (I use hikey960_v4.4 to build a v4.14 kernel...)

Although you're right on the akari thing.

tools/wltests/platforms/xz2_android/build_images
valschneider: I actually tested this by exporting the env variables and it works well enough I think ¯\_(ツ)_/¯

tools/wltests/build
valschneider: I would say no, but the build guide uses the same flags for building the config & for building the kernel image, and seeing how everything breaks when I try not to follow the guide I applied this for the defconfig as well...

tools/wltests/platforms/xz2_android/definitions
qperret: I think we sort of agreed that the kernel version is meant to be the lowest supported (/tested) kernel, even though I agree this isn't bringing a lot of useful information in practice ...

valschneider: Fair enough, I'll add it :)

tools/wltests/platforms/xz2_android/build_images
qperret: Fair enough, I'm happy to keep it that way and not clutter the command line tool with even more parameters than it already has

qperret:
Is the pixel2-style module update procedure required for this device ? I guess we'll need to hack the kernel to figure it out ...

valschneider:
Fixed the naming issues.

valschneider:
Rebased on top of master because Travis.

tools/wltests/platforms/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.img
```
derkling: Mmm... not entirely convinced about defining that for all platforms... moreover, for wahoo we already use the `ARTIFACTS_PATH`:
   https://github.com/ARM-software/lisa/blob/master/tools/wltests/helpers#L32

To collect all the platform generates assets, e.g.:
https://github.com/ARM-software/lisa/blob/master/tools/wltests/platforms/wahoo_android_v4.4/build_images#L64

Maybe you should add something similar for the XZ2?

tools/wltests/platforms/akari_android-v4.9/build_images
```diff
@@ -0,0 +1,75 @@
...
+
+warn="\
+They can be found at \
+https://android.googlesource.com/platform/prebuilts/misc/+/master/linux-x86/\
+"
+from_path=0
+
+for var in DTC_EXT DTC_OVERLAY_TEST_EXT; do
+    if [[ -z ${!var} ]]; then
+	echo "tt"
```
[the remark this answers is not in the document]
valschneider: That's a stray debug message...

tools/wltests/platforms/.gitignore
valschneider: Right, I didn't see that ``ARTIFACTS_PATH` thing, I'll add it in.
As for the gitignore, why aren't you convinced about it ? We shouldn't ever want to commit build artifacts, and if we eventually do we can -f them.

tools/wltests/build
```diff
@@ -190,6 +190,7 @@ c_info "Platform configuration:"
 c_info "   PLATFORM_NAME         : ${PLATFORM_NAME}"
 c_info "   DEFCONFIG             : ${DEFCONFIG}"
 c_info "   CONFIG_CMD            : ${CONFIG_CMD}"
+c_info "   BUILD_FLAGS           : ${BUILD_FLAGS}"
```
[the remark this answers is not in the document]
valschneider: Yes and no. They can be used for CFLAGS, but also for other flags to pass to make (like I'm doing here).

tools/wltests/platforms/akari_android-v4.9/build_images
```diff
@@ -0,0 +1,75 @@
...
+	c_warning "$var not set, will attempt to set it by looking at the PATH"
+	from_path=1
+    fi
+done
+
+if [[ -z $from_path ]]; then
+    for tool in dtc ufdt_apply_overlay; do
+	if [[ -z $(which $tool) ]]; then
```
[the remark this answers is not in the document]
valschneider: Yeah, I just wanted to explicitly warn about "variable X not set, looking into your PATH" so that the user knows which variable to set and he/she can re-run wltest easily without having to look into the actual code. I can try to come up with a middle ground.

valschneider:
- Moved `M-x whitespace-cleanup` to a different commit
- Cleaned up **build_image** following @derkling's comments

tools/wltests/platforms/akari_android-v4.9/build_images
```diff
@@ -0,0 +1,75 @@
...
+	fi
+    done
+    export DTC_EXT=$(which dtc)
+    export DTC_OVERLAY_TEST_EXT=$(which ufdt_apply_overlay)
+fi
+
+export KCFLAGS=-mno-android
+export KBUILD_DIFFCONFIG=akari_diffconfig
+
+# The kernel is riddled with
```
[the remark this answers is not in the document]
valschneider: I added a link to the kernel in the commit message

tools/wltests/android/create_boot_img.sh
```diff
@@ -64,6 +64,12 @@ if [ "${ANDROID_RAMDISK_OFFSET}" ]; then
     ANDROID_RAMDISK_OFFSET="--ramdisk_offset ${ANDROID_RAMDISK_OFFSET}"
 fi
 
+if [ "${NO_RAMDISK}" ]; then
```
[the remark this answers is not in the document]
derkling: The point is that variables are empty by default in Bash... thus, you probably don't need to be "explicit"...

However, looking better into it I was wondering why we cannot use an empty RAMDISK definition to just say "i don't want a ramdisk" ?

If you just set in your definitions file:
```bash
RAMDISK_IMAGE=""
```
this will be turned into a:
```bash
android/create_boot_img.sh:DEFAULT_RAMDISK="${PLATFORM_OVERLAY_PATH}/${RAMDISK_IMAGE}"
```
and finally, in your:
```bash
RAMDISK="${RAMDISK:-$DEFAULT_RAMDISK}"
```

Which, ultimately means that all this patch can be simply reduced to this:

```bash
[ -f ${RAMDISK} ] || ramdisk_cmd="--ramdisk '${RAMDISK}'"
```

Meaning that if we do not provide somehow a ramdisk... we don't want to use it... without any additional explicit NO_RAMDISK variables ;)

tools/wltests/test_series
```diff
@@ -804,7 +804,7 @@ match_sha1() {
 
 	# It may happen that a tag pointing to a given commit has
 	# a different SHA1 (signed tag), so check if this this is
-        # the case.
```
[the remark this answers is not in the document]
derkling: you probably missed this one above?

tools/wltests/platforms/akari_android-v4.9/build_images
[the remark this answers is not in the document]
derkling: Still here we have an inline comment... thus I think we should either move the comment in the changelog or, I think better, add the kernel reference here. At least we explicitly say what we are blaming.
Moreover, I guess once the kernel will be (ever?) updated to fix these odds bits... we will have to update this code too.

tools/wltests/platforms/akari_android-v4.9/build_images
```diff
@@ -0,0 +1,99 @@
...
+IMG_PATH=arch/arm64/boot/"${KERNEL_IMAGE}"
+export BUILD_FLAGS="CONFIG_BUILD_ARM64_DT_OVERLAY=y O=${OUTDIR}"
+
+./build --clean_all --image --silent; ERROR=$?
+[ $ERROR -eq 0 ] || exit $ERROR
+
+# Move the image back
+cp "${OUTDIR}"/"${IMG_PATH}" "${KERNEL_SRC}"/"${IMG_PATH}"
+
+export NO_RAMDISK=y
```
derkling: This can be removed if we defined RAMDISK empty as by my previous comment?

tools/wltests/platforms/akari_android-v4.9/build_images
```diff
@@ -0,0 +1,99 @@
...
+./android/create_boot_img.sh; ERROR=$?
+[ $ERROR -eq 0 ] || exit $ERROR
+
+# Not yet supported
```
derkling: Should we log a warning for the time being?

tools/wltests/platforms/akari_android-v4.9/definitions
```diff
@@ -0,0 +1,52 @@
...
+ lpm_levels.sleep_disabled=1\
+ service_locator.enable=1\
+ swiotlb=2048\
+ androidboot.configfs=true\
+ androidboot.usbcontroller=a600000.dwc3\
+ zram.backend=z3fold\
+ msm_drm.dsi_display0=dsi_panel_cmd_display:config0\
+ buildvariant=userdebug
+}"
+
```
derkling: Here we should add:
```bash
RAMDISK_IMAGE="" # Not available for this target
```

tools/wltests/android/create_boot_img.sh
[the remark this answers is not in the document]
valschneider: The reason I added that explicit `NO_RAMDISK` was to differentiate
- when there is no provided ramdisk but we **need** one
- when there is no provided ramdisk and we **know** it's okay

So while setting `RAMDISK_IMAGE=""` can do that, not specifying `RAMDISK_IMAGE` at all will have the same behaviour - in which case we won't warn the user that he/she forgot to specify it.

tools/wltests/platforms/akari_android-v4.9/build_images
```diff
@@ -0,0 +1,102 @@
...
+
+# The kernel is riddled with
+#     #include <something.h>
+# that should really be
+#     #include "something.h"
+# since those are local driver header files.
+# (see https://github.com/sonyxperiadev/kernel-copyleft/tree/51.1.A.2.xxx)
+#
+# As a tempfix, just use an outputdir
+OUTDIR="${KERNEL_SRC}"/out
```
derkling: Should we set it to be under artifacts?

```bash
OUTDIR="${ARTIFACTS_PATH}/kernel_build"
```

valschneider: Oh yeah actually that could be a good way to have all of the artifacts there.

tools/wltests/android/create_boot_img.sh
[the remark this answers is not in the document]
derkling: I don't see this issue for the definitions file: if you forget to specify a ramdisk in that file for a platform which actually needs it and you are sending a pull request... you deserve to be punished by not getting a warning.

IOW: why you should be wise enough to remember to add `NO_RAMDISK` but, at the same time, stupid enough to not specify a ramdisk if you really need it?

I would really keep it simple as assume that who contributes a platform support does not requires warning for things he supposed to specify... unless I'm missing a use-case, you can better explain be tomorrow over a coffee ;-)

valschneider:
- Changed kernel build outdir to `$ARTIFACTS_PATH`
- Allowed users to override `$ANDROID_OS_VERSION` and `$ANDROID_SECURITY_PATCH_LEVEL`

valschneider:
- Removed the NO_RAMDISK bizz

tools/wltests/android/create_boot_img.sh
```diff
@@ -41,7 +42,13 @@ c_info "Generate BOOT image:"
 c_info "   $BOOT_IMAGE"
 c_info "using this configuration :"
 c_info "  KERNEL                 : $KERNEL"
-c_info "  RAMDISK                : $RAMDISK"
+
+if [ -z "$RAMDISK" ]; then
+    c_warning "  No RAMDISK provided, building image without ramdisk"
```
derkling: what about keeping the same formatting here?
```bash
if [ -z "${RAMDISK}" ]; then
c_warning "  RAMDISK                : NOT PROVIDED"
else
c_info "  RAMDISK                : $RAMDISK"
fi
```

tools/wltests/android/create_boot_img.sh
```diff
@@ -64,6 +71,10 @@ if [ "${ANDROID_RAMDISK_OFFSET}" ]; then
     ANDROID_RAMDISK_OFFSET="--ramdisk_offset ${ANDROID_RAMDISK_OFFSET}"
 fi
 
+if [ ! -z "${RAMDISK}" ]; then
```
derkling: Let's keep these consistent... i like the `"$RAMDISK"` format

tools/wltests/android/create_boot_img.sh
```diff
@@ -41,7 +42,13 @@ c_info "Generate BOOT image:"
 c_info "   $BOOT_IMAGE"
 c_info "using this configuration :"
 c_info "  KERNEL                 : $KERNEL"
-c_info "  RAMDISK                : $RAMDISK"
+
+if [ -z "$RAMDISK" ]; then
```
derkling: Let's keep these consistent and use the (not strictly required) `"${RAMDISK}"` format ?

tools/wltests/test_series
```diff
@@ -893,6 +893,11 @@ test_sha1() {
 	echo
 	c_info "Check current kernel..."
 	if device_in_adb; then
+		export ANDROID_OS_VERSION=${ANDROID_OS_VERSION:-$($ADB shell getprop ro.build.version.release)}
```
derkling: These are horribly long lines... what about these?... which should be slightly shorter:

```bash
[ ! -z $ANDROID_OS_VERSION ] || \
    export ANDROID_OS_VERSION="$($ADB shell getprop ro.build.version.release)"
[ ! -z $ANDROID_OS_PATCH_LEVEL ] || \
    export ANDROID_OS_PATCH_LEVEL="$($ADB shell getprop ro.build.version.security_patch)"
```

valschneider:
Addressed @derkling's comments.
Keeping this on standby until we can sort the module stuff.

derkling:
I've assigned @qperret as responsible for merging this, since he is the one having most fun with the target ;)
Otherwise, The code LGTM so fare

valschneider:
FYI marking this as stale because even though we have a working automation loop, changing the kernel too much (compared to the base) seems to prevent the phone from booting.

PR #610, opened 2018-07-02
libs/utils: Handle 'Disabled' SELinux with non-android userspaces

credp:
On my Juno running Debian Jessie userspace, I have a problem with the
selinux handling code in executor.py. I have selinux infrastructure, but
none of it is enabled.

When you call getenforce, the value is 'Disabled' and return code 0. This
causes us to set self._old_selinux_mode to 'Disabled' and proceed to call
'setenforce 0', which returns 'setenforce: SELinux is disabled' with
return code 1 - which is correctly raised as a TargetError by
target.execute. We can't restore the disabled mode either, setenforce
always returns an error code.

> setenforce 0 ; echo $?
setenforce: SELinux is disabled
1
> setenforce 1 ; echo $?
setenforce: SELinux is disabled
1

Since what we're after is to have selinux in permissive mode, instead
take account of the 'Disabled' _old_selinux_mode to also make no change
to selinux mode, just like we do if getenforce doesn't exist.

### DIFF
--- a/libs/utils/executor.py
+++ b/libs/utils/executor.py
@@ -318,6 +318,30 @@ class Executor():
     def get_run_dir(target):
         return os.path.join(target.working_directory, TGT_RUN_DIR)
 
+    def _selinux_mode_unused(self):
+        return (self._old_selinux_mode is None or 'Disabled' in self._old_selinux_mode)
+
+    def _clear_selinux_mode(self):
+        try:
+            # First, save the old SELinux mode
+            self._old_selinux_mode = self.target.execute('getenforce')
+        except TargetError:
+            # Probably the target doesn't have SELinux. No problem.
+            pass
+        else:
+            if self._selinux_mode_unused():
+                self._log.info('Target SELinux not present or disabled')
+            else:
+                self._log.warning('Setting target SELinux in permissive mode')
+                self.target.execute('setenforce 0', as_root=True)
+
+    def _restore_selinux_mode(self):
+        if not self._selinux_mode_unused():
+            self._log.info('Restoring target SELinux mode: %s',
+                           self._old_selinux_mode)
+            self.target.execute('setenforce ' + self._old_selinux_mode,
+                                as_root=True)
+
     def _setup_rootfs(self, tc):
         # Initialize CGroups if required
         self._cgroups_init(tc)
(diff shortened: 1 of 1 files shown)


libs/utils/executor.py
```diff
@@ -318,6 +318,30 @@ def _setup_sched_features(self, tc):
     def get_run_dir(target):
         return os.path.join(target.working_directory, TGT_RUN_DIR)
 
+    def _selinux_mode_unused(self):
+        return (self._old_selinux_mode is None or 'Disabled' in self._old_selinux_mode)
+
+    def _clear_selinux_mode(self):
+        try:
+            # First, save the old SELinux mode
+            self._old_selinux_mode = self.target.execute('getenforce')
```
valschneider: I'm not too hot on handling that through side effects, but I guess there's no better way to handle it now given it's all within Executor.

credp: Maybe, but this is the only place in the whole codebase where we use it - I could propose it as a devlib module instead, but that seems like a lot of work for a few lines. What do you think?

derkling: I think it could easily fit as a couple of additional target APIs, not necessarily a new module.

Just after the 'adb_*' commands we have here:
   https://github.com/ARM-software/devlib/blob/master/devlib/target.py#L1361
perhaps it's fine to add a new context manager API which allows something like

```python
with selinux_disabled():
   # do your staff... and then SELinux settings will be updated if required.
```

But @valschneider is our context manager expert, is that possible?

valschneider: My endgame view here is to make as many features as possible available to devlib to have a cleaner/simpler interface for the "next generation" of workloads (and to be good open source citizens by pushing our devlib "addons" back to devlib).

I agree it's just a small change but IMHO it's worth the effort in the long run :)

In the end it's your call, we can have this merged and I'll come bother you when we finally get rid of `Executor` :)

valschneider: @derkling That ought to be possible, although I'd prefer a non context-manager version to be added first, and we can easily wrap it around one if needed.

libs/utils/executor.py
```diff
@@ -318,6 +318,30 @@ def _setup_sched_features(self, tc):
     def get_run_dir(target):
         return os.path.join(target.working_directory, TGT_RUN_DIR)
 
+    def _selinux_mode_unused(self):
```
douglas-raillard-arm: nitpicking: it's usually easier to read code with variables that are something instead of "not something", to avoid double negation

PR #611, opened 2018-07-04
Fix cpuhp test

qperret:
The CPU hotplug test is full of useless dependencies which silently talk
to the target (rt-app, freezing userspace, ftrace, ...). When the target
crashes during the hotplug stress, requests sent to the target because
of the above modules can hang indefinitely, or fail miserably. That is
an issue since, in the specific case of the hotplug test, we would like
to classify these crashes as test failures.

Remove all the dependencies by overwritting the runExperiments method of
LisaTest in order to avoid using an Executor. The test spawns a bash
script on the target that plugs CPUs in and out, and does _nothing_ on
the host side during a (configurable) amount of time before checking if
the target is alive.

### DIFF
--- a/tests/eas/cpuhp.py
+++ b/tests/eas/cpuhp.py
@@ -21,44 +21,27 @@ import random
 import json
 import sys
 
+from env import TestEnv
 from test import LisaTest, experiment_test
 from target_script import TargetScript
 from devlib.module.hotplug import HotplugModule
+from devlib.exception import TimeoutError
 
 class _CpuHotplugTest(LisaTest):
     """
     "Abstract" base class for generic EAS tests under CPU hotplug stress
 
-    Subclasses should provide a .workloads member to populate the 'wloads' field
-    of the experiments_conf for the Executor. Furthermore, subclasses may
-    overwrite the .hp_stress member containing the configuration used while
-    generating the random hotplug stress sequence.
-
-    The test will be successful if the system does not halt or kill the
-    workloads because of hotplug stress. It means that the test passes if the
-    workload completes.
+    The test will be successful if the system does not crash during the hotplug
+    stress, and if all CPUs come back online properly.
     """
 
     test_conf = {
-        "ftrace" : {
-            "events" : [
-                "sched_overutilized",
-                "sched_energy_diff",
-                "sched_load_avg_task",
-                "sched_load_avg_cpu",
-                "sched_migrate_task",
-                "sched_switch",
-                "cpu_frequency",
-                "cpu_idle",
-                "cpu_capacity",
-                "cpuhp_enter",
-                "cpuhp_exit",
-                "cpuhp_multi_enter",
-            ],
-        },
-        "modules": ["cgroups", "hotplug"],
-        "tools": ["rt-app", "trace-cmd"],
+        "modules": ["hotplug"],
+        # Remove the modules that silently talk to the target
+        "exclude_modules": ["hwmon", "ftrace"],
     }
+
+    duration_sec = 10                   # Duration of the hotplug stress
 
     hp_stress = {
         'seed' : None,                  # Seed of the random number generator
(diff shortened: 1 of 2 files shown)


tests/eas/cpuhp.py
```diff
@@ -77,18 +61,19 @@ def setUpClass(cls, *args, **kwargs):
     @classmethod
     def _getExperimentsConf(cls, test_env):
```
valschneider: Since we're overriding `runExperiments`, do we still need this?

tests/eas/cpuhp.py
```diff
@@ -109,14 +94,22 @@ def _experimentsInit(cls, *args, **kwargs):
         cls._log.info('Hotpluggable CPUs found on target: ')
         cls._log.info(cls.hotpluggable_cpus)
 
+        # Mark the target as alive for now
+        cls.target_alive = True
+
         # Run random hotplug sequence on target
         cls.cpuhp_seq_script = cls._random_cpuhp_script()
         cls.cpuhp_seq_script.push()
         cls.cpuhp_seq_script.run(as_root=True, background=True)
```
valschneider: Devlib's `execute` has a `timeout` param. What about extending `target_script` to forward this param (it only makes sense for `background=False`, but that's fine)?

Not sure if it's **exactly** the same in terms of end results to launching the script in background and then checking for responsiveness after some amount time though...

tests/eas/cpuhp.py
qperret: Nope, that's a remnant of previous version of this test, I'll remove it

tests/eas/cpuhp.py
qperret: So, as of now, the script on the target is pretty stupid and runs in an infinite loop. So the host has to intervene to stop it, we can't just wait for it to finish or throw an exception ... The reason I originally implemented it that way is because bash on android is 'funny' so I just went for the simplest possible thing to avoid coping with that ...

valschneider: Ok I won't question that :)

qperret: Actually maybe I can do something by using `timeout` http://man7.org/linux/man-pages/man1/timeout.1.html
I'll look into it

qperret:
Updated the patches a bit. Now the script runs on the target for a specific amount of time, not forever. On the host side we consider the target has crashed if we have no sign of life 1 minute after the expected end of execution of the script.
Tested on Juno r0 and H960.

tests/eas/cpuhp.py
```diff
@@ -168,7 +142,13 @@ def _random_cpuhp_script(cls):
             if sleep_max > 0:
                 sleep_dur_sec = random.randint(sleep_min, sleep_max)/1000.0
                 script.append(shift + 'sleep {}'.format(sleep_dur_sec))
-        script.append('done')
+        if timeout:
```
valschneider: `timeout`'s a mandatory parameter now, so could this be unified?
That way, maybe you could just used a time-based loop like this:
```bash
script.append('end=$((SECONDS+{}))'.format(timeout)
script.append('while [ $SECONDS -lt $end ]; do')
# Insert hotplug action here
script.append('done')
```
You could also use `date` instead (Stackoverflow is a wonderful place https://stackoverflow.com/a/11198713/5096023).

tests/eas/cpuhp.py
```diff
@@ -110,44 +84,42 @@ def _experimentsInit(cls, *args, **kwargs):
...
+        cls.target.hotplug.online_all()
 
     @classmethod
     def _cpuhp_path(cls, cpu):
         cpu = 'cpu{}'.format(cpu)
         return cls.target.path.join(HotplugModule.base_path, cpu, 'online')
 
     @classmethod
-    def _random_cpuhp_script(cls):
+    def _random_cpuhp_script(cls, timeout):
```
valschneider: `timeout_s` ?

tests/eas/cpuhp.py
qperret: The time-based loop won't work well because the number of actions in the loop is arbitrary long, so one iteration of the loop can be arbitrary long as well. That why I kill it 'from the outside'.

But yeah I can remove the `if timeout:` part, n/p.

tests/eas/cpuhp.py
qperret: Ack

qperret:
Removed the `if timeout` part of the script generation method and renamed `timeout` to `timeout_s`.

PR #612, opened 2018-07-06
tests: Sanity check on Energy Models

qperret:
This is based on Elieva's #577

I tried to take into account @douglas-raillard-arm's feedback on previous PR (IIUC). I turned `TestStaticEnergyModel` into an abstract class which is sub-classed near the definition of the static energy models. That basically lets us do something like `lisa-test libs/utils/platforms/hikey960_energy.py:SanityCheck`.

I also split some patches into smaller-patches, and fixed minor things to have all functions to work with a simplified Energy Model. Other than that, that's very much the same code.

Tested on a live target, on static models, and also tested the `tests/lisa` part. Everything looks fine.

All the credit goes to @Elieva :)

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -19,6 +19,7 @@ from collections import namedtuple, OrderedDict
 from itertools import product
 import logging
 import operator
+import itertools
 import re
 
 import pandas as pd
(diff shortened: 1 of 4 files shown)


libs/utils/energy_model.py
```diff
@@ -1152,3 +1154,404 @@ def f(input_row):
...
+    :type node: EnergyModelNode
+
+    :returns: The list of capacities associates with the given node.
+    """
+    cap = []
+    for freq in node.active_states.keys():
+        cap.append(node.active_states[freq].capacity)
+    return cap
+
+def _is_list_strickly_increasing(l):
```
valschneider: `pandas.Series` has an `is_monotonic_increasing` property, so that could be turned into:
```python
    def _is_list_monotonic_increasing(l):
        return pd.Series(l).is_monotonic_increasing
```
That brings in a conversion to a Series though (dunno how smart the converstion is). The Numpy way might be a bit better:
```python
    def _is_list_monotonic_increasing(l):
        return all(diff > 0 for diff in np.diff(l))
```

libs/utils/platforms/hikey960_energy.py
```diff
@@ -0,0 +1,119 @@
...
+            idle_states=["cluster-sleep-0"],
+            children=[PowerDomain(idle_states=["WFI", "cpu-sleep-0"], cpu=c)
+                      for c in a73s]),
+        PowerDomain(
+            idle_states=["cluster-sleep-0"],
+            children=[PowerDomain(idle_states=["WFI", "cpu-sleep-0"], cpu=c)
+                      for c in a53s])]),
+    freq_domains=[a53s, a73s])
+
+class SanityCheck(_StaticEnergyModelTest):
```
valschneider: That looks a bit awkward, I suppose that's here because you want to run a test by executing something like `lisa-test libs/utils/platforms/hikey960_energy.py` and you can't give arguments to nosetest other than the test file...

What about having a test in lisa/tests/lisa that just runs the SanityCheck against all of our platform energy files? On the plus side, that'll also make Travis run that against PRs.

tests/lisa/test_energy_model.py
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
+import tests.utils.em as em_utils
```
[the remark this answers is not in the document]
qperret: Oh no actually I had a .pyc file that stayed there and that's why it worked on my host. I fix it anyways :)

libs/utils/energy_model.py
qperret: I would argue that a good old loop if more efficient that those conversions that will require memory re-allocation and all ...

valschneider: Up to you, the more I use Python the more I get the feeling that loops are so Y2K :D

libs/utils/platforms/hikey960_energy.py
qperret: The issue with having Travis look at that is that all platforms have at least one test failing ... For good reasons. And Elieva already implemented some really nice stuff in `test/lisa` I think

Other than that, I quite like the idea of being able to do `lisa-test libs/utils/platforms/hikey960_energy.py:SanityCheck` for example. I don't think the goal of these tests is to actually get them to pass. It is just to report information giving us an idea of how broken a platform is. So I liked the idea to be able to run the test on individual platforms. I assume that most of the time we'll run these tests only once, when we have a new platform

tests/lisa/test_energy_model.py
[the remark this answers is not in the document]
qperret: Ok that _should_ be fixed now

libs/utils/platforms/hikey960_energy.py
valschneider: Okay then. The only bit I don't like is having to redefine the test method for each model, but that's because of how we define those EMs. And that's also why we end up doing all those lines of `import *_energy` in TestEnv.

Anyway, that's nothing this PR is supposed to address, so I guess that's fine.

qperret: We agreed offline that the fact that we don't really want these tests to pass is simply because they're not really tests ... So I'll remove the `_StaticEnergyModelTest` abstract class and add a notebook instead

libs/utils/energy_model.py
```diff
@@ -1152,3 +1153,366 @@ def f(input_row):
...
+def _get_capacities_list(node):
+    """
+    Get the list of capacities for a given node.
+    :param node: A node of the energy model.
+    :type node: EnergyModelNode
+
+    :returns: The list of capacities associates with the given node.
+    """
+    cap = []
+    for freq in node.active_states.keys():
```
douglas-raillard-arm: can be largely simplified into:
```
return [state.capacity for state in node.active_states.values()]
```

libs/utils/energy_model.py
```diff
@@ -1152,3 +1153,366 @@ def f(input_row):
...
+def _get_powers_list(node):
+    """
+    Get the list of powers for a given node.
+    :param node: A node of the energy model. It can be a cluster or cpu.
+    :type node: EnergyModelNode
+
+    :returns: The list of powers associates with the given node.
+    """
+    powers = []
+    for freq in node.active_states.keys():
```
douglas-raillard-arm: same comment as for _get_capacities_list

libs/utils/energy_model.py
```diff
@@ -1152,3 +1153,366 @@ def f(input_row):
...
+
+    :returns: The list of capacities associates with the given node.
+    """
+    cap = []
+    for freq in node.active_states.keys():
+        cap.append(node.active_states[freq].capacity)
+    return cap
+
+def _is_list_strickly_increasing(l):
+    for i in range(len(l)-1):
```
douglas-raillard-arm: For short lists, the fastest measured way is:
sorted(l) == l
This is because sorted is a builtin and avoids part of the bytecode machinery.
You can convince yourself with timeit module.

libs/utils/energy_model.py
```diff
@@ -1152,3 +1153,366 @@ def f(input_row):
...
+    :type node: EnergyModelNode
+
+    :returns: The list of capacities associates with the given node.
+    """
+    cap = []
+    for freq in node.active_states.keys():
+        cap.append(node.active_states[freq].capacity)
+    return cap
+
+def _is_list_strickly_increasing(l):
```
douglas-raillard-arm: typo in "strictly" ("strickly")

qperret:
Ok so I removed all the bits under `tests/eas/` and turned that into a notebook. I kept the `tests/lisa` stuff though because they're nice :smile:

PR #613, opened 2018-07-10
tests fixes

douglas-raillard-arm:
Couple of fixes from @deggeman , one of them is maybe temporary but necessary to be able to test the simplified energy model. Both commits have already been used for a few integrations.

### DIFF
--- a/tests/eas/generic.py
+++ b/tests/eas/generic.py
@@ -16,6 +16,7 @@
 #
 
 import os
+import logging
 import numpy as np
 import pandas as pd
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


qperret:
Ideally I think we'd like to have a test in preliminary.py for the sched_feature. Something simple that checks that the sched_feat is set if it is available.

Other than that, LGTM :)

douglas-raillard-arm:
@qperret if the behavior becomes stable, we can add the check you suggest. This change to the existing tests would still be necessary to allow running them according to the commit message.

qperret:
I'm happy to remove 'ENERGY_AWARE' from the test as suggested, but instead of removing it unilaterally it could just be added to preliminary.py in a smarter way. That would work in all cases.

qperret:
#614 introduces the check on ENERGY_AWARE in preliminary.py. As soon as it's merged, this PR is good to go for me.

Issue #425, opened 2017-06-19
Make new devices work out of the box


valschneider:
Having a new target device for LISA currently requires a bit of work, although the base requirements for a new device aren't that big.

It mostly all comes down to the platform file in **libs/utils/platforms/\<device-name\>.json**. The energy model can be fetched from the target, but post-processing usually requires the json file. This isn't entirely flawed logic, as it should be possible to run post-processing of traces without having access to the device itself, but maybe it should be handled differently. Here's a list of options:

- Do as much post-processing as possible without the platform file - this is what https://github.com/ARM-software/lisa/pull/342 & https://github.com/ARM-software/lisa/issues/427 are about.
- Export the platform description if it's not already there when running tests - we can get all of the information that we need from the device itself, so generating the platform description shouldn't be too hard. The tricky point is cleanly distinguishing big and LITTLE clusters.

joelagnel:
+1

valschneider:
Now that we have automated the devlib module loading, we've done good progress on this side. Now it's mostly about cleaning up analysis to not depend on hard-coded energy model values, but that's https://github.com/ARM-software/lisa/issues/427

PR #615, opened 2018-07-12
Added support for serialization and deserialization of the Energy Model

johnny90:
This is a first step towards making new devices work out of the box (#425).
Take the Energy Model fetched from the target and transform it into a json file.

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -20,6 +20,9 @@ from itertools import product
 import logging
 import operator
 import re
+from ruamel.yaml import YAML, yaml_object
+
+yaml = YAML()
 
 import pandas as pd
 import numpy as np
@@ -67,6 +70,7 @@ class EnergyModelCapacityError(Exception):
     """Used by :meth:`EnergyModel.get_optimal_placements`"""
     pass
 
+@yaml_object(yaml)
 class ActiveState(namedtuple('ActiveState', ['capacity', 'power'])):
     """Represents power and compute capacity at a given frequency
 
@@ -75,6 +79,19 @@ class ActiveState(namedtuple('ActiveState', ['capacity', 'power'])):
     """
     def __new__(cls, capacity=None, power=None):
         return super(ActiveState, cls).__new__(cls, capacity, power)
+
+    # helpers for yaml serialization
+    yaml_tag = u'em_active_state:capacity,power'
+
+    @classmethod
+    def to_yaml(cls, representer, node):
+        return representer.represent_scalar(cls.yaml_tag,
+                u'{.capacity},{.power}'.format(node, node))
+
+    @classmethod
+    def from_yaml(cls, constructor, node):
+        cap, power = (int(x) for x in node.value.split(','))
+        return cls(capacity=cap, power=power)
 
 class _CpuTree(object):
     """Internal class. Abstract representation of a CPU topology.
@@ -128,6 +145,7 @@ class _CpuTree(object):
         """Iterate over leaves"""
         return self._iter(False)
 
+@yaml_object(yaml)
 class EnergyModelNode(_CpuTree):
     """Describes topology and energy data for an EnergyModel.
 
@@ -157,7 +175,7 @@ class EnergyModelNode(_CpuTree):
                  cpu=None, children=None, name=None):
         super(EnergyModelNode, self).__init__(cpu, children)
 
-        self._log = logging.getLogger('EnergyModel')
+        _log = logging.getLogger('EnergyModel')
 
         def is_monotonic(l, decreasing=False):
             op = operator.ge if decreasing else operator.le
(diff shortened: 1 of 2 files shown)


libs/utils/serialization.py
```diff
@@ -0,0 +1,151 @@
+from collections import namedtuple, Iterable, OrderedDict
```
valschneider: This needs a copyright header

libs/utils/serialization.py
```diff
@@ -0,0 +1,151 @@
...
+import numpy as np
+import json
+
+from energy_model import (ActiveState, EnergyModelNode, EnergyModelRoot,
```
valschneider: I think it's alright to go over 80 chars for the imports, so you could have that in a single line and ditch the parentheses.

valschneider:
Travis fails with the following error:
```python
File "/home/travis/build/ARM-software/lisa/libs/utils/serialization.py", line 5, in <module>
    from energy_model import (ActiveState, EnergyModelNode, EnergyModelRoot,
ImportError: cannot import name ActiveState
```

libs/utils/serialization.py
```diff
@@ -0,0 +1,151 @@
...
+                          PowerDomain, EnergyModel)
+
+def serialize(data):
```
douglas-raillard-arm: I'm not sure there is any value in supporting that manually compared to using yaml.dump. The output will be as readable and zero boilerplate to maintain.

libs/utils/energy_model.py
```diff
@@ -1012,6 +1033,11 @@ def from_target(cls, target):
                   reported by the target.
         """
         _log = logging.getLogger('EMReader')
+        yaml.register_class(ActiveState)
```
douglas-raillard-arm: According to the doc, explicit registration is not necessary if the classes are decorated with @yaml_object(yaml)

http://yaml.readthedocs.io/en/latest/dumpcls.html

libs/utils/energy_model.py
```diff
@@ -1056,7 +1082,8 @@ def simplified_em_present_in_cpusysfs(target, directory=None):
             if em_present:
                 _log.info('Attempting to load EM using {}'.format(load.__name__))
                 em = load(target, **args)
-                data_to_json(em)
+                with open("em_from_target.yml", "w") as f:
```
douglas-raillard-arm: This should be moved in a separate method like save_yaml(yaml_path) so it is only done explicitly with the user deciding of the filename

libs/utils/energy_model.py
```diff
@@ -77,6 +80,18 @@ class ActiveState(namedtuple('ActiveState', ['capacity', 'power'])):
...
+    yaml_tag = u'em_active_state:capacity,power'
+
+    @classmethod
+    def to_yaml(cls, representer, node):
+        return representer.represent_scalar(cls.yaml_tag,
+                u'{.capacity},{.power}'.format(node, node))
+
+    @classmethod
+    def from_yaml(cls, constructor, node):
+        return cls(*node.value.split(','))
```
douglas-raillard-arm: This may work, but there is a good chance that the rest of the code is expecting integers instead of strings, so something like that:
cap, power = (int(x) for x in node.value.split(','))
return cls(capacity=cap, power=power)

douglas-raillard-arm:
LGTM, can be merged after squashing commits as appropriate and rebasing on master

libs/utils/energy_model.py
```diff
@@ -1001,6 +1025,16 @@ def read_idle_states(cpu, domain_level):
                    root_power_domain=root_pd,
                    freq_domains=freq_domains)
 
+    @staticmethod
+    def save_em_yaml(em, yaml_path):
+        with open(yaml_path, "w") as f:
```
valschneider: Do we want to encode some sane default name, e.g. "nrg_model.json"? Same goes for the loading, so you can just point it at some directory and if you used the defaults it will just work™

douglas-raillard-arm: TBH, it does not sound like a great idea. We will end up with code expecting that and not propagating the path in the wrapping APIs just because "it is not required" and we end up with the overwritten files nightmare again.

douglas-raillard-arm: Until we have a better view on how this API will actually be used, having defaults does not really make sense.

valschneider: Fair enough

libs/utils/energy_model.py
```diff
@@ -1001,6 +1025,16 @@ def read_idle_states(cpu, domain_level):
                    root_power_domain=root_pd,
                    freq_domains=freq_domains)
 
+    @staticmethod
+    def save_em_yaml(em, yaml_path):
```
valschneider: Naming nit: Could those be named `to_yaml` and `from_yaml` ? That would better match what we currently have (`from_target`). Also those methods are under the EnergyModel class, so IMO there's no need to specify "em" in the method name.

Also, `from_yaml` should be a classmethod, and `to_yaml` can be a regular non-static method (or a classmethod if you want some sort of symmetry, but I'd find it more cumbersome than anything else).

Those are comments that I could have made way back then, but they only hit me as I started testing this PR...

valschneider: Hm the names I suggested interfere with `ruamel.yaml` and how it serializes objects. I guess we can use something like `to/from_disk` instead, and it's even neater because we don't expose what file format we're using :D

douglas-raillard-arm: Unfortunately, these names are used by ruamel.yam:
https://yaml.readthedocs.io/en/latest/dumpcls.html

Ultimately, these helpers should go in a base class so it can be used for other classes as well, so it may not be ready to merge yet.

valschneider: Aye, a base class for those would be nice.

valschneider:
I stole this one too...

PR #616, opened 2018-07-19
WiP: energy_model: Fix loading simplified EM from target w/o 'capacity'

qperret:
The v5 of the simplified EM doesn't expose the 'capacity' file in sysfs.
Fix this by checking if the file exists, and re-compute those values
form within LISA if not.

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -830,7 +830,14 @@ class EnergyModel(object):
             sysfs_em[fd]['cpus'] = cpus
             sysfs_em[fd]['frequency'] = map(int, sysfs_em[fd]['frequency'].split(' '))
             sysfs_em[fd]['power'] =  map(int, sysfs_em[fd]['power'].split(' '))
-            sysfs_em[fd]['capacity'] = map(int, sysfs_em[fd]['capacity'].split(' '))
+
+            # Compute the capacity of the CPUs at each OPP with a linerar
+            # mapping to the frequencies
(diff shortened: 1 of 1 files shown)


qperret:
I leave this as WiP until the patches actually get posted

libs/utils/energy_model.py
```diff
@@ -830,7 +830,18 @@ def from_simplifiedEM_target(cls, target,
             sysfs_em[fd]['cpus'] = cpus
             sysfs_em[fd]['frequency'] = map(int, sysfs_em[fd]['frequency'].split(' '))
             sysfs_em[fd]['power'] =  map(int, sysfs_em[fd]['power'].split(' '))
-            sysfs_em[fd]['capacity'] = map(int, sysfs_em[fd]['capacity'].split(' '))
+
+            if 'capactity' in sysfs_em[fd]:
```
qperret: BROKEN !

credp: When you fix that, LGTM.

valschneider: Hmm thinking about it, do we really want to support <v4 bindings? Or do you want to keep it as it to merge it now, and eventually remove the v4 bits?

qperret: Actually I can just remove that and just recompute the 'capacity' no matter what the EM version is. That'll work just fine for both versions

qperret:
No more special case for EM v4 and previous. We just recompute the capacity no matter what. That seem to work in both cases

PR #618, opened 2018-07-19
Add schedtune hold synthetic test

credp:
Use the frequency behaviour when running rt and cfs tasks
to determine if the schedtune hold functionality is working
properly.

### DIFF
--- a/tests/eas/st_hold.py
+++ b/tests/eas/st_hold.py
@@ -1,0 +1,389 @@
+#    Copyright 2018 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import division
+
+import os
+import logging
+import unittest
+
+from env import TestEnv
+from test import LisaTest
+from trappy import FTrace
+from wlgen import Periodic, RTA
+from bart.common.Utils import area_under_curve
+
+BOOSTED_CGROUP_NAME = '/boosted'
+UNBOOSTED_CGROUP_NAME = '/unboosted'
+FREQ_MARGIN = 20000
+TEST_CPU = 0
+
+class _SchedtuneHoldTest(LisaTest):
+    """
+    Base class for all Schedtune-Hold feature tests. Contains
+    setup code and useful test methods.
+    When a derived class is instantiated, this sequence of events
+    happens:
+      :meth:`setUpClass` called
+          -> connect target object
+          -> call derived_class:child_check_config() if implemented
+          -> call derived_class:populate_params() (must be implemented)
+          -> run workload
+          -> build Trace object in cls.trace_obj
+    Then the actual test function is executed on the collected trace.
+
+    child_check_config is used for the two tests present where only one
+    can be run - which one is OK depends upon the status of the
+    SCHEDTUNE_BOOST_HOLD_ALL sched_feature. This function should raise a
+    :class:`unittest.SkipTest` exception if it wants to prevent a derived
+    class test from running without causing a failure.
+    """
+    boosted_cgroup_name = BOOSTED_CGROUP_NAME
+    unboosted_cgroup_name = UNBOOSTED_CGROUP_NAME
+    frequency_margin = FREQ_MARGIN
+    test_cpu = TEST_CPU
+
+    test_conf = { "modules": ["bl", "cpufreq", "cgroups" ],
+                  "tools": ["rt-app"],
+                  "ftrace" : {
+                      "events" : ["cpu_frequency", "sched_switch", "sched_boost_cpu"],
+                      "buffsize" : 10240
+                  },
+                  "flags" : [ "ftrace" ]
+                }
+
+    @classmethod
+    def setUpClass(cls):
+        cls._log = logging.getLogger('StHoldTest')
+        cls.params = {}
+        # connect the target
+        cls.env = TestEnv(test_conf=cls._getTestConf())
+        cls.trace_file = os.path.join(cls.env.res_dir, '{}_trace.dat'.format(cls.task_name))
+        cls.target = cls.env.target
+        # check that the SCHEDTUNE_HOLD_ALL feature exists in the target
+        cls.check_hold_available()
+        # check any per-class requirements
+        cls.child_check_config()
+        cls.rta_name = 'rt-test-{}'.format(cls.task_name.replace('/',''))
+        # ask derived class to populate workload config etc.
+        cls.populate_params()
+        # ensure required cgroups exist
+        cls.setup_cgroups()
+        # run workload
+        with cls.env.freeze_userspace():
+            cls.run_workload()
+        # collect and parse the trace
+        cls.ftrace_obj = FTrace(cls.trace_file)
+
+    @classmethod
+    def child_check_config(cls):
+        pass
+
+    @classmethod
+    def run_workload(cls):
+        cls.env.ftrace.start()
+        cls.rtapp.run(out_dir=cls.env.res_dir, cgroup=cls.cgroup_name, as_root=True)
+        cls.env.ftrace.stop()
+        cls.env.ftrace.get_trace(cls.trace_file)
+        cls.env.platform_dump(cls.env.res_dir)
+
+    @classmethod
+    def setup_cpuset_cgroup(cls, cgroup, cpus, mems):
+        cpu_g = cls.cpuset_controller.cgroup(cgroup)
+        cpu_g.set(cpus=cpus, mems=mems)
+
+    @classmethod
+    def setup_schedtune_cgroup(cls, cgroup, boost, prefer_idle):
+        st_g = cls.schedtune_controller.cgroup(cgroup)
+        st_g.set(boost=boost, prefer_idle=prefer_idle)
+
+    @classmethod
+    def setup_cgroups(cls):
+        """
+        When we set up the cgroups in a unified hierarchy, we
+        need to ensure that the cpu controller is minimally
+        configured before we can run tasks inside a group.
+        Copy the root group's mems and cpus parameters into the
+        groups we create just in case.
+        """
+        cls.cpuset_controller = cls.target.cgroups.controller('cpuset')
+        root_g = cls.cpuset_controller.cgroup('/')
+        params = root_g.get()
+        cls.schedtune_controller = cls.target.cgroups.controller('schedtune')
+        cls.setup_schedtune_cgroup(cls.boosted_cgroup_name, 100, 0)
+        cls.setup_cpuset_cgroup(cls.boosted_cgroup_name,
+                                params['cpus'], params['mems'])
+        cls.setup_schedtune_cgroup(cls.unboosted_cgroup_name, 0, 0)
+        cls.setup_cpuset_cgroup(cls.unboosted_cgroup_name,
+                                params['cpus'], params['mems'])
+
+    @classmethod
+    def get_task_times(cls):
+        """
+        figure out timestamps where task is running
+        defined here as:
+          start_time = first time we switch to the correctly named task
+          end_time   = last time we switch away from the correctly named task
+        """
+        df = cls.ftrace_obj.sched_switch.data_frame
+        cls.task_start_time = df[df.next_comm == cls.task_name].iloc[0].name
+        cls.task_end_time = df[df.prev_comm == cls.task_name].iloc[-1].name
+
+    @classmethod
+    def avg_cpu_freq(cls):
+        """
+        To calculate average frequency, first obtain the area under the
+        series of frequency transitions for the period of interest. We do
+        this by recomputing the index on the frequency series derived from
+        the cpu_frequency trace objects and then using area_under_curve from
+        bart. Once we have the area, we can divide it by the time span to
+        obtain average frequency.
+        """
+        df = cls.ftrace_obj.cpu_frequency.data_frame
+        cpu_df = df[df.cpu == cls.test_cpu]
+        freq_s = cpu_df.frequency
+        old_index = list(freq_s[cls.task_start_time:cls.task_end_time].index)
+        # the new index should run from the start to the end time, including
+        # any events which occurred in between
+        new_index = [ cls.task_start_time ] + old_index + [ cls.task_end_time ]
+        windowed_freq_s = freq_s.reindex(index=new_index, method='pad')
+        # avg_freq is area / time for the window of interest
+        area = area_under_curve(windowed_freq_s, method='rect')
+        return int((area + 0.5) / (cls.task_end_time - cls.task_start_time))
+
+    @classmethod
+    def check_hold_available(cls):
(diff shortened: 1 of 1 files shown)


tests/eas/st_hold.py
```diff
@@ -0,0 +1,380 @@
+#    Copyright 2018-2018 ARM Limited
```
valschneider: I think "2018" is enough, right?

tests/eas/st_hold.py
```diff
@@ -0,0 +1,380 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import numpy as np
+import pandas as pd
+from env import TestEnv
```
valschneider: Nit: I don't think we have an official style for imports, but the [holy book](https://www.python.org/dev/peps/pep-0008/#imports) states that it should be standard modules first, then third party modules, then local modules.

tests/eas/st_hold.py
```diff
@@ -0,0 +1,380 @@
...
+        self.params = {}
+        self.env = TestEnv(test_conf=test_conf)
+        self.trace_file = os.path.join(self.env.res_dir, '{}_trace.dat'.format(self.task_name))
+        self.target = self.env.target
+        self._testRequiredSchedFeatures()
+        self.doPerClassEarlyCheck()
+        self.rta_name = 'rt-test-{}'.format(self.task_name.replace('/',''))
+        self.populate_params()
+        self.setUpCGroups()
+        with self.env.freeze_userspace():
```
valschneider: Just another proof that we need to go 1789 on Executor :D

tests/eas/st_hold.py
credp: ok, I'm updating this

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+        # ensure required cgroups exist
+        self.setUpCGroups()
+        # run workload
+        with self.env.freeze_userspace():
+            self.run_workload()
+        # collect and parse the trace
+        self.ftrace_obj = trappy.FTrace(self.trace_file)
+
+    @classmethod
+    def doPerClassEarlyCheck(self):
```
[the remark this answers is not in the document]
credp: Just that the test case I started with used it, so I did too, nothing specific

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+        self.ftrace_obj = FTrace(self.trace_file)
+
+    @classmethod
+    def doPerClassEarlyCheck(self):
+        pass
+
+    @classmethod
+    def run_workload(self):
+        self.env.ftrace.start()
+        self.rtapp.run(out_dir=self.env.res_dir, cgroup=self.cgroup_name)
```
qperret: That also wnats `as_root=True` I think

credp: done

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+        cpu_g = self.cpuset_controller.cgroup(cgroup)
+        cpu_g.set(cpus=cpus, mems=mems)
+
+    @classmethod
+    def setup_schedtune_cgroup(self, cgroup, boost, prefer_idle):
+        st_g = self.schedtune_controller.cgroup(cgroup)
+        st_g.set(boost=boost, prefer_idle=prefer_idle)
+
+    @classmethod
+    def setUpCGroups(self):
```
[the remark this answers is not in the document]
credp: I got rid of the camelcase all over

tests/eas/st_hold.py
credp: Changed

tests/eas/st_hold.py
```diff
@@ -0,0 +1,409 @@
...
+        defined here as:
+          start_time = first time we switch to the correctly named task
+          end_time   = last time we switch away from the correctly named task
+        """
+        df = self.ftrace_obj.sched_switch.data_frame
+        self.task_start_time = df[df.next_comm == self.task_name].iloc[0].name
+        self.task_end_time = df[df.prev_comm == self.task_name].iloc[-1].name
+
+    @classmethod
+    def avg_cpu_freq(self):
```
qperret: Is it possible to simplify this one a bit ? I found the following under `analysis/frequency_analysis.py`:

```python
    avg_freq = 0
    if len(_df) > 1:
        timespan = _df.index[-1] - _df.index[0]
        avg_freq = area_under_curve(_df['frequency'], method='rect') / timespan
 ```

`area_under_curve` comes from bart.

Comment (credp) [reply tests/eas/st_hold.py]: Maybe, I'll take a look

Comment (credp) [reply tests/eas/st_hold.py]: It doesn't seem to be able to figure out the area under the curve of an arbitrary slice of data, only a whole series - I could use it, but I think I'd have to build a new series using the frequency change events from the original trace and generate start and end events. I'm not convinced it'd be simpler..

Comment (credp) [reply tests/eas/st_hold.py]: yep, none of the simple slicing methods will result in the desired effect. I can get it (I think) by converting timestamps to actual DateTime objects, then reindexing using a new index generated by the intersection of the original index and the new range, but I think I prefer just iterating the frequency change events

Comment (qperret) [reply tests/eas/st_hold.py]: Yeah I was hoping to make our life easier with this but if that's not the case feel free to just ignore that :wink:

Comment (credp) [reply tests/eas/st_hold.py]: Here's an example of recomputing the index, do you think it's cleaner?
```python
import bart

import pandas as pd

# 0.0->0.5 @ 450 -> 450 * 0.5 -> 225
# 0.5->1.5 @ 600 -> 600 * 1.0 -> 600
# 1.5->2.0 @ 900 -> 900 * 0.5 -> 450
# 2.0->2.5 @ 450 -> 450 * 0.5 -> 225
# 2.5->3.0 @ 600 -> 600 * 0.5 -> 300
# total 1800 / 3.0 -> 600 avg
# average between 0.25 and 2.25
# (112.5 + 600 + 450 + 112.5) / 2.0 -> 637.5
t=[0.0,0.5,1.5,2.0,2.5,3.0]
f=[450,600,900,450,600,450]
#series = pd.Series(data=f,index=pd.to_datetime(t,unit='s'))
series = pd.Series(data=f,index=t)
print "Complete Series:"
print series

# calculate the average frequency over a specific time range
start_time = 0.25
end_time   = 2.25
new_index = [ start_time ]+list(series[start_time:end_time].index)+[ end_time ]
filtered_series = series.reindex(index=new_index, method='pad')
print "\nSeries between {} and {}".format(start_time, end_time)
print filtered_series

print "Average Frequency for complete series: {}".format( bart.common.Utils.area_under_curve(series, method='rect') / (t[-1] - t[0]) )
print "Average Frequency for range: {}".format(bart.common.Utils.area_under_curve(filtered_series, method='rect') / (end_time - start_time))
```

```
Complete Series:
0.0    450
0.5    600
1.5    900
2.0    450
2.5    600
3.0    450
dtype: int64

Series between 0.25 and 2.25
0.25    450
0.50    600
1.50    900
2.00    450
2.25    450
dtype: int64
Average Frequency for complete series: 600.0
Average Frequency for range: 637.5
```

qperret: The updated version LGTM :wink:

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+    unboosted_cgroup_name = UNBOOSTED_CGROUP_NAME
+    frequency_margin = FREQ_MARGIN
+    test_cpu = TEST_CPU
+
+    @classmethod
+    def setUpClass(self):
+        self._log = logging.getLogger('StHoldTest')
+        self.params = {}
+        # connect the target
+        self.env = TestEnv(test_conf=test_conf)
```
douglas-raillard-arm: That would be better to have that done using _getTestConf: TestEnv(test_conf=cls._getTestConf())
Otherwise, it will break the result directory creation on the integration branch

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+    @classmethod
+    def run_workload(self):
+        self.env.ftrace.start()
+        self.rtapp.run(out_dir=self.env.res_dir, cgroup=self.cgroup_name, as_root=True)
+        self.env.ftrace.stop()
+        self.env.ftrace.get_trace(self.trace_file)
+        self.env.platform_dump(self.env.res_dir)
+
+    @classmethod
+    def setup_cpuset_cgroup(self, cgroup, cpus, mems):
```
douglas-raillard-arm: First parameter of class methods is usually called "cls" rather than self by convention

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+        cpu_g = self.cpuset_controller.cgroup(cgroup)
+        cpu_g.set(cpus=cpus, mems=mems)
+
+    @classmethod
+    def setup_schedtune_cgroup(self, cgroup, boost, prefer_idle):
```
douglas-raillard-arm: same comment as above on "cls" instead of "self"

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+        cpu_df = df[df.cpu == self.test_cpu]
+        freq_s = cpu_df.frequency
+        old_index = list(freq_s[self.task_start_time:self.task_end_time].index)
+        # the new index should run from the start to the end time, including
+        # any events which occurred in between
+        new_index = [ self.task_start_time ] + old_index + [ self.task_end_time ]
+        windowed_freq_s = freq_s.reindex(index=new_index, method='pad')
+        # avg_freq is area / time for the window of interest
+        area = area_under_curve(windowed_freq_s, method='rect')
+        return int((area + 0.5) / (self.task_end_time - self.task_start_time))
```
douglas-raillard-arm: Since this code is using division, it should use this statement at the beginning:
from __future__ import division
And then the code needs to be adapted so it works properly. That will ensure the same behavior on Python2 and 3 without modifying the code itself

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+FREQ_MARGIN = 20000
+TEST_CPU = 0
+
+class _SchedtuneHoldTest(unittest.TestCase):
+    """
+    Base class for all Schedtune-Hold feature tests. Contains
+    setup code and useful test methods.
+    When a derived class is instantiated, this sequence of events
+    happens:
+      _SchedtuneHoldTest:setUpClass called
```
valschneider: You can write that as ``:meth:`setUpClass` ``, that way once I get around to having the autodoc generated for the tests we'll get a clickable link to the right method :)

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+                    period_ms=40, duty_cycle_pct=1,
+                    duration_s=2, sched={ 'policy': 'FIFO', },
+                    cpus=self.test_cpu,
+                ).get(),
+            },
+            run_dir='/tmp'
+        )
+
+    def test_cpu_frequency(self):
+        self._test_cpu_frequency(lower_freq_tgt=True)
```
douglas-raillard-arm: should probably add "return" in front to propagate whatever is returned by _test_cpu_frequency

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+    Base class for all Schedtune-Hold feature tests. Contains
+    setup code and useful test methods.
+    When a derived class is instantiated, this sequence of events
+    happens:
+      _SchedtuneHoldTest:setUpClass called
+          -> connect target object
+          -> call derived_class:child_check_config() if implemented
+          -> call derived_class:populate_params() (must be implemented)
+          -> run workload
+          -> build Trace object in self.trace_obj
```
valschneider: Ditto here, I think ``:class:`Trace` `` should work

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+    the boost holding.
+    This test uses a 40ms periodic 1% rt task in a 0% boosted
+    group and verifies that schedutil does not raise the cpu
+    frequency to max for the entire duration of the test.
+    """
+    task_name = "1pct_rt_0"
+    cgroup_name = UNBOOSTED_CGROUP_NAME
+
+    @classmethod
+    def setUpClass(self, *args, **kwargs):
```
douglas-raillard-arm: I haven't checked the exact behavior of nosetest but is that needed ? This code can probably be removed.

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+                ).get(),
+            },
+            run_dir='/tmp'
+        )
+
+    @classmethod
+    def child_check_config(self):
+        self.check_hold_config(want_default=False)
+
+    def test_cpu_frequency(self):
```
douglas-raillard-arm: Should probably be a classmethod like all the other ones

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+                self.task_name: Periodic(
+                    period_ms=40, duty_cycle_pct=1,
+                    duration_s=2, sched={ 'policy': 'FIFO', },
+                    cpus=self.test_cpu,
+                ).get(),
+            },
+            run_dir='/tmp'
+        )
+
+    def test_cpu_frequency(self):
```
douglas-raillard-arm: Should be a classmethod

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
```
douglas-raillard-arm: Should be a classmethod

tests/eas/st_hold.py
douglas-raillard-arm: Given that all classes have this method, it should be implemented in the base class, and overriden only when needed to add the extra parameter

tests/eas/st_hold.py
```diff
@@ -0,0 +1,408 @@
...
+          -> call derived_class:child_check_config() if implemented
+          -> call derived_class:populate_params() (must be implemented)
+          -> run workload
+          -> build Trace object in self.trace_obj
+    Then the actual test function is executed on the collected trace.
+
+    child_check_config is used for the two tests present where only one
+    can be run - which one is OK depends upon the status of the
+    SCHEDTUNE_BOOST_HOLD_ALL sched_feature. This function should raise a
+    unittest.SkipTest exception if it wants to prevent a derived class test
```
valschneider: I think for external stuff  ``:class:`unittest.SkipTest` `` should work. I don't think it's as important to have proper links for those though - people usually already know them. It's more about making it easier for people new to LISA's codebase.

tests/eas/st_hold.py
credp: done :)

tests/eas/st_hold.py
credp: ok :)

tests/eas/st_hold.py
credp: This test is derived from unitTest rather than LisaTest, I based it on the capacity capping test. Is this likely to break something in the integration branch as well?

tests/eas/st_hold.py
credp: ok

tests/eas/st_hold.py
credp: ok

tests/eas/st_hold.py
credp: ok

tests/eas/st_hold.py
credp: ok

tests/eas/st_hold.py
credp: ok

PR #619, opened 2018-08-14
Load tracking cpu migration improvements

qais-yousef:
The 2 patches make the {One,Two}Task[s]CPUMigrationTest more stable by reducing some sources of noise that made the tests fail occasionally at some iterations and during one integration test (20180720) fail 100% of the iterations.

I can't root cause why 20180720 integration was worse than others but changes to NFS might have caused more kworker threads to run randomly on the CPUs the test was running on causing utilization to be slightly off. Combined with rounding/quantization errors in calculating util_mean in _get_util_mean() they all were adding up enough to tip the point to a 100% failure rate.

Depending on the next integration test results with these fixes - more investigation might be necessary.

### DIFF
--- a/libs/utils/test.py
+++ b/libs/utils/test.py
@@ -132,12 +132,17 @@ class LisaTest(unittest.TestCase):
         """
 
     @memoized
-    def get_sched_assert(self, experiment, task):
+    def get_sched_assert(self, experiment, task, pid=0):
         """
         Return a SchedAssert over the task provided
         """
-        return SchedAssert(
-            self.get_trace(experiment).ftrace, self.te.topology, execname=task)
+
+        if pid:
+            return SchedAssert(
+                self.get_trace(experiment).ftrace, self.te.topology, pid=pid)
+        else:
+            return SchedAssert(
+                self.get_trace(experiment).ftrace, self.te.topology, execname=task)
 
     @memoized
     def get_multi_assert(self, experiment, task_filter=""):
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -837,6 +837,11 @@ class OneTaskCPUMigrationTest(_CPUMigrationBase):
             Util            Util
     CPU 4:   35%             10%
     CPU 5:   5%              30%
+
+    Conditions
+    ==========
+    The CPU[s] on the which the test is running must be reasonably quiet or the
+    measured util_mean will be skewed by any creeping tasks/kthreads/softirqs.
     """
     @classmethod
     def _getExperimentsConf(cls, test_env):
@@ -903,6 +908,11 @@ class TwoTasksCPUMigrationTest(_CPUMigrationBase):
             Util            Util
     CPU 4:   20%             50%
     CPU 5:   50%             20%
+
+    Conditions
+    ==========
+    The CPU[s] on the which the test is running must be reasonably quiet or the
+    measured util_mean will be skewed by any creeping tasks/kthreads/softirqs.
     """
     @classmethod
     def _getExperimentsConf(cls, test_env):
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -694,7 +694,7 @@ class _CPUMigrationBase(LisaTest):
                              start + phase_duration_s)
         return window
 
-    def _get_util_mean(self, start, end, df, cpus):
+    def _get_util_mean(self, experiment, phase, start, end, df, cpus):
         """
         Compute the mean utilization per CPU given a time interval
         :param start: start of the time interval
@@ -715,7 +715,26 @@ class _CPUMigrationBase(LisaTest):
         util_mean = {}
         df = df[start:end]
         for cpu in cpus:
-            util = df[(df.cpu == cpu) & (df.path == '/')].util
+            util = df[(df.cpu == cpu) & (df.path == '/')]
+            for task in util['__comm'].unique():
+                if task == "<idle>" or task in self.tasks_desc.keys():
+                    continue
+
+                # The same task might appear under several pids, calculate
+                # the total percentage of time for this task
+                pids = util[(util['__comm'] == task)]['__pid'].unique()
+                creep_time_pct = 0
+                for pid in pids:
+                    sched_assert = self.get_sched_assert(experiment, task, pid)
+                    creep_time_pct += sched_assert.getRuntime((start, end), True)
+
+                # Record any tasks that creeped in during the test and the
+                # percentage of time they ran
+                msg = "task {} creeped into CPU{} for {}% during phase {}"
+                msg = msg.format(task, cpu, creep_time_pct, phase)
+                self._log.info(msg)
+
+            util = util.util
             util_mean[cpu] = area_under_curve(util) / (end - start)
         return util_mean
 
@@ -776,7 +795,8 @@ class _CPUMigrationBase(LisaTest):
                 cpus.update(self._get_cpus(task[1], phase))
 
             # Get the mean utilization per CPU
-            util_mean = self._get_util_mean(window[phase][0], window[phase][1],
+            util_mean = self._get_util_mean(experiment, phase,
+                                            window[phase][0], window[phase][1],
                                             util_df, cpus)
             # Get the expected utilization per CPU
             expected = self._get_util_expected(self.tasks_desc, cpus, phase)
--- a/tests/eas/load_tracking.py
+++ b/tests/eas/load_tracking.py
@@ -39,6 +39,8 @@ UTIL_SCALE = 1024
 UTIL_AVG_CONVERGENCE_TIME = 0.3
 # Allowed margin between expected and observed util_avg value
 ERROR_MARGIN_PCT = 15
+# Allowed error margin % for migration tests
+MIGRATION_ERROR_MARGIN_PCT = 2
 # PELT half-life value in ms
 HALF_LIFE_MS = 32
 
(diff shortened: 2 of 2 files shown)


tests/eas/load_tracking.py
```diff
@@ -692,7 +692,39 @@ def _get_stable_window(self, df, task):
...
+        for cpu in cpus:
+            tasks_count = 0
+            idle_pct = 0
+            for task in tasks.iteritems():
+                [phase_cpu] = self._get_cpus(task[1], phase)
+                if phase_cpu == cpu:
+                    idle_pct += 100 - self._get_duty_cycle(task[1], phase)
+                    tasks_count += 1
+
+            if tasks_count == 0:
```
douglas-raillard-arm: Duck-typing-friendly practice is to never compare to 0.
In that case, the if statement can be simplified into:
```
tasks_count = tasks_count or 1
```

qais-yousef:
If we keep the period values small enough the error in calculations are 'acceptable'/predictible.
I will change the content of this pull request to detect tasks that creep into the CPU while performing the test which is also a known factor of causing false failures. Keeping the system quite will be difficult since kthreads are seen to run randomly on the CPUs we perform the tests on.

qais-yousef:
The updated code will look at any tasks that 'creeped' into the cpu while the test is running. The assert is commented out to help us finalise what would be an acceptable MIGRATION_ALLOWED_INTERFERENCE_PCT to set by collecting more data first.

tests/eas/load_tracking.py
```diff
@@ -715,7 +724,26 @@ def _get_util_mean(self, start, end, df, cpus):
...
+                    # the total percentage of time for this task
+                    pids = util[(util['__comm'] == task)]['__pid'].unique()
+                    creep_time = 0
+                    for pid in pids:
+                        sched_assert = self.get_sched_assert_pid(experiment, pid)
+                        creep_time += sched_assert.getRuntime(window=(start, end), percent=True)
+
+                    # Make sure that any other processes that creeped into the
+                    # cpu we are testing on are insignificant
+                    msg = "task {} creeped into CPU{} for {}% during {}".format(task, cpu, creep_time, phase)
```
derkling: `"... during Phase[{}]"` ?
... just to be explicit we are reporting a phase name?

tests/eas/load_tracking.py
```diff
@@ -776,7 +804,8 @@ def _test_util_per_cpu(self, experiment, tasks):
                 cpus.update(self._get_cpus(task[1], phase))
 
             # Get the mean utilization per CPU
-            util_mean = self._get_util_mean(window[phase][0], window[phase][1],
+            util_mean = self._get_util_mean(experiment, phase,
```
derkling: As a general comment, I would say that at the end we are interested in the overall pct of creepy tasks showing up in the entire trace and I also don't know how much interest we can have in the phase info.

Thus, I would say that we could consider to avoid entirely to pass `experiment` and `phase` to  `_get_util_mean` and run our sanity checks even before computing the mean. Consider that, if the mean is within margins but we have an high noise level... still we cannot trust it and the test should fail anyway.

What do you think?

tests/eas/load_tracking.py
```diff
@@ -41,6 +42,14 @@
...
 # Allowed error margin % for migration tests
 MIGRATION_ERROR_MARGIN_PCT = 2
+# When running migration tests intererence from kthreads could skew the results
+# which we try to detect. This threshold defines the acceptable percentage of
+# intererence that we know won't skew the results enough.
+# Sadly it's hard to make the kthreads completely quite or prevent from running
+# randomly on the CPUs we are running the tests on, and these tests are very
+# sensitive to this kind of 'creep' because of the tight error margin we would
+# like to keep.
+MIGRATION_ALLOWED_INTERFERENCE_PCT = 0.5
```
derkling: `s/_INTERFERENCE_/_NOISE_` ?

tests/eas/load_tracking.py
```diff
@@ -41,6 +42,14 @@
 ERROR_MARGIN_PCT = 15
 # Allowed error margin % for migration tests
 MIGRATION_ERROR_MARGIN_PCT = 2
+# When running migration tests intererence from kthreads could skew the results
```
derkling: This test is quite good... we should also probably add an explicit section in the documentation of each test class to clearly state what are the _expected invariants_ of its tests. In this case: we do not expect other tasks running on the target CPUs for a percentage of time bigger then the configured threshold.

libs/utils/test.py
```diff
@@ -139,6 +139,14 @@ def get_sched_assert(self, experiment, task):
         return SchedAssert(
             self.get_trace(experiment).ftrace, self.te.topology, execname=task)
 
+    @memoized
+    def get_sched_assert_pid(self, experiment, pid):
```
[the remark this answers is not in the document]
qais-yousef: Yes that would be better.

tests/eas/load_tracking.py
qais-yousef: OK I'll update the test docs.

tests/eas/load_tracking.py
qais-yousef: OK

tests/eas/load_tracking.py
```diff
@@ -715,7 +724,26 @@ def _get_util_mean(self, start, end, df, cpus):
         util_mean = {}
         df = df[start:end]
         for cpu in cpus:
-            util = df[(df.cpu == cpu) & (df.path == '/')].util
+            util = df[(df.cpu == cpu) & (df.path == '/')]
+            for task in util['__comm'].unique():
+                if task != "<idle>" and not task in self.tasks_desc.keys():
```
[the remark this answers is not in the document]
qais-yousef: :+1:

tests/eas/load_tracking.py
```diff
@@ -715,7 +724,26 @@ def _get_util_mean(self, start, end, df, cpus):
...
+
+                    # The same task might appear under several pids, calculate
+                    # the total percentage of time for this task
+                    pids = util[(util['__comm'] == task)]['__pid'].unique()
```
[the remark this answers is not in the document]
qais-yousef: OK I'll try that and if it returns something equivalent I'll use it.

tests/eas/load_tracking.py
```diff
@@ -715,7 +724,26 @@ def _get_util_mean(self, start, end, df, cpus):
...
+                    creep_time = 0
+                    for pid in pids:
+                        sched_assert = self.get_sched_assert_pid(experiment, pid)
+                        creep_time += sched_assert.getRuntime(window=(start, end), percent=True)
+
```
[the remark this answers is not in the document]
qais-yousef: :+1:

tests/eas/load_tracking.py
qais-yousef: because phase is the unit of test. The tests are done in steps of phases. So we catch the noise during that step/phase and report if we caught an issue in that phase.

tests/eas/load_tracking.py
qais-yousef: I think we will lose a lot of accuracy in this case.

For example if we have 2 phases and a noise of 0.3% creeps into each phase, it's overall small for each phase and the test should run smoothly. If we have a noise margin of 0.6% each phase will be happy to deal with that. But the combined noise is 0.6% which can look high and we could prematurely decide it's a failure.

Similarly if we increase the margin to be for instance 0.9% since we are looking at the noise value in the entire test duration, then if we have a per phase noise levels of 0.6% + 0.2% then the overall test will say things are OK, but 0.6% at phase one is too high and will cause false negatives.

Makes sense?

tests/eas/load_tracking.py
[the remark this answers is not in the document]
qais-yousef: It seems to me self.get_trace() will cause the trace to be parsed again and generate the optimized map or did I misread the code? In this case we are not really speeding things up but doing more work

valschneider:
In the upcoming API rework I've straight up murdered load_tracking and let it be reborn from its ashes, so the integration of these changes might be a bit different. In any case, I'm not ignoring this, but it'll have to wait for the new stuff to drop in first.

qais-yousef:
Or just incorporate them as part of your changes in one go? I don't mind either way. Assuming the new API rework is going to be merged soon enough though?

valschneider:
Mm well it'll appear in a new branch for some time, we could merge that in the meantime for our integration tests (although from what I've seen in my rework load_tracking is kinda broken anyway).

qais-yousef:
I don't mind. They helped identify some sources of noise that were affecting this test (systemd not been completely frozen) - but if we move to buildroot this will be less of an a problem. So again I don't mind either way :)

qais-yousef:
This is irrelevant now we have buildroot and we're moving to lisa/next. So I am closing it.

douglas-raillard-arm:
For future reference: if somebody wants to implement that, there is now support for "undecided" test result that was introduced in lisa-next to allow tests to have a 3rd choice rather than just failure and success.

valschneider:
> For future reference: if somebody wants to implement that, there is now support for "undecided" test result that was introduced in lisa-next to allow tests to have a 3rd choice rather than just failure and success.

Speaking of that, I was thinking that we could turn @qais-yousef's sanity checking into actual test helpers that would force the result to be "undecided". Something like listing the name of tasks that are part of the test (or some regexp), and if something else ran for more than x threshold then we'd go undecided.

Leaving that here, in case someone wants to have a go at it ;)

douglas-raillard-arm:
We could have that as a decorator for test methods indeed. Task names to track can be read from the `rtapp_profile` attribute of the `TestBundle`, and the trace from the `trace` attribute. The `ResultBundle` can be intercepted and updated to `Result.UNDECIDED` and `self.get_logger().info()` to make the "magic" visible from the log.

qais-yousef:
There's certainly a value in doing that. Though I think trying to tighten the knots to ensure the conditions under which the tests are running are correct is going to be somewhat of a rabbit hole and potentially an energy consuming effort - both in tightening the knots side and dealing with failures due to loose knots.

But since I already did this work, I'd be happy to port it to lisa/next if any of you would like it be ported :-)

PR #621, opened 2018-08-28
error message in lisa_shell is not very instructive, /sbin/ifconfig might not be installed

v0lker:
modern linux systems don't have /sbin/ifconfig installed by default (but /sbin/ip, instead)

### DIFF
--- a/src/shell/lisa_shell
+++ b/src/shell/lisa_shell
@@ -188,9 +188,19 @@ function _lisa-ipython-usage {
 
 function _lisa-ipython-start {
 # Get IP address for the specified interface
-IPADDR=$(/sbin/ifconfig $NETIF 2>/dev/null  | \
-    awk '/inet / {print $2}' | \
-    sed 's/addr://')
+IPADDR=
+
+if [ -x /sbin/ifconfig ]; then
+    IPADDR=$(/sbin/ifconfig $NETIF 2>/dev/null  | \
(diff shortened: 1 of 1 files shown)


src/shell/lisa_shell
```diff
@@ -188,9 +188,19 @@ function _lisa-ipython-usage {
...
-    sed 's/addr://')
+IPADDR=
+
+if [ -x /sbin/ifconfig ]; then
+    IPADDR=$(/sbin/ifconfig $NETIF 2>/dev/null  | \
+        awk '/inet / {print $2}' | \
+        sed 's/addr://')
+fi
+
+if [ "x$IPADDR" == "x" -a -x /sbin/ip ]; then
```
qperret: According to [this](http://tldp.org/LDP/abs/html/fto.html), `-a` is now deprecated and should be replaced by `-e`

v0lker: "this" is assuming that it is a file test operator:
http://tldp.org/LDP/abs/html/fto.html

but in this case it is supposed to be a logical and:

"expr1 -a expr2
True if both expr1 and expr2 are true."

(from bash man page)

qperret: Hrmpf, bash always confuses the heck out of me ... Out of curiosity, is there any benefit of `-a` over the good old `&&` ?

PR #622, opened 2018-08-29
init_env: do not replace the running shell. instead start a new one, …

v0lker:
…so that on exiting lisa, we go back to the shell.


v0lker:
ahhh. ok, i'll make that conditional on currently running interactively...

PR #623, opened 2018-08-29
init_env: do not replace the running shell. instead start a new one, …

v0lker:
…so that on exiting lisa, we go back to the shell. only do this if we were running interactively in the first place.


v0lker:
weirdly, travis seems to run the tests in a way that they seem to run interactively, so the CI test times out at the LISA prompt.

PR #624, opened 2018-08-30
some issues with two examples from the tutorials directory

v0lker:
tutorial/00_LisaInANutshell: should be trying to set ENERGY_AWARE scheduling as root, otherwise permission will be denied (i haven't tested the rest of the file)

tutorial/UseCaseExamples_SchedTuneAnalysis: was not using the path that was set to load the trace file and the ASCII trace file is missing, therefore need to generate it

there are more issues, at least with the latter one.

### DIFF
--- a/ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
+++ b/ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
@@ -183,9 +183,10 @@
    ],
    "source": [
     "import json\n",
+    "import os\n",
     "\n",
     "# Load the platform information\n",
-    "with open('../../results/SchedTuneAnalysis/platform.json', 'r') as fh:\n",
+    "with open(os.path.join(res_dir, 'platform.json'), 'r') as fh:\n",
     "    platform = json.load(fh)\n",
     "print \"Platform descriptio collected from the target:\"\n",
     "print json.dumps(platform, indent=4)"
--- a/ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
+++ b/ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
@@ -568,8 +568,8 @@
    "source": [
     "# Get a list of first 5 \"sched_load_avg_events\" events\n",
     "sched_load_avg_events = !(\\\n",
-    "    grep sched_load_avg_task {trace_file.replace('.dat', '.txt')} | \\\n",
-    "    head -n5 \\\n",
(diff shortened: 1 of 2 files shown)


ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
```diff
@@ -183,9 +183,10 @@
    ],
    "source": [
     "import json\n",
+    "import os\n",
     "\n",
     "# Load the platform information\n",
-    "with open('../../results/SchedTuneAnalysis/platform.json', 'r') as fh:\n",
+    "with open(res_dir + os.sep + 'platform.json', 'r') as fh:\n",
```
derkling: Instead of hardcoded concatenations, we preferably use:
```python
os.path.join(res_dir, 'platform.json')
```

ipynb/tutorial/UseCaseExamples_SchedTuneAnalysis.ipynb
```diff
@@ -568,8 +568,8 @@
    "source": [
     "# Get a list of first 5 \"sched_load_avg_events\" events\n",
     "sched_load_avg_events = !(\\\n",
-    "    grep sched_load_avg_task {trace_file.replace('.dat', '.txt')} | \\\n",
```
derkling: Good catch... I think that's related to:
   https://github.com/ARM-software/trappy/pull/270
... and specifically this commit:
   https://github.com/ARM-software/trappy/pull/270/commits/159f149e8d8e431d44b875e1785532d06670c993
from mr. @valschneider 

Before that we used to generate and keep a txt file every time Trappy pared a trace file.
We then optimised that by avoiding to keep the `txt` file around since we have both the original file... and also a local cache for already parsed events.

v0lker:
i hope i addressed the concerns - ?

v0lker:
it's been two months and nobody seems to be using the tutorial. in that case it's probably better if the tutorials were deleted?

valschneider:
Hi @v0lker, 

Sorry about that. Notebooks are quite a pain to maintain because you can't really add a Travis job to make sure they still work (you *can* translate them to a Python script, but then you have nothing to check that e.g. the plots got rendered correctly).

Some of those Notebooks serve a dual purpose - both API documentation and tutorial/example code. This stems from the fact that currently our API documentation is in a poor shape, and that sometimes writing pure code is simpler than writing text...

A non negligible work that remains to be done on https://github.com/ARM-software/lisa/tree/next is to do a summer cleaning of those notebooks - remove redundant ones, remove those serving as pure API documentation, and only keep a few useful example ones. The rest should live in the documentation, and that can *actually* be CI-tested via https://docs.python.org/3.5/library/doctest.html.

PR #625, opened 2018-09-05
utils/energy_model: Explicitize EM loader type

valschneider:
The logger currently uses the class method name to log which EM was
loaded, which isn't very explicit.

Instead, use some more human-friendly naming.

### DIFF
--- a/libs/utils/energy_model.py
+++ b/libs/utils/energy_model.py
@@ -1042,13 +1042,17 @@ class EnergyModel(object):
         # em_loaders dictionary joins EM loaders and the identifying functions
         # with any associated metadata
         em_loaders = {
-            'sd'    : { 'check': em_present_in_sd,
-                        'load': cls.from_sd_target,
-                        'filename': '/proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/{}' },
-            'sysfs' : { 'check': simplified_em_present_in_cpusysfs,
-                        'load': cls.from_simplifiedEM_target,
-                        'directory': '/sys/devices/system/cpu/energy_model' }
-                     }
+            'sched domain' : {
+                'check': em_present_in_sd,
+                'load': cls.from_sd_target,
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
As agreed when this code was merged, it needs to be replaced with a class instead of a custom re-implementation of a class system based on dictionaries. No edit to that code can be merged until that cleanup is done.

libs/utils/energy_model.py
```diff
@@ -1042,13 +1042,17 @@ def simplified_em_present_in_cpusysfs(target, directory=None):
...
         # with any associated metadata
         em_loaders = {
-            'sd'    : { 'check': em_present_in_sd,
-                        'load': cls.from_sd_target,
-                        'filename': '/proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/{}' },
-            'sysfs' : { 'check': simplified_em_present_in_cpusysfs,
-                        'load': cls.from_simplifiedEM_target,
-                        'directory': '/sys/devices/system/cpu/energy_model' }
-                     }
+            'sched domain' : {
```
douglas-raillard-arm: these dict can be replaced by classes to get rid of some boilerplate

libs/utils/energy_model.py
```diff
@@ -1042,13 +1042,17 @@ def simplified_em_present_in_cpusysfs(target, directory=None):
...
+                'check': em_present_in_sd,
+                'load': cls.from_sd_target,
+                'filename': '/proc/sys/kernel/sched_domain/cpu{}/domain{}/group{}/energy/{}'
+            },
+            'CPU' : {
```
qperret: 'CPU sysfs' ? `'Loading energy model from CPU data` doesn't great as-is IMO. Or ideally just log the directory name ...

valschneider:
I had forgotten the discussion about turning those into classes. I was doing a documentation/logging cleanup and this seemed like an easy win, but I guess I can keep that for a later rework...

PR #626, opened 2018-09-11
env: Freeze all systemd-* cmd except PID 1

douglas-raillard-arm:
Since all commands part of the systemd package are named
systemd-<something>, we need a more refined pattern in order to properly
freeze systemd-journald process for example.

### DIFF
--- a/libs/utils/env.py
+++ b/libs/utils/env.py
@@ -160,7 +160,9 @@ class TestEnv(ShareState):
     critical_tasks = {
         'linux': [
             'init',
-            'systemd',
+            # We want to freeze everything except PID 1, we don't want to let
+            # sysmted-journald or systemd-timesyncd running.
+            'systemd[^-]',
             'dbus',
             'sh',
             'ssh',


douglas-raillard-arm:
Will be tested during upcoming integration cycle

valschneider:
One issue I'm having with this (and which seems recurrent, I've encountered something similar maybe a year ago when trying to freeze more stuff) is that tracing takes *much* longer.

We're talking ~20 seconds added to each trace, where nothing happens and we're just waiting for the tracing to stop.

I don't know where this comes from exactly, but I'd really like to mitigate this.

valschneider:
@douglas-raillard-arm What came out of the integration testing? Is it worth the much longer trace time or should we ditch it?

douglas-raillard-arm:
This commit is still required to have the best pass rate on a systemd-based system. However, I am transitioning to the buildroot based system, so it may matter less now.

valschneider:
I'm marking this as stale since we haven't done much with it in a while - what's the results with buildroot?

douglas-raillard-arm:
No issues with buildroot, that issue has not been investigated yet an is kind of at the bottom of the priority list, since tests still works and we don't really use sytemd-based distro anymore for now (Buildroot).

PR #627, opened 2018-09-14
[RFC] Load tracking tests rework

johnny90:
Refactor of frequency invariance, cpu invariance and PELT tests to the new API. Probably the load tracking file should contain only these, and cgroups tests should be under a separate file.


valschneider:
I've integrated all of that in my local branch, coming to your nearby LISA mirror soon™.

PR #628, opened 2018-09-18
lisa_shell: add a new function to generate buildroot rootfs

qais-yousef:
buildroot rootfs are a better choice to run lisa tests on as they are
quieter and provide a minimal userspace that allows testing kernel
functionality with minimum interference from userland. Also it is less
of a maintanence burden as it's unlikely to find unexpected things
configured without our knowledge that might intefere with some of the
tests or skew their results.

A recent example was found in systemd indirectly causing higher number
of kthreads to run on debian and arch-linux distro's which caused random
number of failure rates in cpu migration load tracking tests.


src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+set -eu
+
+BUILDROOT_URI=git://git.busybox.net/buildroot
+BUILDROOT_VERSION=2018.08
+BUILDROOT_DIR=$LISA_HOME/src/buildroot
+BUILDROOT_PATCHES_DIR=$(dirname $0)/buildroot_patches
```
douglas-raillard-arm: $0 needs quotes: "$0", otherwise word splitting will occur and you could end up passing multiple parameters to dirname, each getting their output line in dirname output.

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+BUILDROOT_CONFIG=$(dirname $0)/buildroot_config
```
douglas-raillard-arm: ditto

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+
+while getopts "hp" opt
+do
+	case $opt in
+		h)
+			print_usage
+			exit 0
+			;;
+		p)
+			rm -rf $BUILDROOT_DIR
```
douglas-raillard-arm: $BUILDROOT_DIR needs to be double-quoted to avoid parameter expansion and word splitting, especially on rm commands since they can be disastrous.

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+		?)
+			print_usage
+			exit -1
+			;;
+	esac
+done
+
+# Executed command @$1 if file @$2 doesn't exist
+function do_if {
+	if [ ! -e $2 ]; then
```
douglas-raillard-arm: [[ ]] bashism is preferred over simple [ ] since no word splitting or pathname expansion is performed. Note that at the end of the day, double quotes would still be needed because:

> tilde expansion, parameter and variable expansion, arithmetic expansion, command substitution, process substitution, and quote removal are performed.
https://linux.die.net/man/1/bash

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+		$1
```
douglas-raillard-arm: Without quotes, bash will perform some expansions a second time, which could be confusing

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+		touch $2
```
douglas-raillard-arm: Needs double quote to ensure only one parameter is passed to touch

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+
+function br_checkout_version {
+	pushd $BUILDROOT_DIR > /dev/null
+	git checkout -b $BUILDROOT_VERSION $BUILDROOT_VERSION
+	popd > /dev/null
+}
+
+function br_patch {
+	pushd $BUILDROOT_DIR > /dev/null
+	for patch in $(ls $BUILDROOT_PATCHES_DIR)
```
douglas-raillard-arm: Would benefit from using shell globing to avoid parsing ls output:
 	for patch in "$BUILDROOT_PATCHES_DIR/*"
https://unix.stackexchange.com/questions/128985/why-not-parse-ls

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+do_if br_patch $BUILDROOT_DIR/.lisa_patched
+
+do_if br_apply_config $BUILDROOT_DIR/.lisa_applied_config
+
+make -C $BUILDROOT_DIR
+
+# Generate ssh-keys so that they persist across multiple runs
+ssh-keygen -A -f $BUILDROOT_DIR/output/target/
+# Enable root login on sshd
+sed -i 's/#PermitRootLogin.*/PermitRootLogin	yes/' $BUILDROOT_DIR/output/target/etc/ssh/sshd_config
```
douglas-raillard-arm: Is matching the "#" necessary (.* instead) ? If the conf file was changed to "PermitRootLogin no", not matching the # would still work

src/shell/lisa_shell
```diff
@@ -527,6 +527,14 @@ $WLTEST_HOME/test_series "$@"
 deactivate
 }
 
+################################################################################
+# LISA Shell buildroot rootfs generation functions
+################################################################################
+
+function lisa-gen-buildroot-rootfs {
+	$LISA_HOME/src/shell/lisa_gen_buildroot_rootfs $@
```
douglas-raillard-arm: double quotes around "$@" are needed
https://stackoverflow.com/questions/37141039/is-there-any-difference-between-and

src/shell/lisa_gen_buildroot_rootfs
douglas-raillard-arm: nit: $BASH_SOURCE is a handy replacement for $0, and has the advantage of always pointing to the current file even when the script is sourced (but it is obviously not POSIX).

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: How could the expansion happen for a second time?

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: Yes it is necessary otherwise we'll replace every occurenance in the file instead.

If the file was changed manually then we probably shouldn't be touching it. I don't think it's nice to override a user change behind their back - it should provide something that works out of the box, and if the user does some other  customisations on top it's better not to undo them.

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: Yeah we are not guaranteed that the user won't have put lisa in a directory with spaces in the name - I'll fix all similar occurrences up

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+	popd > /dev/null
+}
+
+function br_apply_config {
+	cp $BUILDROOT_CONFIG $BUILDROOT_DIR/.config
+	make -C $BUILDROOT_DIR olddefconfig
+}
+
+
+do_if br_clone $BUILDROOT_DIR/.lisa_cloned
```
derkling: I find this `do_if` a bit overkilling and also difficult to undestand without going back to the comment.

What about just using this pattern:
```bash
[[  -f "$BUILDROOT_DIR/.lisa_cloned" ]] || br_clone
```
?

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+	case $opt in
+		h)
+			print_usage
+			exit 0
+			;;
+		p)
+			rm -rf $BUILDROOT_DIR
+			exit 0
+			;;
+		?)
```
derkling: Is not `*)` usually used as default statement ?

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+function br_clone {
+	git clone $BUILDROOT_URI $BUILDROOT_DIR
+}
+
+function br_checkout_version {
+	pushd $BUILDROOT_DIR > /dev/null
+	git checkout -b $BUILDROOT_VERSION $BUILDROOT_VERSION
+	popd > /dev/null
```
derkling: Unless you want to log push/pop folders... you can just have:
 `$(cd $BUILDROOT_DIR ; git checkout -b $BUILDROOT_VERSION $BUILDROOT_VERSION)`

src/shell/lisa_gen_buildroot_rootfs
derkling: nit-pick: let's save one line by keeping the do above: `for (); do` ...

derkling: Maybe `ls $BUILDROOT_PATCHES_DIR/*.patch` just to be more safe in case we should add some doc there in the future?

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
...
+make -C $BUILDROOT_DIR
+
+# Generate ssh-keys so that they persist across multiple runs
+ssh-keygen -A -f $BUILDROOT_DIR/output/target/
+# Enable root login on sshd
+sed -i 's/#PermitRootLogin.*/PermitRootLogin	yes/' $BUILDROOT_DIR/output/target/etc/ssh/sshd_config
+
+make -C $BUILDROOT_DIR
+
+echo "---===::: rootfs was created at $BUILDROOT_DIR/output/images/rootfs.cpio.xz :::===---"
```
derkling: If you wanna play with some colors we have escape defined in `source src/shell/lisa_colors`... to be used as we use them in `src/shell/lisa_shell`

derkling: I would split the status lien and the output product on different lines for better readability, i.e. something like:
```bash
echo "---===::: RootFS Created :::===---"
echo "Path: $BUILDROOT_DIR/output/images/rootfs.cpio.xz"
```

src/shell/lisa_gen_buildroot_rootfs
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+set -eu
+
+BUILDROOT_URI=git://git.busybox.net/buildroot
+BUILDROOT_VERSION=2018.08
+BUILDROOT_DIR=$LISA_HOME/src/buildroot
```
derkling: How do we get buildroot in that folder ?

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: drat; buildroot (or maybe make) doesn' thandle the space in the path anyway - oh well there's so much one can do \o/

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: I've reworked that part to avoid any confusion

src/shell/lisa_gen_buildroot_rootfs
derkling: Can't find where you generate the `.lisa_cloned` token... can't we just use a file which we know existing into the cloned repo, e.g. a makefile.

src/shell/lisa_shell
```diff
@@ -527,6 +527,14 @@ $WLTEST_HOME/test_series "$@"
 deactivate
 }
 
+################################################################################
+# LISA Shell buildroot rootfs generation functions
+################################################################################
+
+function lisa-gen-buildroot-rootfs {
```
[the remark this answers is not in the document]
qais-yousef: I don't mind but I see the long name self documenting and descriptive - with autocomplete typing it shouldn't be a problem..

I'll update LisaShell.txt

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: do_if() touches the file that we pass as a paramter - I can try to rework this to make it clearer. I see the [[]] || command && command; a good short hand but IMHO opinion is also unreadable in this case as you'll pack too much into one line;

I'll see what I can do

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: Yes usually. This is the form I found most examples using so I stuck to it too.

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: ditto. I removed the concept of patching buildroot as we should really do few configs on top only

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: yeah colours would be good, i'll play with that

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: we clone it passing it as the path to clone to

derkling: Yes right... should have cancelled this when I found it out. :)

src/shell/lisa_gen_buildroot_rootfs
douglas-raillard-arm: A first expansion will take place when calling `do_if`, and the second one will take place inside `do_if` when evaluating $1: If you pass the single-quoted parameter `'echo *'` to `do_if`, it should fail since that is not a command name, but in our case it will actually go and expand it. If we intend to support parameters on the function, swapping the file and the command would be a better idea instead of relying on this double expansion:
```
 function do_if {
# usage: do_if FILENAME CMD [[[PARAM1] PARAM2] .....]]
	if [ ! -e "$1" ]; then
                shift
		"$@"
```

src/shell/lisa_gen_buildroot_rootfs
douglas-raillard-arm: Agreed

src/shell/lisa_gen_buildroot_rootfs
qais-yousef: I've enhanced the description of the function in the newer version

```
# Execute function @$1 once
function do_once {
	FILE="$BUILDROOT_DIR/.lisa_$1"
	if [ ! -e "$FILE" ]; then
		$1
		touch "$FILE"
	fi
}
```

it's not meant to be a generic function and it's meant as a little wrapper to reduce the clutter of repetitive code. We can completely remove it if it's confusing. I didn't quote $1 in the newer version as I couldn't see how double expansion could happen from passing the function name. Don't mind to add the quotes to future proof it.

I would prefer not to over generalize the function though.

src/shell/lisa_gen_buildroot_rootfs
derkling: If they are really the same (don't know) then perhaps just check what we most use in other LISA scripts. I guess that consistency in that case could be a way to avoid confusion (and the same question I raised before) in the future.

src/shell/lisa_gen_buildroot_rootfs
derkling: Fine then, ok for me. Maybe a minor nit, I think in LISA we usually write `>/dev/null` (i.e. no space before the redirect)/// but worth a check and a change just if it's really a significant consistency diff.

src/shell/lisa_gen_buildroot_rootfs
derkling: Right, well... I think I was assuming that you always call a function in the `||` case, and in that case you can just move the touch within that function. Actually, if that was the case, perhaps something like this pattern would have worked just fine:
```bash
function fun1()
    GUARDFILE="$BUILDROOT_DIR/<WHATEVER_GUARD_FILE_MAKES_SENSE"
    [[  ! -f "$BUILDROOT_DIR/.touched_file" ]] || return
    touch $GUARDFILE
    # Function business fallowing...
```

But... your new do_once function seems to actually factor off the pattern above in a common function... and I like it more then your previous proposal and mine above.

derkling:
Few possible nit-picks... but that looks fine to me. I don't merge it because I did not actually tested it.

qais-yousef:
Thanks for all the reviews. I believe I addressed all comments and will update the branch shortly.

Certainly someone else other than me needs to try that before it gets merged.

qais-yousef:
Hmm do we care about adding license header here similar to other files?

derkling:
> Hmm do we care about adding license header here similar to other files?

Mmm... good point!... Would say it does not arm to add it. Go for it please.

douglas-raillard-arm:
It would be nice to add udhcp command to the init script, so the board gets an IP when it boots. Since there is no persistence, it's easier if that can be done in the base rootfs that is generated.

qais-yousef:
As per douglas request, added a new function to help updating kernel config to use the new rootfs which would help with automation.

qais-yousef:
Now this is superseded by #654

PR #631, opened 2018-10-17
wa_results_collector.py: fix test_ids being None

qais-yousef:
When using the sched-evaluation-full.ipynb a weird exception about
unable to parse NaN values is thrown. This happens when running
example-jankbench.yaml agenda.

It turns out that since this agenda doesn't define test_ids, the code
tries to read the property anyway and sets it to None - which later
causes the exception when read from the ipython notebook.

Always check that the property is not empty before setting it - and
fallback to the workload name by default.

### DIFF
--- a/libs/utils/wa_results_collector.py
+++ b/libs/utils/wa_results_collector.py
@@ -289,23 +289,20 @@ class WaResultsCollector(object):
             # of the full key=value pairs.
             classifiers = job['classifiers'] or {}
 
+            test = "{}".format(workload)
+
             if 'test' in classifiers:
                 # If the workload spec has a 'test' classifier, use that to
                 # identify it.
-                test = classifiers.pop('test')
+                test = classifiers.pop('test') or test
             elif 'test' in job['workload_parameters']:
                 # If not, some workloads have a 'test' workload_parameter, try
                 # using that
-                test = job['workload_parameters']['test']
+                test = job['workload_parameters']['test'] or test
             elif 'test_ids' in job['workload_parameters']:
                 # If not, some workloads have a 'test_ids' workload_parameter, try
                 # using that
-                test = job['workload_parameters']['test_ids']
-            else:
-                # Otherwise just use the workload name.
-                # This isn't ideal because it means the results from jobs with
-                # different workload parameters will be amalgamated.
-                test = workload
+                test = job['workload_parameters']['test_ids'] or test
 
             rich_tag = ';'.join('{}={}'.format(k, v) for k, v in classifiers.iteritems())
             tag = classifiers.get('tag', rich_tag)


libs/utils/wa_results_collector.py
```diff
@@ -292,14 +292,26 @@ def _read_wa_dir(self, wa_dir):
...
+                if not test:
+                    test = workload
             elif 'test_ids' in job['workload_parameters']:
                 # If not, some workloads have a 'test_ids' workload_parameter, try
                 # using that
                 test = job['workload_parameters']['test_ids']
+
+                # Use the workload name if not set; see comment in else branch
+                if not test:
+                    test = workload
```
valschneider: Sorry for the stall on this one, I unconsciously tend to avoid wltests PRs since I'm not very fluent in it.

Could we clean the handling a bit?

```python
# Use the workload name as a default fallback.
# This isn't ideal because it means the results from jobs with
# different workload parameters will be amalgamated.
test = workload

if 'test' in classifiers:
    test = classifiers.pop('test') or test
elif x in y:
    test = job[x] or test
[...]
```

qais-yousef: This is less readable for me - but it could be only me.

will classifiers.pop() remove the item from the list too or just read it?

qais-yousef:
I think example-jankbench.yaml is meant to trigger everything - so I think it's more maintainable like it is now.

Even if we fix the agenda - I think we need to handle this case here. By either silently fixing it like I did or shout loudly and error. Although the latter is very aggressive in my opinion and we should carry on with the test since we really can.

If we can get the test_id through another mechanism that would be even better.

libs/utils/wa_results_collector.py
valschneider: It removes the item. I don't know why it's done here, I was just commenting on the look of the change.

This was mostly about avoiding a
```python
if not test:
    test = workload
```
in every single condition.


If you don't like the or'ing, you could go with something like this:

```python
test = None

if 'test' in classifiers:
    test = foo()
elif x in y:
    test = bar()
[...]

if not test:
    # blurb about workload name here
    test = workload
```

derkling:
You set the `test_id` to the same value, the `workload` name, for all tests: isn't `workload` going to be always "jankbench" for all tests then?

If that's the case, but I might be wrong so please correct me if it's the case, we will end up with samples of (for example) `list_view` mixed with samples of `shadow_grid`... which are two very different workloads. Isn't it?

libs/utils/wa_results_collector.py
qais-yousef: Ah OK I get you now. So your suggestion to basically append `or workload` whenever we assign test in the if else legs? Then yeah that would be better.

qais-yousef:
I don't think they get mixed - I didn't notice the results were duplicated then. But I'll double check - I might have missed this detail.

What I understood is that each test will have its own entry in the table but its name is unset which caused the sched-evaluation-full.ipynb crash when processing jankbench data. The crash happens somewhere in Panda because the table has a NaN value in a place it didn't expect.

qais-yousef:
> You set the test_id to the same value, the workload name, for all tests: isn't workload going to be always "jankbench" for all tests then?
>
> If that's the case, but I might be wrong so please correct me if it's the case, we will end up with samples of (for example) list_view mixed with samples of shadow_grid... which are two very different workloads. Isn't it?

I think you're right. it looks like it aggregate all tests together in this case.

I am not sure if this is the intended behaviour or a bug. If it's wrong, then wa should have thrown an error that this is not accepted.

It seems I'll have to dig even deeper to get this simple thing fixed, sigh.

qais-yousef:
I updated the branch to address Valentin's comment, but maybe I fixed a symptom rather than the problem...

douglas-raillard-arm:
@qais-yousef Do we still have the issue, or has it been somehow solved in another way ?

qais-yousef:
I haven't revisited this since then to be honest. I am not aware anything changed about this. Looking at the code in next the file hasn't changed there.

douglas-raillard-arm:
Closing as an upcoming refactoring of collector will use WA API and is likely to fix the issue

PR #633, opened 2018-10-17
Add a contribution guideline

valschneider:


douglas-raillard-arm:
LGTM, that may converted to reStructuredText since GitHub handles .rst extension as well.

valschneider:
> LGTM, that may converted to reStructuredText since GitHub handles .rst extension as well.

I gave .rst a try but github wouldn't let me pretty-print it in the web UI, so I stuck with .md

douglas-raillard-arm:
> > LGTM, that may converted to reStructuredText since GitHub handles .rst extension as well.
> 
> I gave .rst a try but github wouldn't let me pretty-print it in the web UI, so I stuck with .md

That's weird, it seems to work perfectly well for that project both for readme.rst and other random files:
https://github.com/ARM-software/arm-trusted-firmware/
https://github.com/ARM-software/arm-trusted-firmware/blob/master/docs/change-log.rst

valschneider:
I guess I must have used some Sphinx-only stuff, since that's what I'm used to.
Otherwise, markdown doesn't really bother me so IMO it's fine to go with it (we already have the README in markdown anyway)

PR #636, opened 2018-10-24
env: TestEnv: Add a from_cli() alternative constructor

valschneider:
This uses Python's argparse library.

I couldn't find a  way to make subparsers work the way I want them to work (I want to be able to specify `--kind=android` or `--kind=linux`, not just `android`/`linux`), so I had to make do.

With this baby it's easy to write a script `dummy.py`:

```python
#!/usr/bin/python3

from lisa.env import TestEnv

te = TestEnv.from_cli()
print(te.target.echo("42"))
```

```bash
./dummy.py 
usage: dummy.py [-h] --kind {android,linux,host}
dummy.py: error: the following arguments are required: --kind/-k
```

```bash
./dummy.py -k android
usage: dummy.py [-h] --kind {android,linux,host} --device DEVICE
dummy.py: error: the following arguments are required: --device/-d
```

```bash
./dummy.py -k linux
usage: dummy.py [-h] --kind {android,linux,host} --hostname HOST
                    --username USERNAME --password PASSWORD
dummy.py: error: the following arguments are required: --hostname/-n, --username/-u, --password/-p
```

```bash
./dummy.py -k host
Module bl is not supported by the target
Module devfreq is not supported by the target
Module fastboot is not supported by the target
Module gem5stats is not supported by the target
Module gpufreq is not supported by the target
Module hotplug is not supported by the target
Module odroidxu3-fan is not supported by the target
Module cgroups is not supported by the target
Failed to initialized "bl" devlib Module
Failed to initialized "cgroups" devlib Module
Failed to initialized "devfreq" devlib Module
Failed to initialized "fastboot" devlib Module
Failed to initialized "gem5stats" devlib Module
Failed to initialized "gpufreq" devlib Module
Failed to initialized "hotplug" devlib Module
Failed to initialized "mbed-fan" devlib Module
Failed to initialized "odroidxu3-fan" devlib Module
Couldn't read target energy model: Unable to probe for energy model on target.
42
```

### DIFF
--- a/lisa/env.py
+++ b/lisa/env.py
@@ -25,6 +25,9 @@ from pathlib import Path
 import shlex
 from collections.abc import Mapping
 import copy
+import sys
+import argparse
+import textwrap
 
 import devlib
 from devlib.utils.misc import which
@@ -210,6 +213,75 @@ class TestEnv(Loggable):
             cls.get_logger().warn('No platform information could be found: {}'.format(e))
             plat_info = None
         return cls(target_conf=target_conf, plat_info=plat_info)
+
+    @classmethod
+    def from_cli(cls, argv=None):
+        """
+        Create a TestEnv from command line arguments.
+
+        :param argv: The list of arguments. ``sys.argv[1:]`` will be used if
+          this is ``None``.
+        :type argv: list(str)
+
+        Trying to use this in a script that expects extra arguments is bound
+        to be confusing (help message woes, argument clashes...), so for now
+        this should only be used in scripts that only expect TestEnv args.
+        """
+        # Subparsers cannot let us specify --kind=android, at best we could
+        # have --android which is lousy. Instead, use a first parser to figure
+        # out the target kind, then create a new parser for that specific kind.
+        kind_parser = argparse.ArgumentParser(
+            # Disable the automatic help to not catch e.g. ./script.py -k linux -h
+            add_help=False,
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            description=textwrap.dedent(
+                """
+                Extra arguments differ depending on the value passed to 'kind'.
+                Try e.g. "--kind android -h" to see the arguments for android targets.
+                """))
+
+        kind_parser.add_argument(
+            "--kind", "-k", choices=["android", "linux", "host"],
+            help="The kind of target to create")
+
+        # Add a self-managed help argument, see why below
+        kind_parser.add_argument("--help", "-h", action="store_true")
+
+        args = kind_parser.parse_known_args(argv)[0]
+
+        # Print the generic help only if we can't print the proper --kind help
+        if not args.kind or (args.help and not args.kind):
+            kind_parser.print_help()
+            sys.exit(2)
+
+        kind = args.kind
+
+        parser = argparse.ArgumentParser()
+        parser.add_argument("--kind", "-k",
+                            choices=[kind],
+                            required=True,
+                            help="The kind of target to create")
+
+        if kind == "android":
+            parser.add_argument("--device", "-d", type=str, required=True,
+                                help="The ADB ID of the target")
+        elif kind == "linux":
+            parser.add_argument("--hostname", "-n", type=str, required=True, dest="host",
+                                help="The hostname/IP of the target")
+            parser.add_argument("--username", "-u", type=str, required=True,
+                                help="Login username")
+            parser.add_argument("--password", "-p",  type=str, required=True,
+                                help="Login password")
+
+        parser.add_argument("--platform", "-pi", type=str,
+                            help="Path to a PlatformInfo yaml file")
+
+        args = parser.parse_args(argv)
+        platform_info = PlatformInfo.from_yaml_map(args.platform) if args.platform else None
+        target_conf = TargetConf(
+            {k : v for k, v in vars(args).items() if k != "platform"})
+
+        return TestEnv(target_conf, platform_info)
 
     def _init_target(self, target_conf, res_dir):
         """


lisa/env.py
```diff
@@ -210,6 +212,46 @@ def from_one_conf(cls, path):
...
+        parser.add_argument("--kind", "-k",
+                            choices=["android", "linux", "host"],
+                            required=True,
+                            help="The kind of target to create")
+
+        # subparsers cannot let us specify --kind=android, at best we could
+        # have --android which is lousy.
+        # Use parse_known_args() instead to look at --kind, and add relevent
+        # arguments
+        args = parser.parse_known_args()[0]
```
douglas-raillard-arm: The comment on subparsers is right. The main issue with parsing twice like this is that --help/-h will be caught by parser.parse_known_args which will not provide a helpful help message and then exit. The solution is to just add all parameters to the parser, and only use the one that are relevant for a given kind (and specify that in the help strings of options).

lisa/env.py
```diff
@@ -210,6 +212,46 @@ def from_one_conf(cls, path):
             plat_info = None
         return cls(target_conf=target_conf, plat_info=plat_info)
 
+    @classmethod
+    def from_cli(cls):
```
douglas-raillard-arm: Being able to specify the argument list can be handy:
```
def from_cli(argv=None):
```
`argv` can be given to `parse_args`, which will by itself default to sys.argv[1:] if `None` is given.

Since a --platform-info option is desirable (see below), that method does not fully belong to TargetConf so it's better for it to stay here in TestEnv.

lisa/env.py
```diff
@@ -210,6 +212,46 @@ def from_one_conf(cls, path):
...
+            parser.add_argument("--hostname", "-n", type=str, required=True, dest="host",
+                                help="The hostname/IP of the target")
+            parser.add_argument("--username", "-u", type=str, required=True,
+                                help="Login username")
+            parser.add_argument("--password", "-p",  type=str, required=True,
+                                help="Login password")
+
+        args = parser.parse_args()
+
+        return TestEnv(TargetConf(vars(args)))
```
douglas-raillard-arm: A --platform-info CLI option would allow creating a PlatformInfo from a file if specified. I don't think it's a good idea to try to build one from CLI options directly, since it is full of objects that are not strings, which means we would need to parse that ourselves. TargetConf is not so much of an issue since it only uses "simple" keys with YAML builtin basic types, and will not change much overtime.
`PlatformInfo.from_yaml_map(FILENAME)` can be used to build it if FILENAME is not None.

lisa/env.py
```diff
@@ -210,6 +212,46 @@ def from_one_conf(cls, path):
             plat_info = None
         return cls(target_conf=target_conf, plat_info=plat_info)
 
+    @classmethod
+    def from_cli(cls):
+        """
+        Create a TestEnv from command line arguments
```
douglas-raillard-arm: The docstring should state that it should only be used to build simple throw away scripts, since even `using parse_know_args` instead of `parse_args` to let other parts of the script use the CLI will lead to confusing --help messages, unless a more complicated API is introduced (which defeats the orginial purpose of a helper for quick&dirty scripts).

Also, there can be argument clash if we add more here, and a helper from a library should not be in charge of defining the CLI of a tool built using it, so it really has to be restricted to quick scripts just like `fileinput` module intended use.

valschneider:
This is how it looks now:

```python
./test_misfit.py -h
usage: test_misfit.py [--kind {android,linux,host}] [--help]

Extra arguments differ depending on the value passed to 'kind'.
Try e.g. "--kind android -h" to see the arguments for android targets.

optional arguments:
  --kind {android,linux,host}, -k {android,linux,host}
                        The kind of target to create
  --help, -h
```

```python
./test_misfit.py --kind android -h
usage: test_misfit.py [-h] --kind {android} --device DEVICE
                      [--platform PLATFORM]

optional arguments:
  -h, --help            show this help message and exit
  --kind {android}, -k {android}
                        The kind of target to create
  --device DEVICE, -d DEVICE
                        The ADB ID of the target
  --platform PLATFORM, -pi PLATFORM
                        Path to a PlatformInfo yaml file

```


```python

```


```python

```

lisa/env.py
```diff
@@ -210,6 +212,83 @@ def from_one_conf(cls, path):
...
+        # have --android which is lousy. Instead, use a first parser to figure
+        # out the target kind, then create a new parser for that specific kind.
+        kind_parser = argparse.ArgumentParser(
+            # Disable the automatic help to not catch e.g. ./script.py -k linux -h
+            add_help=False,
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            description=textwrap.dedent(
+                """
+                Extra arguments differ depending on the value passed to 'kind'.
+                Try e.g. \"--kind android -h\" to see the arguments for android targets.
```
douglas-raillard-arm: \" can be turned into " when inside a triple-quoted string or inside single quotes

lisa/env.py
```diff
@@ -210,6 +212,83 @@ def from_one_conf(cls, path):
...
+
+        # Add a self-managed help argument, see why below
+        kind_parser.add_argument("--help", "-h", action="store_true")
+
+        args = kind_parser.parse_known_args(argv)[0]
+
+        # Print the generic help only if we can't print the proper --kind help
+        if not args.kind or (args.help and not args.kind):
+            kind_parser.print_help()
+            exit(2)
```
douglas-raillard-arm: `sys.exit(2)` is preferred for non-interactive use

valschneider:
@douglas-raillard-arm Done.

PR #637, opened 2018-10-25
WaResultsCollector: fix no kernel path case

vingu-linaro:
If no kernel path has been defined, kernel_refs is empty and
df['kernel_sha1'].replace(kernel_refs) fails

Add sha1 as kernel ref in case no kernel path is provied

Signed-off-by: Vincent Guittot <vincent.guittot@linaro.org>

### DIFF
--- a/libs/utils/wa_results_collector.py
+++ b/libs/utils/wa_results_collector.py
@@ -160,11 +160,12 @@ class WaResultsCollector(object):
         df = df.append(df_list)
 
         kernel_refs = {}
-        if kernel_repo_path:
-            for sha1 in df['kernel_sha1'].unique():
-                ref = Git.find_shortest_symref(kernel_repo_path, sha1)
-                if ref:
-                    kernel_refs[sha1] = ref
+        for sha1 in df['kernel_sha1'].unique():
+            if kernel_repo_path:
+                kernel_refs[sha1] = Git.find_shortest_symref(kernel_repo_path, sha1) or sha1
+            else:
+                kernel_refs[sha1] = sha1
+
 
         common_prefix = os.path.commonprefix(kernel_refs.values())
         for sha1, ref in kernel_refs.iteritems():


libs/utils/wa_results_collector.py
```diff
@@ -165,6 +165,10 @@ def __init__(self, base_dir=None, wa_dirs=".*", platform=None,
                 ref = Git.find_shortest_symref(kernel_repo_path, sha1)
                 if ref:
                     kernel_refs[sha1] = ref
+        else:
```
valschneider: The change looks fine to me, although perhaps we could simplify the whole `kernel_refs` region:

```python
kernel_refs = {sha1 : sha1 for sha1 in df['kernel_sha1'].unique()
if kernel_repo_path:
    kernel_refs = {sha1 : Git.find_shortest_symref(kernel_repo_path, sha1) or sha1 for sha1 in kernel_refs.keys()}
```

vingu-linaro: I feel this implementation a bit more difficult to read . In addition, kernel_refs is uselessly set a 1st time before being overwritten just after when kernel_repo_path is defined.
If you strongly want to minimize the number of line, the below seems more readable
```
if kernel_repo_path:
    kernel_refs = {sha1 : Git.find_shortest_symref(kernel_repo_path, sha1) or sha1 for sha1 in kernel_refs.keys()}
else 
    kernel_refs = {sha1 : sha1 for sha1 in df['kernel_sha1'].unique()
```

valschneider: I don't really mind the number of lines, I mostly wanted to refrain from duplicating the `for sha1 in df['kernel_sha1'].unique()` statement.

vingu-linaro: hmm ... but you still have twice "for sha1 in df['kernel_sha1'].unique()"
would you prefer something like below ?
```
for sha1 in df['kernel_sha1'].unique():
    if kernel_repo_path:
        kernel_refs[sha1] = Git.find_shortest_symref(kernel_repo_path, sha1) or sha1
    else:
        kernel_refs[sha1] = sha1

```

vingu-linaro: if it's ok for everybody, i 'm oing to update the patch with this change

vingu-linaro:
ah yes sorry. I forgot to squash them before pushing
I have merged them

PR #640, opened 2018-10-29
Minor improvements for scheduler/sanity.py

douglas-raillard-arm:
Save the work done on all CPUs for a given capacity, and then reduce
that to the minimum work to avoid comparing with sys.maxsize

Also add annotated `from_testenv` method.

### DIFF
--- a/lisa/env.py
+++ b/lisa/env.py
@@ -173,7 +173,7 @@ class TestEnv(Loggable):
             plat_info = copy.copy(plat_info)
         self.plat_info = plat_info
 
-        logger.info('Pre-configured platform information:\n%s', self.plat_info)
+        logger.info('User-defined platform information:\n%s', self.plat_info)
 
         self.ftrace = None
         self._installed_tools = set()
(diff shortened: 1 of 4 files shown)


lisa/tests/kernel/scheduler/sanity.py
```diff
@@ -25,7 +26,7 @@ class CapacitySanity(TestBundle):
     A class for making sure capacity values make sense on a given target
 
     :param capacity_work: A description of the amount of work done on the
-      target, per capacity value ({capacity : work})
+      target, per capacity value ({capacity : [work_on_cpu0, work_on_cpu1, ...]})
```
valschneider: AFAICT when creating `capa_work` in `_from_testenv()`, we have no guarantee on the CPU iteration order. So instead we should use `{capacity : {cpu : work_on_cpu}}` and we can do the test on

```python
capacity_work =  {
    capa: min(cpus.values())
    for capa, cpus in self.capacity_work.items()
}
```

And then the test will be truly obvious.

douglas-raillard-arm: That's right, I left the CPU ID out since it was not used anywhere for the actual computation, and having the list of values can still be useful to check wildly wrong values (even without the CPU ID).

PR #641, opened 2018-10-30
submodules: progress devlib

qperret:
Devlib has bug fixes we want to have in Lisa, so let's progress it to
the latest master.

Tested with:

  lisa-test -n 1 tests/eas/generic.py:EnergyModelWakeMigration

on Android and Linux targets. No regressions observed.


qperret:
This fixes the ugly colouring issue I had in devlib. Should I push a PR for the lisa-next branch too ? I'm just not really sure how to test it yet :)

valschneider:
I think we're in the annoying timeframe where we still need some fixes/changes to lisa/master and will eventually need to copy those over to lisa/master... On the bright side this is just a submodule update so that's easy to replicate on lisa/next.

PR #645, opened 2018-10-31
doc/overview: move recommendation to use venv

ionela-voinescu:
Move recommendation to use a virtual environment at the beginning of the
installation process.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -5,6 +5,13 @@ Overview
 Installation
 ============
 
+As with any other Python project, we recommend using a virtual python environment::
+
+  python3 -m venv --system-site-packages <venv_folder>
+  source <venv_folder>/bin/activate
+  # do work
+  deactivate
+
 For now, you will need to clone LISA from `github
 <https://github.com/ARM-software/lisa>`_ and use ``setup.py`` to install the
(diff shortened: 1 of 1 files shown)


doc/overview.rst
```diff
@@ -5,6 +5,13 @@ Overview
 Installation
 ============
 
+As with any other Python project, we recommend using a virtual python environment::
+
+  python3 -m venv --system-site-packages <venv_folder>
```
derkling: As discussed with @douglas-raillard-arm, I would suggest to force venv creation at source init_env time.
I would also suggest to use a LISA-related naming... just to make it easy to cross check that users are actually in the venv created by LISA scripts.

If we go the "forced venv creation" way... probably we can also completely drop this bit of instructions and just warn users that LISA will make use of a venv to install all its dependencies.

valschneider: Had a think about that - we might not want to force the use of `venv` in `init_env`. There are some external tools (ReadTheDocs for sure, Travis maybe) that automatically create a venv before using LISA.

douglas-raillard-arm:
I'm actually implementing the venv management in LISA-next, so that will not be required anymore

PR #647, opened 2018-10-31
misc fixes

douglas-raillard-arm:


lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -403,6 +403,20 @@ def test_task_load_avg(self, allowed_error_pct=15) -> ResultBundle:
         """
         return self._test_signal('load_avg', allowed_error_pct)
 
+
+    def test_slack(self, negative_slack_allowed_pct=15) -> ResultBundle:
+        """
+        Same test as :meth:`RTATestBundle.test_slack`, but executed on all
+        the collected traces.
+        """
+        return CombinedResultBundle(
```
valschneider: Ah right, this one needs special attention here.

For now I'd be inclined not to introduce a new `ResultBundle` class. `TestMetric`s can be nested inside a `ResultBundle`, so I would see this as being the same as `RTATestBundle.test_slack` but with an extra metric level (per frequency).


Adding this class could make things more confusing for new users. If a single test depends on several metrics (e.g. `test_slack()` has per-task slack score), what should be used? `CombinedResultBundle`s with `ResultBundle` per task "test"? `ResultBundle` with several metrics?

I'd like to keep things as simple as possible and stick with `ResultBundle`+`TestMetric`. It just so happens here that we're basically doing n iterations of `test_slack()`, so it looks like it makes sense to create a class that handles "iterations of `ResultBundle`s", but this frequency test is an ugly duckling that doesn't fit our usual test pattern (a single trace per workload).

lisa/tests/kernel/test_bundle.py
```diff
@@ -85,6 +86,11 @@ def lower_name(self):
         """Return the name in lower case"""
         return self.name.lower()
 
+class ProcessResultBundle:
```
valschneider: That is exclusively used by Exekall, as such I'd like it not to be visible to LISA. Having it as a "private" class would hide it from autocomplete, but I would rather have the code not live here. Could this live in the glue instead?

lisa/env.py
```diff
@@ -437,22 +472,31 @@ def get_res_dir(self, name=None, append_time=True, symlink=True):
         """
         logger = self.get_logger()
 
+        if isinstance(self._res_dir, ArtifactPath):
```
valschneider: There already is a use of the pathlib API in `get_res_dir()` that slipped my attention, but I would like to refrain from using it (it's probably too late for that in exekall/bisector but it can still be done for LISA).

A large amount of bugs and issues we've had with the next branch was due to this API breaking compatibility between Python 3.5/3.6/3.7, and I'd rather spend my time on other things than to constantly fight those and hack up workarounds.

We might envision embracing it once the API *actually* stabilizes and we bump up the Python 3 version requirement, but until now I believe we'd be better off without it.

lisa/env.py
```diff
@@ -437,22 +472,31 @@ def get_res_dir(self, name=None, append_time=True, symlink=True):
...
             time_str = datetime.now().strftime('%Y%m%d_%H%M%S.%f')
             if not name:
                 name = time_str
             elif append_time:
                 name = "{}-{}".format(name, time_str)
 
-            res_dir = os.path.join(self._res_dir, name)
+            # If we were given an ArtifactPath with an existing root, we
+            # preserve that root so it can be relocated as the caller wants it
+            res_dir = ArtifactPath(root, relative/name)
```
valschneider: As for the use of `ArtifactPath` itself, no direct client of `get_res_dir()` (`TestBundle` & co) have a use for it, but I think that's okay (since it's just a string). You might want to change `TestBundle.from_dir()` though, to make sure we always have the same type for `res_dir`.

lisa/env.py
```diff
@@ -267,61 +267,63 @@ def from_cli(cls, argv=None):
...
+
+        parser.add_argument("--log-level",
+                            choices=('warning', 'info', 'debug'),
+                            help="Verbosity level of the logs.")
+
+        parser.add_argument("--res-dir", "-o",
+                            help="Result directory of the created TestEnv.")
+
+        # Options that are not a key in TargetConf must be listed here
+        not_target_conf_opt = ('platform_info', 'log_level', 'res_dir')
```
valschneider: `target_conf` also needs to be in there.

lisa/env.py
```diff
@@ -267,61 +267,63 @@ def from_cli(cls, argv=None):
...
                 Try e.g. "--kind android -h" to see the arguments for android targets.
                 """))
 
-        kind_parser.add_argument(
-            "--kind", "-k", choices=["android", "linux", "host"],
-            help="The kind of target to create")
 
-        # Add a self-managed help argument, see why below
-        kind_parser.add_argument("--help", "-h", action="store_true")
+        parser = argparse.ArgumentParser()
```
valschneider: So overall I'm okay with the single parser as it's much easier to maintain and develop, but I miss the contextual help that made things crystal clear.

It's not coming back, so could we augment the CLI help to give a few examples?

lisa/env.py
douglas-raillard-arm: All direct clients of `get_res_dir()` have a use for it: that's the only reasonable way they can get their res_dir updated when reloading objects from the exekall serialized database. The glue detects these and will relocate them properly.

I could do the same in `TestBundle.from_dir`, although it will unlikely be serialized again (but for consistency that may be better).

lisa/env.py
douglas-raillard-arm: sure, I'll add some

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: The thing is that if you have a class that provides the `from_testenv()` infra for rt-app, you have everything that's required to run `test_slack()` (running an rt-app workload generates the file required for this test). It might make more sense to override `test_slack()` with just an exception raise (and you can strip out the annotations for exekall).

Then I don't know how the whole `TestBundle` logic would work - you only have one `TestBundle` for the frequency test, but you also get several `TestBundle`s for the slack test?

lisa/utils.py
```diff
@@ -135,17 +144,47 @@ def _init_yaml(cls):
...
         # file instead of being replaced by escape sequence.
-        yaml.allow_unicode = (cls.YAML_ENCODING == 'utf-8')
+        yaml.allow_unicode = ('utf' in cls.YAML_ENCODING)
         yaml.default_flow_style = False
         yaml.indent = 4
         yaml.Constructor.add_constructor('!include', cls._yaml_include_constructor)
         yaml.Constructor.add_constructor('!var', cls._yaml_var_constructor)
         yaml.Constructor.add_multi_constructor('!call:', cls._yaml_call_constructor)
+
+        # Replace unknown tags by a placeholder object containing the data This
```
valschneider: Missing comma? Also, out of curiosity, how did you hit this issue? I don't think I've ever hit that.

lisa/tests/kernel/test_bundle.py
```diff
@@ -141,6 +142,13 @@ def add_metric(self, name, data, units=None):
         """
         self.metrics[name] = TestMetric(data, units)
 
+    def display_and_exit(self) -> type(None):
```
valschneider: So while it's just a few lines I don't see the value of having this in `test_bundle.py`.
If a script wants to do that, it's just 2 lines. If it's about exekall, could you make that the actual default goal? That should be pretty generic enough (print the result and exit).

lisa/env.py
```diff
@@ -49,11 +48,46 @@
 LATEST_LINK = 'results_latest'
 DEFAULT_DEVLIB_MODULES = ['sched', 'cpufreq', 'cpuidle']
 
-class ArtifactPath(str):
+class ArtifactPath(str, Loggable, HideExekallID):
```
valschneider: Shouldn't that be in `utils`?

lisa/utils.py
douglas-raillard-arm: It happens if you load a YAML file containing tags for classes that have not been created yet (you have not imported the module). That comes from the fact that if you call cls._yaml.register_class() on Serializable classes, the default tag will be cls.__qualname__, instead of being a full dotted path including the module name. As a result, the tag is more stable against refactoring, but it is also not possible for ruamel.yaml to find the class if it has not been registered. 

It should not be a big problem since if you haven't imported the corresponding module, chances are you don't care about these objects. Classes for objects used internally by classes that matters are usually imported indirectly, so it's fine. Ultimately, that also allows to inspect the content of a YAML file that cannot be completely parsed instead of a global failure at load time, so it improves robustness

The main reason I call register_class() is because it's the only way to take into account the optional class attribute yaml_tag, and that is desirable to maintain a stable set of tags in case we want to rename the class.

lisa/tests/kernel/test_bundle.py
douglas-raillard-arm: It's very similar to TestEnv.from_cli(), just a helper to write a simple script. It's indeed called by exekall when it is asked for, so even if it is just a few lines it needs to sit somewhere in a function.

Why would you want to make that the default goal ? Having the ResultBundle being the result of expressions makes it easier to post-process the DB, and also allows the glue to print the summary and access the ResultBundle without extra steps. The test expressions are not in charge of managing the flow of execution and decide when it is time to exit, and they are also not in charge of deciding when to display the summary of execution.

PR #653, opened 2018-11-07
misc improvements

douglas-raillard-arm:


lisa/utils.py
```diff
@@ -217,6 +218,33 @@ def _yaml_include_constructor(cls, loader, node):
         with open(path, 'r', encoding=cls.YAML_ENCODING) as f:
             return cls._yaml.load(f)
 
+    @classmethod
+    def _yaml_env_var_constructor(cls, loader, suffix, node):
+        """
+        Provide a !include tag in YAML that can be used to include the content
+        of an environment variable, and converting it to a Python type:
+
+        !env:int MY_ENV_VAR
```
valschneider: I'd like the docstring to mention CI/Jenkins, so people like me don't start thinking they *need* to use env variables to setup TestEnv's.

douglas-raillard-arm: Such documentation does not really belong to here. You also don't *need* !include, !var and !call , they are just features of the configuration "on top" of YAML. This class is used as the base class for basically anything that would use a config file. Maybe we can have a non-docstring documentation on how to do an advanced setup, with examples of these tags to show people what they can use to fit their specific use case (like the fact that you can put all these configurations into one file).

valschneider: Having a few lines of docstring explaining *why* this feature is there when it's not obvious (to me) at a first glance can never hurt.

lisa/env.py
```diff
@@ -46,37 +47,46 @@
...
+            KeyDesc('events', 'FTrace events to trace', [StrList]),
+            KeyDesc('functions', 'FTrace functions to trace', [StrList]),
+            KeyDesc('buffsize', 'FTrace buffer size', [int]),
+        )),
+        LevelKeyDesc('devlib', 'devlib configuration', (
+            LevelKeyDesc('platform', 'devlib.platform.Platform subclass specification', (
+                KeyDesc('class', 'Name of the class to use', [str]),
+                KeyDesc('args', 'Keyword arguments to build the Platform object', [Mapping]),
+            )),
+            KeyDesc('excluded-modules', 'List of devlib modules not *not* load', [StrList]),
```
valschneider: Typo here

lisa/env.py
```diff
@@ -46,37 +47,46 @@
...
-        'devlib': {
-            'platform': {
-                'class': str,
-                'args': Mapping,
-            },
-            'excluded-modules': StrList,
-        }
-    }
+    STRUCTURE = TopLevelKeyDesc(YAML_MAP_TOP_LEVEL_KEY, 'target connection settings', (
+        KeyDesc('board', 'Board name, free-form value only used to embelish logs', [str]),
```
valschneider: Since `board` has changed from affecting `TestEnv` to just being a pretty-print helper, it might be time to rename it to `name` or something similar.

douglas-raillard-arm: makes sense, that is what I actually already did for PlatformInfo

valschneider:
Agreed out of band with @douglas-raillard-arm that documentation will live with `TargetConf` & `PlatformInfo` which will be the actual clients of this.

PR #658, opened 2018-11-16
rework install process

douglas-raillard-arm:
Now creates a venv by default unless LISA_USE_VENV=0, and properly install all dependencies in editable mode when we ship them and LISA_DEVMODE is used, or from PyPI otherwise

Also rename lisa-ipython to lisa-jupyter, since ipython has been a Jupyter kernel for some time now


doc/overview.rst
```diff
@@ -8,22 +8,27 @@ Installation
...
-  python3 -m pip install -e .
-
-Extra (optionnal) packages can also be installed, for instance notebook support::
-
-  python3 -m pip install -e .[notebook]
+  # A few packages need to be installed, like python3 or kernelshark. Python
+  # modules will be installed in a venv at the next step, without touching 
+  # any system-wide install location.
+  ./install_base_ubuntu.sh
+  # On the first run, it will take care of creating a Python venv and populate it
```
valschneider:
```suggestion
  # On the first run, it will take care of creating a Python venv and populating it
```
Or
```suggestion
  # On the first run, it will take care of creating a Python venv and will populate it
```

douglas-raillard-arm:
Since the ``lisa`` package is installed inside a venv, it is necessary to activate it before trying to import it. However a devmode_requirements.txt is available for other tools to install LISA in their own venv/virtualenv (like readthedocs)

valschneider:
> Since the `lisa` package is installed inside a venv, it is necessary to activate it before trying to import it. However a devmode_requirements.txt is available for other tools to install LISA in their own venv/virtualenv (like readthedocs)

Right, I just saw RTD can let us specify a requirements.txt so we should be fine.

valschneider:
I didn't have python3-venv installed and `source init_env` didn't complain. I had to manually issue a
```
lisa-install
```
to get the warning.

douglas-raillard-arm:
Removing --system-site-package, as pip --ignore-installed seems to not actually completely ignore what is already installed. Isolating the venv from the system will make sure that it actually cannot even see what is outside the venv.

PR #660, opened 2018-11-20
improve installation process

douglas-raillard-arm:
Add setup.py for exekall and bisector, so we don't have to manipulate PYTHONPATH anymore


lisa/version.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2016, ARM Limited and contributors.
```
valschneider: [s/2016, ARM/2018, Arm](https://i.kym-cdn.com/entries/icons/original/000/007/784/what-year-is-it-robin-williams.jpg). That reminds me we should add a copyright header test...

PR #661, opened 2018-11-23
Add ppc64le binaries

jlelli:
Add statically compiled binaries for ppc64le, so that LISA can run on
that architecture as well.

This goes together with https://github.com/ARM-software/devlib/pull/348 (already merged).


derkling:
I guess it could make easy the transition to the LISA next branch if you post this against the next branch.

jlelli:
Opened another one based on next.

PR #662, opened 2018-11-23
Add ppc64le binaries

jlelli:
Add statically compiled binaries for ppc64le, so that LISA can run on that architecture as well.

Binaries have been compiled from same tool sources of other binaries already present in LISA.


douglas-raillard-arm:
LGTM, FYI there is an env var used by lisa shell in LISA-next that you can use to add that folder to the PATH:
```
export LISA_HOST_ABI=ppc64le
source init_env
```

LISA-next is not stable yet, so it might change but we are likely to keep something similar so the tools can be used on the host side as well.

qperret:
Should we progress Lisa's devlib as well to get the ppc64 binaries ?

PR #663, opened 2018-11-23
[next] Minor setup improvements

qperret:
Just a bunch of minor things I found out while installing Lisa next on my (relatively fresh) Ubuntu 18.04 setup.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -16,7 +16,7 @@ and then issue these commands::
   # A few packages need to be installed, like python3 or kernelshark. Python
   # modules will be installed in a venv at the next step, without touching 
   # any system-wide install location.
-  ./install_base_ubuntu.sh
+  sudo ./install_base_ubuntu.sh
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
Interesting, it seemed to work on Ubuntu 18.04 bionic in the vagrant VM. We can add these packages if they are required on that install though ...

qperret:
I have absolutely no clue why Travis complains, and I cannot reproduce with `tools/scripts/travis_tests.sh` on my end. Sorry in advance but I will need to experiment by force-pushing this branch directly ...

qperret:
Adding a new package to the dependency list in install_ubuntu_base seemed to do the trick -- turns out Travis was complaining about missing GTK dependencies.

I do not know why this happens only now, nor if my 'fix' makes any sense at all. So, I would recommend to wait for somebody else to confirm the issues I had in the first place before merging this PR ...

PR #669, opened 2018-11-27
misc fixes

douglas-raillard-arm:


lisa/utils.py
```diff
@@ -64,6 +65,20 @@ def get_logger(cls, suffix=None):
             name += '.' + suffix
         return logging.getLogger(name)
 
+class Subclassable:
```
valschneider: What about a static module method that we could import?

To me it feels simpler that way - if you want to get the subclasses of a given class you can just import & call `get_subclasses(cls)` instead of having to understand you need to inherent from this special class.

lisa/utils.py
```diff
@@ -64,6 +65,20 @@ def get_logger(cls, suffix=None):
...
+    @classmethod
+    def get_subclasses(cls, cls_set=None):
+        """Get all indirect sublcasses of the class."""
```
valschneider:
```suggestion
        """Get all indirect subclasses of the class."""
```

lisa/utils.py
douglas-raillard-arm: > static module method
You mean a plain function ;-) ?

Indeed, that can be a plain function in utils.py . I probably wrote that this way as an "extension" of `__subclasses__()`, but there is no real reason.

lisa/env.py
```diff
@@ -748,6 +744,23 @@ def disable_idle_states(self):
             for domain in self.target.cpufreq.iter_domains():
                 self.target.cpuidle.enable_all(domain[0])
 
+    def get_emeter(self, res_dir=None):
```
valschneider: While I'm thinking about it, we need to go through the `EnergyMeter` class and remove the homemade classes that could be replaced by their devlib equivalents (e.g. ACME or hwmon).

PR #670, opened 2018-11-29
[next] Fix HWMon and ACME

qperret:
Switching to Python 3 broke the ACME and HWMon support. The ACME is (currently) entirely managed in Lisa so there's a first patch for that. For HWMon, the fix got merged in devlib so this PR also progresses the subtrees.


lisa/energy_meter.py
```diff
@@ -450,6 +453,14 @@ def report(self, out_dir, out_energy='energy.json'):
         logger = self.get_logger()
         channels_nrg = {}
         channels_stats = {}
+
+        # iio-capture returns and empty string if killed right after its
+        # invocation, so to workaround the issue, make sure reset was long ago
+        # enough
+        delta = time.monotonic() - self.reset_time
+        if delta < 2.0:
```
valschneider: I know I said we'll ditch this and move on to use devlib's ACME, but... Can we use a class attribute instead?
Something like this:

```python
class ACME(EnergyMeter):
    # ...
    REPORT_DELAY_S = 2.0
    """
    iio-capture returns an empty string if killed right after its invocation,
    so we have to enforce a delay between reset() and report()
    """

    # ...

    delta = time.monotonic() - self.reset_time
    if delta < self.REPORT_DELAY_S:
        sleep(self.REPORT_DELAY_S - delta)
```

lisa/energy_meter.py
```diff
@@ -407,12 +407,14 @@ def reset(self):
             csv_file = os.path.join(self._res_dir, 'samples_{}.csv'.format(channel))
 
             # Start a dedicated iio-capture instance for this channel
-            self._iio[ch_id] = Popen([self._iiocapturebin, '-n',
+            self._iio[ch_id] = Popen(['stdbuf', '-i0', '-o0', '-e0',
```
valschneider: Unless it's proven that we really need this incantation, I'd happily do without it.

lisa/energy_meter.py
qperret: Sure

lisa/energy_meter.py
qperret: @douglas-raillard-arm was willing to keep it I think. I don't mind either way.

douglas-raillard-arm: According the the glibc doc, terminating the process will not cause it to flush streams. Since stdout is a stream and `printf()` is equivalent to `fprintf(stdout)` [1], I can't see anything that would force iio-capture to actually write to stdout before dying (unless it explicitly disable buffering on stdout). Maybe iio-capture is made to be robust, or we were just lucky for now.

[1] http://port70.net/~nsz/c/c11/n1570.html#7.21.6.3p2

valschneider:
@douglas-raillard-arm You changed the Emeter code with https://github.com/ARM-software/lisa/pull/669, is this still needed?

douglas-raillard-arm:
@valschneider That PR is orthogonal, I cleaned up the `EnergyMeter` base class and made it available to tests, but haven't touched the subclasses that do the real work.

valschneider:
> @valschneider That PR is orthogonal, I cleaned up the `EnergyMeter` base class and made it available to tests, but haven't touched the subclasses that do the real work.

I see, thanks. @qperret, have fun rebasing this whenever you feel like ;)

qperret:
Rebased and tested :)

douglas-raillard-arm:
LGTM

PR #671, opened 2018-11-29
Misc improvements/fixes

douglas-raillard-arm:
Most visible change: 
* `TargetConf` have been renamed to `TestEnvConf` to match its actual usage. Top-level key in YAML file has changed accordingly to 'testenv-conf'
* Target connection settings are now located under a "target" subkey in `TestEnvConf`:
```
testenv-conf:
    target:
        name: foo
        host: foobar
        password: mypassword
        username: foobar
```

The same changes applies to the dictionary passed to `TestEnvConf.__init__`

### DIFF
--- a/lisa/energy_model.py
+++ b/lisa/energy_model.py
@@ -16,12 +16,12 @@
 #
 
 from collections import namedtuple, OrderedDict
-from itertools import product, groupby
+from itertools import product
 import logging
 import operator
 import re
 
-from lisa.utils import Loggable, Serializable, memoized
+from lisa.utils import Loggable, Serializable, memoized, groupby
 
 import pandas as pd
 import numpy as np
@@ -437,7 +437,7 @@ class EnergyModel(Serializable, Loggable):
         List of lists of CPUs nodes who share the same active state values
         """
         def key(node):
-            return node.active_states
+            return sorted(node.active_states.items())
 
         return [
             list(group)
--- a/lisa/exekall_customize.py
+++ b/lisa/exekall_customize.py
@@ -29,7 +29,7 @@ import traceback
 
 from lisa.env import TestEnv, TargetConf
 from lisa.platforms.platinfo import PlatformInfo
-from lisa.utils import HideExekallID, Loggable, ArtifactPath, get_subclasses, MultiSrcConf, groupby
+from lisa.utils import HideExekallID, Loggable, ArtifactPath, get_subclasses, MultiSrcConf, groupby, Serializable
 from lisa.tests.kernel.test_bundle import TestBundle, Result, ResultBundle, CannotCreateError
 from lisa.tests.kernel.scheduler.load_tracking import FreqInvarianceItem
 
@@ -111,6 +111,14 @@ class LISAAdaptor(AdaptorBase):
                 non_reusable_type_set=non_reusable_type_set
             ))
 
+        # Inject serialized objects as root operators
+        for path in self.args.inject:
+            obj = Serializable.from_path(path)
+            op_list.append(
+                PrebuiltOperator(type(obj), [obj],
+                non_reusable_type_set=non_reusable_type_set
+            ))
+
         return op_list
 
     def get_hidden_callable_set(self, op_map):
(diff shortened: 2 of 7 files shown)


testenv_conf.yml
```diff
@@ -7,30 +7,39 @@
 #
 # Configuration required by devlib to connect to a target.
 #
-target-conf:
-    # Kind of platform
-    # - linux   : accessed via SSH connection
-    # - android : accessed via ADB connection
-    # - host    : run on the local host
-    kind : android
+testenv-conf:
```
valschneider: We have conflicting opinions about this, but I'll spit my bit regardless.

I hate mappings as constructor parameters. Ideally I'd go for something like this for `TestEnv`:

```python
class TestEnv(Loggable):
    def __init__(self, name, kind, host=None, username=None, password=None, port=None,
                 device=None, keyfile=None, workdir=None, ftrace=None, emeter=None, devlib=None,
                 plat_info=None, res_dir=None):
        pass
```

Now, that's so verbose it almost brings me back to my Delphi days. Looking at devlib, they use a `connection_settings` for the target creation, so there's precedence for it. That means we could have something like this instead:

```python
class TestEnv(Loggable):
    def __init__(self, name, kind, connection_settings, workdir=None, ftrace=None, emeter=None, devlib=None,
                 plat_info=None, res_dir=None):
        pass
```

So what you're proposing would make us do this:
```python
testenv_conf = TestEnvConf({
    "target": {
        "kind" : "linux",
        "name" : "hikey960",
        "host": "10.1.195.57",
        "username" : "root",
        "password" : "root"
    }
    "ftrace": {
        "events" : ["sched_switch", "sched_wakeup"]
    }
})

te = TestEnv(testenv_conf)
```

While I'd much rather go for this:
```python
connection_settings = {
    "host": "10.1.195.57",
    "username" : "root",
    "password" : "root"
}

ftrace = {
    "events" : ["sched_switch", "sched_wakeup"]
}

te = TestEnv(kind="linux", name="hikey960", connection_settings, ftrace=ftrace)
```

And I think that is already more reasonable. I sincerely believe that's what a class ``__init__()`` should look like: as many parameters as possible should be exposed. Alternative constructors should be there for shortcuts, not the other way around.

Yes, all of those arguments before `plat_info` are what you want to arrange in a separate `target`  node, but I don't think it's worth the added verbiage. So while adding a `target` level in the configuration feels more `tidy`, I'm not convinced it's worth a) the API breakage b) the extra level just to get a target working.

douglas-raillard-arm: One simple question: what is the added value of TestEnv if that is not to dispatch configuration bits, in which case it's much simpler to achieve by passing the conf as it is done currently ? It's sole current goal is to provide a single point of input validation and logging for common things. 

* `TestEnv.collect_ftrace()` would be best added on a subclass of `devlib.Target`. Same goes for the `disable_idle_states()` and similar.
* Once that is done, the only "service" left in `TestEnv` is `get_res_dir()`, which can easily be provided in another way.

This way, you will be able to build a target by passing it it's parameters directly as you want to do. It's quite pointless to shovel around `Target.__init__` parameters through objects that will do nothing with it apart from building a `Target`, again doing nothing with it apart from giving it to the rest of the code.

If we still want an object that can centralize the configuration of common things to have a single parameter to pass to tests, `TestEnv` can be turned into a dummy namespace that tests will optionally use if no alternative is provided.

> a) the API breakage

There have been no commitment yet on that point, and LISA has undergone an astonishingly low number of refactoring for a library that is supposed to be a solid foundation. `bisector` had at least 4 major refactoring, and `exekall` 5. The point of LISA-next is to do the best of our knowledge and hopefully, break the API only once when transitioning from LISA-legacy to LISA-next. This can only happen if classes have a clearly defined responsibility and concerns, which was never the case for `TestEnv`.

douglas-raillard-arm: Following offline discussion, I'll try to remove `TestEnv` and see how it goes. If that attempt is successful, that change will be irrelevant so I'm removing it from that PR until we have more information to choose which way to go.

PR #673, opened 2018-11-30
Remove TestEnv

douglas-raillard-arm:
* introduce Target: Composition of devlib.Target's instance. It provides additional context manager (freezer etc), a PlatformInfo, and allows transparent access to the composed devlib.Target (using `__getattr__` forwarding). This allows easily managing which class is chosen exactly (based on "kind"), and make explicit all references to the underlying target.
* remove lisa.env.TestEnv, replace it with lisa.target.Targtet
* update wlgen to use Target
* introduce lisa.traceFtraceCollector as a thin wrapper around devlib's one
  This collector is propagated as an optional parameter along the `from_target()` path down to `_run_rtapp()`. Exekall is able to create a custom collector configured with a user conf that is merged with the conf from the TestBundle class.
* Introduce lisa.conf.Configurable class to allow pear a configuration class with a regular class.


lisa/target.py
```diff
@@ -211,61 +210,112 @@ class TestEnv(Loggable, HideExekallID):
...
         # computed when actually needed.
-        self.plat_info.add_target_src(self, fallback=True)
+
+        rta_calib_res_dir = os.path.join(self._res_dir, 'rta_calib')
+        os.makedirs(rta_calib_res_dir)
+        self.plat_info.add_target_src(self, rta_calib_res_dir, fallback=True)
 
         logger.info('Effective platform information:\n%s', self.plat_info)
 
+    def __getattr__(self, attr):
```
valschneider: I remember us having a conversation about this but I can't remember the details. Could you remind me why we do this instead of inheriting from `devlib.target:Target`?

lisa/wlgen/workload.py
```diff
@@ -72,24 +72,18 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False, value=42)
...
     """
 
-    def __init__(self, te, name, res_dir=None):
-        self.te = te
+    def __init__(self, target, name, res_dir):
+        self.target = target
         self.name = name
-
-        # XXX: move this to run() instead?
-        if not res_dir:
```
valschneider: Why that change? I've tried running a simple notebook and I end being forced to specify a directory, I suppose this stems from `get_res_dir()` being removed? It was a very convenient feature.

douglas-raillard-arm: The target is not a res dir manager anymore, that's a target. That responsibility is delegated to ArtifactPathManager, which has a `get_res_dir()` method for interactive use. If needed, we can have an alternate constructor that would take a TestEnv-ish object and use it like we used to do. The old way of passing a TestEnv to `__init__` is akin to having `__init__` taking a configuration object instead of a bunch of parameters: the basic API takes a bunch of parameters, and some alternate constructors provide sugar on top of that if needed.

valschneider: > 
> ```
>         res_dir = res_dir or ArtifactPathManager().get_dir(
>             name=cls.__qualname__,
>             symlink=True,
>         )
> ```

Looks fine to me - I don't care who provides the directory as long as I don't have to care about it if I don't care about it (heh).

douglas-raillard-arm: If we end up having more things like that, it could be a good idea to consolidate all that interactive-style use in a `TestEnv`, and having a `from_testenv` method extracting all the required bits out of it.
The critical part it: that `TestEnv` role is only to provide better interactive support, so it can only be manipulated by toplevel methods dedicated to that. Lower-level APIs must use independent pieces.

lisa/target.py
```diff
@@ -211,61 +210,112 @@ class TestEnv(Loggable, HideExekallID):
...
+        self.name = name
 
-        # That res_dir is for the exclusive use of TestEnv itself, it must not
-        # be used by users of TestEnv
-        self._res_dir = res_dir
-        if self._res_dir:
-            os.makedirs(self._res_dir, exist_ok=True)
-            if os.listdir(self._res_dir):
-                raise ValueError('res_dir must be empty: {}'.format(self._res_dir))
+        # That res_dir is for the exclusive use of Target itself, it must not
```
valschneider: Why is that?

lisa/target.py
valschneider: I deleted my previous question about this, but it might be worth adding either some comments or some docstrings explaining why we're doing this (instead of `devlib.target.Target` composition)

lisa/target.py
douglas-raillard-arm: We don't mix up artifacts anymore. Tests get their own artifact folder when needed, and so does the Target. It should be mentioned that in a typical exekall session, only one Target is created. If tests start to store things there, that boils down to having part of their artifacts stored in the folder dedicated to the first test being run, which is the one for which the Target will be initially created.

douglas-raillard-arm: The reasons the Target has an artifact folder in the first place are:
* GEM5 needs a folder to store its things, since when using GEM5, the Target is not only the connection to the target but the actual virtual board itself.
* RTapp calibration value artifacts, when the Target.plat_info['rtapp']['calib'] is accessed. Since that PlatformInfo instance is deeply linked in a 1-1 fashion with that Target object, it makes sense for it to use a subfolder in the Target's `res_dir`

valschneider: > We don't mix up artifacts anymore. Tests get their own artifact folder when needed, and so does the Target

But then that means we no longer the "all of your experiment results are under that `TestEnv`'s main dir feature" which I found really useful.

Up until recently my results directory was unparsable, but having things stored under that tells me "This is stuff you ran on your HiKey960 and the TestEnv/Target instance was created on X day at Y hour" was really helpful.

douglas-raillard-arm: > But then that means we no longer the "all of your experiment results are under that `TestEnv`'s main dir feature" which I found really useful.

That is already not the case anymore. The ArtifactPathManager is responsible for managing such folder, so the typical case would be to create one per notebook and then call `get_res_dir()` on it to allocate subfolders. One commit in this stack introduces a new-style `TestEnv`, which has its own internal ArtifactPathManager and therefore distributes subfolders this way.

douglas-raillard-arm: So the bottom line is: that feature has definitely nothing to do with connecting to a target, and the new `TestEnv` class is there to centralize some boilerplate that is related to one notebook session.

douglas-raillard-arm: The idea of that series is to remove notebook-specific code that ends up creeping into the lowest level APIs we have. Notebook usecase should be handled as a layer on top of basic entities, by providing a couple of extra classes that are relevant. `exekall` does not have such code sneaking down into `wlgen` for example, it just has a bunch of classes in `exekall_customization` and then use the building blocks of LISA verbatim.

valschneider:
Could we make it so Notebooks are still populated with their output, instead of being wiped clean? Part of their role is to show what kind of output we get.

I did that (and also fixed) for the profiling notebook here: https://github.com/valschneider/lisa/commit/b44b5c303e818ee10ed4caea67430c35c222d0a7

douglas-raillard-arm:
> Could we make it so Notebooks are still populated with their output, instead of being wiped clean

Notebooks I could run are populated, so the only one that has no output is the function profiling one. I'll update the PR with your link, thanks

valschneider:
> Notebooks I could run are populated, so the only one that has no output is the function profiling one. I'll update the PR with your link, thanks

I thought there were more than one but they all look good now :)

lisa/wlgen/workload.py
```diff
@@ -72,24 +72,22 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False, value=42)
...
-
-        self.res_dir = res_dir
-
         self.command = None
         self.output = ""
-        self.run_dir = self.te.target.working_directory
+        self.run_dir = self.target.working_directory
+
+        res_dir = res_dir if res_dir else target.get_res_dir(
+            name='{}-{}'.format(self.__class__.__qualname__, name)
```
valschneider: We could probably drop the "name" requirement at this point, but seeing how big this PR is I'll do that separately.

PR #677, opened 2018-12-06
misc fixes/improvements

douglas-raillard-arm:


lisa/env.py
```diff
@@ -61,10 +61,10 @@ class TargetConf(MultiSrcConf, HideExekallID):
     .. code-block:: python
 
         TargetConf({{
```
valschneider: Hang on a sec, why is that a nested dict? I must have missed something because I always create my TargetConfs like so
```python
conf = TargetConf({
    "foo" : bar
})
```

lisa/__init__.py
```diff
@@ -1,3 +1,16 @@
 from lisa.version import __version__
 
+# Put all the code inside one function, to allow easy cleanup of the namespace
+# at the end. We definitely don't want to expose these things to the outside
+# world
+def f():
+    import os
+    if os.getenv('LISA_DO_NOTEBOOK_SETUP'):
```
valschneider: Could we have that done regardless of notebooks? Say I have a script that uses TestEnv.from_cli(), it'd be nice if the logging setup was done for me. The answer probably contains the words "CI", but I'm asking anyway :p

lisa/tests/kernel/scheduler/eas_behaviour.py
```diff
@@ -476,7 +476,7 @@ class TwoBigThreeSmall(EASBehaviour):
     @classmethod
     def get_rtapp_profile(cls, te):
         # 50% of the smallest CPU's capacity
-        small_duty = cls.unscaled_utilization(cls.min_cpu_capacity(te), 50)
+        small_duty = cls.unscaled_utilization(cls.min_cpu_capacity(te), 25)
```
valschneider: Would be nice to explain why that change is done (IIRC it's to match the values used in master).

lisa/env.py
douglas-raillard-arm: No, it's the way to have a "{" since that string is post-processed for the placeholder.

lisa/__init__.py
douglas-raillard-arm: No, setting up logging is the job of the application, not of libraries. If for example devlib was messing with logging configuration, that would quickly be an issue. Libraries just use the logging functions, that's not up to them to decide what to do with the result

lisa/tests/kernel/scheduler/eas_behaviour.py
douglas-raillard-arm: The reason is to make it smaller. That is indeed closer to what was in master, although it cannot be equal, since it now depends on the board. I cannot really see any fundamental reason why EAS would be allowed to not give the expected result for 50, so it's more to check that there are less issues with that value than another. At the end of the day, both are arbitrary, and there is a good chance the test would be supposed to pass with both all the time.

douglas-raillard-arm:
I can add a sequence number if you prefer, but given the amount of unrelated fixup commits I've spitting out these days, I'm clearly not going to maintain a separate branch and PR for each, with a local integration branch merging all of them in order to test them. I can split commits touching parts that are supposed to be "stable" from the rest though, but so far there was no such classifications of the tree.

lisa/env.py
valschneider: Gotcha, is that why the indentation is now doubled?

valschneider:
> I can add a sequence number if you prefer, but given the amount of unrelated fixup commits I've spitting out these days, I'm clearly not going to maintain a separate branch and PR for each, with a local integration branch merging all of them in order to test them. I can split commits touching parts that are supposed to be "stable" from the rest though, but so far there was no such classifications of the tree.

A sequence number wouldn't bring anything to the human-readableness of the PR.

I've always had a new branch per PR and that has worked fine or me. You should try to split up your changes between relevant PRs as much as possible. If you have somewhat unrelated changes in the lot (e.g. you add some feature in a PR and fix some bugs you found out while working on that feature) that's fine, just name the PR after the main topic.

You could name it something like "Add a LISA setup for jupyter + fixes". At least I can associate changes to this ("oh yeah that came with the jupyter setup").

lisa/env.py
douglas-raillard-arm: No, that looks like a silent rebase issue, I'll make the indentation consistent

lisa/tests/kernel/scheduler/eas_behaviour.py
valschneider: I have 0 issue with the change, I know why we added it - I just meant this kind of change should be explained in the commit ("This brings the test behaviour in line with what we have in lisa/master" would be enough).

PR #681, opened 2018-12-07
Validate hotplug

douglas-raillard-arm:


lisa/tests/kernel/hotplug/torture.py
```diff
@@ -27,6 +28,9 @@
 from lisa.env import TestEnv
 from lisa.utils import ArtifactPath
 
+class CPUHPSequenceError(Exception):
+    pass
+
 class HotplugTorture(TestBundle):
```
qperret: If we're going to extend this to execute other hotplug tests, I would suggest to create a `HotplugTest` parent class, and have `HotplugTorture` as a child. Also, `_random_cpuhp_seq` should be renamed to a more generic name such as `_hotplug_sequence` or something along those lines. I guess that function should raise `NotImplementedError` in the `HotplugTest`, and be overridden by `HotplugTorture`, or something like this.

lisa/tests/kernel/hotplug/torture.py
```diff
@@ -36,6 +40,55 @@ def __init__(self, plat_info, target_alive, hotpluggable_cpus, live_cpus):
         self.hotpluggable_cpus = hotpluggable_cpus
         self.live_cpus = live_cpus
 
+    @classmethod
+    def _check_cpuhp_seq_consistency(cls, nr_operations, hotpluggable_cpus,
+                                     max_cpus_off, sequence):
```
qperret: I'm wondering if we actually want this. I don't think we want it in the parent `HotplugTest` class I mentioned above, simply because trying to offline already offline CPUs could actually be a valid test case, for example.

lisa/tests/kernel/hotplug/torture.py
```diff
@@ -49,6 +102,7 @@ def _random_cpuhp_seq(cls, nr_operations,
         plugged-off before (and vice versa). Moreover the state of the CPUs
         once the sequence has completed should the same as it was before.
 
+        FIXME: is that actually still true ?
```
qperret: Good point, I don't know. I'll have a second look -- this is a bit ugly.

lisa/tests/kernel/hotplug/torture.py
```diff
@@ -37,50 +41,74 @@ def __init__(self, plat_info, target_alive, hotpluggable_cpus, live_cpus):
...
+            2-tuples (CPU and hot plug way)
+        :type sequence: Sequence
+
+        """
+        if len(sequence) != nr_operations:
+            raise CPUHPSequenceError('{} operations requested, but got {}'.fromat(
+                nr_operations, len(sequence)
+            ))
+
+        # Assume als CPUs are plugged in at the beginning
```
valschneider:
```suggestion
        # Assumes all CPUs are plugged in at the beginning
```

PR #685, opened 2018-12-07
lisa: Do some implicit notebook setup

douglas-raillard-arm:
When the lisa package is imported with LISA_DO_NOTEBOOK_SETUP
environment variable is set, some default setup code will be run. This
currently only calls utils.setup_logging() to save some boilerplate in
notebooks.

### DIFF
--- a/lisa/__init__.py
+++ b/lisa/__init__.py
@@ -1,3 +1,16 @@
 from lisa.version import __version__
 
+# Put all the code inside one function, to allow easy cleanup of the namespace
+# at the end. We definitely don't want to expose these things to the outside
+# world
+def f():
+    import os
+    if os.getenv('LISA_DO_NOTEBOOK_SETUP'):
+        from lisa.utils import jupyter_notebook_setup
+        jupyter_notebook_setup()
+
+f()
+del f
+
+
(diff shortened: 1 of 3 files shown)


douglas-raillard-arm:
looks like it needs some more tricks to actually work, this PR does not really work

douglas-raillard-arm:
Now fixed

douglas-raillard-arm:
> then it might be confusing if the notebook is ported as a pure Python script

The thing I also dislike with that approach is that even running a notebook "directly" (i.e. running jupyter directly without lisa-jupyter) will not run that initialization. Since notebooks are self contained (just like a script), it may be a better idea to just factorize the boilerplate like I did, but then call that explicitly. This way, we can add more content to that boilerplate without updating all notebooks.

valschneider:
> it may be a better idea to just factorize the boilerplate like I did, but then call that explicitly. This way, we can add more content to that boilerplate without updating all notebooks.

Yeah that'd work, although for now we just have a call to `setup_logging()` so that can be postponed to if/when we'd need to do more (which I sincerely hope won't happen).

valschneider:
Let's close it to tidy up the PR board - it'll still be visible in the closed section if we ever want to revive it.

PR #688, opened 2018-12-10
Cleanup and add self tests for utils.MultiSrcConf

douglas-raillard-arm:
* Remove redundant information exposed in `YAML_MAP_TOP_LEVEL_KEY` and the actual top-level key descriptor in `STRUCTURE`
* Add self tests for `utils.MultiSrcConf` (along with `TargetConf` and `PlatformInfo`)

### DIFF
--- a/lisa/tests/lisa/test_conf.py
+++ b/lisa/tests/lisa/test_conf.py
@@ -1,0 +1,169 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2018, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+import os
+import copy
+from unittest import TestCase
+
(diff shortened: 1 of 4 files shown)


lisa/tests/lisa/test_conf.py
```diff
@@ -0,0 +1,169 @@
...
+from lisa.tests.lisa.utils import StorageTestCase, HOST_PLAT_INFO, HOST_TARGET_CONF
+
+""" A test suite for the MultiSrcConf subclasses."""
+
+
+class TestMultiSrcConfBase:
+    """
+    A test class that exercise various APIs of MultiSrcConf
+    """
+    def test_serialization(self):
```
valschneider: So at the moment all of our test methods have a one-liner docstring, and this is what nosetest will print when there are failures.

e.g.:

https://github.com/ARM-software/lisa/blob/next/lisa/tests/lisa/test_test_bundle.py#L61
gives
ERROR: Test that creating a bundle from a target works

Removing the docstring gives me:
ERROR: test_target_init (lisa.tests.lisa.test_test_bundle.BundleCheck)

I was about to suggest adding a docstring to your test methods, but now I'm wondering if having no docstring just doesn't make the nosetest ouput clearer...

lisa/utils.py
```diff
@@ -440,37 +433,41 @@ def from_map(cls, mapping):
     def from_yaml_map(cls, path):
         """
         Allow reloading from a plain mapping, to avoid having to specify a tag
-        in the configuration file. The content is hosted under a top-level key
-        specified in :attr:`YAML_MAP_TOP_LEVEL_KEY`.
+        in the configuration file. The content is hosted under the top-level
+        key specified in :attr:`STRUCTURE`.
```
valschneider: That should be
```suggestion
        key specified in ``STRUCTURE``.
```
(the ref can't be resolved)

douglas-raillard-arm: ok, if we have no better way to refer to class attributes let's use the same convention as instance ones

PR #689, opened 2018-12-11
Refactor trace event checking

douglas-raillard-arm:
* Extract event checking infrastructure from analysis, so it can be used in tests as well.
* Improve it so it can cope with expected logical combinations of events (AND and OR)
* Remove a bit of syntactic overhead in the common case
* Allow easy combination of required events from different sources (other methods)

Also do some cleanup in trace.py:
* remove camel case and name mangling in method names to align with the rest of the code


lisa/tests/kernel/scheduler/misfit.py
```diff
@@ -210,6 +210,7 @@ def _trim_state_df(self, state_df):
         return Trace.squash_df(state_df, self.start_time,
                                state_df.index[-1] + state_df.delta.values[-1], "delta")
 
+    @requires_events('sched_switch')
```
valschneider: Those should be something like
```python
@requires_events('sched_switch', TasksAnalysis.df_task_states.required_events)
```

lisa/tests/lisa/test_events_checkers.py
```diff
@@ -0,0 +1,68 @@
...
+            (True, or_2),
+            (False, OrTraceEventChecker.from_events(['arthur', 'lancelot'])),
+
+            (True, AndTraceEventChecker.from_events([and_1, or_1])),
+            (True, AndTraceEventChecker.from_events([and_1, or_2])),
+            (False, AndTraceEventChecker.from_events([and_1, and_2])),
+        ]
+
+
+        for expected_success, checker in checker_list:
```
valschneider: IIUC we're aggregating tests inside a single nosetest test. Shouldn't those above be split into their own tests methods?

valschneider:
> I considered doing that, but the extra boilerplate did not seem appealing. Since I print the checker's expression before testing, if one fails you still know which one failed, so it should not be a big issue

I think reading the test's outcome isn't such an issue, it's more about conforming to the usual test structure

lisa/tests/lisa/test_events_checkers.py
douglas-raillard-arm: I split with one class per checker, so at least the x1.5 size increase brings the ability to easily test other methods than just `check_events()`

PR #690, opened 2018-12-11
conf: move MultiSrcConf into lisa/conf.py

douglas-raillard-arm:
Split MultiSrcConf and its friends in their own module, since it has
become a chunky set of classes.


valschneider:
The Travis error seems to be saying we need to add a sphinx directive for `lisa.conf` somewhere.

PR #691, opened 2018-12-12
Reintroduce CPU migration tests

valschneider:


lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+        Get the per-phase average CPU utilization expected from the rtapp profile
+
+        :returns: A dict of the shape {cpu : {phase_id : expected_util}}
+        """
+        cpu_util = {cpu : {phase_id : 0 for phase_id in range(self.nr_phases)}
+                    for cpu in self.cpus}
+
+        for task in self.rtapp_profile.values():
+            for phase_id, phase in enumerate(task.phases):
+                cpu_util[phase.cpus[0]][phase_id] += UTIL_SCALE * (phase.duty_cycle_pct / 100)
```
derkling: Should not be more safe as:
```python
+= (UTIL_SCALE * phase.duty_cycle_pct) / 100
```

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+        Get the per-phase average CPU utilization read from the trace
+
+        :returns: A dict of the shape {cpu : {phase_id : trace_util}}
+        """
+        cpu_util = {cpu : {phase_id : 0 for phase_id in range(self.nr_phases)}
+                    for cpu in self.cpus}
+        df = self.trace.analysis.load_tracking.df_cpus_signals()
+        sw_df = self.trace.df_events("sched_switch")
+
+        phase_start = sw_df[sw_df.next_comm == list(self.rtapp_profile.keys())[0]].index[0]
```
derkling: `index.min()` ?

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+
+        for phase in range(self.nr_phases):
+            # Start looking at signals once they should've converged
+            start = phase_start + UTIL_AVG_CONVERGENCE_TIME_S
+            # Trim the end a bit, otherwise we could have one or two events
+            # from the next phase
+            end = phase_start + self.phases_durations[phase] * .9
```
derkling: Since we have:
```python
PHASE_DURATION_S = 3 * UTIL_AVG_CONVERGENCE_TIME_S
```
I would say it could be just good to set:
```python
end = phase_start + UTIL_AVG_CONVERGENCE_TIME_S
```

derkling: Actually, giving the following code, where you keep computing (end-start), why not computing directly here:
```python
stable_interval = end - start
```
and use it in L826 ?

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+        return res
+
+class OneTaskCPUMigration(CPUMigrationBase):
+    """
+    Some tasks on two big CPUs, one of them migrates in its second phase.
+    """
+
+    @classmethod
+    def get_migration_cpus(cls, te):
+        cpu_capacities = te.plat_info["cpu-capacities"]
```
derkling: Wondering if we should add in some check/errors/failure in case this info is not available.
Not sure if we have any other pre-check before executing a test that all the required data are there... but I would expect in that case we expose somehow what are the requirements for a test... and we don't do that here AFAIKS.

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+
+        return [cpu for cpu, capa in cpu_capacities.items()
```
derkling: why not just "cap" ? :)

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+                if capa == max(cpu_capacities.values())][:2]
```
derkling: Can we the list comprehension and the return in two lines?

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+
+    @classmethod
+    def check_from_testenv(cls, te):
+        super().check_from_testenv(te)
+
+        cpus = cls.get_migration_cpus(te)
+        if not len(cpus) == 2:
+            raise RuntimeError("This workload requires 2 CPUs of highest capacity")
```
derkling: Mmm... this will not work on tri-gear systems with just one big CPU.
In principle, we can run on 2 medium, provided we configure the tasks wrt their capacity... not sure, do we still have this change with the new wlgen class?

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+
+        for task in ["migrating", "static0", "static1"]:
+            # An empty RTATask just to sum phases up
+            profile[task] = RTATask()
+
+        for i in range(2):
+            # A task that will migrate to another CPU
+            profile["migrating"] += Periodic(
+                duty_cycle_pct=20, duration_s=cls.PHASE_DURATION_S,
+                period_ms=cls.TASK_PERIOD_MS, cpus=[cpus[i]])
```
derkling: can we pass in the calibration cpu here ?
In that case `get_migration_cpus()` could be update to return the highest capacity pair of CPUs in the system...

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -203,7 +203,9 @@ def _test_task_signal(self, signal_name, allowed_error_pct,
                           trace, cpu, task_name, capacity=None):
         exp_util = self.get_expected_util_avg(trace, cpu, task_name, capacity)
         signal_df = self.get_task_sched_signals(trace, cpu, task_name, [signal_name])
-        signal_mean = signal_df[UTIL_AVG_CONVERGENCE_TIME_S:][signal_name].describe()['mean']
+        signal = signal_df[UTIL_AVG_CONVERGENCE_TIME_S:][signal_name]
+
+        signal_mean = area_under_curve(signal) / (signal.index[-1] - signal.index[0])
```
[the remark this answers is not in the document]
valschneider: Indexes are monotonic (https://github.com/ARM-software/trappy/commit/7bb0bd9f29711e2a40e88fa33ae04fcb7df94e56), and I have no idea if Pandas/Numpy records this for optimizing min/max. So directly using an index seems like the better choice to me.

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+    Since we sum up the utilization of each task, make sure not to overload the
+    CPU - IOW, there should always be some idle cycles.
+
+    The code assumes all tasks have the same number of phases, and that those
+    phases are all aligned.
+    """
+
+    PHASE_DURATION_S = 3 * UTIL_AVG_CONVERGENCE_TIME_S
+    """
+    The duration of a single phase
```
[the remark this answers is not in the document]
valschneider: Nope, docstrings are always below definitions.

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: Welcome to the wonderful world of Python3 where `20 / 100` == `0.2` (`20 // 100` == `0`)

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: Same comment as above.

lisa/tests/kernel/scheduler/load_tracking.py
douglas-raillard-arm: That key should always be there since it is computed from the target. If it is not, it will fail, saying that cpu-capacities cannot be found anyway. If we can provide more information by failing early, we could add more pre-tests, but otherwise I don't really see the point (it will never be silent in any case).

lisa/tests/kernel/scheduler/load_tracking.py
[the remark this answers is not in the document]
douglas-raillard-arm: autodoc gives the choice between comments above or a string literal below. We went the string way, to be consistent with functions/classes/modules
http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: > Since we have:
> 
> ```python
> PHASE_DURATION_S = 3 * UTIL_AVG_CONVERGENCE_TIME_S
> ```
> 
> I would say it could be just good to set:
> 
> ```python
> end = phase_start + UTIL_AVG_CONVERGENCE_TIME_S
> ```

IMO the larger the interval is, the better it is in terms of averages.



> Actually, giving the following code, where you keep computing (end-start), why not computing directly here:
> 
> ```python
> stable_interval = end - start
> ```
> 
> and use it in L826 ?

Can do.

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: Because otherwise my mind [gets sidetracked](https://www.youtube.com/watch?v=1dWIPUuTPB4) whenever I read this line.

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: Isn't that already two lines? Or do you mean save the list in a variable and then do the return?

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: Yeah so that's directly copied from the test we have in master (not the sanity check, but the behaviour).

Offline discussion with @derkling suggests that we could just try to find 2 CPUs of same capacities and scale the rtapp workload using https://github.com/ARM-software/lisa/blob/next/lisa/tests/kernel/scheduler/eas_behaviour.py#L374

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: As discussed offline, I'd rather change the duty cycles than affect the calibration, because otherwise 20% duty_cycle != 20% utilization.

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: Mmm yeah, I did add this hook: https://github.com/ARM-software/lisa/blob/next/lisa/tests/kernel/test_bundle.py#L264, but I haven't been a perfect citizen in that I don't think I've used it everywhere I should've...

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+
+    The idea here is to run several rt-app tasks and to have them pinned to
+    a single CPU for a single phase. They can change CPUs in a new phase,
+    and we can then inspect the CPU utilization - it should match the
+    sum of the utilization of all the tasks running on it.
+
+    **Design notes:**
+
+    Since we sum up the utilization of each task, make sure not to overload the
+    CPU - IOW, there should always be some idle cycles.
```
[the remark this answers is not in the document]
valschneider: Yeah, maybe...

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -735,3 +735,218 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+
+        return cpu_util
+
+    def get_trace_cpu_util(self):
+        """
+        Get the per-phase average CPU utilization read from the trace
+
+        :returns: A dict of the shape {cpu : {phase_id : trace_util}}
+        """
+        cpu_util = {cpu : {phase_id : 0 for phase_id in range(self.nr_phases)}
```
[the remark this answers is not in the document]
valschneider: I usually find pre-initing easier on the eyes than using `setdefault()`, but that's personal preference :)

lisa/tests/kernel/scheduler/load_tracking.py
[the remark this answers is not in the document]
derkling: Yes, we had a sanitization step somewhere in trappy which grants that... I just find the code above more "robust" and (maybe) a bit more easy to read... although I've also personally always used your same style. That was just to possibly consider to change our style... but not sure if the "robustness" argument is "robust" enough ;)

lisa/tests/kernel/scheduler/load_tracking.py
derkling: Yes, I guess it was P3 magic... however, we should maybe keep consistency with the Linux kernel world we come from, and the interpretation above is what we use in kernel space... just in case we want to be more friendly with some "low-level guys" looking at our tests ;)

lisa/tests/kernel/scheduler/load_tracking.py
derkling: Fine... although I would have pick a different name (e.g. `cpu_cap`)...

lisa/tests/kernel/scheduler/load_tracking.py
derkling: the second... since with all those parenthesis my mind [gets sidetracked](https://www.youtube.com/watch?v=1dWIPUuTPB4) ...

lisa/tests/kernel/scheduler/load_tracking.py
valschneider: Yeah I feel you, I hated the guts out of that division change when I started doing Python3 stuff. Thing is, you can't hide it all the times with parentheses, so might as well embrace it.

valschneider:
So, I introduced some wilder changes:
* Added a `capacity-classes` entry in plat_info (@douglas-raillard-arm, is there a way to cleanly mark this as "derived" information? There's no need to serialize it as it's computed from CPU capacities)
* Moved `EASBehaviour.unscaled_utilization` to `RTATestBundle` to use it in `xCPUMigration`
* `xCPUMigration` now scans through `capacity-classes` to get 2 CPUs of the same capacity, and uses `unscaled_utilization`

douglas-raillard-arm:
> is there a way to cleanly mark this as "derived" information? There's no need to serialize it as it's computed from CPU capacities

If that is always derived from there, that should probably turned into a function that derives the information from the key that already exists. Having keys derived from other keys brings coherence issues. The user has final control of every key, so it's better to keep the set of controls as small as possible, without redundance when possible (in that case, it is possible, unlike deriving everything from an EM that may not be present for example).

douglas-raillard-arm:
If we really want to have that computed information travel along the `MultiSrcConf` instances, we could introduce a way to define derived keys, that won't be settable from the file and that will not be dumped either. This way, we avoid confusing state duplication, and we can fully reflect that in the documentation. Its source name could be derived from the source names of all "mother" keys, so we still keep track of where things are coming from.

derkling:
I like better the first solution proposed by @douglas-raillard-arm ... if something is derived, what can be a good reason to not compute it using just a (memoized) function ?

lisa/tests/kernel/scheduler/eas_behaviour.py
```diff
@@ -431,7 +393,7 @@ def test_task_placement(self, energy_est_threshold_pct=20, nrg_model:EnergyModel
     @classmethod
     def get_rtapp_profile(cls, te):
         # 50% of the smallest CPU's capacity
-        duty = cls.unscaled_utilization(cls.min_cpu_capacity(te), 50)
+        duty = cls.unscaled_utilization(te, te.plat_info["capacity-classes"][0][0], 50)
```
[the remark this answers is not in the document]
valschneider: Yeeeeah when I wrote that change it striked me as extremely newcomer-unfriendly.

You have to read that `plat_info` access as "The first CPU of the first CPU capacity class" (they're sorted, so it's the group CPUs with smallest capacity).

I could turn that into
```python
@classmethod
def smallest_cpus(te):
    return te.plat_info["capacity-classes"][0]

duty = cls.unscaled_utilization(te, cls.smallest_cpus(te)[0], 50)
```

And write its sibling for biggest CPUs.

lisa/tests/kernel/scheduler/load_tracking.py
```diff
@@ -733,3 +735,244 @@ def test_load_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
...
+
+                # Just some verbose metric collection...
+                phase_str = "phase{}".format(phase)
+
+                expected_metrics[cpu_str].data[phase_str] = TestMetric(
+                    expected_cpu_util[cpu][phase])
+                trace_metrics[cpu_str].data[phase_str] = TestMetric(
+                    trace_cpu_util[cpu][phase])
+                deltas[cpu_str].data[phase_str] = TestMetric(
+                    -100 * (expected_cpu_util[cpu][phase] - trace_cpu_util[cpu][phase]) /
```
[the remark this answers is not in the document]
valschneider: You'll love the next version ;)

lisa/tests/kernel/scheduler/load_tracking.py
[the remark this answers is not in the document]
qais-yousef: I prefer using pandas.Series.mean() function. I find the `area_under_curve()/duration` usage convoluted. Besides the behavior to re-sample and interpolate before it does the integration is semi magic. Most importantly around decay time where there's a big gap of no samples until the task wakes up again.

If it does fix a problem then we need to understand exactly why. Can you add a few debug messages to help with that? I suggest calculate using the 2 methods and print them out in a log.debug() or something.

By the way, why you had `___.describe()['mean']` instead of `___.mean()` in the removed line? Curious what's the difference.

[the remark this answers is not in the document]
valschneider: > I prefer using pandas.Series.mean() function. I find the `area_under_curve()/duration` usage convoluted. Besides the behavior to re-sample and interpolate before it does the integration is semi magic. Most importantly around decay time where there's a big gap of no samples until the task wakes up again.
> 

That was my feeling as well, so I went for `mean()` until I hit issues and reconsidered (see below).

> If it does fix a problem then we need to understand exactly why. Can you add a few debug messages to help with that? I suggest calculate using the 2 methods and print them out in a log.debug() or something.
> 

I did that and the difference is noticeable. With the integral mean I can safely set a 5% threshold (usual error % is < 2%), with the regular mean I need at least 10%. I had a chat with @derkling and the integration *is* the way to go - IIRC that can be verified on the PELT simulator. He'll 100% explain it better than I, but that integration is required.

> By the way, why you had `___.describe()['mean']` instead of `___.mean()` in the removed line? Curious what's the difference.

There's none, that wasn't necessary :)

valschneider:
Updated using derived keys.

lisa/platforms/platinfo.py
```diff
@@ -16,15 +16,37 @@
...
+    Derive the platform's capacity classes from the given conf
+
+    This is intended for the creation of the ``capacity-classes`` key of
+    :class:`PlatformInfo`.
+    """
+    cpu_caps = conf['cpu-capacities']
+    capacities = sorted(set(cpu_caps.values()))
+
+    classes = [
+        [cpu for cpu, cpu_capa in cpu_caps.items() if cpu_capa == capacity]
```
douglas-raillard-arm: The inner list comprehension needs to be sorted as well, as it depends on the iteration order of a dictionary, which cannot be relied upon.

Also, this can be simplified using:
```
def compute_capa_classes(conf): 
     cpu_caps = conf['cpu-capacities']
     return [
         sorted(cpu_group)
         for capa, cpu_group in utils.groupby(cpu_caps.items(), key=operator.itemgetter(1))
      ]
```
`operator.itemgetter()` just returns a function that will get the `N`th item of a sequence, in this case the 2nd item of the 2-tuples returned by `dict.items()`
Could be even worth adding a `group_by_value()` helper in `utils`:
```
def group_by_value(mapping, key_sort=lambda k: 0):
     return {
        val: sorted((k for k, v in key_group), key=key_sort)
         for val, key_group in groupby(mapping.items(), key=operator.itemgetter(1))
     }
```
And then:
```
def compute_capa_classes(conf): 
     return classes = [*group_by_value(conf['cpu-capacities'], key_sort=lambda k: k).values()]
```

lisa/tests/kernel/scheduler/misfit.py
```diff
@@ -127,16 +127,7 @@ def __init__(self, res_dir, plat_info, rtapp_profile):
...
-        cpu_classes = {}
-        for cpu, capacity in cpu_capacities.items():
-            if capacity not in cpu_classes.keys():
-                cpu_classes[capacity] = []
-
-            cpu_classes[capacity].append(cpu)
-
-        capacities = sorted(cpu_classes.keys())
-        self.cpu_classes = [cpu_classes[capacity] for capacity in capacities]
+        self.cpu_classes = plat_info['capacity-classes']
```
douglas-raillard-arm: may be better to refer to self.plat_info where it is actually needed. That allows easier refactoring (the code just depends on a `plat_info` to work, instead of a specific class), and allows better tracking what data is used as the input of each test by looking at the debug log (which logs access to `plat_info`).

lisa/tests/kernel/scheduler/misfit.py
```diff
@@ -127,16 +127,7 @@ def __init__(self, res_dir, plat_info, rtapp_profile):
         super().__init__(res_dir, plat_info, rtapp_profile)
 
         cpu_capacities = plat_info['cpu-capacities']
```
douglas-raillard-arm: this line can be removed as well

lisa/tests/kernel/scheduler/misfit.py
valschneider: I feel much more comfortable storing this value into a variable, especially since it's only ever used in this method - I now see it's only used in the `__init__()` so I could make it a local variable instead of an instance attribute .

douglas-raillard-arm: right, if it's just some local use then there is no issue. What we just want to avoid is ending with quite generic code, that ends up seemingly deeply tied to a class where in fact, it could easily be extracted and put in a helper (or a base class) if needed.

valschneider:
@douglas-raillard-arm All done

valschneider:
@derkling, wanna re-re-review this one? :D

PR #696, opened 2018-12-14
conf: Introduce derived configuration keys

douglas-raillard-arm:
* Allow using a key that is purely defined in term of other keys. That key cannot be added from a source, and is not shown when iterating over the `MultiSrcConf` instance. This avoids inconsistencies when adding it back as a source for example.

* minor cleanup in conf-related classes

### DIFF
--- a/lisa/tests/lisa/test_conf.py
+++ b/lisa/tests/lisa/test_conf.py
@@ -20,7 +20,7 @@ import os
 import copy
 from unittest import TestCase
 
-from lisa.conf import MultiSrcConf, KeyDesc, LevelKeyDesc, TopLevelKeyDesc, IntList
+from lisa.conf import MultiSrcConf, KeyDesc, LevelKeyDesc, TopLevelKeyDesc, IntList, DerivedKeyDesc
 from lisa.tests.lisa.utils import StorageTestCase, HOST_PLAT_INFO, HOST_TARGET_CONF
 
 """ A test suite for the MultiSrcConf subclasses."""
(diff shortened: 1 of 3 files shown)


lisa/tests/lisa/test_conf.py
```diff
@@ -68,10 +68,18 @@ def __init__(self, *args, **kwargs):
...
+
 INTERNAL_STRUCTURE = (
     KeyDesc('foo', 'foo help', [int]),
     KeyDesc('bar', 'bar help', [IntList]),
     KeyDesc('multitypes', 'multitypes help', [IntList, str, None]),
+    LevelKeyDesc('sublevel', 'sublevel help', (
+        KeyDesc('subkey', 'subkey help', [int]),
+    )),
+    DerivedKeyDesc('derived', 'derived help', [int],
+        [['foo'], ['bar'], ['sublevel', 'subkey']], compute_derived),
```
valschneider: (commenting here but this goes to the whole derived key concept)
I'm wondering if there is an actual value in specifying subkeys. IMO sublevels should be sufficient, otherwise we would start sprinkling duplicate key names between the derived key description and the computation method.

If we limit that to sublevels it would still be sensible without duplicating relevant keys.

douglas-raillard-arm: That is unfortunately necessary to:
* Have a clear documentation on what depends on what. Being sloppy on that sounds like a really bad idea. These derived keys cannot be set from the conf file, so the least we can do is to tell the user what needs to be modified to change it.
* Get a useful source name for the derived key. Otherwise, derived keys will become totally opaque and will break one of the core feature of the conf objects, and lisa-next in general: knowing where things come from, which used to be totally obscure.
* Know when the key should be considered as existing in the conf or not. If you iterate (or print) the conf, you will only see keys that have a value defined. Specifying a whole level means the compute function will barf if some keys are not present. There could be a specific exception to raise in that case, but it seems cleaner to just hide it until it can be computed.

valschneider: > Re-reading my last point, it seems there is a bug in that PR

https://en.wikipedia.org/wiki/Rubber_duck_debugging :)

PR #699, opened 2018-12-18
shell: Add check updates to setup.py files

douglas-raillard-arm:
Check that setup.py files have not been modified since last time
lisa-install was run.

Also provide a post-checkout hook for git to carry out the check on each
branch checkout.

### DIFF
--- a/tools/check-setuppy
+++ b/tools/check-setuppy
@@ -1,0 +1,89 @@
+#! /usr/bin/env python3
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2018, Arm Limited and contributors.
+#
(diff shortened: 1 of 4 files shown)


tools/check-setuppy
```diff
@@ -0,0 +1,89 @@
...
+LISA_HOME = os.getenv('LISA_HOME', '.')
+
+def call_git(*args):
+    return subprocess.check_output(('git', *args)).decode('utf-8')
+
+def get_git_sha1(ref):
+    return call_git('rev-parse', ref).strip()
+
+def main():
+    parser = argparse.ArgumentParser(description='''
```
valschneider: We use `textwrap.dedent()` in `TestEnv.from_cli()`, *I think* because otherwise the indentation gets included.

PR #701, opened 2018-12-18
buildroot: enable mdev

qais-yousef:
Use mdev to auto setup usb ethernet dongles.


valschneider:
@douglas-raillard-arm Is that something you used for the integration tests? AFAICT you're already using eth dongles with buildroot, right?

douglas-raillard-arm:
@valschneider We faced the issue on Hikey 960 where the USB network interface is detected a few seconds after the boot sequence completes. This means the dhcp client runs before there is any network interface, and when it actually comes up, the dhcp client is not run again. This solves the issue by getting an IP every time a network interface appears, so it's event-based and we don't miss it.

Issue #702, opened 2018-12-20
ImportError: No module named 'wa_results_collector'


vingu-linaro:
I tried to run sched-evaluation-full.ipynb with next branch but face error:

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-c8c2c71ef9f4> in <module>
      2 from IPython.display import display
      3 
----> 4 from wa_results_collector import WaResultsCollector
      5 import pandas as pd
      6 

ImportError: No module named 'wa_results_collector'

valschneider:
Hi Vincent,

wltests is probably the last block that hasn't been yet updated in lisa/next. It's still on my backlog, but there was an offline chat with @credp && @derkling suggesting that we could restructure it instead.

I merged https://github.com/ARM-software/lisa/pull/692 just today but I now realise I didn't move the wltests notebook into "deprecated". I'll see if I can just get it working instead.

vingu-linaro:
ok.
Our farm uses a python script based on sched-evaluation-full.ipynb (python2.7) to post process WAv3 results.
IIUC, you don't accept new changes in master branch so my plan was to move next branch and push our script (at least the .ipynb version).

valschneider:
I see, thanks for thinking about upstreaming it!

I'll have a look and see if I can unblock that shortly.

valschneider:
So looking into it it seems like `wa_results_collector` should mostly work, I think the issues I'm hitting are with the notebook code itself.

All imports in lisa/next have to be absolute, so to import `wa_results_collector` you should do:
```python
from lisa.wa_results_collector import WaResultsCollector
```
And see how that fares with your script. I'll keep looking at the notebook at my side and fix any issues I find.

valschneider:
https://github.com/ARM-software/lisa/pull/703 should get things working in next.

PR #703, opened 2018-12-21
Fix wa_results_collector + wltests notebook, get last master changes

valschneider:
I had a geekbench wltests folder somewhere, so I used that to test the collector & notebook.

While at it, carry over the last few fixes that went into master that we didn't get in next.

closes https://github.com/ARM-software/lisa/issues/702

### DIFF
--- a/shell/lisa_shell
+++ b/shell/lisa_shell
@@ -396,7 +396,7 @@ EOF
 
     # Check for required external dependencies
 
-    if which realpath &>/dev/null; then
+    if ! which realpath &>/dev/null; then
 cat <<EOF
 ERROR: this script requires the realpath binary!
 
@@ -410,7 +410,7 @@ EOF
 function lisa-wltest-series {
     # Ensure the wltest environment has been configured, and get the relative
     # patch loaded in the environment
-    lisa-wltest-init || exit -1
+    lisa-wltest-init || return -1
 
     # Run the build's provided test_series
     $WLTEST_HOME/test_series "$@"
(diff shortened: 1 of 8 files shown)


valschneider:
I actually got to run wltests and had some issues with the shell, added a commit to fix those.

shell/lisa_shell
```diff
@@ -396,7 +396,7 @@ EOF
 
     # Check for required external dependencies
 
-    if which realpath &>/dev/null; then
+    if [[ -z $(which realpath) ]]; then
```
douglas-raillard-arm: If we are not interested in stderr, that form can result in a slightly shorter/less clunky line:
```
if not which realpath &>/dev/null; then 
```

shell/lisa_shell
```diff
@@ -396,7 +396,7 @@ EOF
```
douglas-raillard-arm: While we are at it, I'm not sure calling `lisa-venv-activate` is really needed anymore in `lisa-wltest-init`, since that is already done when sourcing lisa_shell itself (bottom of the file). That probably dates back to before we revamped the install process.

shell/lisa_shell
valschneider: Seeing as we have `-z`/ `-x` checks everywhere, I figured I'd try to keep this in line.
Also, that `not` command isn't standard, is it?

douglas-raillard-arm: Right, `not` seems to be provided by llvm package on my system, but `!`  definitely is [1]. It only checks the return code of the command, which is exactly what we need here (taken from `man which`):
```
RETURN VALUE
       Which returns the number of failed arguments, or -1 when no `program‐
       name' was given.
```
`-z` checks if a string is empty, which seems equivalent here.

[1]  "2.9.2 Pipelines" in http://pubs.opengroup.org/onlinepubs/009696899/utilities/xcu_chap02.html

valschneider:
@douglas-raillard-arm I think I addressed your comments.

PR #707, opened 2019-01-14
doc: kernel_tests: Add a "Running tests" section

valschneider:
Running `lisa_tests` or `exekall` doesn't give any hints as to how to run LISA tests - I have an older run somewhere than I always go back to.

Add a small section to point out how to run the tests.


douglas-raillard-arm:
Marking as stale until upcoming changes to exekall are pushed/merged

PR #709, opened 2019-01-14
lisa_shell: Check PYTHONPATH

douglas-raillard-arm:
Warn if PYTHONPATH is not empty, since it can interfere with venv. In
the worst case, it can lead to using a wrong version of e.g. bart that
would be installed somewhere else, leading to strange failures in LISA.


douglas-raillard-arm:
@deggeman This should warn in the situation you encountered recently (bart installed in /opt/ and PYTHONPATH set in your .bashrc)

PR #711, opened 2019-01-15
tools: check-setuppy: Display list of commits

douglas-raillard-arm:
Show the list of "offending" commits that modified setup.py files.

### DIFF
--- a/tools/check-setuppy
+++ b/tools/check-setuppy
@@ -77,8 +77,9 @@ def main():
         relevant_commits = call_git('rev-list', rev_range, '--', args.filename_pattern)
 
         if relevant_commits:
-            print('{} files have been modified, re-run lisa-install to update dependencies.'.format(
-                args.filename_pattern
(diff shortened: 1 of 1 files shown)


tools/check-setuppy
```diff
@@ -77,8 +77,9 @@ def main():
         relevant_commits = call_git('rev-list', rev_range, '--', args.filename_pattern)
 
         if relevant_commits:
-            print('{} files have been modified, re-run lisa-install to update dependencies.'.format(
-                args.filename_pattern
+            print('{} files have been modified in the following commits, re-run lisa-install to update dependencies:\n{}'.format(
+                args.filename_pattern,
+                relevant_commits,
```
valschneider: Will this be properly formatted, or do we need a quick `'\n'.join(...)` ?

douglas-raillard-arm: That's the output of git-rev-parse, so it's already formatted

PR #713, opened 2019-01-17
exekall's (early) spring cleanup

douglas-raillard-arm:
* Deep refactoring/cleanup of exekall
* rework of lisa-test, so it's slightly easier to use than exekall directly (see update to shell/README.txt)
* Generate the list of kernel tests in the Sphinx documentation using exekall. This allows tying things together as ID components are clickable.

### DIFF
--- a/doc/kernel_tests.rst
+++ b/doc/kernel_tests.rst
@@ -21,6 +21,16 @@ In our case, the data usually consists of
 `Ftrace <https://www.kernel.org/doc/Documentation/trace/ftrace.txt>`_ traces
 that we then postprocess using :mod:`trappy`.
 
+Available tests
+===============
+
+The following tests are available. They can be used as:
+  * direct execution using ``lisa-test`` command (``LISA shell``) and ``exekall``
+  * the individual classes/methods they are composed of can be used in custom
+    scripts/jupyter notebooks (see ipynb/tests/synthetics_example.ipynb)
+
+.. include:: test_list.rst
+
 Writing tests
 =============
 
@@ -59,17 +69,17 @@ execution of `rt-app <https://github.com/scheduler-tools/rt-app>`_ workloads.
 It is very useful for scheduler-related tests, as it makes it easy to create
 tasks with a pre-determined utilization.
 
+API
+===
+
 Base API
-========
+++++++++
 
 .. automodule:: lisa.tests.kernel.test_bundle
    :members:
    :private-members: _from_testenv
 
 .. TODO:: Make those imports more generic
-
-Implemented tests
-=================
 
 Scheduler tests
 +++++++++++++++
@@ -107,3 +117,4 @@ Cpufreq tests
 
 .. automodule:: lisa.tests.kernel.cpufreq.sanity
    :members:
+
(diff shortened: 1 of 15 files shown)


doc/kernel_tests.rst
```diff
@@ -71,6 +71,10 @@ Base API
 Implemented tests
 =================
 
+The following tests are implemented:
+
+.. include:: test_list.rst
```
valschneider: That's pretty nifty! However this makes me wonder if we shouldn't change the hierarchy a bit. Maybe something like this?

```
* Introduction
* Writing tests
** Basics
** Where to start
(nothing changed up till now)

* Available tests
(this new fancy list)

* API
(base + implemented tests API)
```

shell/README.txt
```diff
@@ -33,5 +33,9 @@ lisa-report  - Pretty format results of last test
 .:: Test commands
 --------------------------------------
 
-lisa-test    - Run tests and assert behaviours
+lisa-test    - Run tests and assert behaviours.
+               This is just a wrapper around exekall that selects all tests
+               modules and use positional arguments as --select patterns.
+               Also the default configuration file will be used if available
```
valschneider: Just to make it absolutely clear to the user - could we mention that the conf file is "target_conf.yml"?

doc/kernel_tests.rst
douglas-raillard-arm: yes, I was also wondering if we should put that list a bit more under the spotlight somehow. A user may not necessarily be interested in the APIs directly, but more about "what can I run with lisa-test/exekall and what can I expect to test using that LISA software ?"

doc/kernel_tests.rst
```diff
@@ -21,6 +21,53 @@ In our case, the data usually consists of
...
+===============
+
+The following tests are available. They can be used as:
+  * direct execution using ``lisa-test`` command (``LISA shell``) and ``exekall``
+  * the individual classes/methods they are composed of can be used in custom
+    scripts/jupyter notebooks (see ipynb/tests/synthetics_example.ipynb)
+
+.. include:: test_list.rst
+
+Scheduler tests
```
valschneider: I'd like to keep the "generic" bits before the API description - the "Writing tests" section is sandwiched between the tests implementation and the base API.

I'd rather see something like this:

```
* Introduction
* Available tests
* Writing tests
* API
** Base API
** Scheduler tests
** blah blah
```

lisa/tests/kernel/scheduler/eas_behaviour.py
```diff
@@ -372,6 +373,41 @@ def test_task_placement(self, energy_est_threshold_pct=5, nrg_model:EnergyModel=
         res.add_metric("energy threshold", threshold, 'bogo-joules')
         return res
 
+    def test_slack(self, negative_slack_allowed_pct=15) -> ResultBundle:
```
valschneider: Agreed with the change, we might want to have that in a RTATestBundle helper/decorator class but since it's the only user for now we can leave it as is.

douglas-raillard-arm: actually, we can keep it in RTATestBundle if it has any value there, and just not annotate it. Then we put an annotated stub in eas_behavior and off we go

doc/kernel_tests.rst
douglas-raillard-arm: sounds good, the new version has that order. The test list acts as an index at the beginning, which allows us to have a more logical order for the actual descriptions.

PR #715, opened 2019-01-22
Cleanup in lisa shell

douglas-raillard-arm:


derkling:
LGMT

qperret:
Sorry I'm a bit late for the fight but this didn't update the `lisa-help` command which still reports the old stuff, including `lisa-report`, `lisa-ipython`, and so on ... :/

douglas-raillard-arm:
I'll address that soon, the current plan is to turn that into a chunk of reStructuredText and have Sphinx turn it into a manpage, and also include it in the main documentation.

PR #720, opened 2019-01-24
doc: Fix autodoc options for Sphinx < 1.8

douglas-raillard-arm:


douglas-raillard-arm:
Marking as stale as we may be able to choose the right version of Sphinx in readthedocs

douglas-raillard-arm:
fixed by https://github.com/ARM-software/lisa/pull/721

PR #726, opened 2019-01-25
doc: Comprehensive user & contributor guides

valschneider:
The LISA transition guide is a cleaned up version of the changelog I had written when the `next` branch was first created. It's probably missing some bits (we did change a *lot* of things), do point them out. Also, if you've struggled with some specific points, do bring it up so we can make sure others don't go through the same pain. You can have a preview of it here: https://my-lisa.readthedocs.io/en/lisa-next/users_guide.html#transitioning-from-lisa-legacy

Contributor's guide needs some filling - just enough to get people going with subtrees and that should be it.


qais-yousef:
LGTM in general.

One issue I faced but I still have to re-test again is I get some errors when trying to run my notebooks on lisa/next. One annoying error was related to logging module. I'll compile a list and poke you :)

valschneider:
> LGTM in general.
> 
> One issue I faced but I still have to re-test again is I get some errors when trying to run my notebooks on lisa/next. One annoying error was related to logging module. I'll compile a list and poke you :)

Ah yes @douglas-raillard-arm mentioned (offline) the logging setup, I'll add that. Thanks!

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
++++++++
+
+LISA legacy used relative imports with a bit of dark magic to hold everything
+together. Say you want to import LISA the :class:`lisa.trace.Trace` class found
+in ``lisa/trace.py``, previously you would do it like so::
+
+  from trace import Trace
+
+However, relative imports are dangerous - did you know :class:`trace.Trace` exists
+in Python's standard library? This means that the above import would import done
```
douglas-raillard-arm: this sentence needs some fixes to be parseable :)
Also, `from trace import Trace` will still give the stdlib `Trace`, even in LISA environment, since that is not a relative import. Imports exist in 3 flavors:

1) explicit: full name from top-level package, e.g. `import lisa.foo.bar`
2) relative: equivalent to full name, but less obvious and potentially confusing error messages when they are broken, e.g. 'from . import bar' (assuming the current module is at the same level as `lisa.foo`)
3) implicit relative import: Python2-only evil, which introduces ambiguity, e.g. `import bar`. It is not obvious that `bar` is actually `lisa.foo.bar`, and it can shadow an stdlib module named `bar` (just like `trace.Trace` that was shadowed, as it already existed in Python 2). That is very nasty as it becomes impossible to import some modules from the stdlib, which is quite broken.

To sum up:
* really dangerous imports (implicit relative) are not possible in Python 3
* explicit relative imports are still not recommended, for code clarity, grepping and sensible error messages when something goes wrong.
* absolute imports are the way to go, unless there is a good justification the other types should be banned from the code base.

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+in a non-LISA environment would import something completely different!
+
+.. tip::
+
+  This can help you figure out what you are really importing:
+
+    >>> import trace
+    >>> print(trace.__file__)
```
douglas-raillard-arm: `__file__` may not exist on packages (if there is no `__init__.py`, so we could add as well:
```
print(package.__path__)
```

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+    /usr/lib/python3.5/trace.py
+
+We now enforce the use of absolute imports, which look like this::
+
+  from lisa.trace import Trace
+
+.. warning::
+
+  Do make sure you haven't kept some ``PYTHONPATH`` tweaking in your ``.bashrc``
+  that would lead to an older LISA/devlib/trappy being imported.
```
douglas-raillard-arm: s/would/could/

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+* ``"platform"`` is now ``"kind"``
+* ``"board"`` used to load some target-specific settings, which we got rid of.
+  The closest thing to it is ``"name"`` which is just a pretty-printing name and
+  has no extra impact.
+* You don't have to specify devlib modules to load anymore. All (loadable)
+  modules are now loaded. If you find some module too slow to load, you can
+  specify a list of modules to exclude.
+* LISA used to have ``target.config`` JSON file at its root. Its equivalent is
+  now ``target_conf.yml``, which is in YAML.
+
```
douglas-raillard-arm: should also mention `PlatformInfo`, since it is the place to modify for `rtapp` calib values. While at it, we can also mention `!var package.module.global` YAML tag we have to reference Python objects directly. This unifies the long-standing JSON-vs-module issue, e.g. `lisa/platforms/juno_r0.yml`

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+.. admonition:: Cool new feature
+
+   :class:`lisa.env.TestEnv` instances can now be easily be  created
+   :meth:`via CLI<lisa.env.TestEnv.from_cli>`!
```
douglas-raillard-arm: worth mentioning  `lisa.env.TestEnv.from_default_conf()`, which will read `target_conf.yml` (actually, whatever is set in `LISA_CONF` env var).

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+  wload = RTA.by_profile(te, "foo", profile, res_dir, calibration)
+
+Kernel tests
+++++++++++++
+
+The ``Executor`` from LISA legacy has been entirely removed, and a new test
+framework has been put in place. Tests are now coded as straight Python classes,
+which means they can be imported and executed in scripts/notebooks without any
+additionnal effort. See :ref:`kernel-testing-page`.
+
```
douglas-raillard-arm: could be worth mentioning the data-collection-then-process split approach.

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+We now enforce the use of absolute imports, which look like this::
+
+  from lisa.trace import Trace
+
+.. warning::
+
+  Do make sure you haven't kept some ``PYTHONPATH`` tweaking in your ``.bashrc``
+  that would lead to an older LISA/devlib/trappy being imported.
+
```
douglas-raillard-arm: could be worth mentioning the new `setup_logging()` import, since it will break in almost every lisa-legacy notebook.

doc/contributors_guide.rst
```diff
@@ -0,0 +1,13 @@
...
+
+Handling subtrees
+=================
+
+
+
+Contribution rules
+==================
+
+See `this document <https://github.com/ARM-software/lisa/blob/next/CONTRIBUTING.md>`__.
```
douglas-raillard-arm: not sure if that was already discussed, but github can render reStructuredText when named `.rst` (there are plain reST, no Sphinx extension). That way, we could `.. include::` the source here directly.

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+The ``Executor`` from LISA legacy has been entirely removed, and a new test
+framework has been put in place. Tests are now coded as straight Python classes,
+which means they can be imported and executed in scripts/notebooks without any
+additionnal effort. See :ref:`kernel-testing-page`.
+
+Notebooks
++++++++++
+
+In LISA legacy, notebooks served as documentations and where the main source of
+exemples. We now have a proper documentation (you're reading it!), so we
```
qperret: s/exemples/examples

doc/contributors_guide.rst
valschneider: I gave that a try and Github rendered a very plain textfile (I had no syntax highlight for the python snippet). Maybe I missed something - being able to inline the file would be nice indeed.

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+greatly trimmed down the number of notebooks we had.
+
+We've kept older notebooks in ``ipynb/deprecated``, but they have not been ported
+over to the new APIs (or even to Python3) so they won't work. They are there in
+case we find a reason to bring back some of them.
```
qperret: Thinking about it, this is a bit odd actually no ? Maybe we should just remove them ? Keeping things that are totally borked feels kinda wrong. I mean, they will still exist in the 'old' Lisa I suppose, and we can always go back there to pick them up if we want to ... Thoughts ?

doc/users_guide.rst
valschneider: Yep, although ``lisa.env.TestEnv.from_default_conf()`` doesn't have a docstring :fearful:

doc/users_guide.rst
valschneider: I think that's a bit that could go into a proper kernel test documentation rather than the transition guide.

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+
+* ``"platform"`` is now ``"kind"``
+* ``"board"`` used to load some target-specific settings, which we got rid of.
+  The closest thing to it is ``"name"`` which is just a pretty-printing name and
+  has no extra impact.
+* You don't have to specify devlib modules to load anymore. All (loadable)
+  modules are now loaded. If you find some module too slow to load, you can
+  specify a list of modules to exclude.
+* LISA used to have ``target.config`` JSON file at its root. Its equivalent is
+  now ``target_conf.yml``, which is in YAML.
```
qperret: If we give an example for the old version, should we give an example for the new one too ?

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+(much needed) changes. If you are used to using LISA before the refactoring came
+into place, this guide is for you.
+
+Project structure
++++++++++++++++++
+
+* ``lisa/libs/utils`` is now ``lisa/lisa/``. ``lisa/libs/wlgen`` has also been
+  moved to that location.
+* devlib, trappy and bart are now under ``lisa/external/``. Git subtrees are
+  now used instead of submodules.
```
qperret: I guess we could add a pointer to the contribution guide once it explains how to work with those subtrees.

doc/users_guide.rst
```diff
@@ -0,0 +1,306 @@
...
+
+  wload = RTA.by_profile(te, "foo", profile, res_dir, calibration)
+
+Kernel tests
+++++++++++++
+
+The ``Executor`` from LISA legacy has been entirely removed, and a new test
+framework has been put in place. Tests are now coded as straight Python classes,
+which means they can be imported and executed in scripts/notebooks without any
+additionnal effort. See :ref:`kernel-testing-page`.
```
qperret: Should we mention bisector somewhere in this section ?

doc/users_guide.rst
valschneider: Mmm so the main reason I kept them is to not lose their history. That way we still have the right contributors associated to the right notebooks - I tried to keep that consistent with as many files as possible, no matter how much we moved them. Deleting them, then reverting the delete + changing them would be a bit of a pain (plus I wonder if the revert wouldn't cover the whole file when blaming it :/)

doc/users_guide.rst
valschneider: Yup!

doc/users_guide.rst
valschneider: I've written this with the mindset of "the old version isn't going to change, the new one maybe...". I try to limit myself as much as possible to just links to the documentation (though I do single out changes when I feel it's helpful). And we have a full example in the new doc.

doc/users_guide.rst
valschneider: I have another PR that adds some bits to the kernel testing page about how to run tests and all. Since I point it here we should be good.

doc/contributors_guide.rst
douglas-raillard-arm: You can try to start from that source file:
https://github.com/ARM-software/arm-trusted-firmware/blob/master/readme.rst

And then update the content one piece at a time, maybe there is a directive that is a sphinx extension that make it fail, or something like that

doc/users_guide.rst
douglas-raillard-arm: I'm under the process of streamlining integration between bisector and exekall, when it's done I'll write a howto page with a workflow from bisector to notebook visualization.

doc/users_guide.rst
douglas-raillard-arm: right, let's fix that ...

valschneider:
Removed the [RFC] tag cause I think it's mergeable now :)

doc/users_guide.rst
```diff
@@ -0,0 +1,368 @@
...
+
+  from trace import Trace
+
+However, implicit relative imports are dangerous - did you know :class:`trace.Trace`
+exists in Python's standard library? This means that with the previous setup, the LISA
+module would shadow the standard library's. The above import done in a non-LISA
+environment would import something completely different!
+
+
+We now enforce the use of absolute imports, which look like this::
```
douglas-raillard-arm: s/enforce/mandate/

doc/users_guide.rst
```diff
@@ -0,0 +1,368 @@
...
+LISA legacy used implicit relative imports with a bit of dark magic to hold
+everything together. Say you want to import the :class:`lisa.trace.Trace` class
+found in ``lisa/trace.py``, previously you would do it like so::
+
+  from trace import Trace
+
+However, implicit relative imports are dangerous - did you know :class:`trace.Trace`
+exists in Python's standard library? This means that with the previous setup, the LISA
+module would shadow the standard library's. The above import done in a non-LISA
+environment would import something completely different!
```
douglas-raillard-arm: s/would import/would have imported/

That only applied to lisa-legacy, Python 3 does not have implicit relative import anymore so in lisa-next, it would always import the stdlib one.

valschneider:
@douglas-raillard-arm Done

PR #727, opened 2019-01-28
env: Add docstrings to some TestEnv methods

douglas-raillard-arm:


lisa/env.py
```diff
@@ -268,11 +268,21 @@ def __init__(self, target_conf:TargetConf, plat_info:PlatformInfo=None, res_dir:
 
     @classmethod
     def from_default_conf(cls):
+        """
+        Create a :class:`TestEnv` from the YAML configuration file pointed by
+        ``LISA_CONF`` environment variable.
```
valschneider: That's setup for us in `lisa_shell`, could we mention that somehow?

douglas-raillard-arm: We should list all env var in `lisa_shell` that are read from (and not just assigned to) and document them indeed. The reason it was not done before is that the set of env var was not stable and subject to change without notice, but nowadays it should be ok to do it.

PR #729, opened 2019-01-30
Unify invariance tests

douglas-raillard-arm:
Unify CPU invariance and frequency invariance. The rational is:

* simplify the hierarchy of class in the test, and simplify the signature of functions by removing some defaults that are not always used depending on the execution path.
* Avoid misleading assumptions: if CPU invariance fail, that could just mean that invariance is broken for the highest frequency, since working freq invariance is a hardcoded assumption in the CPU invariance test. The new layout allows unbiased access to data.
* Allow running the test for any combination of CPU and frequency without arbitrary constraint
* The previous point also means that some duplicated was removed

Also, make sure only one set of InvarianceItem is collected and use that for both the aggregated and non-aggregated version of each test method.

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -17,6 +17,7 @@
 
 import os
 from collections import OrderedDict
+from statistics import mean
 
 import matplotlib.pyplot as plt
 import pylab as pl
@@ -53,13 +54,64 @@ class LoadTrackingHelpers:
     """
     Common bunch of helpers for load tracking tests.
     """
-    @staticmethod
-    def get_max_capa_cpu(te):
+
+    MAX_RTAPP_CALIB_DEVIATION = 3/100
+    """
+    Blacklist CPUs that have a RTapp calibration value that deviates too much
+    from the average calib value in their capacity class.
+    """
+
+    @classmethod
+    def _get_blacklisted_cpus(cls, plat_info):
         """
-        :returns: A CPU with the highest capacity value
+        Consider some CPUs as blacklisted when the load would not be
+        proportionnal to utilization on them.
+
+        That happens for CPUs that are busy executing other code than the test
+        workload, like handling interrupts. It is detect that by looking at the
+        RTapp calibration value and we blacklist outliers.
         """
-        cpu_capacities = te.plat_info['cpu-capacities']
-        return max(cpu_capacities.keys(), key=lambda cpu: cpu_capacities[cpu])
+        rtapp_calib = plat_info['rtapp']['calib']
+        blacklisted = set()
+        # For each class of CPUs, get the average rtapp calibration value
+        # and blacklist the ones that are deviating too much from that
+        for cpu_class in plat_info['capacity-classes']:
+            calib_mean = mean(rtapp_calib[cpu] for cpu in cpu_class)
+            calib_max = (1 + cls.MAX_RTAPP_CALIB_DEVIATION) * calib_mean
+            blacklisted.update(
+                cpu
+                for cpu in cpu_class
+                # exclude outliers that are too slow (i.e. calib value too small)
+                if rtapp_calib[cpu] > calib_max
+            )
+        return sorted(blacklisted)
+
+    @classmethod
+    def filter_capacity_classes(cls, plat_info):
+        """
+        Filter out capacity-classes key of ``plat_info`` to remove blacklisted CPUs.
+        .. seealso:: :meth:`_get_blacklisted_cpus`
+        """
+        blacklisted_cpus = set(cls._get_blacklisted_cpus(plat_info))
+        return [
+            sorted(set(cpu_class) - blacklisted_cpus)
+            for cpu_class in plat_info['capacity-classes']
+        ]
+
+    @classmethod
+    def get_max_capa_cpu(cls, te):
+        """
+        :returns: A CPU with the highest capacity value that is not blacklisted.
+        """
+        # capacity-classes is sorted by capacity, last class therefore contains
+        # the biggest CPUs
+        candidates = cls.filter_capacity_classes(te.plat_info)[-1]
+
+        if not candidates:
+            raise RuntimeError('All CPUs of that class have been blacklisted: {}'.format(
+                te.plat_info['capacity-class'][-1]
+            ))
+        return candidates[0]
 
     @classmethod
     def get_task_duty_cycle_pct(cls, trace, task_name, cpu):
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
If needed, we can add a couple of test methods on `Invariance` (the aggregated object):
* `test_cpu_invariance`: check that items using the max freq on each CPU is passing. There could be false positives, but they are expected to be relatively rare.
* `test_freq_invariance`: check that at least one CPU has all items passing.

item passing = InvarianceItem.test_XXX_avg are passing

lisa/tests/scheduler/load_tracking.py
```diff
@@ -53,13 +55,64 @@ class LoadTrackingHelpers:
     """
     Common bunch of helpers for load tracking tests.
     """
-    @staticmethod
-    def get_max_capa_cpu(te):
+
+    MAX_RTAPP_CALIB_DEVIATION = 3/100
```
valschneider: I gave this a try on my HiKey960 and CPU0 is still picked up. FYI the calibration I get is:
```python
{0: 308, 1: 302, 2: 302, 3: 303, 4: 155, 5: 155, 6: 155, 7: 155}
```

So looking at this, maybe it would be simpler to get rid of it and ditch the load tests. I've never really trusted them anyway - a tiny (timewise) task with a huge priority could mess the whole test up.

douglas-raillard-arm: For now I'd rather not ditch the test until it's proven to be unstable (i.e. going through one round of integration). That said we might want to turn that threshold into an absolute one instead of a percentage. On my experiments, I got no issues with load testing on hikey 960 at all frequencies on all capacity classes.

valschneider: They all passed for me, but I just don't really see a value in asserting the load values.

It would make more sense to have something like run the same task but change its niceness and assert that it did change the load. Or voluntarily create an imbalanced scenario and make sure there is a significant load difference. In this case (load for a rq with a single test task) it's bound to be more painful than helpful.

I'm not saying this is to rip out now, but my 2cents for "the future".

lisa/tests/scheduler/load_tracking.py
```diff
@@ -192,30 +294,46 @@ def get_expected_util_avg(self, trace, cpu, task_name, capacity=None):
         # Find duty cycle of the workload task
         duty_cycle_pct = self.get_task_duty_cycle_pct(trace, task_name, cpu)
         cpu_capacities = self.plat_info['cpu-capacities']
-        capacity = capacity or cpu_capacities[cpu]
```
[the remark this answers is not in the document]
douglas-raillard-arm: We don't, what we always pass is the "effective capacity", which takes into account frequency and CPU:
```
            capacity = self.plat_info['cpu-capacities'][cpu] * capacity_scale
```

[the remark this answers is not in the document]
douglas-raillard-arm: `capacity_scale` would be better renamed to `scaling_factor` actually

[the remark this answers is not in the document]
valschneider: Just saw that bit... :(

lisa/tests/scheduler/load_tracking.py
douglas-raillard-arm: > I just don't really see a value in asserting the load values.

It could break independently from util signal, so still worth testing if that's easy. In our case, it is easy (unless proven to be really flaky in practice), so we might as well do it.

valschneider:
I had to add these changes to make the computed signals be closer to reality:

```diff
diff --git a/lisa/tests/scheduler/load_tracking.py b/lisa/tests/scheduler/load_tracking.py
index 7c6dc349..d970be3a 100644
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -293,11 +293,9 @@ class InvarianceItem(LoadTrackingBase):
         """
         # Find duty cycle of the workload task
         duty_cycle_pct = self.get_task_duty_cycle_pct(trace, task_name, cpu)
-        cpu_capacities = self.plat_info['cpu-capacities']
 
-        # Scale the relative CPU/freq capacity into the range 0..1
-        scaling_factor = capacity / max(cpu_capacities.values())
-        return UTIL_SCALE * (duty_cycle_pct / 100) * scaling_factor
+        # Scale the relative CPU/freq capacity
+        return (duty_cycle_pct / 100) * capacity
 
     def _test_task_signal(self, signal_name, allowed_error_pct,
                           trace, cpu, task_name, capacity):
@@ -306,15 +304,10 @@ class InvarianceItem(LoadTrackingBase):
         signal = signal_df[UTIL_AVG_CONVERGENCE_TIME_S:][signal_name]
 
         signal_mean = area_under_curve(signal) / (signal.index[-1] - signal.index[0])
-        # Scale the signal to the capacity of the CPU we are running on. We
-        # have to do that to account for the difference in CPUs processing
-        # speed.
-        scale_factor = self.plat_info['cpu-capacities'][cpu] / UTIL_SCALE
-        # For the load, re-apply the factor again since the kernel scales it
-        # according to the capacity.
+
         if signal_name == 'load':
-            scale_factor *= scale_factor
-        signal_mean *= scale_factor
+            # Load isn't scaled with capacity
+            exp_util /= capacity
 
         ok = self.is_almost_equal(exp_util, signal_mean, allowed_error_pct)
 
@@ -327,25 +320,17 @@ class InvarianceItem(LoadTrackingBase):
 
         freq_str = '@{}'.format(self.freq) if self.freq is not None else ''
 
-        capacity_scale = (
-            # Scale the capacity linearly according to the frequency
-            (self.freq / max(self.freq_list))
-            *
-            # scale for that CPU
-            (self.plat_info['cpu-capacities'][self.cpu] / UTIL_SCALE)
-        )
+        # Scale the capacity linearly according to the frequency & capacity_orig
+        capacity = (self.freq / max(self.freq_list)) * self.plat_info['cpu-capacities'][self.cpu]
 
         for name, task in self.rtapp_profile.items():
-            cpu = task.phases[0].cpus[0]
-            capacity = self.plat_info['cpu-capacities'][cpu] * capacity_scale
-
             ok, exp_util, signal_mean = self._test_task_signal(
-                signal_name, allowed_error_pct, self.trace, cpu, name, capacity)
+                signal_name, allowed_error_pct, self.trace, self.cpu, name, capacity)
 
             if not ok:
                 passed = False
 
-            metric_name = 'cpu{}{}'.format(cpu, freq_str)
+            metric_name = 'cpu{}{}'.format(self.cpu, freq_str)
             expected_data[metric_name] = TestMetric(exp_util)
             trace_data[metric_name] = TestMetric(signal_mean)
 

```

One thing I started doing but that'd need finishing: we can ditch the `cpu` parameters since we now have `self.cpu`

valschneider:
Duuh this just failed, hang on a sec...

douglas-raillard-arm:
> One thing I started doing but that'd need finishing: we can ditch the cpu parameters since we now have self.cpu

IMHO it's a better idea to keep it, so that a subclass can reuse that helper if needed even in the case it has to catter with multiple CPUs. It also helps knowing what are the inputs of the function, otherwise the flow can quickly become extremely confusing.

douglas-raillard-arm:
> Eventually we might want to add a knob in the test code checking for the kernel version and deciding whether to do the scaling or not, but for now I think it's fine as it is.

The kernel version is in PlatformInfo so its directly available

valschneider:
> > Eventually we might want to add a knob in the test code checking for the kernel version and deciding whether to do the scaling or not, but for now I think it's fine as it is.
> 
> The kernel version is in PlatformInfo so its directly available

That's not an issue, it's just that the change is only in tip/sched/core so we don't yet have a kernel release to use as a reference.

PR #730, opened 2019-01-30
Tests location shuffling

valschneider:
Yet another one...

We now have self tests at the root of the repo (like most Python projects), and I took this opportunity to remove one level from the kernel tests (`$repo/lisa/tests/kernel/x` -> `$repo/lisa/tests/x`). Self-tests are passing, doc & notebooks have been updated.

Based on top of https://github.com/ARM-software/lisa/pull/726 because then I can update the new shiny doc in one go :)


doc/contributors_guide.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Submitting your subtree changes
++++++++++++++++++++++++++++++++
+
+Our changes to subtrees are often developped conjointly with LISA, so we write our
+modifications directly in the subtrees. You can commit these changes in the LISA
+repository, then shape those modifications into a git history ready to be pushed
+to e.g. :mod:`devlib` by running::
+
+  git subtree split --prefix=external/devlib -b my-devlib-feature
```
douglas-raillard-arm: following the issue I faced yesterday, that command should be run first:
```
git fetch devlib-remote
```
Otherwise, `git subtree split` will be unable to find the commit on which it is supposed to rebase your commit series.

doc/lisa_tests.rst
```diff
@@ -12,9 +12,9 @@ From the root of LISA, you can run those tests like so:
 
 >>> python3 -m nose
 >>> # You can also target specific test modules
->>> python3 -m nose --tests=lisa/tests/lisa/test_test_bundle.py
+>>> python3 -m nose tests/test_test_bundle.py
```
douglas-raillard-arm: is it working without --tests now ? It used to not work.

valschneider: Yeah, because we had to specify it in the config file and I think it would always use it, so we'd be forced to override it. It works now.

doc/contributors_guide.rst
douglas-raillard-arm: When we update the subtree, we also do the --fetch. But that will only help the person that updated the subtree. All other developers that want to push a change to devlib will face the issue, unless they update devlib first (which is a sensible thing to do, but is not mandatory).

valschneider: Ah gotcha

valschneider:
> Nit: I guess the `tests_TO_DISPATCH_IN_kernel` folder needs renaming too ? Other than that LGTM.

Actually we might as well remove it... Thoughts?

qperret:
Hmm most of it is garbage, but not all. There is at least the stune_hold test from @credp that I run on the product branch every once in a while. Not sure about the heavy_load test either (some of these things were supposed to mimic geekbench no ?). Probably worth porting at some point.

But I don't mind moving them somewhere else. Or even just removing them and keeping a list somewhere so we don't forget. They'll still be part of the old Lisa anyway so it's OK.

valschneider:
Okay, I'll just rename the whole thing `deprecated` to mimic what I did with the notebooks and keep them here.

valschneider:
Also renamed `test_bundle.py` while I still can break APIs :D

valschneider:
@douglas-raillard-arm I realized I had to change some of exekall's glue code, which is now done.

PR #735, opened 2019-01-31
travis: check for commits touching external

douglas-raillard-arm:
This still allows merge commits touching that location. The squashed
commit produced by git subtree is fine since it does not refer to
"external/" (path rewriting is done in the merge commit).

Also fixes various Travis issues.

### DIFF
--- a/tools/scripts/travis_tests.sh
+++ b/tools/scripts/travis_tests.sh
@@ -20,9 +20,6 @@
 # Script run by Travis. It is mostly a workaround for Travis inability to
 # correctly handle environment variable set in sourced scripts.
 
-# Failing commands will make the script return with an error code
-set -e
-
 illegal_location="external/"
 illegal_commits=$(find "$illegal_location" -mindepth 1 -maxdepth 1 -type d -print0 | xargs -0 git log --no-merges --oneline)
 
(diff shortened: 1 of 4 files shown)


tools/scripts/travis_tests.sh
```diff
@@ -31,7 +28,12 @@ if [[ -n "$illegal_commits" ]]; then
     exit 1
 fi;
 
-source init_env
+# Some commands are allowed to fail in init_env, e.g. to probe for installed
+# tools. However, the overall script has to succeed.
+source init_env || exit 1
```
valschneider: Shouldn't we make sure sourcing `init_env` always work? I suppose if it breaks the rest of the script will fail, but then it'll be less obvious where the failure came from.

douglas-raillard-arm: That's exactly why I use `|| exit 1`: if init_env fails, then travis_tests.sh will fail as well.
We just do not fail at every command that fails inside init_env. Some of them are expected to fail under normal circumstances since they probe for existing things (like jupyter being installed).

douglas-raillard-arm:
Rebased, let's see if Travis is happy ..

PR #748, opened 2019-02-08
buildroot: fix lisa-buildroot-update-kernel-config

douglas-raillard-arm:
Use KConfig fragments, which are not trying to detect features of GCC
when merged. Using make oldconfig can break the configuration if the
host version of GCC is wildly different than the one used to
cross-compile. KConfig fragment also has the added benefit of reporting
what was modified in the config.

### DIFF
--- a/tools/lisa-buildroot-update-kernel-config
+++ b/tools/lisa-buildroot-update-kernel-config
@@ -51,8 +51,27 @@ if [[ -z "$1" || -z "$2" ]]; then
 	exit -1
 fi
 
-set -u
+# Automatically detect the arm arch we are building for
+export ARCH=${ARCH:-$(grep -owE "(arm|arm64)" "$KERNEL_CONFIG")}
 
-# Update kernel .config to use the provided rootfs as builtin initramfs
-sed -i "s#[.]*CONFIG_INITRAMFS_SOURCE.*#CONFIG_INITRAMFS_SOURCE=\"$ROOTFS\"#" "$KERNEL_CONFIG"
-make -C "$KERNEL_DIR" olddefconfig
+# That env var is not needed, but wrong values can break the make command
+# executed by merge_config.sh
+unset CROSS_COMPILE
+
+if [ -z "$ARCH" ]; then
+	echo "ERROR: set ARCH environment variable"
+	exit -1
+fi
+
+echo "Assuming ARCH=$ARCH"
+
+echo "Merging Kconfig fragment into $KERNEL_CONFIG"
+fragment=$(mktemp)
+trap "rm "$fragment"" EXIT
+
+# Use that CPIO archive to be used as builtin initramfs
+tee "$fragment" << EOF
+CONFIG_INITRAMFS_SOURCE="$ROOTFS"
+EOF
+
+KCONFIG_CONFIG="$KERNEL_CONFIG" "$KERNEL_DIR/scripts/kconfig/merge_config.sh" "$KERNEL_CONFIG" "$fragment"


tools/lisa-buildroot-update-kernel-config
```diff
@@ -53,6 +53,19 @@ fi
...
 set -u
 
+
+echo "Merging Kconfig fragment into $KERNEL_CONFIG"
+fragment=$(mktemp)
+
 # Update kernel .config to use the provided rootfs as builtin initramfs
-sed -i "s#[.]*CONFIG_INITRAMFS_SOURCE.*#CONFIG_INITRAMFS_SOURCE=\"$ROOTFS\"#" "$KERNEL_CONFIG"
-make -C "$KERNEL_DIR" olddefconfig
+# Setting root UID and GID also seems to be necessary.
```
qais-yousef: Can you expand on this comment please? Why is it necessary, what breaks if this is not set?

qais-yousef:
Sorry for the late response - just seen this.

If you can expand the comment about the UID and GID then I'm ok with this change in general. Although I am still surprised that the GCC version thingy is causing a break - I think there's something else fishy. But if this is giving more reliable results then yeah let's go for it.

Although if you can expand in the commit message on how the break happens (paste an error message or something) that'd be great.

tools/lisa-buildroot-update-kernel-config
douglas-raillard-arm: `Kconfig` asks you in an interactive way what you want, in a similar way as `make oldconfig`. Not sure why exactly, since we just need the default value though. I haven't investigated deeply since I could not find any information on the exact expected behavior of that Kconfig fragment script.

douglas-raillard-arm:
> Although I am still surprised that the GCC version thingy is causing a break - I think there's something else fishy

It's just that Kconfig runs some autodections on the version of GCC that is being used and has a notion of GCC plugins somehow, that gets enabled/disabled. It's not a purely lexical operation on the config file, it does other things in the background as well

douglas-raillard-arm:
@qais-yousef Updated with sample output of Kconfig issues

tools/lisa-buildroot-update-kernel-config
qais-yousef: `make olddefconfig` doens't ask interactive questions. It keeps the current configs but updates it to be correct in case you modified something and there are some dependencies on it like in your case. `make olddefconfig` is the right method to deal with these cases.

qais-yousef:
`make olddefrconfig` requries ARCH=arm64 yes because you must always specify the arch when running any config config including menuconfig. Why this is bad? You can't skip this really as it is by design required. Otherwise it'll pick the host architecture by default and apply x86 configs - which will screw things up

qais-yousef:
My original script had setting the arch bit - not sure why it disappeared in next. I think this is the real problem you're trying to fix here?

tools/lisa-buildroot-update-kernel-config
douglas-raillard-arm: Sorry, I meant running `make Image`, the issue does not trigger when running `make oldconfig` but at the next step.

douglas-raillard-arm:
Looks like kconfig fragments just try to merge config files when called with `-m`, without trying to understand what is supposed to be there or not, or what are the possible dependencies. Haven't found a lot of documentation around that, and grepping for `fragment` or `merge_config` in Documentation/ does not give any relevant result.

> My original script had setting the arch bit - not sure why it disappeared in next. I think this is the real problem you're trying to fix here?

So we have multiple solutions here:
1) As you suggest, define ARCH=arm64 when calling `merge_config.sh` without `-m`. That misdetects GCC version (it picks up the host compiler version), but that seems fixed under the hood quietly when running `ARCH=arm64 make Image`. `CROSS_COMPILE` is not needed when doing that, which is ideal.
2) Set a couple of options that have a default as done in that PR, which is not great since we have to maintain that copy/paste of Kconfig defaults.
3) Set `CONFIG_INITRAMFS_SOURCE`, call `merge_config.sh` with`-m`, and when building the kernel, call make in an equivalent way to what `merge_config.sh` would do without `-m`:
```
make olddefconfig Image dtbs
```

Options 2) and 3) have the benefit of not having to set either ARCH or CROSS_COMPILE when updating the conf file, but a base config is needed, which itself would need ARCH anyway.

Option 1) is probably the best one as it fully decouples config file creation from building the kernel itself, I'll make the corresponding change in that PR to bring back ARCH autodection.

tools/lisa-buildroot-update-kernel-config
```diff
@@ -51,8 +51,26 @@ if [[ -z "$1" || -z "$2" ]]; then
...
+
+echo "Assuming ARCH=$ARCH"
+
+echo "Merging Kconfig fragment into $KERNEL_CONFIG"
+fragment=$(mktemp)
 
 # Update kernel .config to use the provided rootfs as builtin initramfs
-sed -i "s#[.]*CONFIG_INITRAMFS_SOURCE.*#CONFIG_INITRAMFS_SOURCE=\"$ROOTFS\"#" "$KERNEL_CONFIG"
-make -C "$KERNEL_DIR" olddefconfig
+# Setting root UID and GID also seems to be necessary, otherwise `make Image`
```
qais-yousef: This whole comment is stale now

tools/lisa-buildroot-update-kernel-config
```diff
@@ -51,8 +51,26 @@ if [[ -z "$1" || -z "$2" ]]; then
...
+# starts asking interactive questions like that one:
+# User ID to map to 0 (user root) (INITRAMFS_ROOT_UID) [0] (NEW)
+tee "$fragment" << EOF
+CONFIG_INITRAMFS_SOURCE="$ROOTFS"
+EOF
+
+KCONFIG_CONFIG="$KERNEL_CONFIG" "$KERNEL_DIR/scripts/kconfig/merge_config.sh" "$KERNEL_CONFIG" "$fragment"
+
+rm "$fragment"
```
qais-yousef: nit: I suggest doing this in `trap "rm $fragment"` EXIT or something like that to avoid with accidental left over - but that's a bit too picky. I'm OK with it if it stays and you don't feel like pushing a new PR with this improvement :)

qais-yousef:
When dealing with configs ARCH is compulsory and usually inferred to be the host's arch if not specified by the user. Cross compile is not important if you're just executing config related comamnds ie: defconfig, menuconfig, olddefconfig, savedefconfg etc. So re-instating the ARCH setup is the right thing for sure as it's not really an optional argument.

Hopefully I looked at the right version of your latest commit - mind updating the commit message to reflect the new findings?

Thanks for taking care of it :)

douglas-raillard-arm:
Yes, ARCH is definitely needed. CROSS_COMPILE is not needed but defining it to something stupid seems to make it fail, even though it's not supposed to be involved in the process:
```
>>> CROSS_COMPILE=foobar lisa-buildroot-update-kernel-config ~lisa/integration-lisa-next/src/buildroot/output/images/rootfs.cpio .
Assuming ARCH=arm64
Merging Kconfig fragment into ./.config
CONFIG_INITRAMFS_SOURCE="/home/dourai01/Data/Git/lisa/integration-lisa-next/src/buildroot/output/images/rootfs.cpio"
Using ./.config as base
Merging /tmp/tmp.3ImcjxrQfO
Value of CONFIG_INITRAMFS_SOURCE is redefined by fragment /tmp/tmp.3ImcjxrQfO:
Previous value: CONFIG_INITRAMFS_SOURCE=""
New value: CONFIG_INITRAMFS_SOURCE="/home/dourai01/Data/Git/lisa/integration-lisa-next/src/buildroot/output/images/rootfs.cpio"

/bin/sh: foobargcc: command not found
/bin/sh: foobargcc: command not found
scripts/kconfig/conf  --alldefconfig Kconfig
./scripts/gcc-version.sh: line 26: foobargcc: command not found
./scripts/gcc-version.sh: line 27: foobargcc: command not found
./scripts/gcc-version.sh: line 29: foobargcc: command not found
./scripts/gcc-version.sh: line 26: foobargcc: command not found
./scripts/gcc-version.sh: line 27: foobargcc: command not found
./scripts/gcc-version.sh: line 29: foobargcc: command not found
init/Kconfig:17: syntax error
init/Kconfig:16: invalid statement
init/Kconfig:17: invalid statement
./scripts/clang-version.sh: line 15: foobargcc: command not found
./scripts/gcc-plugin.sh: line 11: foobargcc: command not found
make[1]: *** [scripts/kconfig/Makefile:69: alldefconfig] Error 1
make: *** [Makefile:538: alldefconfig] Error 2
Value requested for CONFIG_INITRAMFS_SOURCE not in final .config
Requested value:  CONFIG_INITRAMFS_SOURCE="/home/dourai01/Data/Git/lisa/integration-lisa-next/src/buildroot/output/images/rootfs.cpio"
Actual value:     CONFIG_INITRAMFS_SOURCE=""
```

tools/lisa-buildroot-update-kernel-config
douglas-raillard-arm: Interestingly it is hard to get rid of a race condition. The best we can do without knowing the file name before creating it is:
```
fragment=$(mktemp)
trap "rm "$fragment"" EXIT
```
If we interrupt the script at the right moment, the file won't be cleaned up. If we create the filename without creating the file, we introduce another race condition where the same name could be used by 2 different processes. Python's `tempfile.NamedTemporaryFile` also seems to be affected by this kind of race condition between creation and cleanup setup.

qais-yousef:
OK that is weird; maybe a regression in kernel build system somehow..

I'm happy with the change but if you can afford rewording the commit message to mention that the missing ARCH in make olddefconfig was causing the problem as I believe it's the rootcause that would be great.

PR #749, opened 2019-02-08
analysis/tasks: Create an explicitly human readable df_task_states

valschneider:
The current df_task_states forces the user to use the

```python
df[df.curr_state == TaskState.TASK_*.char]
```

pattern since we store the state's character into the dataframe to be
readable. Analysis code shouldn't have to deal with this, and using
pure ints will make filtering a bit faster.

Change the existing `df_task_states` to use pure ints, and add a
stringified sibling that uses the char representation of `TaskState`.

### DIFF
--- a/lisa/analysis/tasks.py
+++ b/lisa/analysis/tasks.py
@@ -200,12 +200,15 @@ class TasksAnalysis(AnalysisBase):
         return rt_tasks
 
     @requires_events('sched_switch', 'sched_wakeup')
-    def df_task_states(self, task):
+    def df_task_states(self, task, stringify=False):
         """
         DataFrame of task's state updates events
 
         :param task: The task's name or PID
         :type task: int or str
+
+        :param stringify: Use `TaskState`'s string representation (instead of int)
+        :type stringify: bool
 
         :returns: a :class:`pandas.DataFrame` with:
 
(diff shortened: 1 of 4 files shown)


lisa/analysis/tasks.py
```diff
@@ -256,11 +256,52 @@ def stringify_row_state(row):
...
         task_state_df = task_state_df[['target_cpu', 'cpu', 'curr_state']]
-        task_state_df['next_state'] = task_state_df.curr_state.shift(-1)
+        # Make the last "next_state" the same as the last "curr_state"
+        task_state_df['next_state'] = task_state_df.curr_state.shift(
+            -1, fill_value=task_state_df['curr_state'].values[-1])
         self.trace.add_events_deltas(task_state_df, inplace=True)
 
         return task_state_df
 
+    def df_task_states_stringify(self, df_task_states, inplace=False):
```
douglas-raillard-arm: The name would benefit from:
* get an "internal" name prefixed by `_`, since it's only used by a similarly-named method.
* be turned into static/classmethod to make it explicit it does not touch the instance itself but is just a helper

valschneider: >  get an "internal" name prefixed by `_`, since it's only used by a similarly-named method.

I wanted to have it exposed publicly to make debugging things easier - if some tests is doing a lot of operations of a `task_states` dataframe, you could just do a `log.debug(df_task_states_stringify(df))` instead of trying to redo the same operations with the `_stringified` version.

douglas-raillard-arm: makes sense, in that case we could have it as a standalone function like `stringify_task_states_df` that is not especially related to a class, and more importantly that does not use `df_` prefix that is by convention reserved for "regular" `df_` functions that extract a df from the current trace (so that autocomplete suggestions are relevant in notebooks).

valschneider:
@douglas-raillard-arm Renamed the helper to `stringify_df_task_states` and changed it to a `classmethod`

lisa/analysis/tasks.py
```diff
@@ -256,11 +256,53 @@ def stringify_row_state(row):
...
+            except ValueError:
+                return TaskState.sched_switch_str(state)
+
+        for col in ['curr_state', 'next_state']:
+            df[col] = df[col].apply(stringify_state)
+
+        return df
+
+    @df_task_states.used_events
+    def df_task_states_stringified(self, task):
```
derkling: Why not keeping just `df_task_states()` and add there a `stringify=false` param ?

valschneider: True, I did not even consider this... Dunno which is better

douglas-raillard-arm: `stringify_state=False` would avoid duplicating the docstring indeed

derkling: I personally like better polymorphic APIs... then multiple APIs... :)

valschneider:
@derkling Done

valschneider:
The doc build is broken on my end but Travis didn't seem to catch that, don't merge it yet please.

valschneider:
Should be good now

derkling:
LGTM.

PR #751, opened 2019-02-11
Fix wlgen Ramp

douglas-raillard-arm:
Make sure we get the expected number of steps, even if the `delta_pct` is not an exact divider of `(end_pct-start_pct)` but is slightly too large instead.  If the `delta_pct` is not an exact divider of `(end_pct-start_pct)`, we'll end up with a final phase that has a duty cycle that does not match `end_pct`. Force the last phase to have a duty cycle of `end_pct`.

Also add a debugging aid.

### DIFF
--- a/lisa/utils.py
+++ b/lisa/utils.py
@@ -63,6 +63,26 @@ class Loggable:
             name += '.' + suffix
         return logging.getLogger(name)
 
+    @classmethod
+    def log_locals(cls, var_names=None, level='debug'):
+        """
+        Debugging aid: log the local variables of the calling function
+
+        :param var_names: List of variable names to display, or all of them if
+            left to default.
+        :type var_names: list(str)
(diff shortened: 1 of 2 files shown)


valschneider:
> The step for the last stage will be smaller than the one before to compensate for "drift" in order to honour `end_pct` and not overshoot.

I got confused by that sentence but the code speaks for itself. Here you use "step" as in "the duty cycle delta between two phases", but it happens that we usually call each phase of a Ramp a step (because it's staircase-shaped) which is why I got confused.

Saying something along those lines would be better IMO:
```
If the `delta_pct` is not an exact divider of `(end_pct-start_pct)`, we'll end up
with a final phase that has a duty cycle that does not match `end_pct`. Force
the last phase to have a duty cycle of `end_pct`.
```

PR #754, opened 2019-02-12
tests: eas_behaviour: Rename test_slack metric

douglas-raillard-arm:
Rename "slack" metric to "bad activations", since that metric is not a
mesure of time, but the proportion of bad activations.

### DIFF
--- a/lisa/tests/scheduler/eas_behaviour.py
+++ b/lisa/tests/scheduler/eas_behaviour.py
@@ -366,7 +366,7 @@ class EASBehaviour(RTATestBundle, abc.ABC):
         """
         pa = PerfAnalysis(self.res_dir)
 
-        slacks = {}
+        bad_activations = {}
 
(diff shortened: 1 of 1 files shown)


lisa/tests/scheduler/eas_behaviour.py
```diff
@@ -378,13 +378,15 @@ def test_slack(self, negative_slack_allowed_pct=15) -> ResultBundle:
...
+            bad_activations[task] = bad_activations_pct
 
         res = ResultBundle.from_bool(passed)
 
-        for task, slack in slacks.items():
-            res.add_metric("{} slack".format(task), slack, '%')
-
+        for task, bad_activations_pct in bad_activations.items():
+            res.add_metric(
+                "{} bad activations".format(task),
```
valschneider: Might as well go for 
```suggestion
                "{} delayed activations".format(task),
```

Otherwise LGTM

PR #755, opened 2019-02-12
lisa: PerfAnalysis: cleanup

douglas-raillard-arm:
* Move to lisa.analysis.rta
* Give "action" name to methods where appropriate ('get_' prefix)
* turn tasks() into a property
* Streamline `__init__` so it takes the common denominator between all
  ways of building the object.
* Add alternative factories:
    * from_log_files
    * from_task_names
    * from_dir

    note: from_dir() has a similar behavior to the former constructor.

* Rename camelcase methods to PEP-8 naming style like the rest of the
  code
* Remove leading double undescores from method names to avoid name
  mangling.
* Split plot_perf into 3 separate plot methods, like it is done in other
  analysis classes
* Remove dead code here and there
* Use pandas.read_csv instead of read_table to avoid FutureWarning

Changes to AnalysisBase:
* Split into AnalysisHelpers and AnalysisBase
* Improve AnalysisHelpers.save_to() to infer the image format from
  filepath.

EDIT: updated to match latest commit message


douglas-raillard-arm:
`plot_perf` is currently creating wild matplotlib figures, so it will need to be aligned with the revamped analysis methods

lisa/wlgen/rta.py
```diff
@@ -720,4 +725,187 @@ def __init__(self, barrier, time_s=1, delay_s=0, loops=1, sched_policy=None,
...
+                'df': self._parse_df(logfile),
+            }
+            for task_name, logfile in task_log_map.items()
+        }
+
+    @classmethod
+    def from_log_files(cls, rta_logs):
+        """
+        Build a :class:`RTALogAnalysis` from a sequence of RTApp log files
+        """
```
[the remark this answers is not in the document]
douglas-raillard-arm: Does that really make sense here ? It is actually documented by the one line description of the function, so I'm not sure it really adds anything to restate the same thing again in a `:param:` option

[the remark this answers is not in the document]
valschneider: Even if the one-liner seems sufficient, I'd go for the proper parameter documentation.

lisa/analysis/rta.py
```diff
@@ -0,0 +1,245 @@
...
+            :class:`lisa.analysis.base.AnalysisHelper` would use a `trace`
+            attribute that does not exists otherwise.
+        """
+        return super().save_plot(
+            figure=figure,
+            filepath=filepath,
+            img_format=img_format
+        )
+
+    def plot_perf(self, task, filepath=None, axis=None):
```
derkling: On this simple example:

```python
from lisa.analysis.rta import RTALogAnalysis
rta = RTALogAnalysis.from_dir('/data/Code/lisa/results/20180209_UtilEst_JunoR2/')
rta.tasks
rta.plot_perf('jumper')
```

I get this error:

```python
AttributeError                            Traceback (most recent call last)
<ipython-input-6-9d6013597ea8> in <module>
----> 1 rta.plot_perf('jumper')

/data/Code/lisa2/lisa/analysis/rta.py in plot_perf(self, task, filepath, axis)
    202             axis.set_ylim(0, 2)
    203 
--> 204         return self._do_plot(plotter, filepath, axis)
    205 
    206     def plot_latency(self, task, filepath=None, axis=None):

/data/Code/lisa2/lisa/analysis/base.py in _do_plot(self, plotter, filepath, axis)
    154 
    155         if local_fig:
--> 156             self.save_plot(fig, filepath)
    157         return axis
    158 

/data/Code/lisa2/lisa/analysis/rta.py in save_plot(self, figure, filepath, img_format)
    183             figure=figure,
    184             filepath=filepath,
--> 185             img_format=img_format
    186         )
    187 

/data/Code/lisa2/lisa/analysis/base.py in save_plot(self, figure, filepath, img_format)
    138             caller = inspect.stack()[1][3]
    139             filepath = os.path.join(
--> 140                 self.trace.plots_dir,
    141                 "{}.{}.{}".format(module, caller, img_format))
    142 

AttributeError: 'RTALogAnalysis' object has no attribute 'trace'
```

which also breaks loops like:

```python
for task in rta.tasks:
    rta.plot_latency(task)
```

derkling: Same warnings for the following two methods

lisa/analysis/rta.py
```diff
@@ -0,0 +1,245 @@
...
+import os
+import re
+import glob
+
+import pandas as pd
+
+from lisa.utils import Loggable
+from lisa.analysis.base import AnalysisHelpers
+
+class RTALogAnalysis(AnalysisHelpers):
```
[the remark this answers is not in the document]
derkling: and potentially we can have other PerfAnalysis classes in other scopes... all doing the same things "report/plot" performance figures...

douglas-raillard-arm:
> Can we come up with a "new" concept of "report" or whatever you like to name it, where the idea is that you get a single meaningful resuming set of information?

We could do that, however in the current state, there is a diversity of API in plot functions that we may want to rationalize, before we add a layer of sugar on top. Some of them take axes, some of them don't, some always call `save_plot()`, some don't etc. That change would belong to another PR that aligns all analysis and add a thin layer on top as you suggest.

douglas-raillard-arm:
@valschneider Are you happy with the `_do_plot` helper I introduced, or do you see room for removing more boilerplate ?

douglas-raillard-arm:
@derkling @valschneider comments addressed

lisa/analysis/rta.py
```diff
@@ -0,0 +1,243 @@
+# SPDX-License-Identifier: Apache-2.0
```
valschneider: IIRC you had issues with making this live in `lisa.wlgen.rta.py`, which is why it's in `lisa.analysis`. However, the analysis module has this fancy proxy. Doesn't that mean we'll get a `trace.analysis.rta` member when creating `Trace` objects?

douglas-raillard-arm: Currently no, the proxy only gives access to subclasses of AnalysisBase. PerfAnalysis does not inherit from it, since it does not use a trace at all, and is using a different set of data. Therefore it would not make sense to make it available as an attribute of lisa.Trace instances.

PR #758, opened 2019-02-13
lisa/tests: misfit: Remove test_migration_delay()

valschneider:
This test is just a subtest of test_throughput(), and was only added
because we initially had high failure rates. With the proper scaling
of the allowed idle time, both tests have a near identical pass rate,
so let's remove the subtest.

### DIFF
--- a/lisa/tests/scheduler/misfit.py
+++ b/lisa/tests/scheduler/misfit.py
@@ -298,47 +298,6 @@ class StaggeredFinishes(MisfitMigrationBase):
         return res
 
     @TasksAnalysis.df_task_states.used_events
-    def test_migration_delay(self, allowed_idle_time_s=None) -> ResultBundle:
-        """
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
Can you check with `bisector report -oshow-dist` on a bisector report first, to make sure they effectively fail during the same iteration ?

valschneider:
> EDIT: looks like they do fail at the same time (20190208 integration on TC2):
> 
> ```
>     StaggeredFinishes[board=pwrsft-TC2-1]:test_migration_delay:
>     	..........F.....F.........F...........................F....F..........F.......F.........F.....................F......
>     
>     StaggeredFinishes[board=pwrsft-TC2-1]:test_throughput:
>     	..........F.....F.........F...........................F....F..........F.......F.........F.....................F......
> ```

Thanks for checking this!

PR #760, opened 2019-02-14
shell: Updated README lisa-ipython -> lisa-jupyter

msrasmussen:
Trivial fix for lisa-help.


qperret:
Maybe for another PR, but lisa-help needs some love. lisa-report and all need to be fixed/removed too ...

douglas-raillard-arm:
> Maybe for another PR, but lisa-help needs some love. lisa-report and all need to be fixed/removed too ...

That is planned, and will likely end up in a manpage rendered with Sphinx

PR #761, opened 2019-02-14
doc/users_guide: Add bits about updating LISA

valschneider:


doc/users_guide.rst
```diff
@@ -91,7 +82,24 @@ access to many LISA related functions (done automatically by Vagrant)::
...
+.. tip:: Run ``lisa-help`` to see an overview of the provided LISA commands.
+
+Updating
+========
+
+Over time, we might change/add some dependencies to LISA. As such, if you
+update your LISA repository, you should make sure your locally-installed
+packages still match those dependencies. Issuing a ``lisa-install`` from
+the LISA shell should suffice, although re-source ``init_env`` would be
+preferable in case there were also updates to the shell commands.
```
douglas-raillard-arm: > Issuing a ``lisa-install`` from the LISA shell should suffice

We cannot sanely support anything that does not involve re-sourcing. If somebody wants to be a cow boy and not re-source, that's fine, but we should not encourage that in the doc nor give the impression that it is supposed to work (it is not supposed to and actually does not work, leading to weird issue that we won't deal with).

I would rephrase:
```
Sourcing ``init_env`` from a new shell  should suffice, which will hint the user if running ``lisa-install`` again is needed.

valschneider:
@douglas-raillard-arm done

PR #762, opened 2019-02-15
lisa.trace: Remove Trace.df()

douglas-raillard-arm:
This was a useless alias to Trace.df_events().
Also update the transition guide.


derkling:
LGTM

PR #764, opened 2019-02-18
lisa.wlgen.rta: Cleanup RTA.by_str()

douglas-raillard-arm:
Recurse into the nested mapping instead of doing text-based tricks on
the JSON representation. That ensures we don't need to manage any kind
of quoting and allows better control to what path in the JSON file we
would allow updating.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -262,10 +262,6 @@ class RTA(Workload):
 
         :returns: a JSON dict
         """
-        # pload can either be a string like "CPU1" or an integer, if the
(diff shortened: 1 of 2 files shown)


valschneider:
IIRC this doesn't work because the tokens we replace are not proper JSON, so we can't do `json.loads(template)`. Do you still want to work on this?

douglas-raillard-arm:
If we go the JSON way, I'll keep that PR as it is, otherwise I'll respin it with a plain simplification of the code (it boils down to a `str.replace()`, but current implementation takes 20+ lines for that).

PR #767, opened 2019-02-20
lisa.env: Issue adb root when connecting to AndroidTarget

douglas-raillard-arm:
Since we need it in all cases, let's do it in LISA rather than having to
do it manually.

### DIFF
--- a/lisa/env.py
+++ b/lisa/env.py
@@ -509,6 +509,9 @@ class TestEnv(Loggable, HideExekallID):
 
         target.setup()
 
+        if isinstance(target, devlib.AndroidTarget):
+            target.adb_root(force=True)
+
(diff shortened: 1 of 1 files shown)


lisa/env.py
```diff
@@ -509,6 +509,10 @@ def _init_target(self, target_conf, res_dir):
 
         target.setup()
 
+        # Only available for Android targets
+        with contextlib.suppress(AttributeError):
```
valschneider: Any reason for doing this instead of that:
```python
if target_kind == "android":
    target.adb_root(force=True)
```
?

douglas-raillard-arm: Not especially, the one ignoring the exception does not have to make any assumption on the exact class used

PR #768, opened 2019-02-21
utils/energy_model: Fix loading EM from recent android-4.14 targets

credp:
Since commit 9001b8f1a,
"ANDROID: sched: Add support for frequency/power energy model"
android-4.14 has been able to load a frequency-based energy model.

Adding this extended the capacity_state structure in
sched_group_energy with a new frequency member. This is automatically
reflected in the energy model tuples printed in procfs and so Lisa can
no longer load sensible Energy models from any target running
android-4.14 later than the above commit, although the error is not
detected so you do get an EnergyModel object which contains junk
capacity and power values.

We can figure out which kind of kernel we have by reading the number
of capacity states for a cpu and check how many items are in the
cap_state tuples so that we can handle 2 or 3 values per state
transparently.

### DIFF
--- a/lisa/energy_model.py
+++ b/lisa/energy_model.py
@@ -31,6 +31,15 @@ from devlib.utils.misc import mask_to_list, ranges_to_list
 from devlib.exception import TargetStableError
 from trappy.stats.grammar import Parser
 
+#TODO: This should be moved into a utility library somewhere if its useful elsewhere
+from itertools import zip_longest
+def grouper(iterable, n, fillvalue=None):
+    """Collect data into fixed-length chunks or blocks"""
+    # grouper('ABCDEFG', 3, 'x') --> ABC DEF Gxx"
+    # Since the same iterator is used, it will yield a new item every time zip call next() on it
+    args = [iter(iterable)] * n
+    return zip_longest(*args, fillvalue=fillvalue)
+
 """Classes for modeling and estimating energy usage of CPU systems"""
 
 def read_multiple_oneline_files(target, glob_patterns):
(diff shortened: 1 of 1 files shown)


lisa/energy_model.py
```diff
@@ -1034,13 +1035,43 @@ def read_sge_file(path):
...
+                raise TargetStableError('Unsupported cap_states format '
+                                  'cpu={} domain_level={} path={}'.format(cpu, domain_level, cap_states_path))
 
+            # In a two-value system (no frequency model),
             # cap_states lists the capacity of each state followed by its power,
             # in increasing order. The `zip` call does this:
             #   [c0, p0, c1, p1, c2, p2] -> [(c0, p0), (c1, p1), (c2, p2)]
-            cap_states = [ActiveState(capacity=int(c), power=int(p))
+            if (not freq_model_supported):
+                cap_states = [ActiveState(capacity=int(c), power=int(p))
```
credp: @douglas-raillard-arm mentioned in #657 against master that we should probably update the grouping. He wrote:
iterating N items at a time seems to be a common pattern, it may be worth adding the "grouper" from itertools recipes to a helper module (or just locally here with a #TODO to move it, to ease forward porting to LISA-next):
https://docs.python.org/3/library/itertools.html#itertools-recipes

```
from itertools import zip_longest
def grouper(iterable, n, fillvalue=None):
    "Collect data into fixed-length chunks or blocks"
    # grouper('ABCDEFG', 3, 'x') --> ABC DEF Gxx"
    # Since the same iterator is used, it will yield a new item every time zip call next() on it
    args = [iter(iterable)] * n
    return zip_longest(*args, fillvalue=fillvalue)
```

_Originally posted by @douglas-raillard-arm in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMTQxODU0MzMzOnYy/pull_request_review_threads/discussion_

valschneider:
Could you please squash your last commit into its ancestor?

lisa/energy_model.py
```diff
@@ -1034,13 +1035,43 @@ def read_sge_file(path):
...
+            # In a three-value system (frequency model supported),
+            # cap_states lists the capacity of each state followed by its freq,
+            # then the power, in increasing order. The `zip` call does this:
+            #   [c0, p0, c1, p1, c2, p2] -> [(c0, f0, p0), (c1, f1, p1), (c2, f2, p2)]
+            if (freq_model_supported):
+                cap_states = [ActiveState(capacity=int(c), power=int(p))
+                          for c, f, p in zip(cap_states_strs[0::3],
+                                          cap_states_strs[1::3],
+                                          cap_states_strs[2::3])]
             freqs = target.cpufreq.list_frequencies(cpu)
```
qperret: Then I'm wondering if you actually need that ? There is a risk it is misaligned with the EM no ? Perhaps you could use the EM's frequencies directly when `freq_model_supported == True` or something ?

qperret: Hmm that doesn't really work with your second commit I guess, so nevermind.

credp: There is a risk that we don't match, you're right. I think if the EM doesn't match the frequencies at this stage though, we're lost already and I'm not sure we can do anything about it. It would be possible to take the capacities of the OPPs from the EM and check it matches using the reported cpu_capacity and available_frequencies nodes I think.

PR #769, opened 2019-02-21
Wltest on P3

qperret:
Marking this as WiP since it needs further testing to make sure it is robust enough.

This is basically the least horrible way I could find to integrate the Google build system thing into the wltest infrastructure. The basic idea is to set `WLTEST_REPO_TARGET=y` in the platform's `definitions` file, and then check that in `test_series` to know if we're dealing with a 'normal' kernel source tree managed with git, or with a 'repo' tree using manifests and all.

The nice thing is, the API stays the same. You just need to make `--kernel_src` point at your repo, give a series of sha1s for the repo manifest, and things should 'just work'.

The part I don't like is that we're making `test_series` play with  _manifest_ sha1s, not kernel sha1s. And so that makes it a pain to check if we've flashed the right kernel on the target since obviously the kernel and manifest sha1s are different. So I basically disabled the checks for now.

The alternative would be make `--kernel_src` point at the actual kernel source tree, and to have an additional `--repo_path` parameter, or something like that. But that makes the command line even worst than it already is so I'm not sure if it's worth it.

The instructions on how to use that thing are in one of the commit messages. We should probably try and find a better place to store them, I guess.

Feedback welcome !


tools/wltests/test_series
```diff
@@ -118,7 +124,9 @@ Additional arguments:
 	--adb			The ADB binary to use.
 				Default: ADB=[ANDROID_HOME/platform-tools/adb|\$(which adb)]
 	--fastboot		The FASTBOOT binary to use.
-				Default: FASTBOOT=[ANDROID_HOME/platform-tools/fatboot|\$(which fatboot)]
+				Default: FASTBOOT=[ANDROID_HOME/platform-tools/fastboot|\$(which fastboot)]
+	--repo			The REPO binary to use.
```
valschneider: Does it make sense to override `repo`? It's a bit like specifying which `git` binary to use, no?

qperret:
> > The nice thing is, the API stays the same. You just need to make `--kernel_src` point at your repo, give a series of sha1s for the repo manifest, and things should 'just work'.
> > The part I don't like is that we're making `test_series` play with _manifest_ sha1s, not kernel sha1s.
> 
> IIUC, we have to craft as many manifests as kernels we want to test?

That's right. The .repo/manifest/ folder in a repo-managed project is basically a normal git tree with manifests. In the case of P3, the manifest is used to checkout the toolchain, the build scripts, the modules, and the kernel source.

> > The alternative would be make `--kernel_src` point at the actual kernel source tree, and to have an additional `--repo_path` parameter, or something like that. But that makes the command line even worst than it already is so I'm not sure if it's worth it.
> 
> We already have half a gazillion parameters, what's one extra?
> I think specifying a repo_path is better (or at least, less bad) since we don't have to juggle with manifests and we can verify the sha1 of the flashed kernel.

We pretty much _have_ to juggle with manifests anyways, the repo need to be kept in sync. Maybe a kernel change imply to change something in the modules too. So you need to checkout 'matching' commits in both the kernel and modules trees, which is basically what 'repo' is about. And supporting that actually paves the way to test loads more things (different toolchains, potentially userspace changes, ...) for free, even though we probably care less about that.

Thinking more about it, what I could do is specify in the platform's definition file a `$REPO_KERNEL_PATH` variable in the platform's `definitions` file, relative to the root folder of the repo, so that I know where to find the kernel in `test_series`. And then I can actually implement the thing that checks the kernel's sha1 properly, I think.

I hope that sort of makes sense. I'll try to implement something to see where I can get.

> We're nearing the point where wltests would also need a nice facelift, but ain't nobody's got time for that.

Yeah it'd be nice to use the standard target config, emeter config, ... instead of having a wltest-specific one. And bash sucks. But that's for another time I guess.

> > The instructions on how to use that thing are in one of the commit messages. We should probably try and find a better place to store them, I guess.
> 
> We definitely need a wltests section in the LISA documentation, but that's not a trivial task :(

+1

tools/wltests/test_series
qperret: Right, perhaps I can remove that one :)

valschneider:
> We pretty much _have_ to juggle with manifests anyways, the repo need to be kept in sync. Maybe a kernel change imply to change something in the modules too. So you need to checkout 'matching' commits in both the kernel and modules trees, which is basically what 'repo' is about.

Urgh :(

> And supporting that actually paves the way to test loads more things (different toolchains, potentially userspace changes, ...) for free, even though we probably care less about that.
> 
> Thinking more about it, what I could do is specify in the platform's definition file a `$REPO_KERNEL_PATH` variable in the platform's `definitions` file, relative to the root folder of the repo, so that I know where to find the kernel in `test_series`. And then I can actually implement the thing that checks the kernel's sha1 properly, I think.
> 

That could work.

qperret:
So, I went for the `REPO_KERNEL_PATH` option. Basically, for a platform using a repo-managed tree, the `definitions` file must have the following:

```
WLTEST_REPO_TARGET"y
REPO_KERNEL_PATH="relative/path/to/kernel/tree"
```

`test_series` will checkout manifests and `sync` the repo. And after the kernel/modules have been flashed, it tests that the kernel is actually on the target by looking at the current HEAD under `$REPO_KERNEL_PATH`.

In the end, the command looks like: 
```lisa-wltest-series -k /path/to/repo -s /path/to/manifests.sha1 -p bluecross_android_v4.9 --wa_agenda /path/to/agenda.yaml -d mydevice```
which is nice and short, I think :smile:. And things _should_ just work.

Anyway, I moved some stuff from the wahoo folder so please hold on before merging this, we need testing first.

valschneider:
> In the end, the command looks like:
> `lisa-wltest-series -k /path/to/repo -s /path/to/manifests.sha1 -p bluecross_android_v4.9 --wa_agenda /path/to/agenda.yaml -d mydevice`

That looks sane to me.

qperret:
FWIW I was trying to look at a better way of implementing this thing, and it turns out using `repo start <branch name>` (or `repo checkout <branch name>`) might be preferable over doing `git -C .repo/manifests checkout <sha1> && repo sync`. Indeed, `repo sync` does a whole lot of things like rebasing branches behind your back and things like that. But in order to do that, we'd need to put branch names into the series file and not sha1s (repo doesn't understand sha1 since it manages different git trees).

I'm still figuring out what exactly can be done and what can't, so yeah, this is still very much WiP :/

qperret:
Yeah I think the `repo start <branchname> --all` thingy is probably better. Or maybe just `repo sync -l` at least. Not sure yet. So please hang on before hitting the merge button :wink:

qperret:
OK, I ended up using `repo sync -l` which doesn't rebase things behind your back. So I'm relatively happy with this now. I tested it quite extensively (ran tests for several days, flashing 8 different kernels) and things look good.

PR #770, opened 2019-02-21
lisa/tests: load_tracking: Fix get_task_sched_signals uses

valschneider:
I somehow missed that the "signals" parameter is unused, and sort of
broke the plotting (we plot the whole df). The assertions themselves
do fetch the right signal from the df however.

Remove that parameter and cleanup the remaining places where we didn't
do df[signal_name].


valschneider:
I ran
```
exekall run lisa/tests/scheduler/load_tracking.py --conf target_conf.yml
```

and all tests passed (on an older integration branch so time scaling doesn't get in the way), so I think I didn't break everything :)

PR #771, opened 2019-02-21
document testing workflows

douglas-raillard-arm:
Add user-guide style documentation on synthetic tests workflows using bisector and exekall. That is meant to be a kind of reference flow, with documentation and command line examples, and without too much advanced content. More details about the concepts of exekall and bisector will be documented in their own documentation to come.

### DIFF
--- a/doc/users_guide.rst
+++ b/doc/users_guide.rst
@@ -84,6 +84,27 @@ access to many LISA related functions (done automatically by Vagrant)::
 
 .. tip:: Run ``lisa-help`` to see an overview of the provided LISA commands.
 
+Running tests
+=============
+
+The shortest path to executing a test is:
+
+1. Update the ``target_conf.yml`` file located at the root of the repository
+   with the credentials to connect to the development board (see
+   :class:`~lisa.env.TargetConf` keys for more information)
+2. Run the following:
+
+  .. code-block:: sh
+
+    # To run all tests
+    lisa-test
+    # To list available tests
+    lisa-test --list
+    # To run a test matching a pattern
+    lisa-test '*test_task_placement'
+
+More advanced workflows are described at :ref:`automated-testing-page`.
+
 Updating
 ========
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -34,6 +34,7 @@ Contents:
    users_guide
    contributors_guide
    kernel_tests
+   automated_testing
    wlgen
    internals
    analysis
--- a/doc/internals.rst
+++ b/doc/internals.rst
@@ -67,3 +67,6 @@ Miscellaneous utilities
 
 .. automodule:: lisa.target_script
    :members:
+
+.. automodule:: lisa.regression
+   :members:
--- a/doc/pedantic_build.sh
+++ b/doc/pedantic_build.sh
@@ -19,7 +19,7 @@
 
 # The documentation of those modules causes some issues,
 # be permissive for warnings related to them
-IGNORE_PATTERN="devlib|bart|wa"
+IGNORE_PATTERN="devlib|bart|wa|exekall.engine"
 
 make SPHINXOPTS='-n' html 2>doc_build.log
 all_warns=$(cat doc_build.log)
--- a/lisa/regression.py
+++ b/lisa/regression.py
@@ -60,7 +60,7 @@ class RegressionResult:
     def from_result_list(cls, testcase_id, old_list, new_list, alpha=None):
         """
         Build a :class:`RegressionResult` from two list of
-        :class:`lisa.tests.kernel.test_bundle.Result`, or objects that can be
+        :class:`lisa.tests.base.Result`, or objects that can be
         converted to `bool`.
 
         .. note:: Only ``FAILED`` and ``PASSED`` results are taken into account,
@@ -70,10 +70,10 @@ class RegressionResult:
         :type testcase_id: str
 
         :param old_list: old series
-        :type old_list: list(lisa.tests.kernel.test_bundle.Result)
+        :type old_list: list(lisa.tests.base.Result)
 
         :param new_list: new series
-        :type new_list: list(lisa.tests.kernel.test_bundle.Result)
+        :type new_list: list(lisa.tests.base.Result)
 
         :param alpha: Alpha risk of the statistical test
         :type alpha: float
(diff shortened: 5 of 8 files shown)


doc/users_guide.rst
```diff
@@ -84,6 +84,27 @@ access to many LISA related functions (done automatically by Vagrant)::
 
 .. tip:: Run ``lisa-help`` to see an overview of the provided LISA commands.
 
+Running tests
```
valschneider: So I'm torn between having this here or in the kernel testing section.

Maybe we could have a "LISA use cases" section in the users guide that briefly points to kernel tests / trace analysis?

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
+.. _workflows-page:
```
valschneider: I just want to say this is pretty damn awesome :D

doc/users_guide.rst
douglas-raillard-arm: Initially, I had that paragraph in the workflows.rst introduction. After that, I realized that the user guide provided no clue about running the tests. So the reasoning was: if you just look at the user guide, you get the quickest path to running tests in the simplest way possible. This way the user-guide is self-contained, especially for somebody coming to LISA the first time and who just wants to get a feeling of what it can do.

If you plan on using LISA a bit more, then you are probably ready to read a bit more to see what kind of workflow suits your need. Maybe it would make sense to have a toplevel "LISA use cases" page, with a subpage for exekall/bisector and another one for notebooks.

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
...
+
+  # The energy_est_threshold_pct parameter of functions with a name matching 
+  # '*test_task_placement' will take the following values all values from 0 to 15
+  # by increments of 5.
+  exekall run lisa/tests/ --conf target_conf.yml --sweep '*test_task_placement' energy_est_threshold_pct 0 15 5
+
+When something went wrong
+-------------------------
+
+``--replay`` provides a simple way of re-executed the last few stages of a test
```
valschneider:
```suggestion
``--replay`` provides a simple way of re-executing the last few stages of a test
```

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
...
+that had an error. That can be used to reproduce a bug in the test code that
+makes it raise an exception while working on a fix. ``--replay`` takes the UUID
```
valschneider: I'd add something to point out that this will not re-run the workload:
```suggestion
makes it raise an exception while working on a fix without having to re-execute the workload on a board (and thus without needing a board at all). ``--replay`` takes the UUID
```

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
...
+them an infinite number of times (by default). These steps can involve flashing
+the board, rebooting it, and running a test command. If one step goes wrong,
+bisector implements the logic to retry, abort it, mark it as good or bad
+depending on the type of step used. By now, you may have noticed some
+similarities between ``bisector`` behaviour and what is expected of the command
+executed by ``git bisect run`` [#]_: that is no coincidence, as both ``bisector
+run`` and ``bisector report`` can be used as a ``git bisect run``-compliant
+script.
+
+``bisector run`` then records all the output of the steps in a machine-readable
```
valschneider:
```suggestion
``bisector run`` records all the output of the steps in a machine-readable
```

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
...
+
+  shell/shell (shell) [GOOD]
+      command: sshpass -p password ssh-copy-id -i $HOME/.ssh/id_rsa "$USER@$HOSTNAME"
+      #1 : OK
+      #2 : OK
+      #3 : OK
+      #4 : OK
+      #5 : OK
+
+  test/behaviour (exekall-LISA-test) [BAD]
```
valschneider: Why [BAD]? Seems like everything went right on this run.

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
...
+  Overall bisect result: good commit
+
+There is one section per step, reflecting the steps configuration. Each step
+will aggregate the results of all its iterations. The header is formatted as
+*<step name>/<step category> (step class name) [<step result>]*. The overall
+bisect result is the combination of the result of each steps.
+
+``exekall-LISA-test`` has special support for inspecting ``exekall``'s database
+collected during each iteration of ``bisector``, and can display a summary
+table. By default, a *passed* label will only appear if all iteration
```
valschneider: Italic doesn't catch the eye very well, so I'd rather go with bold:
```suggestion
table. By default, a **passed** label will only appear if all iteration
```

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
...
+successfully passed.  Otherwise, an appropriate combination of *FAILED*,
+*ERROR*, *SKIPPED* and *UNDECIDED* lines will be displayed with the
```
valschneider: Ditto for italic

doc/workflows.rst
```diff
@@ -0,0 +1,505 @@
...
+corresponding count.
+
+Various options can affect what is displayed and taken into account. For
+example, ``--skip my-other-test`` will remove the contribution of that step to
+the final result. Step-specific report options are documented in ``bisector
+step-help``. Some of the options allow exporting collected artifacts from the
+report, like ``-o export-logs``. In the case of ``exekall-LISA-test`` step,
```
valschneider: Über-nit: We use the ``-o <option>`` format here, but further down we use ``-o<option>``.

doc/workflows.rst
douglas-raillard-arm: That is not necessarily true. If the exception happened when parsing the target configuration or platform info, it will restart from there and will use the board. It depends where the failure happened, the only guarantee you have is that all that can be "salvaged" will be.

You could even come up with trickier cases where the last stage wants a reference to the TestEnv. Since that cannot be serialized, it would have to be rebuilt, even if "intermediate" stages  like the TestBundle are available .

valschneider: True, I wanted to add that bit as an example rather than a forever true statement (the sentence does start with `that *can* be used to`)

doc/workflows.rst
douglas-raillard-arm: Actually both are allowed, since it is a regular gnu-style one-letter-option.

douglas-raillard-arm:
Added "More on steps options" and "I don't want a config file" subsections for bisector. Not sure if they really belong to that guide or if that should only appear in bisector's own documentation, but I'm just producing content here (that will likely be reshuffled anyway).

doc/workflows.rst
valschneider: Yeah I know, which is why it's an über-nit - it's just to keep some form of consistency :)

doc/workflows.rst
douglas-raillard-arm: got it

douglas-raillard-arm:
Renamed the page and title to "Automated testing" to match the coming doc reorg.

PR #773, opened 2019-02-22
shell: Introduce LISA_PRESERVE_SHELL

qperret:
Sourcing init_env basically provides a set of bash functions (the lisa-*
commands) to the user and activates the python venv. However, that also
changes the PS1 and the colorscheme of the current shell, although this
isn't strictly necessary.

In order to give users the option of keeping their original shell
configuration, make the Lisa PS1 and colors depend on $LISA_SHELL_QUIET
not being set. While at it, list the env variables exported by init_env.

This results in:
  queper01@queper01-lin:~/lisa$ LISA_SHELL_QUIET=1 source init_env
  Activating LISA Python 3.6 venv (/data/work/lisa-next/.lisa-venv-3.6) ...
  LISA version:
      /data/work/lisa-next
      branch: heads/lisa-shell-quiet
      commit: 7f07c99ecec

  Python version: 3.6

  JupyterLab v0.35.4
  No installed extensions
  Environment variables:
  LISA_SHELL_QUIET=1
  LISA_CONF=/data/work/lisa-next/target_conf.yml
  LISA_HOME=/data/work/lisa-next
  LISA_RESULT_ROOT=/data/work/lisa-next/results
  LISA_USE_VENV=1
  LISA_HOST_ABI=x86_64
  LISA_DEVMODE=1
  LISA_VENV_PATH=/data/work/lisa-next/.lisa-venv-3.6
  queper01@queper01-lin:~/lisa$

without changing colors.

### DIFF
--- a/shell/lisa_shell
+++ b/shell/lisa_shell
@@ -30,6 +30,9 @@ LISA_PYTHON=${LISA_PYTHON:-python3}
 # By default use internal libraries
 export LISA_DEVMODE=${LISA_DEVMODE:-1}
 
+# By default use Lisa's PS1 and colorscheme for the shell
+export LISA_PRESERVE_SHELL=${LISA_PRESERVE_SHELL:-0}
+
 # Setup colors
 source "$LISA_HOME/shell/lisa_colors"
 
@@ -426,21 +429,24 @@ EOF
 # LISA Shell MAIN
 ################################################################################
 
-# Dump out a nice LISA Shell logo
-clear
-echo -e "$LISASHELL_BANNER"
+if [[ $LISA_PRESERVE_SHELL == 0 ]]; then
+    # Dump out a nice LISA Shell logo
+    clear
+    echo -e "$LISASHELL_BANNER"
 
-echo "                                                                               "
-echo "                        .:: LISA Shell ::.                                     "
-echo "                                                                               "
-echo -ne "$LISASHELL_RESET"
-echo
+    echo "                                                                               "
+    echo "                        .:: LISA Shell ::.                                     "
+    echo "                                                                               "
+    echo -ne "$LISASHELL_RESET"
+    echo
+fi
 
 # Activate the venv unless it was explicitely disabled
 lisa-venv-activate
 
-echo -ne "$LISASHELL_RESET$LISASHELL_BLUE"
-cat <<EOF
+if [[ $LISA_PRESERVE_SHELL == 0 ]]; then
+    echo -ne "$LISASHELL_RESET$LISASHELL_BLUE"
+    cat <<EOF
 
 Welcome to the Linux Integrated System Analysis SHELL!
 
@@ -449,7 +455,11 @@ $(lisa-version)
 Type "lisa-help" for on-line help on available commands
 EOF
 
-# Setup default SHELL text color
-echo -e "$LISASHELL_DEFAULT"
-
+    # Setup default SHELL text color
+    echo -e "$LISASHELL_DEFAULT"
+else
+    lisa-version
+    echo "Environment variables:"
+    printenv | grep --color=no -E 'LISA|EXEKALL' | sort
+fi
 # vim: set tabstop=4 shiftwidth=4 textwidth=80 expandtab:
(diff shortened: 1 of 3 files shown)


shell/lisa_shell
```diff
@@ -31,7 +31,9 @@ LISA_PYTHON=${LISA_PYTHON:-python3}
 export LISA_DEVMODE=${LISA_DEVMODE:-1}
 
 # Setup colors
-source "$LISA_HOME/shell/lisa_colors"
+if [ -z $LISA_SHELL_QUIET ]; then
```
derkling: Can we use the same "pattern" we already have for `LISA_DEVMODE` ?

```python
export LISA_SHELL_QUIET=${LISA_SHELL_QUIET:-0}
#...
if [[ $LISA_SHELL_QUIET == 1]]; then
else
fi
```

qperret: Yes, much better. That should be done now.

shell/lisa_shell
```diff
@@ -449,7 +457,11 @@ $(lisa-version)
...
 
-# Setup default SHELL text color
-echo -e "$LISASHELL_DEFAULT"
-
+    # Setup default SHELL text color
+    echo -e "$LISASHELL_DEFAULT"
+else
+    lisa-version
+    echo "Environment variables:"
+    printenv | grep --color=no LISA
```
valschneider: :heart:

shell/lisa_shell
```diff
@@ -30,8 +30,13 @@ LISA_PYTHON=${LISA_PYTHON:-python3}
 # By default use internal libraries
 export LISA_DEVMODE=${LISA_DEVMODE:-1}
 
+# By default use Lisa's PS1 and colorscheme for the shell
+export LISA_SHELL_QUIET=${LISA_SHELL_QUIET:-0}
```
valschneider: Being pedantic here, but the name irks me somewhat. Yes, you don't get the banner dumped, but you still get some other information dumped. It's about impacting the current shell.

Naming is hard though. `LISA_NO_COLORS`, `LISA_SHELL_PRISTINE`...?

shell/lisa_shell
qperret: Yeah, I stole the idea from AOSP's lunch thingy. But I just figured, this is in fact half broken. We export more stuff than just `LISA_` variables. We also have stuff for exekall and wltests. So maybe I should try and do a diff on `printenv` before/after sourcing init_env or something ...

valschneider: That could work, but this is already a good step forward anyway.

shell/lisa_shell
qperret: Yeah I suck at naming things, so whatever you think is best will work for me ;)

valschneider: The latest word shuffle that came out of my head is `LISA_PRESERVE_SHELL`, shall we go for that?

qperret: Yep, sounds good

shell/lisa_shell
qperret: OK, I'll keep things simple for now then. We can always extend later if need be.

shell/lisa_shell
```diff
@@ -30,8 +30,13 @@ LISA_PYTHON=${LISA_PYTHON:-python3}
 # By default use internal libraries
 export LISA_DEVMODE=${LISA_DEVMODE:-1}
 
+# By default use Lisa's PS1 and colorscheme for the shell
+export LISA_PRESERVE_SHELL=${LISA_PRESERVE_SHELL:-0}
+
 # Setup colors
-source "$LISA_HOME/shell/lisa_colors"
+if [[ $LISA_PRESERVE_SHELL == 0 ]]; then
```
douglas-raillard-arm: we can source unconditionally and move the check inside lisa_colors, alongside the other 2 checks we have

shell/lisa_shell
douglas-raillard-arm: For exekall and bisector, env var are only used as defaults values, so it's probably better to use the CLI option instead of updating these in 3rd party scripts. (one exception are Artifactorial token and folder).

A plain diff will also show things that should not be modified related to the venv, so IMO it's better to have an explicit grep like you did. It can be expanded to cover bisector, exekall and wltest.

qperret:
I moved the check inside lisa_colors and made the printenv stuff display the exekall variable too. The WA and wltests variables are exported only upon invocation of lisa-wltest-series, so we just can't show them at this point.

PR #777, opened 2019-02-25
readme: Fix broken link

msrasmussen:


valschneider:
@douglas-raillard-arm I was thinking of adding a test for this, say a simple script that gets the URLs out of the README and verifies that they don't end on a 404. This wouldn't have caught this issue though, since the page still existed but not longer held the right content :/

douglas-raillard-arm:
@valschneider will be fixed when that README goes away, replaced by an actual sphinx page included in the doctree and rendered as a manpage :)

PR #779, opened 2019-02-25
analysis/tasks: Compute all task_states dataframes at once

valschneider:
This was benchmarked against a 400MB trace::

```
  %%time
  df = trace.analysis.tasks.df_tasks_states()

  CPU times: user 1.18 s, sys: 100 ms, total: 1.28 s
  Wall time: 1.28 s
```
This gave me a DataFrame with 807089 rows weighing a total of
~45MB.

Note that adding the ``next_state`` column costed me ~0.25s
extra (compared to only computing the deltas). It simplifies the
stringification of the multi & single tasks dataframes, but I could
move it to be done only for single task dataframes.


lisa/analysis/tasks.py
```diff
@@ -199,27 +199,28 @@ def df_rt_tasks(self, min_prio=100):
 
         return rt_tasks
 
+    @memoized
     @requires_events('sched_switch', 'sched_wakeup')
-    def df_task_states(self, task, stringify=False):
+    def df_tasks_states(self, stringify=False):
```
derkling: I would keep stringy True by default.

I think that's the most simple API for "normal users" while, advanced users which wants to do more fancy stuff are never scared to add a parameter.

lisa/analysis/tasks.py
```diff
@@ -228,47 +229,106 @@ def df_task_states(self, task, stringify=False):
...
+        all_sw_df = prev_sw_df.append(next_sw_df, sort=False)
+        all_sw_df.sort_index(inplace=True)
 
         # Integer values are prefered here, otherwise the whole column
         # is converted to float64
-        task_switches_df['target_cpu'] = -1
+        all_sw_df['target_cpu'] = -1
+
+        df = all_sw_df.append(wk_df, sort=False)
+        df.sort_index(inplace=True)
```
derkling: cannot we drop the sort at L245 and do just that once we already appended `next_`'s and `wk_`'s content to all_sw_df ?

derkling: Did you try to use mergesort instead of quicksort?... it's sometimes more cache efficient and perhaps it can buy you something especially on small DFs or on very big ones not fitting in memory.

lisa/analysis/tasks.py
valschneider: Well, I changed that in https://github.com/ARM-software/lisa/pull/749  because this forces us to use `TaskState.TASK_ACTIVE.char` for comparisons instead of `TaskState.TASK_ACTIVE`.

IMHO the most common use case is slicing the dataframe depending on certain task states, and using`TaskState.TASK_*` instead of `TaskState.TASK_*.char` is more instinctive.

lisa/analysis/tasks.py
valschneider: > cannot we drop the sort at L245 and do just that once we already appended `next_`'s and `wk_`'s content to all_sw_df ?

Right, that should work.

> Did you try to use mergesort instead of quicksort?... it's sometimes more cache efficient and perhaps it can buy you something especially on small DFs or on very big ones not fitting in memory.

I just did and didn't notice any notable difference.

valschneider:
* Removed two stray `sort_index()`s.

lisa/analysis/tasks.py
derkling: Have we ever considered to add both the state ID and the state label?

This way we can also get rid of the oddly stringify for just a small memory overhead and make happy both coders and humans?

valschneider: Doing this after-the-fact, IOW:
```python
df["curr_state_str"] = stringify_task_state_column(df["curr_state"])
df["next_state_str"] = stringify_task_state_column(df["next_state"])
```
more than doubles the execution time (and increases the memory footprint of the df by ~15%).

I'd rather keep things as is and maybe change the `stringify_df_task_states` method to add new columns instead of modifying the existing ones.

valschneider:
In the end I changed the `stringify` logic to always **add** new columns instead of modifying existing ones.

I removed the `stringify` parameter from `df_tasks_states` though because it would mean we'd memoize both versions of the dataframe (with and without stringified columns) which seems like a waste.

I'll need to update an example notebook to reflect this, but I'll wait for https://github.com/ARM-software/lisa/pull/673 to get in first

lisa/analysis/tasks.py
```diff
@@ -365,6 +365,32 @@ def stringify_df_task_states(cls, df, columns, inplace=False):
...
+    def df_tasks_runtime(self):
+        """
+        DataFrame of the time each task spent in TASK_ACTIVE (:class:`TaskState`)
+
+        :returns: a :class:`pandas.DataFrame` with:
+
+          * PIDs as index
+          * A ``runtime`` column (the time that task spent running)
+        """
+        df = self.df_tasks_states()
```
derkling: I think for this `df` it could be very useful to add a column with the task's `comm`... I guess we will end up most of the time wondering _"what task is that PID"_ ?

valschneider: We can always get the PID from `get_task_pid()` but this dataframe should always be relatively small, so we can add that info.

PR #783, opened 2019-02-28
lisa/tests: Handle noisy tasks

valschneider:
This is based on top of https://github.com/ARM-software/lisa/pull/779

This involved me losing many brain cells fighting with trace.py. Details in the commits logs.

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -367,11 +367,36 @@ class RTATestBundle(TestBundle, abc.ABC):
     definitions.
     """
 
+    def trace_window(self, trace):
+        """
+        The time window to consider for this :class:`RTATestBundle`
+
+        :returns: a (start, stop) tuple
+
+        Since we're using rt-app profiles, we know the name of tasks we are
+        interested in, so we can trim our trace scope to filter out the
+        setup/teardown events we don't care about.
+
+        Override this method if you need a different trace trimming.
+
+        .. warning::
+
+          Don't call ``self.trace`` in here unless you like infinite recursion.
+        """
+        sdf = trace.df_events('sched_switch')
+
+        # Find when the first task starts running
+        rta_start = sdf[sdf.next_comm.isin(self.rtapp_profile.keys())].index[0]
+        # Find when the last task stops running
+        rta_stop = sdf[sdf.prev_comm.isin(self.rtapp_profile.keys())].index[-1]
+
+        return (rta_start, rta_stop)
+
     @property
     @memoized
     def trace(self):
         """
-        :returns: a Trace
+        :returns: a :class:`lisa.trace.TraceView`
 
         Having the trace as a property lets us defer the loading of the actual
         trace to when it is first used. Also, this prevents it from being
@@ -380,7 +405,9 @@ class RTATestBundle(TestBundle, abc.ABC):
         match a different folder structure.
         """
         path = os.path.join(self.res_dir, 'trace.dat')
-        return Trace(path, self.plat_info, events=self.ftrace_conf["events"])
+
+        trace = Trace(path, self.plat_info, events=self.ftrace_conf["events"])
+        return trace.get_view(self.trace_window(trace))
 
     def __init__(self, res_dir, plat_info, rtapp_profile):
         super().__init__(res_dir, plat_info)
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -16,11 +16,14 @@
 #
 
 import enum
+import functools
 import os
 import os.path
 import abc
 import sys
+import textwrap
 from collections.abc import Mapping
+from inspect import signature, Parameter
 
 from devlib.target import KernelVersion
 
@@ -467,6 +470,74 @@ class RTATestBundle(TestBundle, abc.ABC):
         res.add_metric("noisiest task", metric)
 
         return res
+
+    @classmethod
+    #pylint: disable=unused-argument
+    def check_noisy_tasks(cls, noise_threshold_pct=None, noise_threshold_ms=None):
+        """
+        Decorator that applies :meth:`test_noisy_tasks` to the trace of the
+        :class:`TestBundle` returned by the underlying method. The :class:`Result`
+        will be changed to :attr:`Result.UNDECIDED` if that test fails.
+
+        We also expose :meth:`test_noisy_tasks` parameters to the decorated
+        function.
+        """
+        def decorator(func):
+            @cls.test_noisy_tasks.used_events
+            @functools.wraps(func)
+            def wrapper(self, *args,
+                        noise_threshold_pct=noise_threshold_pct,
+                        noise_threshold_ms=noise_threshold_ms,
+                        **kwargs):
+                res = func(self, *args, **kwargs)
+
+                noise_res = self.test_noisy_tasks(
+                    noise_threshold_pct, noise_threshold_ms)
+                res.metrics.update(noise_res.metrics)
+
+                if not noise_res:
+                    res.result = Result.UNDECIDED
+
+                return res
+
+            # https://stackoverflow.com/a/33112180
+            # The wrapper has all of `func`'s parameters plus `test_noisy_tasks`',
+            # but since we use `wraps(func)` we'll only get doc/autocompletion for
+            # `func`'s. Expose the extra parameters to the decorated function to
+            # make it more user friendly.
+            func_sig = signature(func)
+            dec_params = signature(cls.check_noisy_tasks).parameters
+
+            # We want the default values of the new parameters for the
+            # *decorated* function to be the values passed to the decorator,
+            # which aren't the default values of the decorator.
+            new_params = [
+                dec_params["noise_threshold_pct"].replace(default=noise_threshold_pct),
+                dec_params["noise_threshold_ms"].replace(default=noise_threshold_ms)
+            ]
+            wrapper.__signature__ = func_sig.replace(
+                parameters=list(func_sig.parameters.values()) + new_params
+            )
+
+            # Make it obvious in the doc where the extra parameters come from
+            merged_doc = textwrap.dedent(cls.test_noisy_tasks.__doc__).splitlines()
+            # Replace the one-liner func description
+            merged_doc[1] = textwrap.dedent(
+                """
+                **Added by** :meth:`~{}.{}.{}`:
+
+                The returned ``ResultBundle.result`` will be changed to
+                :attr:`~lisa.tests.base.Result.UNDECIDED` if the environment was
+                too noisy:
+                """.format(cls.__module__, cls.__name__, cls.check_noisy_tasks.__name__)
+            )
+
+            #pylint: disable=no-member
+            wrapper.__doc__ = textwrap.dedent(wrapper.__doc__) + "\n".join(merged_doc)
+
+            return wrapper
+        return decorator
+
     @classmethod
     def unscaled_utilization(cls, plat_info, cpu, utilization_pct):
         """
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -24,8 +24,8 @@ from collections.abc import Mapping
 
 from devlib.target import KernelVersion
 
-
-from lisa.trace import Trace
+from lisa.analysis.tasks import TasksAnalysis
+from lisa.trace import Trace, requires_events
 from lisa.wlgen.rta import RTA
 
 from lisa.utils import Serializable, memoized, ArtifactPath
@@ -367,6 +367,13 @@ class RTATestBundle(TestBundle, abc.ABC):
     definitions.
     """
 
+    NOISE_IGNORED_PIDS = [0]
+    """
+    PIDs to ignore in :meth:`test_noisy_tasks`.
+
+    PID 0 is the idle task, don't count it as noise.
+    """
+
     def trace_window(self, trace):
         """
         The time window to consider for this :class:`RTATestBundle`
@@ -413,6 +420,53 @@ class RTATestBundle(TestBundle, abc.ABC):
         super().__init__(res_dir, plat_info)
         self.rtapp_profile = rtapp_profile
 
+    @TasksAnalysis.df_tasks_runtime.used_events
+    def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
+        """
+        Test that no non-rtapp ("noisy") task ran for longer than the specified thresholds
+
+        :param noise_threshold_pct: The maximum allowed runtime for noisy tasks in
+          percentage of the total rt-app execution time
+        :type noise_threshold_pct: float
+
+        :param noise_threshold_ms: The maximum allowed runtime for noisy tasks in ms
+        :type noise_threshold_ms: float
+
+        If both are specified, the smallest threshold (in seconds) will be used.
+        """
+        if noise_threshold_pct is None and noise_threshold_ms is None:
+            raise ValueError('Both "{}" and "{}" cannot be None'.format(
+                "noise_threshold_pct", "noise_threshold_ms"))
+
+        # No task can run longer than the recorded duration
+        threshold_s = self.trace.time_range
+
+        if noise_threshold_pct is not None:
+            threshold_s = noise_threshold_pct * self.trace.time_range / 100
+
+        if noise_threshold_ms is not None:
+            threshold_s = min(threshold_s, noise_threshold_ms * 1e3)
+
+        df = self.trace.analysis.tasks.df_tasks_runtime()
+        test_tasks = list(map(self.trace.get_task_pid, self.rtapp_profile.keys()))
+        df_noise = df[~df.index.isin(test_tasks + self.NOISE_IGNORED_PIDS)]
+
+        if df_noise.empty:
+            return ResultBundle.from_bool(True)
+
+        pid = df_noise.index[0]
+        comm = df_noise.comm.values[0]
+        duration_s = df_noise.runtime.values[0]
+        duration_pct = duration_s * 100 / self.trace.time_range
+
+        res = ResultBundle.from_bool(duration_s < threshold_s)
+        metric = {"pid" : pid,
+                  "comm": comm,
+                  "duration (abs)": TestMetric(duration_s, "s"),
+                  "duration (rel)" : TestMetric(duration_pct, "%")}
+        res.add_metric("noisiest task", metric)
+
+        return res
     @classmethod
     def unscaled_utilization(cls, plat_info, cpu, utilization_pct):
         """
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -357,7 +357,10 @@ class RTATestBundle(TestBundle, abc.ABC):
     """
 
     ftrace_conf = FtraceConf({
-        "events" : ["sched_switch"],
+        "events" : [
+            "sched_switch",
+            "sched_wakeup"
+        ],
     }, __qualname__)
     """
     The FTrace configuration used to record a trace while the synthetic workload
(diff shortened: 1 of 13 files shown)


lisa/trace.py
```diff
@@ -305,16 +280,52 @@ def has_events(self, events):
...
+            end = self._trace_start
+            for trace_class in trace_classes:
+                df = trace_class.data_frame[:t_max]
+                if not df.empty:
+                    end = max(end, df.index[-1])
+            t_max = end
+        else:
+            t_max = self._trace_end
+
+        self.window = (t_min, t_max)
```
valschneider: I'm not entirely sure if we this or separate `start_time`/`end_time` members. Or if that shouldn't be named `time_window`.

lisa/trace.py
```diff
@@ -305,16 +280,52 @@ def has_events(self, events):
         else:
             return set(events).issubset(set(self.available_events))
 
+    def set_window(self, t_min=None, t_max=None):
```
douglas-raillard-arm: If `set_window()` returned a new object that is just a view over the trace, the memoized business is not needed anymore, as the view object would rightly carry that information. That also allows good compositions, so functions can further trim down the trace without having to care about who is going to use the object next, and if it's ok to change the window size.

valschneider: I like that approach, but I need to wrap my head around all the implications that has on the uses of the trace view instances.

valschneider: Mmm it seems to be working okay. This is what I had to add to the new `TraceView` class:
```python
        # <window computation here>
        # Import here to avoid a circular dependency issue at import time
        # with lisa.analysis.base
        from lisa.analysis.proxy import AnalysisProxy
        self.analysis = AnalysisProxy(self)

    def __getattr__(self, name):
        return getattr(self.trace, name)

    def df_events(self, event):
        """
        Get a dataframe containing all occurrences of the specified trace event
        in the parsed trace.

        :param event: Trace event name
        :type event: str
        """
        df = self.trace.df_events(event)
        if not df.empty:
            df = df[self.start:self.end]

        return df
```

Since we re-create an `AnalysisProxy`, that takes care of `memoized` for us.

qais-yousef:
apologies - I missed this notification; i'll get around the review tomorrow - if I don't poke me!

valschneider:
> apologies - I missed this notification; i'll get around the review tomorrow - if I don't poke me!

No worries, it's still a WIP and needs some more care put into it - especially the TraceView part that might break some tests. But in the meantime if you can have a look at the `check_noisy_tasks()` logic that'd be great!

lisa/tests/base.py
```diff
@@ -456,4 +457,42 @@ def _from_target(cls, target, res_dir, ftrace_coll=None):
...
+                    res = func(self, *args, **kwargs)
+
+                    if not ok:
+                        res.result = Result.UNDECIDED
+
+                    metric = {"pid" : pid,
+                              "comm": comm,
+                              "duration": TestMetric(duration, "s"),
+                              "duration_pct" : TestMetric(duration_pct, "%")}
+                    res.add_metric("noisiest task", metric)
```
qais-yousef: s/noisiest/noisy/?

You're not filtering for the noisiest and just creating a list of noisy tasks, right?

lisa/tests/scheduler/misfit.py
```diff
@@ -295,6 +295,7 @@ def _test_cpus_busy(self, task_state_dfs, cpus, allowed_idle_time_s):
         return res
 
     @TasksAnalysis.df_task_states.used_events
+    @check_noisy_tasks(1)
```
qais-yousef: Hmm that's not very intuitive way to check for noisy tasks. Not for me at least - I am not a big python user in general so I could be the exception here.

What does this do? Why a simple function call from within this function isn't a better alternative?

lisa/tests/base.py
valschneider: No we are actually just picking the noisiest one. There's no need to list all of them (there can be plenty), just the worst offender.

lisa/tests/scheduler/misfit.py
valschneider: In short, the `check_noisy_tasks` decorator turns a call to `func()` into:

```
# This is the function to which that decorator is attached to
res = func(<right parameters dumped here>)

pid, comm, duration = self.find_noisiest_task()
ok = 100 * duration / self.trace.time_range < threshold_pct
add_relevant_metrics(res)

if not ok:
    res.result = Result.UNDECIDED

return res
```

The good thing is it prevents us from having to add that `noisy_task_pct` threshold parameter to every single test method that we want to guard against noisy tasks, and calling the right function just before returning.

But that's also a bad thing - it means we can't easily re-run that test offline while playing with the noise threshold like the other test thresholds.

ATM it's in the form of a decorator because it was the least intrusive way and that made for a simpler RFC, I'll have to think more about it :)

lisa/tests/base.py
qais-yousef: OK so the list is sorted and index[0] contains the task with the highest runtime value?

It depends on the sequence of events but it can happen that the noisiest ran during 'idle' time of the rt-app task but the 2nd busiest is the culprit, no? I don't expect the list of noisy task to be high, have you seen a lot? Was that on Debian or Buildroot? In my testing previously (Debian) I saw 3 or 4 max IIRC.

valschneider: > OK so the list is sorted and index[0] contains the task with the highest runtime value?
> 

Yeah the dataframe is sorted on runtime values, so any value at index 0 will be the one with the highest runtime.

> It depends on the sequence of events but it can happen that the noisiest ran during 'idle' time of the rt-app task but the 2nd busiest is the culprit, no? I don't expect the list of noisy task to be high, have you seen a lot? Was that on Debian or Buildroot? In my testing previously (Debian) I saw 3 or 4 max IIRC.

Right, tasks running during the "active" part of our rt-app workload would have a higher impact, but I think focusing on the longest-running (aka noisiest) undesired task is a safe proxy. It gives us a "worst case scenario" value that we can easily compare against a threshold (which ATM is a % of the total execution time of our rt-app workload).

Also, even if the tasks run while the rt-app tasks are sleeping, they can affect the test in some way or another: frequency increase we wouldn't see otherwise, unexpected load/util increases (w/ blocked load that could impact the rt-app workload even if it's not being coscheduled with the noisy task), etc.

qais-yousef: >Also, even if the tasks run while the rt-app tasks are sleeping, they can affect the test in some way or another: frequency increase we wouldn't see otherwise, unexpected load/util increases (w/ blocked load that could impact the rt-app workload even if it's not being coscheduled with the noisy task), etc.

This further confirms that the timing of when the task run is as important as the duration. But I am happy with this as is if that's what you feel is the best for now. We can improve on this as we go along of we need to :-)

lisa/tests/scheduler/misfit.py
qais-yousef: Cool. Thanks for explanation. If you ended up staying in this route - please add this info in the commit message. It's an important design decision that is worth explaining in the commit message for whoever comes up with a reason to need to change this in the future.

By the way, if the function has a named parameter with a default value - would this go around the need to pass the threshold with every call? Using a function will make for a simpler code if it is as convenience as this approach.

lisa/tests/base.py
valschneider: I'm curious as to what you have in mind - would you rather do something like check the noisiness in each time window where we have an active rt-app task (thus ignoring time windows where we are not running any rt-app task, but are still between first & last rt-app events)?

I do believe taking idle times into account is just as important (as I've explained above). I also have an extra example:

I still haven't ported to our test framework (only as notebook), but I have a test that makes sure we correctly decay nohz blocked load (here's some blurb about it: https://lore.kernel.org/lkml/b1efd041-2faa-76e0-7d9b-3d2917e03012@arm.com/). We *really* don't want anything to run during that decay phase, and during this phase we don't have any rt-app task running (since we want to observe this decay). Arguably that's a bit of a special case, but it does help back the way I've done this check.

lisa/tests/scheduler/misfit.py
douglas-raillard-arm: The main thing the decorator offers is that it can intercept the return value of the original function and update it the way it wants. An alternative way would be to "hijack" explicitly the return value like that:
```
def mytest(..., threshold_pct=42):
    ...
    res = ResultBundle()
    ...
    return check_noisy_tasks(res, threshold_pct)
```
This way exposes `threshold_pct` to the user, the decorator way allows getting the feature without even looking for `return` in the function. Either ways are good to me here, as we don't need to update the function itself (unlike for `requires_events`, where we need to set attributes to be picked up by the documentation).

valschneider:
I've tweaked the decorator a bit to allow for offline tweaking. This means you decorate a test that needs to do some noise checking like so:

```
@RTATestBundle.check_noisy_tasks(noise_threshold_pct=1)
    def test_throughput(self, allowed_idle_time_s=None) -> ResultBundle:
        pass
```

And you can still tweak that offline like this:

```
bundle = StaggeredFinishes.from_target(target)
bundle.test_troughput(noise_threshold_ms=50)
```

The generated doc looks like this: 
image

valschneider:
I finally have something that has been tested and that looks sane. I've propagated the noisiness checking to misfit & eas_behaviour where relevant (e.g. I don't apply it to `test_slack()` but to `test_task_placement()`). We can do that for load_tracking later on.

valschneider:
Woops, forgot to include an update to eas_behaviour `ftrace_conf`, hang on...

lisa/tests/base.py
qais-yousef: No what I was saying is that the noisiest doesn't necessarily correlate to the problematic interfering task. Another noisy task that happens to run at the wrong time but for a much shorter period could be the one that break assumptions. e.g: a kworker waking up at the same time with another task for 20us but causing the rt-app task to migrate. In the same test there could be another noisy task that ran for 0.2 ms which you'd detect as the noisiest but it's completely harmless.

So what I had in mind is keeping a list of how many noisy tasks were there in total and possibly there names/pids to make it easier to look for them in kernelshark. Noisy task is more than just about duration and while I don't think it's worth doing complex checks to define noisy but keeping a full list is helpful IMO as there can be several modes of interference.

valschneider: Right, you've got a point. FWIW, if you execute `trace.analysis.tasks.df_tasks_runtime()` you'll get that list (comm & pid) sorted by total runtime (the noise test just filters that dataframe). It doesn't help for the test but it can help to figure out what to look for in kernelshark

douglas-raillard-arm: Theses metrics are primarily meant for quick inspection of the `ResultBundle` (`-oshow-details` in bisector will show them for example). If we can think of something that is better for human consumption and that would not provide huge output (if that cannot be avoided), it may become a better candidate as a metric. As Valentin pointed, the data are there in any case if somebody needs to sort out the wheat from chaff on a particularly bad set of `TestBundle`.

lisa/tests/base.py
```diff
@@ -466,6 +469,75 @@ def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
         res.add_metric("noisiest task", metric)
 
         return res
+
+    @classmethod
+    #pylint: disable=unused-argument
+    def check_noisy_tasks(cls, noise_threshold_pct=None, noise_threshold_ms=None):
+        """
+        Decorator that applies :meth:`test_noisy_tasks` to the underlying function
```
douglas-raillard-arm: Would reword that as:
```
Decorator that applies :meth:`test_noisy_tasks` to the trace of the `TestBundle` the underlying method is called on.
```

lisa/tests/base.py
```diff
@@ -466,6 +469,75 @@ def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
...
+        def decorator(func):
+            @cls.test_noisy_tasks.used_events
+            @functools.wraps(func)
+            def wrapper(self, *args,
+                        noise_threshold_pct=noise_threshold_pct,
+                        noise_threshold_ms=noise_threshold_ms,
+                        **kwargs):
+                res = func(self, *args, **kwargs)
+
+                print(noise_threshold_pct)
```
douglas-raillard-arm: remove debug print or turn it into a `cls.get_logger().XXX` call

lisa/tests/base.py
```diff
@@ -466,6 +469,75 @@ def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
...
+            # The wrapper has all of `func`'s parameters plus `test_noisy_tasks`',
+            # but since we use `wraps(func)` we'll only get doc/autocompletion for
+            # `func`'s. Expose the extra parameters to the decorated function to
+            # make it more user friendly.
+            func_sig = signature(func)
+            dec_params = signature(cls.check_noisy_tasks).parameters
+            new_params = []
+            # Propagate the default values given to the decorator
+            for name, param in dec_params.items():
+                new_params.append(param.replace(default=locals()[name]))
```
douglas-raillard-arm: is `replace()` needed ? The parameter from `check_noisy_tasks` should already have a `.default` attribute. Accessing the `locals()` seems a bit dangerous since the same name could have been reused/modified.

lisa/tests/base.py
```diff
@@ -466,6 +469,75 @@ def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
...
+            separator = textwrap.dedent(
+                """
+                **Added by** :meth:`~{}.{}.{}`:
+
+                The returned ``ResultBundle.result`` will be changed to
+                :attr:`~lisa.tests.base.Result.UNDECIDED` if the environment was
+                too noisy:
+                """.format(cls.__module__, cls.__name__, cls.check_noisy_tasks.__name__)
+            )
+            separator = textwrap.indent(separator, indent)
```
douglas-raillard-arm: do we really need to re-indent ? Generated docs will look the same,  and the user will probably not care a lot as long as the first column used is consistent. That would simplify into:
```
doc = textwrap.dedent(cls.test_noisy_tasks.__doc__)
separator = textwrap.dedent(...)
doc = [separator] + doc.splitlines()[1:]
wrapper.__doc__ = textwrap.dedent(wrapper_doc) + '\n' + '\n'.join(doc)
```

The last `textwrap.dedent` is necessary since it cannot be assumed that the docstring of the wrapper has the same indendation than the one we just created, so we dedent everything to ensure proper alignment.

lisa/trace.py
```diff
@@ -310,16 +370,19 @@ def has_events(self, events):
         else:
             return set(events).issubset(set(self.available_events))
 
+    def get_view(self, window):
+        return TraceView(self, window)
+
```
douglas-raillard-arm: Could be interesting to also expose it as:
```
    def __getitem__(self, window):
        if not isinstance(window, slice):
             raise TypeError('window must be a slice')
        if slice.step is not None:
            raise ValueError('step is not handled')
        return TraceView(self, (slice.start, slice.stop))
```
This enables slicing on the `Trace` object:
```
view = Trace(...)[start:stop]
```

note: that also needs to be implemented on `TraceView` (ideally shared in a base class like `TraceHelpers`), since `__getattr__` will not handle lookup of `__getitem__` when it is done by the interpreter (same goes for all the special methods).

lisa/trace.py
```diff
@@ -397,10 +397,15 @@ def get_task_pid(self, task):
         """
         if isinstance(task, str):
             pid_list = self.get_task_by_name(task)
+
+            if not pid_list:
+                raise ValueError("trace does not have any task named \"{}\".format(task)")
```
douglas-raillard-arm: nit: double quotes can be used as-is in single-quoted strings, without the need to escape them (and vise versa)

douglas-raillard-arm:
That PR seems in a pretty good shape, glad to finally have a straightforward way of cropping the trace to rtapp execution time on top of the robustness improvement of tests :)

lisa/tests/base.py
valschneider: Woops!

lisa/tests/base.py
valschneider: Yeah sadly it is. They do have default values, but it's not the "right ones". If you look at the doc screenshot I shared earlier, that was before this `replace()`. The doc shows that the default value is `None` for both of them, even though the default value specified in the decorator for `test_throughput` is `1`.

To make sense, the default values have to be the ones given to the decorator (they are, after all, the default values for this function). Hence the comment "Propagate the default values *given* to the decorator".

lisa/tests/base.py
valschneider: From what I've experienced Sphinx/rst go out of whack where there are mixed indentation levels. The text is still readable, but things like formating for parameters gets broken :(

lisa/trace.py
valschneider: Hooo yes that would be pretty damn great!

lisa/tests/base.py
douglas-raillard-arm: Right, in that case could be build a dict with the right variables in them ? The main issue `locals()` brings it that somebody editing the code and looking for usage of a variable before changing it will not find it, and the chances of this being somehow silent are quite high.

lisa/tests/base.py
douglas-raillard-arm: I experienced that as well, that's why I came up with always dedenting everything, so there is no mixed indentation in the output. The solution above dedents every string it uses before stitching them together for that reason.

lisa/trace.py
douglas-raillard-arm: Now that I think about it, the non-slice case could be used as an alias to `df_events()`

lisa/tests/base.py
qais-yousef: >As Valentin pointed, the data are there in any case if somebody needs to sort out the wheat from chaff on a particularly bad set of TestBundle.

We already do this :-)

Anyway my point was that we need to be aware that noisiest doesn't necessarily correlate to problematic. As long as we agree on that and ready to react to this as experience teaches us more I have no issue with the current implementation. It's all good stuff :-)

FWIW when I had this noise detection code in the past I did encounter cases where aggregated sum of noisy tasks was causing the problem. But I think with buildroot where things are quiet this is hard to happen and generally I expect a very small number of noisy tasks. If it's large then this is a separate issue that needs investigating.

lisa/trace.py
valschneider: I'm sold on the slice (and I think I'll remove `get_window` from `Trace`), but I'm not sure if we want to depart from `df_events()`. @derkling ?

lisa/tests/base.py
douglas-raillard-arm: > FWIW when I had this noise detection code in the past I did encounter cases where aggregated sum of noisy tasks was causing the problem

Right, if we start seeing that issue as well on some setups, it should be straightforward to also check the sum. At this point we can add optional parameters to the decorator to select the flavor of filtering that is more appropriate for a given test. The current decorator implementation is not trivial, but it brings a high level of flexibility in specifying defaults and letting the caller of the test override these defaults :-)

lisa/tests/base.py
valschneider: Right, that's the other way around but it is simpler, I'll go with that.

lisa/trace.py
douglas-raillard-arm: I would keep `df_events` as well, it's would be just some kind of shorthand (exposing things a non-magic that are always looked up as expected is a good idea from my experience).

lisa/tests/base.py
valschneider: At that point I might as just get rid of the loop and explicitely create the list. That way we don't get any surprises for the parameter ordering too.

lisa/trace.py
valschneider: I'll go for just the slice for now, and we can always add the `df_events` alias later on.

valschneider:
@douglas-raillard-arm I think I addressed all of your comments. The only thing is for the `TraceHelpers` - since the slicing relies on `get_view()`, I had to add it in there, so I ended up naming the class `SubscriptableTrace`. We'll have to add `df_events` to it if we go down that route.

PR #790, opened 2019-03-01
[RFC] lisa.wlgen.rta: Honnor logdir in JSON str RTapp conf

### LINKED ISSUE #659: wlgen/rta.py: rt-app log files are being pulled from the wrong directory
specifying a JSON config with `'logdir' : '/some/path'`, i would expect the log files to be retrieved from `/some/path`. instead, the `run_dir` is used:

RTA::__postrun():
``` 
logfile = self.target.path.join(self.run_dir, '*{}*.log'.format(task))
self.target.pull(logfile, destdir)
```

this seems to suggest that the `logdir` parameter effectively only works if it is set to the `run_dir`.

douglas-raillard-arm:
Fetch the RTapp logs from the given logdir if specified, otherwise use
run_dir as it used to do.

Fixes #659

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -70,13 +70,15 @@ class RTA(Workload):
 
         self.command = '{0:s} {1:s} 2>&1'.format(rta_cmd, self.remote_json)
 
-    def _late_init(self, calibration=None, tasks_names=None):
+    def _late_init(self, calibration=None, tasks_names=None, log_dir=None):
         """
         Complete initialization with a ready json file
 
         :parameters: Attributes that have been pre-computed and ended up
           in the json file. Passing them can prevent a needless file read.
         """
+        self.log_dir = log_dir if log_dir else self.run_dir
+
         if not os.path.exists(self.local_json):
             raise RuntimeError("Could not find {}".format(self.local_json))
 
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
@v0lker Would that solve the issue you encountered here ?
https://github.com/ARM-software/lisa/issues/659

lisa/wlgen/rta.py
```diff
@@ -107,7 +109,7 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False):
         for task in self.tasks:
             # RT-app appends some number to the logs, so we can't predict the
             # exact filename
-            logfile = self.target.path.join(self.run_dir, '*{}*.log'.format(task))
```
valschneider: I'm not sure there's any value in splitting the `run_dir` and the `log_dir`. I can understand wanting to honor what's given in the rtapp conf, but even at the time of the above bug I'd say you were supposed to use https://github.com/ARM-software/lisa/blob/master/libs/wlgen/wlgen/rta.py#L237 rather than specify your own. 

My gripe with this mostly is about adding an extra parameter that will have abysmal uses (I know I'll never use it).

douglas-raillard-arm: If we expect to only do that with __LOGDIR__ placeholder, let's close that PR and #659 .

PR #792, opened 2019-03-04
install_base_ubuntu: Add graphviz

douglas-raillard-arm:
Exekall can make use of the "dot" command line tool to render a graph of
the testcases. Install it even though it's optional.


valschneider:
Is that something exekall can do after-the-fact? IOW, can it reload a given DB and then spit out the graph? If that is the case, maybe we should keep it as an optional dependency.

PR #795, opened 2019-03-04
Fix lisa/test PELTTask test cases

msrasmussen:
Cleaned up PELTTask test cases. Behaviour test case need larger margins due to inaccurate PELT simulation, and range test is made much more accurate by dropping the broken mean comparison. Further details in the commit messages.

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -632,6 +632,11 @@ class PELTTask(LoadTrackingBase):
 
     task_prefix = 'pelt_behv'
 
+    TASK_PERIOD_MS = 16 * (1024/1000)
+    """
+    The task period must be n*1.024 due to :class:`bart.pelt.Simulator` restrictions
+    """
+
     @classmethod
     def from_target(cls, target:Target, res_dir:ArtifactPath=None, ftrace_coll:FtraceCollector=None) -> 'PELTTask':
         """
@@ -685,10 +690,11 @@ class PELTTask(LoadTrackingBase):
         )
 
         phase = self.rtapp_profile[self.task_name].phases[0]
-        pelt_task = pelt.PeriodicTask(period_samples=phase.period_ms,
-                                      duty_cycle_pct=phase.duty_cycle_pct)
         peltsim = pelt.Simulator(init_value=init_value,
                                  half_life_ms=HALF_LIFE_MS)
+        period_samples = int(phase.period_ms*1e3/peltsim._sample_us)
+        pelt_task = pelt.PeriodicTask(period_samples=period_samples,
+                                      duty_cycle_pct=phase.duty_cycle_pct)
         df = peltsim.getSignal(pelt_task, 0, phase.duration_s)
 
         return peltsim, pelt_task, df
@@ -743,6 +749,23 @@ class PELTTask(LoadTrackingBase):
         phase = task.phases[0]
         cpu = phase.cpus[0]
 
+        # Note: This test-case is only valid if executed at capacity == 1024.
+        # The below assertion is insufficient as it only checks the CPU can potentially
+        # reach a capacity of 1024.
+        assert self.plat_info["cpu-capacities"][cpu] == UTIL_SCALE
+
+        # Due to simplifications in the PELT simulator :class:`bart.pelt.Simulator` and
+        # general misalignment between the simulated PELT signal and the traced one the
+        # error margin has to be fairly generous. The error sources include:
+        #
+        # 1. Misaligned PELT period boundaries (error <= 22)
+        # 2. Inaccurate duty cycle of rt_app task which can't be compensated for in the
+        #    PELT simulator as it only speaks integer PELT periods.
+        # 3. Nearest sample can be up to 512us off (error <= 11).
+        #
+        # The combined error depends on the properties of the task being traced/modelled.
+        # For a 50% 16ms period task that amounts to 5-7% error on TC2.
+
         peltsim, _, sim_df = self.get_simulated_pelt(cpu, signal_name)
         signal_df = self.get_task_sched_signals(cpu)
 
@@ -761,19 +784,13 @@ class PELTTask(LoadTrackingBase):
         plt.close()
 
         # Compare actual PELT signal with the simulated one
-        period_s = phase.period_ms / 1e3
-        sim_period_ms = phase.period_ms * (peltsim._sample_us / 1e6)
         errors = 0
 
         for entry in signal_df.iterrows():
             trace_val = entry[1][signal_name]
             timestamp = entry[0]
-            # Next two instructions map the trace timestamp to a simulated
-            # signal timestamp. This is due to the fact that the 1 ms is
-            # actually 1024 us in the simulated signal.
-            n_periods = timestamp / period_s
-            nearest_timestamp = n_periods * sim_period_ms
-            sim_val_loc = sim_df.index.get_loc(nearest_timestamp,
+
+            sim_val_loc = sim_df.index.get_loc(timestamp,
                                                method='nearest')
             sim_val = sim_df.pelt_value.iloc[sim_val_loc]
 
@@ -806,10 +823,11 @@ class PELTTask(LoadTrackingBase):
         """
         return self._test_range('load', allowed_error_pct)
 
-    def test_util_avg_behaviour(self, error_margin_pct=5, allowed_error_pct=5)\
+    def test_util_avg_behaviour(self, error_margin_pct=7, allowed_error_pct=5)\
         -> ResultBundle:
         """
-        Validate every utilization signal event
+        Validate every utilization signal event by comparing traced events with
+        simulated ones based on :class:`bart.pelt.Simulator`.
 
         :param error_margin_pct: How much the actual signal can stray from the
           simulated signal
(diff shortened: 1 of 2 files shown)


lisa/tests/scheduler/load_tracking.py
```diff
@@ -738,11 +739,24 @@ def _plot_behaviour(self, ax, df, title, task_duty_cycle_pct):
         ax.legend()
 
     def _test_behaviour(self, signal_name, error_margin_pct, allowed_error_pct):
+	# Note: This test-case is only valid if executed at capacity == 1024
```
qais-yousef: Can we assert for this condition here? It seems we have a list of cpu capacities in self.plat_info['cpu-capacities'], and we read the cpu values below - so we can check that the capacity of that cpu == 1024?

lisa/tests/scheduler/load_tracking.py
```diff
@@ -738,11 +739,24 @@ def _plot_behaviour(self, ax, df, title, task_duty_cycle_pct):
...
+        # the simulated PELT signal and the traced one the error margin has to be fairly
+	# generous. The error sources include:
+	#
+	# 1. Misaligned PELT period boundaries (error <= 22)
+	# 2. Inaccurate duty cycle of rt_app task which can't be compensated for in the
+        #    PELT simulator as it only speaks integer PELT periods.
+        # 3. Nearest sample can be up to 512us off (error <= 11).
+        #
+        # For a 50% 16ms period task that amounts to 5-7% error on TC2.
+
```
qais-yousef: Is 16ms the only period we care about? If the period was changed to 20 or 32 ms, will the error margin be different then?

msrasmussen: The error is still there for other periods. The various error sources scale differently with period so it is non-trivial to work out what the true margin should be. Most likely as complicated as writing a more accurate PELT simulator.

lisa/tests/scheduler/load_tracking.py
```diff
@@ -650,7 +650,7 @@ def get_rtapp_profile(cls, plat_info):
         rtapp_profile["{}_cpu{}".format(cls.task_prefix, cpu)] = Periodic(
             duty_cycle_pct=50,
             duration_s=2,
-            period_ms=cls.TASK_PERIOD_MS,
+            period_ms=16.384, # Must be n*1.024 due to PELTsim restrictions.
```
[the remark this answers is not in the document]
msrasmussen: 16ms is just an arbitrary value set in the base class isn't it? I don't see why we should attempt to fix it. We need a specific period for this particular test case as we compare with the PELT simulator. Other test cases that don't use that can still use the default value.

lisa/tests/scheduler/load_tracking.py
msrasmussen: IIUC, plat_info just tells us if which capacities that are available. The PELT test-cases require that the actual capacity is 1024 throughout the trace. To create an assertion for that we need to trace the capacity. We can do that as separate value-add if needed. I just put the note there to remind people of a reason why this test-case could break. I should have added it to _test_range() too as it applied to that as well.

lisa/tests/scheduler/load_tracking.py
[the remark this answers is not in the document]
valschneider: So if it makes sense to freeze that test to a 16.384 ms period I'm fine with that. However that should be reflected in `PELTTask.TASK_PERIOD_MS`, as such I'd rather see it changed like so:
```diff
git diff lisa/tests/scheduler/load_tracking.py
diff --git a/lisa/tests/scheduler/load_tracking.py b/lisa/tests/scheduler/load_tracking.py
index d9d2e234..a5cf53ce 100644
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -630,6 +630,11 @@ class PELTTask(LoadTrackingBase):
     the signal should be very similar to simulated one.
     """
 
+    TASK_PERIOD_MS = 16.384
+    """
+    The task period must be n*1.024 due to :class:`bart.pelt.Simulator` restrictions
+    """
+
     task_prefix = 'pelt_behv'
 
     @classmethod
```

And we'd still use `cls.TASK_PERIOD_MS` in `get_rtapp_profile()`

lisa/tests/scheduler/load_tracking.py
valschneider: That should be fairly easy to do, the question is whether we want to raise an exception or return a failed test (@douglas-raillard-arm ?).

Since we get the CPU associated with the task a few lines below, we could do this:
```python
cpu = phase.cpus[0]

if self.plat_info["cpu-capacities"][cpu] != UTIL_SCALE:
    # raise RuntimeError or return a ResultBundle.from_bool(False) or something
```

lisa/tests/scheduler/load_tracking.py
qais-yousef: OK. What I was trying to get to if we changed the test to use a different period other than 16ms, will we need to change the error margin too then?

lisa/tests/scheduler/load_tracking.py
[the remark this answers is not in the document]
qais-yousef: >6ms is just an arbitrary value set in the base class isn't it? I don't see why we should attempt to fix it. We need a specific period for this particular test case as we compare with the PELT simulator. Other test cases that don't use that can still use the default value.

OK. If we use cls.TASK_PERIOD_MS then as Valentine suggested that'd be great.

lisa/tests/scheduler/load_tracking.py
msrasmussen: Yes, most likely.

lisa/tests/scheduler/load_tracking.py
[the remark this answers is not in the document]
msrasmussen: I'm fine with redefining the constant.

lisa/tests/scheduler/load_tracking.py
msrasmussen: But as said above, that only checks the max capacity of the CPU. We need to verify that the CPU did indeed run at that capacity throughout the test.

valschneider: Right, I get it now. Since we'd need extra trace events, that's fine as is.

lisa/tests/scheduler/load_tracking.py
qais-yousef: OK I was hoping we can explicitly mention this in the comment but I see you already pushed an updated version and not sure if it's worth a re-spin just for that. Up to you.

lisa/tests/scheduler/load_tracking.py
msrasmussen: It is fine to add the check to catch at least some cases.

lisa/tests/scheduler/load_tracking.py
msrasmussen: I can make it explicit that the error margin is workload dependent if that helps. I have already specified the errors that have an upper bound.

lisa/tests/scheduler/load_tracking.py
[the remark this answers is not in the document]
msrasmussen: Done

lisa/tests/scheduler/load_tracking.py
msrasmussen: Done

lisa/tests/scheduler/load_tracking.py
msrasmussen: Done for both range tests and behaviour tests.

lisa/tests/scheduler/load_tracking.py
```diff
@@ -743,6 +749,23 @@ def _test_behaviour(self, signal_name, error_margin_pct, allowed_error_pct):
         phase = task.phases[0]
         cpu = phase.cpus[0]
 
+	# Note: This test-case is only valid if executed at capacity == 1024.
```
qais-yousef: Broken indentation here?

lisa/tests/scheduler/load_tracking.py
```diff
@@ -743,6 +749,23 @@ def _test_behaviour(self, signal_name, error_margin_pct, allowed_error_pct):
...
+        # The below assertion is insufficient as it only checks the CPU can potentially
+        # reach a capacity of 1024.
+        assert self.plat_info["cpu-capacities"][cpu] == UTIL_SCALE
+
+        # Due to simplifications in the PELT simulator :class:`bart.pelt.Simulator` and
+        # general misalignment between the simulated PELT signal and the traced one the
+        # error margin has to be fairly generous. The error sources include:
+	#
+	# 1. Misaligned PELT period boundaries (error <= 22)
+	# 2. Inaccurate duty cycle of rt_app task which can't be compensated for in the
```
qais-yousef: And the three lines above?

lisa/tests/scheduler/load_tracking.py
```diff
@@ -704,27 +705,25 @@ def _test_range(self, signal_name, allowed_error_pct):
         task = self.rtapp_profile[self.task_name]
         cpu = task.phases[0].cpus[0]
 
+	# Note: This test-case is only valid if executed at capacity == 1024.
```
qais-yousef: Broken indentation?

lisa/tests/scheduler/load_tracking.py
msrasmussen: Yes, it looks like I can't do indentation in python right. It looks fine in vim but it is caused by using tabs instead of whitespace.

msrasmussen: Fixed.

PR #798, opened 2019-03-05
analysis: Fix plat_info uses

valschneider:


lisa/analysis/frequency.py
```diff
@@ -308,7 +308,11 @@ def plot_cpu_frequencies(self, cpu, filepath=None, axis=None):
         if local_fig:
             fig, axis = self.setup_plot()
 
-        frequencies = self.trace.plat_info['freqs'][cpu]
+        try:
+            frequencies = self.trace.plat_info['freqs'][cpu]
```
douglas-raillard-arm: Could we split indexing on `['freqs']` (inside the `try`) and `[cpu]` (outside or in `else` clause) ?
This way we would still fail if a platform info is provided with incomplete frequencies, which would probably be the sign of a user error (if somebody wants auto-detection, then one can omit the `freqs` keys altogether).

valschneider: Would this work for you then?

```python
        if "freqs" in self.trace.plat_info:
            frequencies = self.trace.plat_info['freqs'][cpu]
        else:
            self.get_logger().info("Estimating CPU frequencies from trace")
            frequencies = df.frequency.unique()
```

douglas-raillard-arm: yes, that's equivalent (less efficient but nobody cares in that case)

valschneider:
@douglas-raillard-arm Done

lisa/trace.py
```diff
@@ -143,6 +143,7 @@ def cpus_count(self):
         # If we don't know the number of CPUs, check the trace for the
         # highest-numbered CPU that traced an event.
         except KeyError:
+            self.get_logger().info("Estimating CPU count from trace")
```
douglas-raillard-arm: Can we log that after the estimation and show the estimated number as well ? That can speed up debugging if the log already contains the info straight away

lisa/analysis/frequency.py
```diff
@@ -308,7 +308,11 @@ def plot_cpu_frequencies(self, cpu, filepath=None, axis=None):
         if local_fig:
             fig, axis = self.setup_plot()
 
-        frequencies = self.trace.plat_info['freqs'][cpu]
+        if "freqs" in self.trace.plat_info:
+            frequencies = self.trace.plat_info['freqs'][cpu]
+        else:
+            self.get_logger().info("Estimating CPU frequencies from trace")
```
douglas-raillard-arm: same as for the other logging

valschneider:
@douglas-raillard-arm Done

PR #802, opened 2019-03-06
Improve conf key access logging

douglas-raillard-arm:
* Avoid showing internal helpers of lisa.conf
* Log calls to `add_src` the same way key usage is logged.

### DIFF
--- a/lisa/conf.py
+++ b/lisa/conf.py
@@ -24,7 +24,9 @@ import inspect
 import itertools
 import textwrap
 
-from lisa.utils import Serializable, Loggable, get_nested_key, set_nested_key
+from lisa.utils import (
+    Serializable, Loggable, get_nested_key, set_nested_key, get_call_site
+)
 
 class DeferredValue:
     """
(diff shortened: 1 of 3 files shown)


lisa/conf.py
```diff
@@ -877,7 +895,11 @@ def __getstate__(self):
 
         return state
 
-    def get_key(self, key, src=None, eval_deferred=True):
+    def _quiet_get_key(self, *args, **kwargs):
+        kwargs['quiet'] = True
+        return self.get_key(*args, **kwargs)
+
+    def get_key(self, key, src=None, eval_deferred=True, quiet=False):
```
valschneider: Isn't that going to be too verbose for standard LISA usage?
IMHO outputting this into `logger.debug()` would make more sense - at the very least to be togglable by the end user. I don't think we are able to change `quiet` on `plat_info` uses since AFAICT we go through `__getitem__()`.

douglas-raillard-arm: The current behavior is not being quiet, and outputting a lot more "irrelevant" accesses :-)
Quiet only gates creating the debug entry, it will not turn debug into info.

valschneider: > The current behavior is not being quiet, and outputting a lot more "irrelevant" accesses :-)
> Quiet only gates creating the debug entry, it will not turn debug into info.

Duh I completely missed the `logger.debug()` calls, for some reason I thought this was under `info()`. My bad.

PR #803, opened 2019-03-06
tools: Introduce extract-platinfo

douglas-raillard-arm:
Introduce a dummy CLI tool to get a plat_info.yml file suitable for
loading using either:
* PlatformInfo.from_yaml_map('plat_info.yml')"
* exekall run --conf plat_info.yml"

The CLI is a bit crippled as it does not allow using a different file
name than plat_info.yml, and it does not give the choice to evaluate
deffered values or not (i.e. RTapp calibration will be run, which can be
lengthy and useless depending on the use case).

We may want to introduce a `Target.add_argparse_args(parser)` function that would add the same arguments as `from_cli()`, but on an existing parser. That would allow simple scripts to be built on top of that, giving a chance to register a couple of parameters like the ones missing here.

Also, we may want to change the default log-level, to avoid having to specify `--log-level info` all the time.

### DIFF
--- a/tools/extract-platinfo.py
+++ b/tools/extract-platinfo.py
@@ -1,0 +1,40 @@
+#! /usr/bin/env python3
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2019, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


derkling:
Simple yet great and useful! Tested and works fine for me!

valschneider:
Nit: other python tools in there end with `.py`
Also nit: we have some other tools in `tools/scripts` (travis and that sd flag thingie), maybe we could move them out with this.

The tool is fine, I'm just abusing this PR to point at the annoying things in our `tools` directory :)

douglas-raillard-arm:
> Nit: other python tools in there end with .py

Now fixed, it makes sense to keep such an extension for a basic tool that also has a value as an API usage example.

PR #806, opened 2019-03-07
lisa.analysis: Add hints for autocompletion

douglas-raillard-arm:
Since jedi autocompletion is mostly oriented toward static code
analysis, it does not pick up dynamic attributes. Add a function to give
hints to jedi, so it knows what attributes are available and their type
as well.

That's an ugly PR, but it should be relatively robust against more breakages in jedi, and we care specifically about good autocompletion on that part of LISA, so let's bite the bullet. Should not be a big maintenance hassle anyway, and can go away at anytime if jedi becomes more stable.

### DIFF
--- a/lisa/analysis/proxy.py
+++ b/lisa/analysis/proxy.py
@@ -94,4 +94,24 @@ class AnalysisProxy(Loggable):
                 self._instance_map[attr] = analysis_instance
                 return analysis_instance
 
+    # Hints for Jedi autocomplete library, which uses static code analysis
+    # to figure out the available attributes
+    def _hints_for_jedi(self):
+        raise RuntimeError('do not run that method !')
+        # That is enough to make jedi understand what modules we refer to
+        import lisa.analysis
+
+        # List of analysis classes.
+        # Note: This is dummy code that won't work, it's just hints for jedi
+        self.idle = lisa.analysis.idle.IdleAnalysis()
+        self.latency = lisa.analysis.latency.LatencyAnalysis()
+        self.frequency = lisa.analysis.frequency.FrequencyAnalysis()
+        self.load_tracking = lisa.analysis.load_tracking.LoadTrackingAnalysis()
+        self.tasks = lisa.analysis.tasks.TasksAnalysis()
+        self.functions = lisa.analysis.functions.FunctionsAnalysis()
+        self.status = lisa.analysis.status.StatusAnalysis()
+        self.thermal = lisa.analysis.thermal.ThermalAnalysis()
+        self.cpus = lisa.analysis.cpus.CpusAnalysis()
+        self.rta = lisa.analysis.rta.PerfAnalysis()
+
 # vim :set tabstop=4 shiftwidth=4 expandtab textwidth=80


valschneider:
Seeing as we have dynamic attributes all over the place (worst offender probably is our `Target` class), can we not just disable jedi? AFAIR, like @derkling said, we didn't use to have these issues.

douglas-raillard-arm:
> can we not just disable jedi ?

That can be done, but that will probably not buy us a long life of working autocompletion. Builtin autocompletion is being deprecated, although IPython maintainers seem to still keep it alive until other `jedi` issues have been fixed. That builtin autocompletion apparently had a number of issues on its own, and it seems that maintainers would be happy to delegate that to another lib. Unfortunately, `jedi` is mostly oriented on static analysis to avoid running code partly for security reasons.
https://ipython.readthedocs.io/en/stable/whatsnew/version6.html

> worst offender probably is our Target class

That's right, forgot about this one. It could be solved by just inheriting from devlib's `Target` generic class, which should not create issues (unless they rely on special methods called through builtins, since they are not resolved using `__getattr__`, but that is quite unlikely as it does not define any)

derkling:
Doesn't this patch kind of invalidate the dynamic loading?
Every time we add a new "named" analysis method we have to remember to add it to this list too, isn't it?

douglas-raillard-arm:
>  Every time we add a new "named" analysis method we have to remember to add it to this list too, isn't it?

If we want completion to work with `jedi` yes. It seems to require that (at least in its current state), I've tried many combination of `__dir__` and `setattr()` that gave no results. That makes sense since it is based on parsing the code, not dealing with built objects, even in its Interpreter mode which seems to only replace type inference by the actual type apparently (haven't checked the code, only based on my experiments).

valschneider:
> Doesn't this patch kind of invalidate the dynamic loading?
> Every time we add a new "named" analysis method we have to remember to add it to this list too, isn't it?

That's a shame, really. I'd be tempted to say I'd rather have no autocompletion than to have to maintain this dynamic members list...

douglas-raillard-arm:
> I'd rather have no autocompletion than to have to maintain this dynamic members list

That's your call, I don't mind either. Since that list is relatively stable and short, it's not a big deal to maintain. IMHO it's better to end up with such list than fighting IPython's default, which will prevent from using what good things `jedi` can bring and is bound to being deprecated one day. We can track how that issue evolves, as it is likely to be the same that we have:
https://github.com/ipython/ipython/issues/11609

derkling:
Cannot we make that list dynamically created whenever the module is first used?
That's what we used to do in lisa master before re-organizing all the analysis code...

derkling:
IOW, cannot we build those bindings as part of the loop we have in:
   https://github.com/ARM-software/lisa/blob/next/lisa/analysis/__init__.py#L25

so fare it seems we have a pretty straightforward translation from module name to the Analys class it provides, e.g.
```
   self.MODULE = lisa.analysis.MODULE.MODULEAnalysis()
```
there `MODULE` is just all small case and `_` separated apart from the last usage which is camel case.

douglas-raillard-arm:
> Cannot we make that list dynamically created whenever the module is first used?

That is what we currently do as well. `jedi` find the files and parses them, without executing anything. As I said before, tricks based on `setattr()` (lisa legacy) give the same results as implementing `__getattr__()` (I tried it): `jedi` does not care, since it won't execute anything anyway. The link to Ipython issue tracker (https://github.com/ipython/ipython/issues/11609) is most_ likely caused by the same issue: they want `jedi` to look at what a function returns (a property), but `jedi` won't execute it so it does not know what that is (and does not try to be smart in that case apparently).

> That's what we used to do in lisa master before re-organizing all the analysis code...
`ipython` and `jedi` changed in the meantime, I've seen no evidence so far that this reorganization is the source of the issue. On the other hand, I've seen evidences that `jupyter` nowadays do not complete on things set with `setattr()` anymore by default, which is what LISA-legacy used to do. Completion would be as broken using LISA-legacy style as LISA-next using a march 2019 installation of `jupyter`.

> IOW, cannot we build those bindings as part of the loop we have in

The only loop that can buy us anything would be a loop creating new files, or worse, patching existing files. `jedi` is basically blind on anything that involves meta-programming, since it won't execute that loop (or anything else). It's also a very bad idea, as python modules may not even be files at all (executed from a zip file, bundled using pyfreeze etc). The location where the package is installed is also typically not writeable as a regular user  (/usr/lib/python3.7/site-packages/), and it would introduce race conditions when multiple processes importing lisa are started.

PR #810, opened 2019-03-08
tests: eas_behaviour: Fix TwoBigThreeSmall duty cycle

qperret:
The big tasks in TwoBigThreeSmall are 80%. Because of this, EAS is kept
overutilized consistently, hence not giving the small tasks an
opportunity to be placed appropriately on little CPUs.

Although this is arguably something that could be improved on the kernel
side, it should also be noted that these tasks used to be 70% in the old
Lisa implementation, hence avoiding the problem altogether. Since the
change to 80% was not intended, and since the current test failures
don't teach us anything that we didn't know before, let's revert to 70%
tasks and avoid unnecessary pain.

### DIFF
--- a/lisa/tests/scheduler/eas_behaviour.py
+++ b/lisa/tests/scheduler/eas_behaviour.py
@@ -480,9 +480,9 @@ class TwoBigThreeSmall(EASBehaviour):
         # 50% of the smallest CPU's capacity
         small_duty = cls.unscaled_utilization(
             plat_info, plat_info["capacity-classes"][0][0], 50)
-        # 80% of the biggest CPU's capacity
+        # 70% of the biggest CPU's capacity
         big_duty = cls.unscaled_utilization(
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
That would be fine for me. TwoBigTasks seems to also be using 80%, should it be modified the same way to trigger the (EAS) paths we want to actually test in the kernel ?

qperret:
TwoBigTasks is less of a problem -- we rely on non-EAS paths to put big tasks on big CPUs somehow. If TwoBigTasks failed with 80% tasks we'd like to know it I guess.

douglas-raillard-arm:
> we rely on non-EAS paths to put big tasks on big CPUs somehow

Maybe it would be interesting to also have a `TwoBigTasks2` version that also checks the EAS paths by using 70% util instead

qperret:
TwoMedTasks ? 110% of a little CPU or something ?

qperret:
Could make sense for tri-gear too

PR #813, opened 2019-03-11
[RFC] lisa.tests: load_tracking: Invariance: Treat load like util

douglas-raillard-arm:
Do not rescale the signal specifically when it is the load.

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -309,9 +309,18 @@ class InvarianceItem(LoadTrackingBase):
 
         signal_mean = area_under_curve(signal) / (signal.index[-1] - signal.index[0])
 
-        if signal_name == 'load':
-            # Load isn't CPU invariant
-            exp_signal /= (self.plat_info['cpu-capacities'][cpu] / UTIL_SCALE)
+        # Since load is now CPU invariant in recent kernel versions, we don't
+        # rescale it back. To match the old behavior, that line is
+        # needed:
+        #  exp_signal /= (self.plat_info['cpu-capacities'][cpu] / UTIL_SCALE)
+        kernel_version = self.plat_info['kernel']['version']
+        if (
+            signal_name == 'load'
+            and kernel_version.parts[:2] < (5, 1)
+        ):
+            self.get_logger().warning('Load signal is assumed to be CPU invariant, which is true for recent mainline kernels, but may be wrong for {}'.format(
+                kernel_version,
+            ))
 
         ok = self.is_almost_equal(exp_signal, signal_mean, allowed_error_pct)
 


lisa/tests/scheduler/load_tracking.py
```diff
@@ -309,9 +309,8 @@ def _test_task_signal(self, signal_name, allowed_error_pct,
 
         signal_mean = area_under_curve(signal) / (signal.index[-1] - signal.index[0])
 
-        if signal_name == 'load':
-            # Load isn't CPU invariant
-            exp_signal /= (self.plat_info['cpu-capacities'][cpu] / UTIL_SCALE)
+        # Since load is now CPU invariant, we don't rescale it back
```
valschneider: Don't we want to guard this against a kernel version check, or did I misunderstand the last discussion we had about this?

At the very least we could have a `logger.warning()` that warns about kernel version < 5.0

douglas-raillard-arm: @ionela-voinescu was more keen on just supporting the current mainline kernel. I don't mind either.

ionela-voinescu: I don't believe checking for the kernel version is always safe. It's possible people might backport these patches to older ( < 5.0 ) android kernels so at least controlling test functionality (cpu scaled or not scaled load) based on kernel version seems wrong to me. But I doubt newer kernels would be missing this functionality so it seems right to me to warn people that use older kernels that our estimation of load can be wrong.

douglas-raillard-arm: Ok, let's put a warning if version < 5.0 and unconditionally use the new computation

qperret: This was introduced in 5.1 I believe

douglas-raillard-arm:
It will now log it at the warning level for kernels < 5.1

lisa/tests/scheduler/load_tracking.py
```diff
@@ -309,9 +309,18 @@ def _test_task_signal(self, signal_name, allowed_error_pct,
...
-            # Load isn't CPU invariant
-            exp_signal /= (self.plat_info['cpu-capacities'][cpu] / UTIL_SCALE)
+        # Since load is now CPU invariant in recent kernel versions, we don't
+        # rescale it back. To match the old behavior, that line is
+        # needed:
+        #  exp_signal /= (self.plat_info['cpu-capacities'][cpu] / UTIL_SCALE)
+        kernel_version = self.plat_info['kernel']['version']
+        if (
+            signal_name == 'load'
+            and kernel_version.parts < (5, 1, 0)
```
qperret: Just wondering, how does that work with rc versions ? Is 5.1-rc1 < 5.1.0 ?

douglas-raillard-arm: Good point, "rc" is not in `.parts` tuple, but will lead to `minor=None`, so the comparison will raise an exception if we get to that. Instead, I can replace it with `kernel_version.parts[:2] < (5, 1)` to just ignore the rc and minor numbers (comparing tuples with different number of parts works but seems to give surprising results wrt equality).

douglas-raillard-arm:
Updated to ignore minor version number and rc number (i.e. in 5.1.2, only 5.1 will be taken into account)

PR #815, opened 2019-03-11
lisa.energy_meter: Cleanup EnergyMeter configuration

douglas-raillard-arm:
Use one configuration class per type of EnergyMeter, so that all keys
are properly described, and each of them will get an independent
configuration format to be used with `EnergyMeter.from_conf()`

Disclaimer:  that is untested code
Disclaimer 2: the change has only been done for HWMon as an example, the same would need to be done for the other classes as well.

Using it would give:
```
target = ...
res_dir = ...
conf = HWMonConf.from_yaml_map('path/to/hwmon_conf.yml')
emeter = HWMon.from_conf(target, conf, res_dir)
```
With `hwmon_conf.yml`:
```
hwmon-conf:
     channel-map: ...
```
OR 
```
target = ...
res_dir = ...
emeter = HWMon(target, channel_map=..., res_dir=res_dir)
```

`HWMonConf` documentation would describe the available keys just like `FtraceCollectorConf` does.

### DIFF
--- a/doc/transition_guide.rst
+++ b/doc/transition_guide.rst
@@ -252,3 +252,32 @@ framework has been put in place. Tests are now coded as pure Python classes,
 which means they can be imported and executed in scripts/notebooks without any
 additionnal effort. See :ref:`kernel-testing-page` for more details about
 using/writing tests.
+
+
+Energy Meter
+++++++++++++
+
+Energy meters are all subclasses of :class:`lisa.energy_meter.EnergyMeter`.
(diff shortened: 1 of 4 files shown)


douglas-raillard-arm:
It could also be interesting to add a method like `from_conf()`, but taking a conf file path directly to avoid having to manually create the conf object:
```
from_yaml_map(target, path, res_dir=None)
```

valschneider:
@qperret If you're okay with this I think we can go ahead and merge it.

PR #816, opened 2019-03-12
lisa/test: Adjust PELTTask.test_{util,load}_avg_range margin

msrasmussen:
The margin was recently reduced from 15% to 1% however it turns out that
this is a little bit too tight as minor shifts in time and interference
from kthreads regularly causes the 1% margin to be breached. 1.5% should
be sufficient to accommodate for those inaccuracies.

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -807,7 +807,7 @@ class PELTTask(LoadTrackingBase):
 
         return res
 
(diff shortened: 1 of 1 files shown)


msrasmussen:
> Eventually we might reduce this once #783 gets in.

For some cases it is indeed interference which we could detect, however I don't see interference in all the failed cases. It could be down to just timing of PELT updates in some cases.

PR #819, opened 2019-03-12
Documentation cleanup & reorg

valschneider:
This PR brings in the last bits we needed from the wiki (EM doc), and makes the doc structure more sane.

You can see what it looks like here: https://my-lisa.readthedocs.io/en/lisa-next/


doc/conf.py
```diff
@@ -416,20 +416,49 @@ def get_test_id_list():
...
+
+    rst_list = []
+
+    for subclass in get_subclasses(TraceAnalysisBase):
+        class_path = "{}.{}".format(subclass.__module__, subclass.__qualname__)
+        analysis = subclass.__module__.split(".")[-1]
+        meth_list = [name for name, member in inspect.getmembers(subclass, callable)
+                     if name.startswith(prefix)]
+
+        rst_list += [":class:`{0}<{1}>`::meth:`~{1}.{2}`".format(analysis, class_path, meth)
```
douglas-raillard-arm: can we use named placeholders instead of positionals ? That makes the template more readable:
```
         rst_list += [":class:`{analysis}<{cls_path}>`::meth:`~{cls_path}.{meth}`".format(analysis=analysis, cls_path=class_path, meth=meth) 
```
```

File: doc/lisa_shell.rst [hunk 2]
```diff
@@ -0,0 +1,18 @@
...
+LISA shell
+**********
+
+.. TODO:: Have the shell manpage here
+
+Once you have all of the required dependencies installed, you can use the LISA
+shell, which provides a convenient set of commands for easy access to many LISA
+related functions::
+
+  source init_env
```
douglas-raillard-arm: nit: shouldn't that be in a `.. code-block:: sh` ?

doc/index.rst
```diff
@@ -30,16 +30,33 @@ Contents:
...
-   automated_testing
-   wlgen
-   internals
-   analysis
-   lisa_tests
+   trace_analysis
+   energy_analysis
+   misc_utilities
+
+..
```
valschneider: @douglas-raillard-arm Do you want me to add placeholders for `bisector` & `exekall`?

doc/trace_analysis.rst
```diff
@@ -0,0 +1,143 @@
...
+Trace
+=====
+
+Our :class:`~lisa.trace.Trace` takes an Ftrace ``trace.dat`` file as input (Systrace
+``trace.html`` are also accepted), and provides access to both the raw trace events,
+as well as some new dataframes built from analysing and aggregating trace events.
+
+You can create one like so::
+
+  trace = Trace("path/to/trace.dat", events=["sched_switch", "sched_wakeup"])
```
douglas-raillard-arm: For a simple example, we may even omit `events` since it parses all events by default.

doc/index.rst
douglas-raillard-arm: Could be nice, thanks :)

doc/trace_analysis.rst
valschneider: Does it? @derkling had some issues with that lately IIRC

douglas-raillard-arm: That's right, there are some outdated comments, but by looking at trappy's code, I can't figure out why it is not the case.

doc/conf.py
valschneider: Hmm after giving it a go I find it makes the template less readable actually. Can I keep it as is?

douglas-raillard-arm: It's not very straightforward to check that the template is sticking the right things together, but if it is for you then it's ok. For example sticking {analysis}.{meth} would not be correct but it's not so obvious when we only have numeric placeholders.

doc/trace_analysis.rst
douglas-raillard-arm: Looks like it is related to scope == 'custom', which will only select a handful of events for some reasons, but not the one we usually care about (nothing from sched.py).

valschneider:
If everyone's fine with it, I'll fix the merge conflicts and merge it.

PR #821, opened 2019-03-13
doc: Fix ignored warning pattern

douglas-raillard-arm:
Fix ignored warning patterns, and make more obvious why a warning is ignored or not

### DIFF
--- a/doc/pedantic_build.sh
+++ b/doc/pedantic_build.sh
@@ -23,7 +23,7 @@ IGNORE_PATTERN="devlib|bart|wa|exekall.engine"
 
 make SPHINXOPTS='-n' html 2>doc_build.log
 all_warns=$(cat doc_build.log)
-warns=$(cat doc_build.log | grep WARNING | grep -E -v "$IGNORE_PATTERN")
+warns=$(cat doc_build.log | grep WARNING | grep -E -w -v "$IGNORE_PATTERN")
 
 echo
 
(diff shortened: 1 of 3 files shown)


doc/pedantic_build.sh
```diff
@@ -21,25 +21,30 @@
 # be permissive for warnings related to them
 IGNORE_PATTERN="devlib|bart|wa|exekall.engine"
 
-make SPHINXOPTS='-n' html 2>doc_build.log
-all_warns=$(cat doc_build.log)
-warns=$(cat doc_build.log | grep WARNING | grep -E -v "$IGNORE_PATTERN")
+make SPHINXOPTS='-n --no-color' html 2>doc_build.log
```
valschneider: Why use `--no-color` and then `grep --color=always` ?

douglas-raillard-arm: When letting sphinx-build use colors and then cutting some lines, we end up with the escape sequence to enable red, but we cut out the sequence to go back to normal color. Also, if sphinx-build gives a warning line that is all red, we cannot see easily what grep has selected. By doing that, we can see why a warning is selected or not, so it's easy to spot wrong matches like "wa" unexpectedly matching "was".

douglas-raillard-arm:
> On a sidenote, the doc fixes you've done are already in #819 but I can drop them.

With the fix to pedantic_build.sh and without the fix to the doc, Travis is not happy anymore, so they kinda need to go together, otherwise the CI will fail until they are merged

PR #823, opened 2019-03-14
Collect dmesg when running tests

douglas-raillard-arm:
EDIT: ready to merge, devlib's PR has been merged


external/devlib/devlib/trace/dmesg.py
```diff
@@ -0,0 +1,187 @@
...
+            cmd = 'dmesg'
+        else:
+            cmd = 'dmesg --facility={facility} --force-prefix --decode --level={levels}'.format(
+                levels=','.join(levels_list),
+                facility=self.facility,
+            )
+
+        self.dmesg_out = self.target.execute(cmd)
+
+    def get_trace(self, outfile):
```
qais-yousef: can this be called get_log() instead?

external/devlib/devlib/trace/dmesg.py
```diff
@@ -0,0 +1,187 @@
...
+            prefix = ''
+
+        return '{prefix}[{timestamp}] {msg}'.format(
+            timestamp=self.timestamp.total_seconds(),
+            msg=self.msg,
+            prefix=prefix,
+        )
+
+
+class DmesgCollector(TraceCollector):
```
qais-yousef: This inheritance relationship doesn't sound right to me. From naming point of view at least. What facilities in the parent class you care about here?

If it is to auto start/stop then it seems to me TraceCollector and DmesgCollector should have a common parent class. 

e.g
```
                   BufferCollector
                         |
          +------------- +----------------+
          |                               |
      TraceCollector                  DmesgCollector
```

external/devlib/devlib/trace/dmesg.py
douglas-raillard-arm: That was named to match the "abstract" method defined in the base class (`TraceCollector`), but it would indeed make more sense as `get_log()`. Can you post that comment on devlib's PR as well, so we can have the opinions of the maintainers there ? 
https://github.com/ARM-software/devlib/pull/376

external/devlib/devlib/trace/dmesg.py
douglas-raillard-arm: There are already a number of other collectors in devlib like `LogcatCollector`, which already inherit from `TraceCollector`. That hierarchy of names probably needs to be preserved for backward compatibilities, but if you think it's worth a rename, having aliases is possible if devlib maintainers are happy with that.

douglas-raillard-arm: > What facilities in the parent class you care about here?

Mostly the context manager (`__enter__`/`__exit__`) and the "promise on the API" (all collectors can be expected to behave roughly the same way without looking at each specific case). I initially tried to implement it as an instrument, but it was not suited at all. `TraceCollector` seemed to have a number of derived classes doing more or less the same thing as I do.

PR #824, opened 2019-03-15
ftrace events: use @requires_events on top of hand-specified ftrace_conf

douglas-raillard-arm:
Introduce a metaclass for RTATestBundle that makes sure that the `ftrace_conf` of each subclass get at least all the events that are needed by their methods when they are decorated using `@requires_events`.

That avoids duplication of that information and also avoids writing it explicitly in docstrings, since sphinx will add the list of required events using the same information. That also allows setting the required events deep in the analysis functions that actually use it, and then pointing to that through the call chain using `@the_analysis_method.used_events`.

Also, use that for --list-trace-events that has been revamped and removed at the occasion (removed=there is no CLI option anymore since it is always done by default).

In this PR, I've "ported" synthetic tests to use that system, so they don't need any manual `ftrace_conf` class attribute to be setup anymore.

### DIFF
--- a/lisa/analysis/load_tracking.py
+++ b/lisa/analysis/load_tracking.py
@@ -20,6 +20,7 @@
 import pandas as pd
 
 from lisa.analysis.base import TraceAnalysisBase
+from lisa.trace import requires_one_event_of
 
 
 class LoadTrackingAnalysis(TraceAnalysisBase):
@@ -87,6 +88,7 @@ class LoadTrackingAnalysis(TraceAnalysisBase):
 
         raise RuntimeError("Trace is missing one of either events: {}".format(events))
 
+    @requires_one_event_of('sched_load_cfs_rq', 'sched_load_avg_cpu')
     def df_cpus_signals(self):
         """
         Get the load-tracking signals for the CPUs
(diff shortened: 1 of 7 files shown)


lisa/analysis/load_tracking.py
```diff
@@ -87,6 +88,7 @@ def _df_either_event(self, events):
 
         raise RuntimeError("Trace is missing one of either events: {}".format(events))
 
+    @requires_one_event_of('sched_load_cfs_rq', 'sched_load_avg_cpu')
```
valschneider: You should remove the mention of those in the docstring, since that's now automagic.

lisa/analysis/load_tracking.py
```diff
@@ -20,6 +20,7 @@
 import pandas as pd
 
 from lisa.analysis.base import TraceAnalysisBase
+from lisa.trace import requires_one_event_of
 
 
 class LoadTrackingAnalysis(TraceAnalysisBase):
```
valschneider: I can't tag those lines on the review, but the users of `df_cpus_signals` and `df_tasks_signals` should also get the appropriate decorators.

douglas-raillard-arm:
@valschneider Updated with all synthetic tests updates and comments addressed.

The events decorators can now be used on any callable, even if it is not a method of instances that have a `trace` attribute. It should also work on functions without any parameter at all.

lisa/trace.py
```diff
@@ -1115,11 +1126,27 @@ def __call__(self, f):
...
+                    pass
+                else:
+                    available_events = set(trace.available_events)
+                    checker.check_events(available_events)
+
+                return f(self, *args, **kwargs)
+        # If the decorated object takes no parameters, we cannot check anything
+        else:
+            @wraps(f)
+            def wrapper(*args, **kwargs):
```
valschneider: Can't we just `return f` here?

valschneider: Oh no we can still add the `used_events` - don't mind me!

douglas-raillard-arm: we could return `f` and just stick the events on it, but I'd rather keep the same number of decorator levels in all paths, to avoid later surprises.

PR #826, opened 2019-03-18
TraceView fixups

valschneider:


valschneider:
Argh, stray devlib change in here, hang on

lisa/trace.py
```diff
@@ -152,6 +152,9 @@ def df_events(self, event):
 
         return df
 
+    def add_events_deltas(self, df, col_name='delta', inplace=True):
```
douglas-raillard-arm: Would be better to put that in a shared base class of both Trace and TraceView. "regular" inheritance lookup will take place before `__getattr__` is tried, so you should get the same behaviour.

valschneider: Right, then that `SubscriptableTrace` can become a `TraceBase`.

PR #828, opened 2019-03-19
Fixes for wltests on P3

qperret:
So it turns out `repo sync -l` _can_ rebase stuff behind your back ... I'm a bit tired of playing around with this so I took the decision to stop messing around with `repo sync` and to just use proper repo branches (which is what I should have done from the start). That makes it a bit harder to test different toolchains and things like that, but that actually works properly for what we really want to do, that is, testing kernel changes.

I understand this whole thing is becoming more confusing than necessary (since I change my mind every week or so), so this PR also creates some basic wltest doc to explain how to use it for 'repo' targets.

For the future implementation of wltest we should drop the support for !repo targets I think -- Google now ship build scripts for all platforms we have in wltest (https://source.android.com/setup/build/building-kernels#downloading) but I think that's for later.

### DIFF
--- a/doc/workflows/wltest.rst
+++ b/doc/workflows/wltest.rst
@@ -2,6 +2,69 @@
 Wltest
 ******
 
-.. TODO:: write me
+Wltest enables the comparison of power and performance impacts of kernel
+changes on Android devices. Wltest takes a list of kernel commits as input,
+build and flashes each one on a device, and runs a `workload-automation` agenda
+in order to collect power and performance metrics. After the execution, the
+results can be parsed and analyized using the ipython agenda provided in Lisa
+under ``ipynb/wltests/sched-evaluation-full.ipynb``
 
-**(This page is being worked on, come back later!)**
+The ``lisa-wltest-series --help`` command provides the full list of parameters
+for wltest.
+
+Target types
+============
+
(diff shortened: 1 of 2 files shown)


doc/workflows/wltest.rst
```diff
@@ -2,6 +2,69 @@
...
+
+ 2. the `repo` targets, for which the build procedure is entirely managed
+    by a repo containing the kernel sources, modules, toolchains and build
+    scripts. The repo targets have ``WLTEST_REPO_TARGET="y"`` set in their
+    `definitions` file.
+
+Standard targets
+++++++++++++++++
+
+For standars targets, such as Hikey 960, the wltest workflow goes as follows:
```
valschneider:
```suggestion
For standard targets, such as Hikey 960, the wltest workflow goes as follows:
```

doc/workflows/wltest.rst
```diff
@@ -2,6 +2,69 @@
...
+
+ 1. Clone the kernel sources with
+    ``git clone https://android.googlesource.com/kernel/hikey-linaro``
+
+ 2. Hack the kernel and commit you changes
```
valschneider:
```suggestion
 2. Hack the kernel and commit your changes
```

PR #829, opened 2019-03-19
Android tests

qperret:
We don't have tests for the Android-specific features in Lisa yet, which is unfortunate. I have been using a notebook to test the eas-dev branch for some time, but it breaks regularly with API changes and it can't be run in CI. So, this PR tries to do 'the right thing' and creates proper tests for the Android-specific stuff, which should help fix both problems.

This is still WiP since I have ideas for more tests (sugov tests for RT tasks, up/down sugov thresholds, sync'd tasks, ...), but I figured we don't have to review them all in one go. So this PR only adds tests for schedtune for now. (Yes, we'd like to drop schedtune eventually but it is still there ATM and will be around for a few more years for 4.14 and 4.19 devices, so there's value in having a test I think).

There's quite a few `# XXX` comments in there for the things I'm not too sure about. Also, one important question is how to configure cgroups for RTATestBundle. I have a suggestion in there, but this can probably be done differently.

Feedback welcome :smile: !


lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,290 @@
...
+    }, __qualname__)
+
+    @classmethod
+    def check_from_target(cls, target):
+        # XXX: better way to do that ?
+        if not target.file_exists('/sys/kernel/debug/tracing/available_events'):
+            raise CannotCreateError("Ftrace not enabled")
+
+        # XXX: better way to do that ?
+        for m in ['cpufreq', 'cgroups', 'sched']:
```
valschneider: That's another one we don't explicitely check (and we probably should). Maybe we could add a class attribute that lists the required devlib modules and automagically check for them? @douglas-raillard-arm ?

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,290 @@
...
+        if 'SCHEDTUNE_BOOST_HOLD_ALL' not in target.sched.get_features():
+            raise CannotCreateError("Schedtune boost hold for CFS unavailable")
+
+    @classmethod
+    def _from_target(cls, target, res_dir, ftrace_coll=None):
+        # The boost hold feature for CFS should help keep the frequency stable
+        state = target.sched.get_features()['SCHEDTUNE_BOOST_HOLD_ALL']
+
+        # XXX - Context manager for this ? Move it to RTATestBundle ?
+        target.sched.set_feature('SCHEDTUNE_BOOST_HOLD_ALL', True)
```
valschneider: I <3 context managers! You could add this one to the `sched` module in devlib.

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,290 @@
...
+            if c not in target.cgroups.controllers:
+                raise CannotCreateError("'{}' cgroup controller not available".format(c))
+
+    @classmethod
+    def get_rtapp_profile(cls, plat_info):
+        cpu = plat_info['capacity-classes'][-1][0]
+        rtapp_profile = {}
+        rtapp_profile['rta_stune'] = Periodic(
+            duty_cycle_pct = 1,
+            duration_s = 3,
```
valschneider: Unless I missed something, this profile is very similar to the one in the previous test class - the only diff I see is the duration. Would it make sense to have these merged into a single test class with two test methods?

qperret: There's another difference, for the frequency tests the task is pinned to a big CPU, and for the task placement it's not ... :/

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,290 @@
...
+    This class provides :meth:`get_cgroup` to create in the SchedTune
+    controller a `lisa_test` cgroup configured with the `boost` and
+    `prefer_idle` attributes.
+    """
+
+    boost = 0
+    prefer_idle = False
+
+    @classmethod
+    def get_cgroup(cls):
```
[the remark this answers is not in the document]
qperret: Yeah I think we could use target (or at least plat_info) in there, I just forgot it here. You can imagine tests that play with cpusets for example, and in this case you'll want to know what are the capacity classes or so. So I guess I should just align this protoype with `get_rtapp_profile()` ...

lisa/tests/base.py
```diff
@@ -572,24 +573,71 @@ def get_rtapp_profile(cls, plat_info):
         pass
 
     @classmethod
-    def _run_rtapp(cls, target, res_dir, profile, ftrace_coll=None):
+    def get_cgroup(cls, target):
```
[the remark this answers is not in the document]
qperret: Sounds good to me

lisa/tests/base.py
```diff
@@ -572,24 +573,71 @@ def get_rtapp_profile(cls, plat_info):
...
+        """
+        :returns: a :class:`dict` representing the configuration of a cgroup
+          in a specific controller.
+
+        The returned dict must follow this example:
+          {
+            'controller': 'schedtune',
+            'name': 'lisa_test',
+            'values': {
```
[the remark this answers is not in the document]
qperret: You mean the `values` stuff ? The `boost` and `prefer_idle` attributes exist only for the `schedtune` controller AFAICT. At least so long as we use cgroups v1. Not sure how cgroups v2 would deal with that, but `schedtune` doesn't support it, I think ...

lisa/tests/staging/sched_android.py
valschneider: So I did miss something :)

lisa/tests/base.py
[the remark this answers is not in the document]
valschneider: Yeah I was thinking of something like this:
```
CgroupConf:
    def __init__(self, controller, name, values):
        [...]
```

(maybe defined in devlib's cgroup.py ?)

I strive to limit dictionaries-as-API as much as possible, but the class might be a bit overkill here...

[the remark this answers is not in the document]
qperret: Then I should make sure it serializes-deserializes properly ...

How about I try and restrict things with the API directly like so:
``` python
def get_cgroup_cfg(cls, plat_info):
    # Do stuff ...

    return controller, name, values
```
And then I pass these 3 to `_target_set_cgroup()` and make it return the dict ?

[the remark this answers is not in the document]
valschneider: It does look like a more elegant API for `_target_set_cgroup()`, but you still have to shove that in a dict (or something) to pass it to the `TestBundle`'s `__init__()` - adding three parameters just for that functionality feels like a big no-no.

Maybe a pure dict is fine for that, after all that's what our `rtapp_profile` is. I'll think about it and see if I come up with a smarter idea.

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,290 @@
...
+import os.path
+import abc
+
+from lisa.wlgen.rta import RTA, Periodic
+from lisa.tests.base import ResultBundle, CannotCreateError, RTATestBundle
+from lisa.trace import Trace, FtraceCollector, FtraceConf, requires_events
+from lisa.target import Target
+from lisa.utils import ArtifactPath
+
+class SchedTuneTest(RTATestBundle, abc.ABC):
```
valschneider: @douglas-raillard-arm I'm fairly sure I already asked this in the past, but `RTATestBundle` already being abstract, do we need to re-inherit from `abc.ABC`?

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,290 @@
...
+    """
+    Abstract class for SchedTune testing
+
+    :param boost: The SchedTune `boost` level to apply the rtapp's cgroup
```
valschneider: Those aren't parameters but class attributes (otherwise they'd show up in `__init__()`), so you should document them as such. Here's an example: https://github.com/ARM-software/lisa/blob/next/lisa/tests/base.py#L242

Also it looks like you use them as class constants, so they should be uppercase.

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,290 @@
...
+                'controller': 'schedtune',
+                'name': 'lisa_test',
+                'values': {
+                    'boost': cls.boost,
+                    'prefer_idle': int(cls.prefer_idle)
+                }
+             }
+
+        return cg
+
```
valschneider: You have to define and annotate `from_target()` here for exekall to pick it up (which I think is documented nowhere... On the list it goes!). I got exekall to pick them up with this:

```diff
diff --git a/lisa/tests/staging/sched_android.py b/lisa/tests/staging/sched_android.py
index 8d8678f2..ebb3055d 100644
--- a/lisa/tests/staging/sched_android.py
+++ b/lisa/tests/staging/sched_android.py
@@ -57,6 +57,11 @@ class SchedTuneTest(RTATestBundle, abc.ABC):
 
         return cg
 
+    @classmethod
+    def from_target(cls, target:Target, res_dir:ArtifactPath=None,
+                    ftrace_coll:FtraceCollector=None) -> 'SchedTuneTest':
+        return super().from_target(target, res_dir, ftrace_coll=ftrace_coll)
+
 class SchedTuneFreqBoost(SchedTuneTest, abc.ABC):
     """
     Abstract class for testing SchedTune's frequency biasing feature
```
(when no docstring is specified and we're overriding a method, we use the docstring of the super() method, so you don't have to write one).

lisa/tests/base.py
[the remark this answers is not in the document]
qperret: Yeah agreed, passing 3 parameters to `TestBundle`'s constructor is horrible. But basically I was thinking about creating the dict _inside_ `_target_set_cgroup()`, so basically somethink like the below (totally untested):

``` python

def get_cgroup_configuration(cls, plat_info):
    # Do stuff ...

    return controller, name, values

def _target_set_cgroup(cls, target, controller, name, values):
    if not controller:
        return None, None
    
    ret  = {}
    ctrl =  target.cgroups.controllers[controller]
    ret['controller'] = ctrl.kind

    cg = ctrl.cgroup(name)
    ret['name'] = cg.name

    ret['values'] = values.items()
    for k, v in values.items():
        cg.set(**{k: v})
     
    return ret, cg.name

def _from_target(cls, target, res_dir, ftrace_coll=None):
     plat_info = target.plat_info
     rtapp_profile = cls.get_rtapp_profile(plat_info)
     ctrl, name, values = cls.get_cgroup_configuration(plat_info)
     cg_cfg, cg_path = cls._target_set_cgroup(target, ctrl, name, values)
     cls._run_rtapp(target, res_dir, rtapp_profile, cgroup=cg_path, 
                    ftrace_coll=ftrace_coll)
                                                                                 
     return cls(res_dir, plat_info, rtapp_profile, cg_cfg)
```

That would at least contrain sub-classes to return `controller`, `name` and `values` properly from `get_cgroup_configuration()`, and the `dict` mess is dealt with inside `RTATestBundle` entirely

valschneider:
Also, you need to reference this in the doc:
```diff
diff --git a/doc/kernel_tests.rst b/doc/kernel_tests.rst
index 8f8e636b..2b1dd96d 100644
--- a/doc/kernel_tests.rst
+++ b/doc/kernel_tests.rst
@@ -160,3 +160,6 @@ Staged tests
 
 Those are tests that have been merged into LISA but whose behaviour are being
 actively evaluated.
+
+.. automodule:: lisa.tests.staging.sched_android
+   :members:
```

lisa/tests/staging/sched_android.py
qperret: Yeah I was wondering that too, but that's what we do for `EASBehaviour` so I though I should just do the same here

valschneider: We don't do that for `MisfitMigrationBase` so I think that's superfluous.

lisa/tests/base.py
```diff
@@ -572,24 +573,72 @@ def get_rtapp_profile(cls, plat_info):
...
+            cg.set(**{k: v})
+
+        return '/' + cg_name
+
     @classmethod
     def _from_target(cls, target, res_dir, ftrace_coll=None):
         plat_info = target.plat_info
         rtapp_profile = cls.get_rtapp_profile(plat_info)
-        cls._run_rtapp(target, res_dir, rtapp_profile, ftrace_coll)
+        cgroup_cfg = cls.get_cgroup_configuration(plat_info)
```
douglas-raillard-arm: In the current structure, it can be expected that:

* subclasses will override `_from_target()` completely, without calling `super()._from_target()`
* subclasses may "extend" `_run_rtapp()`, but will always call `super()._run_rtapp()` at some point.

As a consequence, if all this business can be done purely inside `RTATestBundle_run_rtapp()` and `get_cgroup_configuration()`, that should avoid creating any new parameter. 

Now that I'm thinking about it, having `rtapp_profile` as a `RTATestBundle.__init__` parameter is useless, since it can be fully derived from plat_info. We could remove the parameter and add a property if we want. We could follow the same pattern if the cgroups conf can be decided purely based on plat_info. That would allow "offline" code (test methods) to access that configuration, without actually recording it (since we record plat_info)

lisa/tests/staging/sched_android.py
douglas-raillard-arm: Since RTATestBundle already inherits from it, it is superfluous. It's only useful as a remainder "that class contains abstractmethods", which will remain abstract even if one day we decide RTATestBundle is not an ABC anymore (otherwise, it could silently become non-abstract).

lisa/tests/staging/sched_android.py
douglas-raillard-arm: We could consider having that "basic" annotated version directly in RTATestBundle, but so far we rejected that on the ground that if you need additional parameters, it will fail strangely at runtime (missing parameter for `_from_target()` exception). I don't feel strongly against or for any of these options.

lisa/tests/base.py
qperret: Right, the problem is you can't assume the cgroup config can be directly re-constructed using `plat-info`, sometime it'll be hardcoded. That's already not the case for the schedtune tests for example.

But it is true that I have `cls.boost` and `cls.prefer_idle` already for `SchedTuneTest`, so the cgroup config dict is somewhat superfluous ... I'll drop it and see how things go if I merge everything in `_run_rtapp()` directly :wink:

lisa/tests/staging/sched_android.py
douglas-raillard-arm: If we provide a base `RTATestBundle.check_from_target()` it could go in it. That said, it still feels like information duplication: if a required module is not present, `from_target` will fail. If the error it gives is not deemed explicit enough, we should work on that rather than maintaining duplicates that don't introduce any concrete benefit other than presenting the thing in a slightly different way. In real life, we will forget about that list and since all modules are loaded, we will never realize if that list is in sync with the code.

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,204 @@
...
+    :param prefer_idle: The SchedTune `prefer-idle` flag to apply to the rtapp's
+        cgroup
+    :type prefer_idle: bool
+
+    This class provides :meth:`get_cgroup` to create in the SchedTune
+    controller a `lisa_test` cgroup configured with the `boost` and
+    `prefer_idle` attributes.
+    """
+
+    boost = 0
```
douglas-raillard-arm: boost needs to be defined as an abstractmethod (since abstract attributes don't exist for `abc.ABC`, but it does not matter, it's the same at the end):
```
@abc.abstractmethod
def boost():
    """Your docstring"""
    pass
```

Otherwise, that class will not be `inspect.isabstract()` and exekall will try to create instances of it. Inheriting from `abc.ABC` does not make it abstract if all abstract methods are implemented (otherwise it would be impossible to be concrete, since there would be an ABC in your ancestors).

lisa/tests/base.py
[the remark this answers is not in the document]
qperret: So as discussed with @douglas-raillard-arm in another thread, I'll move the cgroup management stuff from `_from_target()` to `_run_rtapp()`, which means I can ditch the dict entirely :)

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,204 @@
...
+
+        return res
+
+class BoostedMaxFreq(SchedTuneFreqBoost):
+    """
+    A tiny task with 100% boost
+    """
+    boost = 100
+
+class BoostedFreq80(SchedTuneFreqBoost):
```
douglas-raillard-arm: Depending on the way we want to analyze the results, it may be advantageous to use the same pattern as `load_tracking.Invariance`. One class is in charge of deciding how many individual items are created, and the tests are done on each item individually. If necessary, there can be an "aggregated" version of the test that checks all items as well.

This avoids the proliferation of classes and ABCs, and more importantly, allows deciding "on the fly" (based on plat_info) how many items are needed and what are their `from_target()` parameters.

lisa/tests/staging/sched_android.py
```diff
@@ -201,4 +201,90 @@ class UnboostedFreq(SchedTuneFreqBoost):
...
+    def test_task_placement(self, bad_cpu_margin_pct=10) -> ResultBundle:
+        """
+        Test that the task placement satisfied the boost requirement
+
+        Check that top-app tasks spend no more than ```bad_cpu_margin_pct```
+        of their time on CPUs that don't have enough capacity to serve their
+        boost.
+        """
+
+        task = list(self.rtapp_profile.keys())[0]
```
douglas-raillard-arm: could we add `assert len(self.rtapp_profile) == 1` ? That would make sure any unexpected change to number of tasks is not silently ignored by that code.

lisa/tests/staging/sched_android.py
```diff
@@ -201,4 +201,90 @@ class UnboostedFreq(SchedTuneFreqBoost):
     """
     boost = 0
 
+
+class SchedTuneTaskPlacement(SchedTuneTest, abc.ABC):
+    """
+    Abstract class for testing SchedTune's task-placement biasing feature
```
douglas-raillard-arm: similar comment to another one: solely inheriting from abc.ABC does not make the class abstract (and in that case, it already inherits from it). Inheriting from abc.ABC *and* not implementing methods/attributes decorated with `@abc.abstractmethod` make a class abstract.

lisa/tests/staging/sched_android.py
qperret: Right, if we just want to rely on `from_target` to fail, I think I can remove this function safely and just use the default. We could also remove the schedutil check in `EASBehaviour.check_from_target()` I assume.

lisa/tests/staging/sched_android.py
qperret: Yeah it is somewhat similar, all those results need to be looked at as a whole. I'll try and see if I can re-use the `load_tracking.Invariance` pattern.

lisa/tests/staging/sched_android.py
valschneider: Hmm not sure I agree with @douglas-raillard-arm. AFAIR that checking step was added to fail early (and explicitely) rather than explode late. So those early checks make sense (at least the cpufreq governors / sched features) IMO.

If that is now deemed superfluous (and I'm not entirely convinced of it), then we should remove the hooks altogether.

lisa/tests/base.py
douglas-raillard-arm: > Right, the problem is you can't assume the cgroup config can be directly re-constructed using plat-info, sometime it'll be hardcoded

Not sure to get it, as the current implementation seems to already take `plat_info` and just ignores it, which is perfectly fine. My point was that since it can be derived from either nothing (hardcoded) or something already saved (`plat_info`), we can just generate it on the fly in a property, so it becomes useless to pass it to `__init__` and we keep as "real" instance attributes the minimal state everything can be derived from.

The reason we pass `rtapp_profile` to `__init__()` is that it used to be built out of a Target, which is obviously not part of the bundle, so it had to be saved. Now that it is derived from `plat_info`, I propose to stop saving it as an attribute and just turn it into a property as well (saving the json in the artifacts for manual inspection is totally orthogonal to that and we will keep doing it regardless of what happens to that `rtapp_profile` attribute).

note: TestBundle attributes are the things that in the end are saved in the serialized objects for the tests. If something is only needed to create the bundle (i.e. in `from_target()` and its callees), and never accessed again in any instance method, there is no need to record it other than for manual inspection. That can be achieved by:

* exposing it as a getter method/property if it can be derived from existing attributes, 
* creating artifacts files
* If none of the above can be done, saving "unnecessary" attributes

lisa/tests/staging/sched_android.py
valschneider: Then if we follow that logic, we should get rid of `check_from_target`, no? I think it's nice to have some way of checking if running something is going to give me an exception (which is the point of `can_create_from_target()`)

lisa/tests/staging/sched_android.py
douglas-raillard-arm: To map Invariance to your new class, you can apply this correspondance:
* `boost` becomes a parameter of your `XXXItem.from_target()` factory, just like `cpu` and `freq` for `InvarianceItem.from_target()`.
* Look at `Invariance.test_task_util_avg` and `Invariance._test_all_freq` for the aggregated version. Ignore `test_cpu_invariance` and `test_freq_invariance` since they do more than just checking an XXXItem test is passing on all items.
* Once we are done, we can add the bit in exekall_customize to tag the XXXItem objects, so they show up with their `boost` value in the non-aggregated results.

lisa/tests/staging/sched_android.py
douglas-raillard-arm: Generally speaking I would be for removing `check_from_target` if that means keeping an unmaintained list of modules, with a check that will always succeed in practice since we load them all. AFAIK, "early" means between Target creation and running rtapp. Most if not all TestBundle will access the said features by the time they start running lengthy things. If we really need to do these checks without running rtapp, we can introduce a `check` parameter in `from_target()`:

```
def can_create_from_target(self, *args, **kwargs):
    try:
        self.from_target(*args, **kwargs)
    except CheckPassedException:
        return True
    except Exception:
        return False
    else:
        return True


def check_barrier(raise_):
    if raise_:
        raise CheckPassedException

def from_target(..., check=False):
    # Do most of the setup, using all features that could raise
    ...
    check_barrier(check)
    # Run lengthy things here
    ...
```

That will leave the target set up, but tests need to be able to recover from that anyway, so it would give an easy way to make sure all setup code can work on a non-clean target as it should.

note: there is nothing to be afraid of regarding exceptions, and general python wisdom even says "it's easier to ask for forgiveness than to get permission" == "don't spend time duplicating code and introducing race conditions, just catch exceptions when they occur".

qperret: I think we should keep `check_from_target()` for things that won't prevent `from_target()` from running but will create unusable artifacts for the test methods.

For example, we can remove the schedutil check from `EASBehaviour.check_from_target()` because `from_target()` will fail anyways if it's not present. But we should keep the `nrg-model` check, because not having it will not prevent `from_target()` from running, but that will break `test_task_placement()`.

So in short, if `from_target()` can run but it's useless because we won't have what we need for the test afterwards, we can just stop early.

qperret:
OK, I pushed an update that should address most of the review comments. The comments / doc is probably utterly broken and travis will complain -- I'll fix it later. In the meantime we can discuss the structure since it changed a bit.

As suggested by @douglas-raillard-arm this now uses the same pattern as the Invariance tests. There are two 'bases' classes:
- `SchedTuneItemBase` which inherits from `RTATestBundle` and just runs an rtapp task in a schedtune cgroup;
- `SchedTuneBase` which runs a bunch of `SchedTuneItemBase` with various boost/prefer idle configuration and aggregates the results. This class also has a bunch of helpers that make the code a bit cleaner (e.g. `_merge_res_bundles()`)

And then both classes are sub-classed for the frequency and placement tests.

I kept the cgroup config stuff as a dict just to keep it consistent with the `rtapp_profile` stuff, although I don't have a strong opinion on that one as long as we keep things simple.

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+    def __init__(self, res_dir, plat_info, test_bundles):
+        super().__init__(res_dir, plat_info)
+
+        self.test_bundles = test_bundles
+
+    @classmethod
+    def from_target(cls, target:Target, res_dir:ArtifactPath=None,
+            ftrace_coll:FtraceCollector=None) -> 'SchedTuneBase':
+        """
+        XXX: Warning if nothing is documented here ...
```
qperret: I don't understand why I need to have something there. With no docstring there is a 'Warning that must be fixed' with `pedantic_build.sh` ...

valschneider: I think I got it - the warning is `WARNING: py:meth reference target not found: _from_target`

If you don't define a docstring, you inherit one, and it happens that the one you inherent from `TestBundle` uses 
```
:meth:`_from_target`
```
instead of
```
:meth:`lisa.tests.base.TestBundle._from_target`
```

so it tries to find a local definition of `lisa.tests.staging.sched_android.SchedTuneBase._from_target`. Using full paths in the docstring of `TestBundle` should fix it (but then the doc would link to `TestBundle._from_target` even if you redefine the docstring of `_from_target` in your own class.... Bah).

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+        pass
+
+    @classmethod
+    def _create_test_bundle(cls, target, res_dir, ftrace_coll, item_cls, boost, pi):
+        """
+        Creates and returns a TestBundle for a given item class, and a given
+        schedtune configuration
+        """
+        item_dir = os.path.join(res_dir, 'boost_{}'.format(boost))
```
derkling: Here you don't encode in the `pi` value... thus we can end up with some overwrites in case we test same boost values but different `pi` values...

nit-pick: should we use for consistency always the extended form, i.e. `s/pi/prefer_idle/`?

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+    checks the frequency selection was performed accordingly.
+    """
+
+    ftrace_conf = FtraceConf({
+        "events" : ["sched_switch", "cpu_frequency"]
+    }, __qualname__)
+
+    @classmethod
+    def get_rtapp_profile(cls, plat_info):
+        cpu = plat_info['capacity-classes'][-1][0]
```
derkling: Can we grant `capacity-classes` is always ordered?
Otherwise, something like:
```python
big_cpus = [cpu for cpu,cap
            in target.plat_info['cpu-capacities'].items()
            if cap==1024]
big_cpu = big_cpus[-1]
```
Should give your the topmost big cpu.

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+        return rtapp_profile
+
+    def trace_window(self, trace):
+        """
+        Set the boundaries of the trace window to ``cpu_frequency`` events
+        before/after the task's start/end time
+        """
+        rta_start, rta_stop = super().trace_window(trace)
+
+        cpu = self.plat_info['capacity-classes'][-1][0]
```
derkling: Perhaps move this code (or the version suggested above) into an accessor method, just to be sure we have only one consistent definition?

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+        fdf = trace.df_events('cpu_frequency')
```
derkling: Don't we have an informal code convention to call DF related symbols with a `_df` suffix?

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+        fdf = fdf[fdf.cpu == cpu]
+
+        # Find the frequency events before and after the task runs
+        freq_start = fdf[fdf.index < rta_start].index[-1]
+        freq_stop = fdf[fdf.index > rta_stop].index[0]
+
+        return (freq_start, freq_stop)
```
derkling: What about turning this into something generic and offered by `lisa.trace::Trace`?
Something like:
```python
def get_window(self, event):
```
where you return the window for the first and last specified `event` in `trace`?

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+            self.get_logger().warning('This test requires the RT boost hold, but it may be disabled in {}'.format(kernel_version))
+
+        cpu = self.plat_info['capacity-classes'][-1][0]
+        freqs = self.plat_info['freqs'][cpu]
+        max_freq = max(freqs)
+
+        # Estimate the target frequency, including sugov's margin, and round
+        # into a real OPP
+        boost = self.boost
+        target_freq = min(max_freq, max_freq * boost / 80)
```
derkling: You sure about this formulation?

Is not overestimating a bit the target frequency, even before considering OPPs I mean?

Why not using the exact SPC formula instead?
Assuming a negligible utilization, something like:
```python
target_freq = 1.2 * max_freq * boost / 100 = max_freq * boost * 0.012
target_freq = min(target_freq, max_freq)
```
should give us the exact value with a 20% margin on top.

lisa/tests/staging/sched_android.py
derkling: Just noticed now that, since in `RTATestBundle` we defined `trace` as a `@memoized` `@property` we also end up applying only one time `trace_window()` slicer.

@valschneider perhaps we should call this out better in the documentation of `RTATestBundle::trace`, it's not obvious if you don't read the code or the documentation of `RTATestBundle::trace_window`...

It also means that, if we override the `trace_window` we can do it only one time for each `RTATestBundle` derived class.
@valschneider: are we really happy about the need to create a complete new `RTATestBundle` derived class in case we wanna test different windows for the same trace?

lisa/tests/base.py
```diff
@@ -580,7 +587,46 @@ def get_rtapp_profile(cls, plat_info):
...
+        ctrl_name = cfg['controller']
+        if ctrl_name not in target.cgroups.controllers:
+            raise CannotCreateError('"{}" cgroup controller unavailable'.format(ctrl_name))
+        ctrl = target.cgroups.controllers[ctrl_name]
+
+        cg = ctrl.cgroup(cfg['name'])
+        for k, v in cfg['values'].items():
+            cg.set(**{k: v})
+
+        return '/' + cg.name
```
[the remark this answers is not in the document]
qperret: No it's not it seems.

lisa/tests/staging/sched_android.py
qperret: Yeah `capacity-classes` is always sorted, we already rely on that all over

lisa/tests/staging/sched_android.py
qperret: Yeah but as per `map_util_freq()` what we apply is actually +25%:

 ```C
static inline unsigned long map_util_freq(unsigned long util,
                                        unsigned long freq, unsigned long cap)
{
        return (freq + (freq >> 2)) * util / cap;
}
```

So dividing by 0.8 is correct.

File: lisa/tests/staging/sched_android.py (hunk 16 above)
Comment (qperret) [reply lisa/tests/staging/sched_android.py]: Good catch, I missed that one. I'll fix it in the next version.

File: lisa/tests/staging/sched_android.py (hunk 20 above)
Comment (qperret) [reply lisa/tests/staging/sched_android.py]: > where you return the window for the first and last specified event in trace?

I don't really want that though. I want the `cpu_frequency` event just before the task starts, and the one just after it ends. The rest of the trace doesn't really interest me TBH

Comment (valschneider) [reply lisa/tests/staging/sched_android.py]: I don't really see how multiple `TraceView` would work with our flow. In the end we take a single `Trace` (or `TraceView`) and run our tests on it. Why would you want several `TraceView`s?

You can refine a specific view if you want (e.g. use the `super()` one and nest a `TraceView` on it), but that's about it.

This also goes with your @memoized comment - I doesn't make sense to compute a different `trace_window()` since it would use the same slicing code on the same trace (one `RTATestBundle` == one `Trace`).

File: lisa/tests/base.py (hunk 22 above)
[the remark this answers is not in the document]
Comment (derkling) [reply lisa/tests/base.py]: Strange, I'm using the same API and it's working for me... I also remember a recent devlib fix just to ensure that we always have that leading `/`...

File: lisa/tests/staging/sched_android.py (hunk 20 above)
Comment (derkling) [reply lisa/tests/staging/sched_android.py]: What about cases where you wanna test different conditions on the same trace, which requires different slicing. Don't have examples on hand but I think there could be cases such that.

Comment (valschneider) [reply lisa/tests/staging/sched_android.py]: Conceptually that doesn't really fit the current framework. What you could have perhaps is a `RTATestBundle` subclass that take a slicing window as parameter, and have a base class that generates as many as you need (kind of like the `Invariance` thingy in load_tracking).

File: lisa/tests/base.py (hunk 22 above)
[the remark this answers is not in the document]
Comment (qperret) [reply lisa/tests/base.py]: Ah, I started to work on this some time ago so perhaps that has been fixed since then. I'll give it another go ;)

File: lisa/tests/staging/sched_android.py (hunk 20 above)
Comment (douglas-raillard-arm) [reply lisa/tests/staging/sched_android.py]: > What about cases where you wanna test different conditions on the same trace, which requires different slicing

The slicing you get through the `trace` property just crops what is outside of rtapp execution. AFAIK, we are never interested by whatever is around. It is possible to have sub-slices of that AFAIK. If we really need to look at what happens before and after rtapp, then it is always possible to override `trace` property with one that would be specific to a test.

File: lisa/tests/scheduler/load_tracking.py [hunk 23]
```diff
@@ -889,15 +889,15 @@ class CPUMigrationBase(LoadTrackingBase):
     """
 
     @classmethod
-    def _run_rtapp(cls, target, res_dir, profile, ftrace_coll):
+    def _run_rtapp(cls, target, res_dir, profile, ftrace_coll, cgroup=None):
```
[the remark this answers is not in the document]
douglas-raillard-arm: > Did not followed all the previous discussion thus get this with a pinch of salt and just let me know if we already agree on the following points:

My main point previously was regarding where the cgroup conf lives. It does not need to be passed to `TestBundle.__init__` (just like `rtapp_profile` which is now just a property computed out of `plat_info`, which is part of the "core state" of the object).

> Are we happy about adding yet another param this call?

Mostly yes, `_run_rtapp()` is a "standalone" class method that runs the thing and collects the stuff. It's quite easy to spot which class is making a direct use of `_run_rtapp()`, so it's easy to spot which class is "deviating" from the basic usage (in this case by using cgroups).

> How do we know in which order ftrace_coll or cgroups should be called? Are we forcing an ordering?

Ordering is the order in which the context managers are passed to `with` statement. We should probably only start tracing after cgroups are setup, since there is little value in tracing devlib's scripts.

> There is not the risk we keep adding new params to this call whenever we need to pass in something more?

Yes, as long as they are optional like this one it is not an issue. It's just that `_run_rtapp()` will offer more services if needed.

> Cannot we come up with a more generic approach, e.g. ordered list of context managers?

That could be an interesting structure, as long as the context manager abstraction is enough to describe the feature (i.e. it is fully self-contained). In the cgroup case, it is unfortunately not enough, so at least for this one we cannot put it in a CM black box:
```
as_root = cgroup is not None
```
FWIW the idea of an ordered list of CM is easy to implement using `contextlib.ExitStack`: https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+        return res
+
+class SchedTuneFrequencyTest(SchedTuneBase):
+    """
+    Runs multiple ``SchedTuneFreqItem`` tests at various boost levels ranging
+    from 20% to 100%, then checks all succedeed.
+    """
+    @classmethod
+    def _create_test_bundles(cls, target, res_dir, ftrace_coll):
+        for boost in range(20, 101, 20):
```
douglas-raillard-arm: is the sequence `20, 101, 20` intended, or should the last one be something else ?

lisa/tests/staging/sched_android.py
```diff
@@ -236,4 +236,78 @@ def test_stune_frequency(self, freq_margin_pct=10) -> ResultBundle:
...
+        total_time = 0
+        for cpu in cpu_caps:
+            t = df['runtime'][cpu]
+            if cpu in ko_cpus:
+                time_ko += t
+            total_time += t
+
+        pct_ko = time_ko * 100 / total_time
+        res = ResultBundle.from_bool(pct_ko < bad_cpu_margin_pct)
+        res.add_metric("time spent on inapproriate CPUs", pct_ko, '%')
```
douglas-raillard-arm: s/inapproriate/inappropriate/

lisa/tests/staging/sched_android.py
```diff
@@ -236,4 +236,78 @@ def test_stune_frequency(self, freq_margin_pct=10) -> ResultBundle:
...
+    """
+    Runs multiple ``SchedTunePlacementItem`` tests with prefer_idle set and
+    typical top-app boost levels, then checks all succedeed.
+    """
+    @classmethod
+    def _create_test_bundles(cls, target, res_dir, ftrace_coll):
+        # Typically top-app tasks are boosted by 10%, or 50% during touchboost
+        for boost in [10, 50]:
+            yield cls._create_test_bundle(target, res_dir, ftrace_coll,
+                    SchedTunePlacementItem, boost, 1)
```
douglas-raillard-arm: s/1/True 
I haven't checked the other places where it is used but they should all be aligned on booleans.

lisa/tests/staging/sched_android.py
qperret: I want `20,40,60,80,100`. Perhaps I should list them explicitly rather than using `range()` here

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+    @classmethod
+    @abc.abstractmethod
+    def _create_test_bundles(cls, target, res_dir, ftrace_coll):
+        """
+        Collects and returns a ResultBundle per test item.
+        """
+        pass
+
+    @classmethod
+    def _create_test_bundle(cls, target, res_dir, ftrace_coll, item_cls, boost, pi):
```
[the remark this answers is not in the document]
qperret: I'm happy to rename it, but yes, there is another caller (see the task placement test) so it's just nicer to have this factored out here IMO

lisa/tests/staging/sched_android.py
```diff
@@ -236,4 +236,78 @@ def test_stune_frequency(self, freq_margin_pct=10) -> ResultBundle:
...
         return self._merge_res_bundles(res_bundles)
 
+
+class SchedTunePlacementItem(SchedTuneItemBase):
+    """
+    Runs a tiny RT-App task marked 'prefer_idle' at a given boost level and
+    tests if it was placed on big-enough CPUs.
+    """
+
+    ftrace_conf = FtraceConf({
```
douglas-raillard-arm: hand-specified `ftrace_conf` is not necessary anymore since #824 as long as all tests methods are decorated with the required events.

lisa/tests/staging/sched_android.py
```diff
@@ -236,4 +236,78 @@ def test_stune_frequency(self, freq_margin_pct=10) -> ResultBundle:
...
+    typical top-app boost levels, then checks all succedeed.
+    """
+    @classmethod
+    def _create_test_bundles(cls, target, res_dir, ftrace_coll):
+        # Typically top-app tasks are boosted by 10%, or 50% during touchboost
+        for boost in [10, 50]:
+            yield cls._create_test_bundle(target, res_dir, ftrace_coll,
+                    SchedTunePlacementItem, boost, 1)
+
+    def test_stune_task_placement(self, margin_pct=10) -> ResultBundle:
```
douglas-raillard-arm: Make sure to forward the required events, so that the `FtraceConf` gets populated with the right events for this class:
```
    @SchedTunePlacementItem.test_stune_task_placement.used_events
     def test_stune_task_placement(self, margin_pct=10) -> ResultBundle: 
```

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+    Runs multiple ``SchedTuneFreqItem`` tests at various boost levels ranging
+    from 20% to 100%, then checks all succedeed.
+    """
+    @classmethod
+    def _create_test_bundles(cls, target, res_dir, ftrace_coll):
+        for boost in range(20, 101, 20):
+            yield cls._create_test_bundle(target, res_dir, ftrace_coll,
+                    SchedTuneFreqItem, boost, 0)
+
+    def test_stune_frequency(self, freq_margin_pct=10) -> ResultBundle:
```
douglas-raillard-arm: In order to forward the used events:
```
    @SchedTuneFreqItem.test_stune_frequency.used_events
     def test_stune_frequency(self, freq_margin_pct=10) -> ResultBundle: 
```

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+        fdf = trace.df_events('cpu_frequency')
+        fdf = fdf[fdf.cpu == cpu]
+
+        # Find the frequency events before and after the task runs
+        freq_start = fdf[fdf.index < rta_start].index[-1]
+        freq_stop = fdf[fdf.index > rta_stop].index[0]
+
+        return (freq_start, freq_stop)
+
+    @requires_events('cpu_frequency')
```
douglas-raillard-arm: Better to forward the events from the function that actually uses them:
```
@FrequencyAnalysis.get_average_cpu_frequency.used_events
```

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+        ])
+        return overall_bundle
+
+class SchedTuneFreqItem(SchedTuneItemBase):
+    """
+    Runs a tiny RT rtapp task pinned to a big CPU at a given boost level and
+    checks the frequency selection was performed accordingly.
+    """
+
+    ftrace_conf = FtraceConf({
```
douglas-raillard-arm: ftrace_conf is not needed anymore as long as all methods are decorated with required events

lisa/tests/staging/sched_android.py
```diff
@@ -0,0 +1,239 @@
...
+            duration_s = 10,
+            period_ms = 16,
+            cpus = [cpu], # pin to big CPU, to focus on frequency selection
+            sched_policy = 'FIFO' # RT tasks have the boost holding feature so
+                                  # the frequency should be more stable, and we
+                                  # shouldn't go to max freq in Android
+        )
+        return rtapp_profile
+
+    def trace_window(self, trace):
```
douglas-raillard-arm:
```
@requires_events(SchedTuneItemBase.trace_window.used_events, "cpu_frequency")
```

valschneider:
I haven't looked at this in a while, but last time I did it looked okay. We should get this in to give it some coverage, and it'll be in `staging` anyway. So if everyone is happy with the !`staging` changes, I'd say this should go in (maybe just fixup the last comments from @douglas-raillard-arm).

qperret:
Argh, sorry, it's been a while since the last update, I'll try and do something about it soon-ish

lisa/tests/base.py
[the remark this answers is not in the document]
qperret: So, I tested again, and no, I don't have a leading `/`. I get `ValueError: cgroup name "lisa_test" must start with "/"`.

qperret:
I addressed most of the review comments and rebased on master but this is now utterly broken for some reason. So please hang on before merging.

The `trace_window()` stuff seems to get stuck in an infinite loop for some reason. If I interrupt it get something like this going on forever:

```
data/work/lisa-next/lisa/tests/base.py in trace(self)
    493         path = os.path.join(self.res_dir, self.TRACE_PATH)
    494         trace = Trace(path, self.plat_info, events=self.ftrace_conf["events"])
--> 495         return trace.get_view(self.trace_window(trace))
    496 
    497     @property

/data/work/lisa-next/lisa/trace.py in wrapper(self, *args, **kwargs)
   1147             def wrapper(self, *args, **kwargs):
   1148                 try:
-> 1149                     trace = self.trace
   1150                 # If there is no "trace" attribute, silently skip the check. This
   1151                 # allows using the decorator for documentation and chaining purpose

/data/work/lisa-next/lisa/tests/base.py in trace(self)
    493         path = os.path.join(self.res_dir, self.TRACE_PATH)
    494         trace = Trace(path, self.plat_info, events=self.ftrace_conf["events"])
--> 495         return trace.get_view(self.trace_window(trace))
    496 
    497     @property

/data/work/lisa-next/lisa/trace.py in wrapper(self, *args, **kwargs)
   1147             def wrapper(self, *args, **kwargs):
   1148                 try:
-> 1149                     trace = self.trace
   1150                 # If there is no "trace" attribute, silently skip the check. This
   1151                 # allows using the decorator for documentation and chaining purpose

/data/work/lisa-next/lisa/tests/base.py in trace(self)
    493         path = os.path.join(self.res_dir, self.TRACE_PATH)
    494         trace = Trace(path, self.plat_info, events=self.ftrace_conf["events"])
--> 495         return trace.get_view(self.trace_window(trace))
    496 
    497     @property

/data/work/lisa-next/lisa/trace.py in wrapper(self, *args, **kwargs)
   1147             def wrapper(self, *args, **kwargs):
   1148                 try:
-> 1149                     trace = self.trace
   1150                 # If there is no "trace" attribute, silently skip the check. This
   1151                 # allows using the decorator for documentation and chaining purpose

/data/work/lisa-next/lisa/tests/base.py in trace(self)
    492         """
    493         path = os.path.join(self.res_dir, self.TRACE_PATH)
--> 494         trace = Trace(path, self.plat_info, events=self.ftrace_conf["events"])
    495         return trace.get_view(self.trace_window(trace))
    496 
```

valschneider:
For some reason that seems to go through the trace event checker wrapper (which itself calls `self.trace`, thus giving us a lovely loop), but I don't see any reason to do that here nor what changed introduced this. @douglas-raillard-arm ?

qperret:
That seems to fix it:

```
diff --git a/lisa/tests/staging/sched_android.py b/lisa/tests/staging/sched_android.py
index df4910b455a4..1b0ff92da111 100644
--- a/lisa/tests/staging/sched_android.py
+++ b/lisa/tests/staging/sched_android.py
@@ -163,13 +163,16 @@ class SchedTuneFreqItem(SchedTuneItemBase):
         )
         return rtapp_profile
 
-    @requires_events("cpu_frequency")
     def trace_window(self, trace):
         """
         Set the boundaries of the trace window to ``cpu_frequency`` events
         before/after the task's start/end time
         """
-        rta_start, rta_stop = super().trace_window(trace)
+        sdf = trace.df_events('sched_switch')
+        # Find when the first task starts running
+        rta_start = sdf[sdf.next_comm.isin(self.rtapp_profile.keys())].index[0]
+        # Find when the last task stops running
+        rta_stop = sdf[sdf.prev_comm.isin(self.rtapp_profile.keys())].index[-1]
 
         cpu = self.plat_info['capacity-classes'][-1][0]
         freq_df = trace.df_events('cpu_frequency')
```

douglas-raillard-arm:
If the method is not decorated, `cpu_frequency` will neither be collected nor parsed from the trace by default. I'd rather avoid having an explicit `ftrace_conf` for something internal like that if possible, since it does not play super-well with inheritance (subclasses overriding it will need to extend its content explicitely, or we would need to merge the `FtraceConf` along the MRO, which the possibility to wipe out any configuration "inherited" from a base class for debugging and special uses).

One way to avoid the issue would be to introduce a decorator that does not carry out the check, but just exposes the data to the rest of the code (auto `FtraceConf`, doc building etc).

qperret:
Sorry, fat fingers...

I believe I addressed all review comments, and all test results on Juno with `lisa-test` and in a notebook look good.

PR #830, opened 2019-03-20
Add exekall/bisector doc

douglas-raillard-arm:
* Add some new reStructuredText directives to ease inclusion of generated doc by running python snippets or external commands 
* Add a man page for exekall
* Add a man page and API doc for bisector
* Add exekall and bisector doc to the main html doc


doc/man1/exekall.1
```diff
@@ -0,0 +1,460 @@
+.\" Man page generated from reStructuredText.
```
valschneider: Mmm don't we want to *not* commit this? Otherwise we'll have to manually keep it in sync with the .rst version in the git repo.

tools/exekall/exekall/main.py
```diff
@@ -780,7 +780,7 @@ def handle_cycle(path):
     if rst_expr_list:
         id_kwargs['style'] = 'rst'
         for expr in expr_list:
-            out(expr.get_id(**id_kwargs))
+           out('* {}'.format(expr.get_id(**id_kwargs)))
```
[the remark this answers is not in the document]
douglas-raillard-arm: not really, that bullet marker used to be added by code that is removed by that commit and replaced by an `exec` reST directive. Otherwise, this commit will generate a paragraph without new lines in the output instead of a bullet list.

doc/man1/exekall.1
douglas-raillard-arm: If we don't commit that, `man exekall` will not work out of the box. I'd rather not have all users need to build the doc just to get access to man pages, especially since we will turn `lisa-help` in a man page as well pretty soon using that infrastructure.

In order to avoid too much issues, `pedantic_build.sh` will always generate the man page from scratch, so it will follow changes of `exekall --help` and friends without troubles (it's fast to build anyway).

douglas-raillard-arm:
@valschneider spelling mistakes addressed, I consider it as good to go. There are a couple of other pages I want to add (mainly on exekall), but it can wait after this PR has been merged.

PR #831, opened 2019-03-20
Doc: class diagrams

valschneider:
Output can be previewed here: https://my-lisa.readthedocs.io/en/lisa-next/kernel_tests.html

### DIFF
--- a/install_base_ubuntu.sh
+++ b/install_base_ubuntu.sh
@@ -10,7 +10,7 @@ SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
 cd "$SCRIPT_DIR"
 
 usage() {
-    echo Usage: "$0" [--install-android-sdk]
+    echo Usage: "$0" [--install-android-sdk] [--install-doc-extras]
 }
 
 latest_version() {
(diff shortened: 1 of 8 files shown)


valschneider:
Hmph, the `TestBundle` diagram needs plantuml & graphviz to be installed. I might just generate that locally and commit the image.

derkling:
Generating the images locally implies you have to commit in binary images, which means we kind of lost the benefits in having just textual representations to version control, isn't it?

I probably like better the idea to add the required dependencies to the `lisa-install` script, maybe guarded by an explicit flag, something similar to what we do for the SDK.

valschneider:
If there's no objections then I'll add these dependencies.

install_base_ubuntu.sh
```diff
@@ -59,13 +59,20 @@ install_nodejs() {
     apt-get install -y nodejs npm
 }
 
+install_doc_extras() {
```
valschneider: I hid the extra deps behind this. Dunno if it would make sense it install it unconditionally.

derkling: I would keep it optional... as per the SDK in the CI loop I don't think we need those bits and thus we can save a bit of download/installation  time, isn't it?

valschneider: Okay, I'll try to tweak the doc build to make it obvious that you need those extra bits when plantuml fails

PR #835, opened 2019-03-25
lisa.tests: Turn rtapp_profile in a property

douglas-raillard-arm:
* Move get_rtapp_profile as an abstractmethod of RTATestBundle instead
of EASBehaviour
* Turn it into a property, to clean up __init__ signature, since it only
depends on plat_info attribute. That will make it clear what is part of
the minimal state of the object, and what is just derived information.

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -350,11 +350,7 @@ class TestBundle(Serializable, abc.ABC):
 
 class RTATestBundle(TestBundle):
     """
-    "Abstract" class for :class:`lisa.wlgen.rta.RTA`-powered TestBundles
-
-    :param rtapp_profile: The rtapp parameters used to create the synthetic
-      workload. That happens to be what is returned by :meth:`get_rtapp_profile`
-    :type rtapp_profile: dict
+    Abstract Base Class for :class:`lisa.wlgen.rta.RTA`-powered TestBundles
     """
 
     TRACE_PATH = 'trace.dat'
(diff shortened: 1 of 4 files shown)


lisa/tests/base.py
```diff
@@ -348,13 +348,9 @@ def to_dir(self, res_dir):
         """
         super().to_path(self._filepath(res_dir))
 
-class RTATestBundle(TestBundle):
+class RTATestBundle(TestBundle, abc.ABC):
```
valschneider: AFAICT That's not needed - I removed some of those to make the class diagrams less confusing, and `TestBundle` is likely to stay abstract for the foreseeable future.

lisa/tests/base.py
```diff
@@ -431,9 +427,22 @@ def trace(self):
...
+    @property
+    def rtapp_profile(self):
+        """
+        Compute the RTapp profile based on ``plat_info``.
+        """
+        return self.get_rtapp_profile(self.plat_info)
+
+    @classmethod
+    @abc.abstractmethod
+    def get_rtapp_profile(cls, plat_info):
```
valschneider: I can't highlight it but it's already defined at line 582 of that file. It's the bit that surprised me from the PR message - it *is* already defined in this class, otherwise it wouldn't make a lot of sense (misfit tests would have to reimplement that logic for instance)

douglas-raillard-arm: That's right, I was mislead by a redundant definition in EASBehaviour probably dating from before a refactoring.

> otherwise it wouldn't make a lot of sense 
hence the commit :)

> misfit tests would have to reimplement that logic for instance
Having an abstractmethod or not is orthogonal, there are many codebases that don't use ABC at all and rely on "informal ABC" that do not enforce implementation, but where it is still expected.

lisa/tests/base.py
douglas-raillard-arm: +1 for class diagram

douglas-raillard-arm:
Merging and handle doc build issue in a separate PR (intersphinx problem with wlauto)

PR #845, opened 2019-04-01
doc: Fix workload automation intersphinx inventory

douglas-raillard-arm:


doc/conf.py
```diff
@@ -338,7 +338,7 @@ def handle_item(fieldarg, content):
     'devlib' : ('https://pythonhosted.org/devlib/', None),
```
valschneider: Can you do devlib's while you're at it?

PR #849, opened 2019-04-02
lisa/tests: eas_behaviour: Make rtapp profile construction less obfuscated

valschneider:
This is where it would've been nice to operate on an instance of a class, so that we could use e.g. `self.{min,max}_cap_cpus[0]`, but we
have to live with classmethods (or
`cls.get_{min,max}_cap_cpus(plat_info)[0]`, which is too wordy for me).

### DIFF
--- a/lisa/tests/scheduler/eas_behaviour.py
+++ b/lisa/tests/scheduler/eas_behaviour.py
@@ -371,8 +371,8 @@ class OneSmallTask(EASBehaviour):
 
     @classmethod
     def get_rtapp_profile(cls, plat_info):
-        # 50% of the smallest CPU's capacity
-        duty = cls.unscaled_utilization(plat_info, plat_info["capacity-classes"][0][0], 50)
+        littles = plat_info["capacity-classes"][0]
+        duty = cls.unscaled_utilization(plat_info, littles[0], 50)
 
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
> cls.get_{min,max}_cap_cpus(plat_info)[0]

As an alternative, it could also be possible to have:
```
plat_info.get_{min,max}_cap_cpus()
```

qais-yousef:
I'm very sorry I just noticed I was added a reviewer to this. The notifications from github are not obvious as I checked the emails from this thread and none contained an explicit request in the body.

douglas-raillard-arm:
> none contained an explicit request in the body

Github is not great at automatic notifications, it only sends them when somebody did an explicit action, so if you get the email, it means you are part of the requested reviewers (or you subscribed to the PR yourself). From this point, you only receive an email for new comments, or if the developer explicitely requested a new review (there is a button somewhere). Merely pushing an updated branch does not trigger any sort of automatic notification, which is why people usually write a comment with "@" references to ping the reviewers when comments are addressed (unlike gerrit which tends to flood with emails unless carefully configured)

qais-yousef:
I added myself as a watcher to this project. So I pretty much get everything which is very verbose.

PR #850, opened 2019-04-08
energy_model: Reflect 80% capacity margin in optimal placement

qperret:
The Linux scheduler always tries to provide at least 20% of idle time to
all CPUs by not placing big tasks on small CPUs, and by selecting
appropriate frequencies. This is implemented by the 'capacity_margin'
coefficient in CFS, and schedutil's map_util_freq() (from 5.0 onwards).

However, this is not yet reflected in Lisa's optimal placement algorithm
which will assume a 39% task fits on a 40% CPU, for example. This
difference is the root cause for many task_task_placement failures on
some boards (TC2) although there is not much we can actually do to 'fix'
the kernel. One possibility would be to make the energy margin larger,
but that would lower the bar for all platforms so an alternative
solution is preferable.

In order to solve this, reflect the kernel's 80% capacity margin in
Lisa's algorithm to ensure an apple to apple comparison in the task
placement test. This arguably lowers the level of abstraction of the
test by making it deal with 'implementation caveats' rather than doing
pure functional testing. But we already do this for other things (e.g.
load tracking) and that should make the test actually work more
reliably, hence enabling better regression testing.

### DIFF
--- a/lisa/energy_model.py
+++ b/lisa/energy_model.py
@@ -531,14 +531,19 @@ class EnergyModel(Serializable, Loggable):
         return [s or list(c.idle_states.keys())[0]
                 for s, c in zip(states, self.cpu_nodes)]
 
-    def _guess_freqs(self, cpu_utils):
+    def _guess_freqs(self, cpu_utils, capacity_margin_pct):
         overutilized = False
         # Find what frequency each CPU would need if it was alone in its
         # frequency domain
         ideal_freqs = [0 for _ in self.cpus]
         for node in self.cpu_nodes:
             [cpu] = node.cpus
-            required_cap = cpu_utils[cpu]
+
+            # A capacity margin should be provided to meet the demand at a
+            # given utilizaton level as per the scheduler's 'capacity_margin'
+            # coefficient so reflect that here
+            margin = 100 / (100 - capacity_margin_pct)
+            required_cap = cpu_utils[cpu] * margin
 
             possible_freqs = [f for f, s in node.active_states.items()
                               if s.capacity >= required_cap]
@@ -559,7 +564,7 @@ class EnergyModel(Serializable, Loggable):
 
         return freqs, overutilized
 
-    def guess_freqs(self, cpu_utils):
+    def guess_freqs(self, cpu_utils, capacity_margin_pct=0):
         """Work out CPU frequencies required to execute a workload
 
         Find the lowest possible frequency for each CPU that provides enough
@@ -568,10 +573,11 @@ class EnergyModel(Serializable, Loggable):
 
         :param cpu_utils: Utilization distribution, see
                              :ref:`cpu_utils <cpu-utils>`
+        :param capacity_margin_pct: Capacity margin before overutilizing a CPU
         :returns: List ``ret`` where ``ret[N]`` is the frequency that CPU N must
                   run at
         """
-        freqs, _ = self._guess_freqs(cpu_utils)
+        freqs, _ = self._guess_freqs(cpu_utils, capacity_margin_pct)
         return freqs
 
     def _estimate_from_active_time(self, cpu_active_time, freqs, idle_states,
@@ -669,7 +675,7 @@ class EnergyModel(Serializable, Loggable):
         return self._estimate_from_active_time(cpu_active_time,
                                                freqs, idle_states, combine=True)
 
-    def get_optimal_placements(self, capacities):
+    def get_optimal_placements(self, capacities, capacity_margin_pct=0):
         """Find the optimal distribution of work for a set of tasks
 
         Find a list of candidates which are estimated to be optimal in terms of
@@ -698,6 +704,7 @@ class EnergyModel(Serializable, Loggable):
                            have a single static utilization value. A set of
                            single-phase periodic RT-App tasks is an example of a
                            suitable workload for this model.
+        :param capacity_margin_pct: Capacity margin before overutilizing a CPU
         :returns: List of ``cpu_utils`` items representing distributions of work
                   under optimal task placements, see
                   :ref:`cpu_utils <cpu-utils>`. Multiple task placements
@@ -729,7 +736,7 @@ class EnergyModel(Serializable, Loggable):
                 continue
 
             if util not in candidates:
-                freqs, overutilized = self._guess_freqs(util)
+                freqs, overutilized = self._guess_freqs(util, capacity_margin_pct)
                 if overutilized:
                     # This isn't a valid placement
                     excluded.append(util)
--- a/tests/test_energy_model.py
+++ b/tests/test_energy_model.py
@@ -182,6 +182,11 @@ class TestOptimalPlacement(TestCase):
         self.assertRaises(EnergyModelCapacityError,
                           em.get_optimal_placements, {'task0' : 401})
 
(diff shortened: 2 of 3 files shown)


qperret:
I ran 50 iterations of all `eas_behaviour.py` tests on TC2 and have 100% passing (with a few undecided ones, but no failures). We should probably try and run it on more boards before merging.

douglas-raillard-arm:
That's confirmed to fix the issue we had on RampDown:test_task_placement on TC2, I'll now check that it did not introduce any issue on other tests/boards.
If that proves to work well (which I expect), it might be worth adding a boolean option to `get_optimal_placement()` to select which behavior we want, and to make it obvious what piece of the code is affected by that capacity margin.

qperret:
Right, perhaps we should have a knob. But maybe not a boolean. Perhaps we should expose it as a parameter of `test_task_placement` ? Some partners change the value of `capacity_margin` to 95% for example. So they would probably appreciate a knob like that

douglas-raillard-arm:
Seems like a good idea, setting it to 0% would make it behave "optimally", which can be called out in the docstring if somebody wants the old behaviour.

lisa/energy_model.py
```diff
@@ -538,7 +538,12 @@ def _guess_freqs(self, cpu_utils):
         ideal_freqs = [0 for _ in self.cpus]
         for node in self.cpu_nodes:
             [cpu] = node.cpus
-            required_cap = cpu_utils[cpu]
+
+            # At least 25% of capacity margin should be provided to meet the
+            # demand at a given utilizaton level as per the scheduler's
+            # 'capacity_margin' coefficient and schedutil's map_util_freq(),
+            # so reflect that here
+            required_cap = cpu_utils[cpu] * 1.25
```
valschneider: Maybe I don't have my thoughts straight, but aren't we mixing two concepts here?

One is the `capacity_margin` (20% default) which determines the overutilized threshold, and which we use in `task_fits_capacity()`.

The other is `map_util_freq()` which relates to schedutil trying to give us a frequency 25% higher than what the `utilization` requires. So IIUC with this we're using `map_util_freq()` for capacity selection, which applies to both CPU and frequencies.

In the end I think my rambling is that we don't actually see `capacity_margin` here.

qperret: Well IMO both are exactly the same thing. At some point I actually wanted to send a patch that makes sugov use `capacity_margin` directly. And I still think we should do that. Both for capacity awareness and frequency selection we try to have at least 20% of idle time.

douglas-raillard-arm:
This has been exhaustively verified on last integrations's data to:
* fix all failing iterations of `RampDown:test_task_placement` on TC2 during 
* not introduce any regression in `*:test_task_placement` on both Juno R0 and hikey 960

qperret:
This should now implement the tunable part. Testing required before merging.

lisa/energy_model.py
```diff
@@ -669,7 +674,7 @@ def estimate_from_cpu_util(self, cpu_utils, freqs=None, idle_states=None):
         return self._estimate_from_active_time(cpu_active_time,
                                                freqs, idle_states, combine=True)
 
-    def get_optimal_placements(self, capacities):
+    def get_optimal_placements(self, capacities, capacity_margin_pct=20):
```
valschneider: What about having the default value in `energy_model` be 0 (i.e. not used), and only keep the 20 for the `eas_behaviour` tests?

tests/test_energy_model.py
```diff
@@ -167,9 +167,9 @@ def test_single_small(self):
                                                    [0, 1, 0, 0]])
 
     def test_single_big(self):
-        placements = em.get_optimal_placements({'task0': 350})
-        self.assertPlacementListEqual(placements, [[0, 0, 350, 0],
-                                                   [0, 0, 0, 350]])
+        placements = em.get_optimal_placements({'task0': 300})
```
valschneider: Now that the knob is exposed, it would be nice to test it here. So we could keep the old test with `capacity_margin_pct=0`, and then re-run that test with e.g. `capacity_margin_pct=20` and make sure that it raises an exception.

lisa/energy_model.py
qperret: Right, that sounds like a good idea.

tests/test_energy_model.py
qperret: Sounds good. I added `test_capacity_margin_single()`.

qperret:
> we should expose that new parameter to the other definitions of `test_task_placement`

Done.

qperret:
I'm running a last set of test using @douglas-raillard-arm 's `check-test-fix.sh`, just to be on the safe side. Things look good so far, I'll let you know when I have the full results.

qperret:
I applied this and ran again `test_task_placement()` against the artifacts of the last integration.

For TC2:
```
testcase                                       old%   new% delta%      pvalue changed
-------------------------------------------------------------------------------------
EnergyModelWakeMigration:test_task_placement   0.0%   0.0%   0.0%    1.00e+00 
OneSmallTask:test_task_placement               0.0%   0.0%   0.0%    1.00e+00 
RampDown:test_task_placement                  53.9%   0.0% -53.9%    3.54e-58 *
RampUp:test_task_placement                     1.4%   0.0%  -1.4%    1.24e-01 
ThreeSmallTasks:test_task_placement            0.7%   0.7%   0.0%    1.00e+00 
TwoBigTasks:test_task_placement                0.0%   0.0%   0.0%    1.00e+00 
TwoBigThreeSmall:test_task_placement          10.8%  10.8%   0.0%    1.00e+00
```

For Juno r0
```
testcase                                       old%   new% delta%      pvalue changed
-------------------------------------------------------------------------------------
EnergyModelWakeMigration:test_task_placement   0.5%   0.0%  -0.5%    4.99e-01 
OneSmallTask:test_task_placement               0.3%   0.0%  -0.3%    1.00e+00 
RampDown:test_task_placement                   0.3%   0.0%  -0.3%    1.00e+00 
RampUp:test_task_placement                     0.3%   0.0%  -0.3%    1.00e+00 
ThreeSmallTasks:test_task_placement            1.1%   0.8%  -0.3%    1.00e+00 
TwoBigTasks:test_task_placement                0.3%   0.0%  -0.3%    1.00e+00 
TwoBigThreeSmall:test_task_placement           3.7%   3.4%  -0.3%    1.00e+00
```

For Hikey 960:
```
testcase                                       old%   new% delta%      pvalue changed
-------------------------------------------------------------------------------------
EnergyModelWakeMigration:test_task_placement   0.0%   0.0%   0.0%    1.00e+00 
OneSmallTask:test_task_placement               0.3%   0.3%   0.0%    1.00e+00 
RampDown:test_task_placement                   0.0%   0.0%   0.0%    1.00e+00 
RampUp:test_task_placement                     0.0%   0.0%   0.0%    1.00e+00 
ThreeSmallTasks:test_task_placement            4.2%   4.2%   0.0%    1.00e+00 
TwoBigTasks:test_task_placement                0.0%   0.0%   0.0%    1.00e+00 
TwoBigThreeSmall:test_task_placement           3.6%   3.6%   0.0%    1.00e+00
```

These look pretty good. If there is nothing else to do on the coding side I believe this is good to go.

PR #851, opened 2019-04-08
RTATestBundle: per-task noisiness threshold values

valschneider:
This implements what @douglas-raillard-arm and I discussed offline. 

Briefly tested on a trace from last week's mainline integration.


valschneider:
FYI on a previousled `UNDECIDED` test I got this output:
```
2019-04-08 15:00:30,738 INFO    : lisa.tests.scheduler.eas_behaviour.TwoBigThreeSmall : Ignored PIDs for noise contribution: 2265 (big_0), 2264 (small_2), 2263 (small_1), 2262 (small_0), 2266 (big_1), 0 (swapper/4), 86 (sugov:2), 84 (sugov:0)

PASSED: noisiest task={duration (abs)=0.004312999999967815 s, comm=sshd, pid=175, duration (rel)=0.4276368626314915 %}, energy threshold=1572.1895813953488 bogo-joules, estimated energy=1517.9204315027 bogo-joules
```

lisa/tests/base.py
```diff
@@ -462,8 +472,37 @@ def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
...
+        df["runtime_pct"] = df.apply(compute_duration_pct, axis=1)
+
+        # Figure out which PIDs to exclude from the thresholds
+        for key, threshold in self.NOISE_ACCOUNTING_THRESHOLDS.items():
+            # Find out which task(s) this threshold is about
+            if isinstance(key, int):
+                pids = [key]
+            elif isinstance(key, str):
+                pids = df[df.comm == key].index.values
+            elif isinstance(key, re._pattern_type):
```
douglas-raillard-arm: `_pattern_type` is a private attribute of `re` module, which seems undocumented and has been removed at some point between Python 3.5 and 3.7. It would probably be better to restrict to int and str, and just use `re.match()`. The `re` module has a compiled regex cache anyway, so if call that in a loop that should be fast enough (or we can call `.compile()` just before the loop if needed, but let's keep that local to this function.

lisa/tests/base.py
```diff
@@ -462,8 +472,37 @@ def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
...
+                comms = [comm for comm in df.comm.values if re.match(key, comm)]
+                pids = df[df.comm.isin(comms)].index.values
+            else:
+                pids = []
+
+            # For those tasks, check the threshold
+            ignored_pids += [pid for pid in pids if df.loc[pid].runtime_pct < threshold]
```
douglas-raillard-arm: nit: We could check `<=` instead of `<`

lisa/tests/base.py
valschneider: Argh, I didn't know about that removal - I thought that was the canonical way of checking this.

I'll simplify this as you've suggested.

valschneider:
@douglas-raillard-arm Done

lisa/tests/base.py
douglas-raillard-arm: as a bonus that should make the Sphinx doc look better, since it displays the content of class attributes :)

lisa/tests/base.py
```diff
@@ -462,8 +472,35 @@ def test_noisy_tasks(self, noise_threshold_pct=None, noise_threshold_ms=None):
...
+
+            # For those tasks, check the threshold
+            ignored_pids += [pid for pid in pids if df.loc[pid].runtime_pct <= threshold]
+
+        log_pids = ["{} ({})".format(pid, df.loc[pid].comm) for pid in ignored_pids]
+        self.get_logger().info(
+            "Ignored PIDs for noise contribution: %s", ", ".join(log_pids))
+
+        # Filter out unwanted tasks (rt-app tasks + thresholds)
+        df_noise = df[~df.index.isin(ignored_pids)]
```
douglas-raillard-arm: I can't comment on the other lines not touched here, but shouldn't we check the sum of noise tasks, instead of just the noisiest ? If we have 20 tasks that contribute 0.9% each, that will not be caught but will still be an issue. 

Reporting the noisiest task makes sense to avoid bloating the metrics, but we should probably take into account the sum:
```
duration_s = df_noise.runtime.values.sum()
```

douglas-raillard-arm:
LGTM

PR #852, opened 2019-04-08
tools: Add check-test-fix.py

douglas-raillard-arm:
tools: Add check-test-fix.py

Add a tool to re-run a given test against data collected using bisector,
like for the fortnightly integration. It will show a
regression/improvement table at the end to compare the reference data in
the bisector report with the new ones that were just generated. It can
be called as follow to check that a fix to a
RampDown:test_task_placement has been fixed on the first 10 available
traces:
$ check-test-fix.py -s 'RampDown:test_task_placement' bisector_report.yml.gz -oiterations=1-10 --only behaviour

Note: this will spawn one `exekall run` per CPU to speed up the
computation. That means the logs are output only when `exekall run`
command finishes to avoid interleaved output.

### DIFF
--- a/doc/man1/bisector.1
+++ b/doc/man1/bisector.1
@@ -353,12 +353,16 @@ Inherits from: shell
             consider only tests that failed or had an error
 
       \-o ignore\-testcase= (comma\-separated list) 
-            completely ignore test cases matching one of the patterns in the
-            comma\-separated list. * can be used to match any part of the name.
+            completely ignore untagged test cases matching one of the patterns
+            in the comma\-separated list. * can be used to match any part of the
+            name.
 
       \-o iterations= (comma\-separated list of integer ranges) 
             comma\-separated list of iterations to consider. Inclusive ranges can
             be specified with <first>\-<last>
+
+      \-o result\-uuid= (comma\-separated list) 
+            show only the test results with a UUID in the comma\-separated list.
 
       \-o show\-artifact\-dirs= (bool) 
             show exekall artifact directory for all iterations
@@ -383,8 +387,9 @@ Inherits from: shell
             tests
 
       \-o testcase= (comma\-separated list) 
-            show only the test cases matching one of the patterns in the comma\-
-            separated list. * can be used to match any part of the name
+            show only the untagged test cases matching one of the patterns in
+            the comma\-separated list. * can be used to match any part of the
+            name
 
       \-o upload\-artifact= (bool) 
             upload the artifact directory to Artifactorial and update the in\-
--- a/doc/workflows/automated_testing.rst
+++ b/doc/workflows/automated_testing.rst
@@ -444,6 +444,25 @@ collected ones. This can then be compared with another one for regressions:
   without relying on other files/folders being available (locally or over
   HTTP).
 
+Fixing regressions
+------------------
+
+``check-test-fix.py`` tool can be used to check that a fix to a test resolved
+errors or a regression, provided that the test can be re-executed on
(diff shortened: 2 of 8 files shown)


tools/check-test-fix.sh
```diff
@@ -0,0 +1,75 @@
+#! /usr/bin/env bash
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2018, ARM Limited and contributors.
```
valschneider:
```suggestion
# Copyright (C) 2019, ARM Limited and contributors.
```
https://www.youtube.com/watch?v=ta41xU-tkFA

douglas-raillard-arm:
Since the DB resulting from calling `exekall run --load-db` will contain both new and existing data, the regression currently computed will exhibit wrong failure rate for new data.

douglas-raillard-arm:
This PR benefits from having this one applied first: https://github.com/ARM-software/lisa/pull/881

tools/check-test-fix.sh
```diff
@@ -0,0 +1,75 @@
...
+mkdir -p "$extracted_artifacts_folder" &&
+find  "$bisector_logs_folder" -name '*.tar.xz' -exec \
+    tar -C "$extracted_artifacts_folder" -xvf '{}' ';'
+
+# Run the test against all DBs that we extracted.
+# Running on one DB at a time instead of merging first allows constant memory
+# usage, instead of having to accumulate all results.
+find "$extracted_artifacts_folder" -name 'VALUE_DB.pickle.xz' -print0	\
+    | xargs -0 -n1 -I '{}'	\
+	exekall run "$LISA_HOME/lisa/tests/" \
```
qperret: This step really takes forever on large data sets :/. One possibility could be to parallelize this a bit. I tried using GNU parallel and got the following results (10 iterations of all `test_task_placement` on TC2):

```
$ time find TC2.report.yml.gz.artifacts/extracted_exekall_artifacts \
    -name VALUE_DB.pickle.xz -print0 | xargs -0 -n1 -I '{}' \
       exekall run "$LISA_HOME/lisa/tests/" --load-db "{}" \
       --artifact-root TC2.report.yml.gz.artifacts/new_exekall_artifacts \
       --select '*:test_task_placement' --load-type '*TestBundle

real	2m27.946s
user	2m52.986s
sys	1m12.864s
```
And with parallel
```
$ time find TC2.report.yml.gz.artifacts/extracted_exekall_artifacts \
    -name VALUE_DB.pickle.xz | parallel -I '{}' --max-args 1 \
       exekall run "$LISA_HOME/lisa/tests/" --load-db "{}" \
       --artifact-root TC2.report.yml.gz.artifacts/new_exekall_artifacts \
       --select '*:test_task_placement' --load-type '*TestBundle'

real	0m24.793s
user	3m53.745s
sys	1m9.134s
```

`parallel` is in the ubuntu repo, so perhaps it's OK to add it as a dependency ?

douglas-raillard-arm:
`exekall run` commands are now run in parallel, which speeds up a lot the processing.

tools/check-test-fix.sh
douglas-raillard-arm: got it working with `xargs -P`. Also made uncompressing parallel, with as many processes as possible

douglas-raillard-arm:
@valschneider @qperret The script is now implemented in Python and will not re-download+extract the artifacts if it has already been done for a given set of `bisector report` options. I consider it as ready, so I removed the [RFC] tag in the PR.

PR #861, opened 2019-04-11
trace.py: improve add_events_deltas()

qais-yousef:
It's completely untested as my next branch is screwed up at the moment.

### DIFF
--- a/lisa/trace.py
+++ b/lisa/trace.py
@@ -91,10 +91,8 @@ class TraceBase(abc.ABC):
         if not inplace:
             df = df.copy()
 
-        df.loc[df.index[:-1], col_name] = df.index.values[1:] - df.index.values[:-1]
-        # Fix the last event, which will have a NaN duration
-        # Set duration to trace_end - last_event
-        df.loc[df.index[-1], col_name] = self.end - df.index[-1]
+        df[col_name] = df.index
+        df[col_name] = df[col_name].diff().fillna(0)
 
         return df
 


lisa/trace.py
```diff
@@ -91,10 +91,8 @@ def add_events_deltas(self, df, col_name='delta', inplace=True):
         if not inplace:
             df = df.copy()
 
-        df.loc[df.index[:-1], col_name] = df.index.values[1:] - df.index.values[:-1]
-        # Fix the last event, which will have a NaN duration
-        # Set duration to trace_end - last_event
-        df.loc[df.index[-1], col_name] = self.end - df.index[-1]
+        df[col_name] = df.index
+        df[col_name] = df[col_name].diff()
```
valschneider: AFAICT to maintain current behaviour this should be 
```suggestion
        df[col_name] = df[col_name].diff().shift(-1)
```

And you need to keep the `df.loc[df.index[-1], col_name]` fixup

lisa/trace.py
```diff
@@ -91,10 +91,8 @@ def add_events_deltas(self, df, col_name='delta', inplace=True):
         if not inplace:
             df = df.copy()
 
-        df.loc[df.index[:-1], col_name] = df.index.values[1:] - df.index.values[:-1]
-        # Fix the last event, which will have a NaN duration
-        # Set duration to trace_end - last_event
-        df.loc[df.index[-1], col_name] = self.end - df.index[-1]
+        df[col_name] = df.index
+        df[col_name] = df[col_name].diff().fillna(0)
```
derkling: Mmm.. wondering if this could somehow bias some analysis.
Technically for the last even we don't really know how long it lasted.
If we wanna keep it, is not better to set it as the delta between its timestamp and the end of the trace?
Otherwise, I would probably better like to play save and discard it as we do now...

valschneider: We still need the last event fixup - no pandas method can do it for us, because it depends on data not present in the dataframe (the end of the trace).

The `fillna(0)` is not needed, we do need a `shift(-1)` however otherwise we get misaligned deltas:
```
# df["delta"].diff()
Time
0.000009          NaN
0.000059     0.000050
0.000094     0.000035
```

```
# df["delta"].diff().shift(-1)
Time
0.000009     0.000050
0.000059     0.000035
0.000094     NaN
```

qais-yousef:
Okay. I'm afraid I didn't dig deep enough into how this function is used and by the face value of it I don't expect shifts etc. I'd rather drop this and let you guys handle it then :-)

PR #862, opened 2019-04-11
trace: Improve/fix add_events_deltas()

valschneider:
While we're not supposed to get duplicate timestamps (see
trappy.utils::handle_duplicate_index), these seem to still happen and
they throw pandas.DataFrame.loc[] off balance.

Use diff() + shift() instead of handcrafted equivalents. Performance
difference is insignificant, but at least we rely on a standard pandas
API.

### DIFF
--- a/lisa/trace.py
+++ b/lisa/trace.py
@@ -91,7 +91,8 @@ class TraceBase(abc.ABC):
         if not inplace:
             df = df.copy()
 
-        df.loc[df.index[:-1], col_name] = df.index.values[1:] - df.index.values[:-1]
+        df[col_name] = df.index
+        df[col_name] = df[col_name].diff().shift(-1)
         # Fix the last event, which will have a NaN duration
         # Set duration to trace_end - last_event
         df.loc[df.index[-1], col_name] = self.end - df.index[-1]


lisa/trace.py
```diff
@@ -91,7 +91,8 @@ def add_events_deltas(self, df, col_name='delta', inplace=True):
         if not inplace:
             df = df.copy()
 
-        df.loc[df.index[:-1], col_name] = df.index.values[1:] - df.index.values[:-1]
+        df[col_name] = df.index
+        df[col_name] = df[col_name].diff().shift(-1)
```
qais-yousef: So with the shift what you're actually implementing is

`delta[n] = index[n+1] - index[n]`

While from the face value of the function name I'd expect the standard behavior implemented by diff() itself, which is:

`delta[n] = index[n] - index[n-1]`

You're preserving the current behavior so you're probably alright I haven't checked all the call sites to see the current users but at least we should document what this function is returning.

Note if this is used with sched_switch() and I look at the delta of the prev_comms then this is wrong.

In case what I'm saying is not clear consider this event train

```
     |           |             |
     |           |             |
-----+-----------+-------------+----------
   [n-1]        [n]          [n+1]
```

`sched_switch.prev_comm.delta[n] = index[n] - index[n-1]` *is* the right duration period since at n the task *finished* running.

Using the implementation in this function the `sched_switch.prev_comm.delta[n] = index[n+1] - index[n]` is *wrong* since this is the duration of the next running task.

In the same sched_switch event if one looks at sched_switch.next_comm then `sched_switch.next_comm.delta[n] = index[n+1] - index[n]` is the right delta you're implementing indeed.

So that's the ambiguity that's confuses me and to be honest judging by the face value of the function I would have expected it to return the same standard behavior of diff().

So if you're 100% sure the current behavior is OK with all users then I suggest updating the function doc at least or maybe look at extending the function to accept a new `method='{backward|forward}'` arg; naming it something more creative maybe.

valschneider: > `sched_switch.prev_comm.delta[n] = index[n] - index[n-1]` _is_ the right duration period since at n the task _finished_ running.
> 
> Using the implementation in this function the `sched_switch.prev_comm.delta[n] = index[n+1] - index[n]` is _wrong_ since this is the duration of the next running task.
> 
> In the same sched_switch event if one looks at sched_switch.next_comm then `sched_switch.next_comm.delta[n] = index[n+1] - index[n]` is the right delta you're implementing indeed.
> 

You're absolutely right, `sched_switch` is a "special" case in that you get to see `prev` & `next` in the same trace event. The most common scenario however is that you only get `next` (`cpu_frequency`, `cpu_idle`, `clock_set_rate`, `thermal_temperature`, `thermal_power_{cpu, devfreq}_limit`, and many many more), which is what this helper caters to.

> So that's the ambiguity that's confuses me and to be honest judging by the face value of the function I would have expected it to return the same standard behavior of diff().
> 
> So if you're 100% sure the current behavior is OK with all users then I suggest updating the function doc at least or maybe look at extending the function to accept a new `method='{backward|forward}'` arg; naming it something more creative maybe.

I don't think a backward delta makes sense for our uses of it, but if you think you can improve the docstring you're more than welcome to have a go at it.

qais-yousef: ¯\_(ツ)_/¯

PR #863, opened 2019-04-11
trace.py: improve add_events_deltas() docs

qais-yousef:
To remove the ambiguity about what delta it exactly returns.

### DIFF
--- a/lisa/trace.py
+++ b/lisa/trace.py
@@ -68,6 +68,17 @@ class TraceBase(abc.ABC):
         """
         Store the time between each event in a new dataframe column
 
+        This function assumes that at time [n] the event starts and at [n+1]
+        the event stops, so the formula for the returned value is::
+
+                  |        |            |
+                  |        |            |
+                  |        |            |
+            ------+--------+------------+------
+                [n-1]     [n]         [n+1]
+
+            delta[n] = index[n+1] - index[n]
+
         :param df: The DataFrame to operate one
         :type df: pandas.DataFrame
 


derkling:
Yes, in general we are mostly interested in knowing how long a certain condition last, for example:
- for how long a task switching in a CPU will run
- for how long an idle state we entered will be used
- for how long a frequency we have activated will be used

I don't see cases for the opposite unless you are correlating different event classes. However, since this function is generic it's worth to call out the way it works. I'm not sure if make sense to specify in its documentation how to compute something different then what it computes. Thus, I would say, either we add a parameter to compute the two different deltas or we remove those additional doc bits.

qais-yousef:
But I did argue that for task_switch at least this is not strictly true. I'm sure there will be something else that will not want this extra shift and for sure if I was going to use this function I would have been caught out by this behavior.

> I'm not sure if make sense to specify in its documentation how to compute something different then what it computes.

True. But a function that abstracts pd.DataFrame.diff().shift() is an overkill as we have added 10 lines of code to replace 2 lines of code (that probably can be squashed into one too). That's why I wasn't keen on extending it. I'll drop the hint to using diff() anyways.

lisa/trace.py
```diff
@@ -68,6 +68,17 @@ def add_events_deltas(self, df, col_name='delta', inplace=True):
...
+                  |        |            |
+                  |        |            |
+            ------+--------+------------+------
+                [n-1]     [n]         [n+1]
+
+        This function assumes that at time [n] the event starts and at [n+1]
+        the event stops, so the formula for the returned value is:
+
+            delta[n] = index[n+1] - index[n]
+
```
valschneider: Sphinx is getting confused by the diagram because it's not in an example field. I'd do something like this:

```
This function [...] the returned value is ::

  <diagram here>

  delta[n] = index[n+1] - index[n]
```

qais-yousef: Thanks for the tip! I was trying to figure this out

PR #864, opened 2019-04-12
Fastmodel fixes

qais-yousef:
The two commit fix issues I've seen when running on fastmodels. The common cause is duplicated timestamps which seem to happen more often on fastmodels. I think this is an artefact of the simulated time in fastmodels so probably the fixes are valid but it'd be good to have more critical eyes at this.

### DIFF
--- a/external/bart/bart/sched/functions.py
+++ b/external/bart/bart/sched/functions.py
@@ -139,6 +139,25 @@ def sanitize_asymmetry(series, window=None):
                 if window:
                     series.index.values[-1] = window[1]
 
+            # Remove repeated entries - which could happen if a task switch in
+            # then immediately switches out; ie: time stamp is exactly the same
+            n = 0
+            next = n + 1
+            while next < len(series):
+                if not series.values[n]:
+                    n = next
+                    next = next + 1
+                    continue
+
+                while not series.values[next]:
+                    next = next + 1
+
+                if series.values[n] == series.values[next]:
+                    series = series.drop(series.index[next])
+                else:
+                    n = next
+                    next = next + 1
+
         # No point if the series just has one value and
         # one event. We do not have sufficient data points
         # for any calculation. We should Ideally never reach
--- a/lisa/tests/scheduler/eas_behaviour.py
+++ b/lisa/tests/scheduler/eas_behaviour.py
@@ -139,6 +139,7 @@ class EASBehaviour(RTATestBundle):
 
         df = self.trace.ftrace.sched_switch.data_frame[['next_comm', '__cpu']]
         df = df[df['next_comm'].isin(tasks)]
+        df = df[~df.index.duplicated()]
         df = df.pivot(index=df.index, columns='next_comm').fillna(method='ffill')
         cpu_df = df['__cpu']
         # Drop consecutive duplicates


external/bart/bart/sched/functions.py
```diff
@@ -139,6 +139,25 @@ def sanitize_asymmetry(series, window=None):
                 if window:
                     series.index.values[-1] = window[1]
 
+            # Remove repeated entries - which could happen if a task switch in
+            # then immediately switches out; ie: time stamp is exactly the same
+            n = 0
+            next = n + 1
```
douglas-raillard-arm: Could we rename `next` to another name (or `next_` if we lack inspiration) ? `next()` is a builtin function so it's better to avoid overriding these names.

external/bart/bart/sched/functions.py
```diff
@@ -139,6 +139,25 @@ def sanitize_asymmetry(series, window=None):
...
+            while next < len(series):
+                if not series.values[n]:
+                    n = next
+                    next = next + 1
```
douglas-raillard-arm: can be abbreviated `next += 1`

external/bart/bart/sched/functions.py
```diff
@@ -139,6 +139,25 @@ def sanitize_asymmetry(series, window=None):
...
+                    continue
+
+                while not series.values[next]:
```
douglas-raillard-arm: This may be faster depending on how indexing is actually implemented as it increments without loosing the current position (something like a linked list would benefit a lot from that):
```
for i, val in enumerate(series.values[next:]):
    if val:
        break
next = next + i
```

valschneider:
I haven't had time to go through the details, but how come we get duplicate events? I thought we were already handling them in trappy via https://github.com/ARM-software/trappy/blob/master/trappy/utils.py#L31 ?

qais-yousef:
>I haven't had time to go through the details, but how come we get duplicate events? I thought we were already handling them in trappy via https://github.com/ARM-software/trappy/blob/master/trappy/utils.py#L31 ?

Interesting. Are you suggesting it should be already used or I should use it instead?

It seems it's exported as utils in trappy and in our code only analysis/idle.py and wa_results_collector.py are using it

external/bart/bart/sched/functions.py
qais-yousef: Sure good point

external/bart/bart/sched/functions.py
qais-yousef: Okay I'll try that

qais-yousef:
Following Valentine's recommendation the below seem to fix it indeed

```
diff --git a/external/trappy/trappy/base.py b/external/trappy/trappy/base.py                                                                                                                                                                            
index 4b05de98..d9205867 100644
--- a/external/trappy/trappy/base.py
+++ b/external/trappy/trappy/base.py
@@ -29,6 +29,7 @@ import warnings
 from resource import getrusage, RUSAGE_SELF
 
 from trappy.exception import TrappyParseError
+from trappy.utils import handle_duplicate_index
 
 def _get_free_memory_kb():
     try:
@@ -279,6 +280,7 @@ class Base(object):
 
         time_idx = pd.Index(self.time_array, name="Time")
         self.data_frame = pd.DataFrame(self.generate_parsed_data(), index=time_idx)
+        self.data_frame = handle_duplicate_index(self.data_frame)
 
         self.time_array = []
         self.line_array = []
```

qais-yousef:
The fix to trappy in https://github.com/ARM-software/trappy/pull/292 was accepted so we no longer need this. Closing.

PR #875, opened 2019-04-15
lisa.target: Update the cricital tasks list

douglas-raillard-arm:
Since all commands part of the systemd package are named
systemd-<something>, we need a more refined pattern in order to properly
freeze systemd-journald process for example.

This commit is ported from LISA legacy:
263dec9685b79351ff16020130fc67e99c82e166


douglas-raillard-arm:
indeed

PR #878, opened 2019-04-16
lisa.regression: Add Regression.fix_validation_min_iter_nr

douglas-raillard-arm:
Compute the number of iterations required to observe the effects of a
fix to a test case, compared to the new failing rate. This gives an
indication on how many iterations are needed to have `exekall compare`
answer the question "is my fix fixing this regression ?", assuming that
you actually fixed it. If `exekall compare` says that there is no
significant change after that number of iterations, chances are that the
fix is only partial (or not fixing anything at all).

### DIFF
--- a/doc/workflows/automated_testing.rst
+++ b/doc/workflows/automated_testing.rst
@@ -109,15 +109,17 @@ iterations, out of which 80 passed and 20 failed.
 
 The output of ``exekall compare`` looks like that:
 
+.. Comparison of 20190222 and 20190412 integration
+
 ::
 
-    testcase                                                           old%   new% delta%      pvalue
-    -------------------------------------------------------------------------------------------------
-    PELTTask:test_load_avg_behaviour                                   3.8%   6.7%   2.9%    9.26e-02
-    ThreeSmallTasks:test_task_placement                                0.0%   1.8%   1.8%    4.13e-03
-    TwoBigTasks:test_slack                                             0.0%   3.1%   3.1%    1.03e-04
-    TwoBigThreeSmall:test_slack                                       82.1%  95.7%  13.6%    1.69e-06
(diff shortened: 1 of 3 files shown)


doc/workflows/automated_testing.rst
```diff
@@ -105,19 +105,35 @@ That would represent an ``old`` test session with 20 iterations of the test, 15
...
+  The total number of results on one column (columns marginal total) is fixed,
+  since a test either has to pass or fail. However, the row marginal totals are
+  not fixed, since the experiment does not constrains the total number of
+  success and total number of failures. This kind of experiment would be best
+  analysed using Barnard's test. That said, Fisher's exact test is just less
+  powerful than Barnard's test, which means its only issue is to be too
+  conservative, i.e. will sometimes fail to spot a failure rate change although
+  there actually was one. Another way to express that is that Fisher's exact
+  test will require more iterations before detecting a failure rate change than
+  strictly required. Barnard's test is unfortunately not widely implement, so
```
valschneider:
```suggestion
  strictly required. Barnard's test is unfortunately not widely implemented, so
```

doc/workflows/automated_testing.rst
```diff
@@ -105,19 +105,35 @@ That would represent an ``old`` test session with 20 iterations of the test, 15
 of which passed and 5 failed. The ``new`` session would have had 100
 iterations, out of which 80 passed and 20 failed.
 
+.. note:: This kind of experiments only fixes some marginal totals and
```
valschneider: Nit: might want a line break or two in there to avoid having a Big Scary Wall of Text.

douglas-raillard-arm: It's now a big scary wall of statistics gibberish :-)

PR #887, opened 2019-04-24
lisa-test: Create latest_results symlink

douglas-raillard-arm:
Update $LISA_HOME/results_latest symlink to point to exekall artifact
folder when running lisa-test.


derkling:
LGTM

PR #890, opened 2019-04-25
Update subtrees

qais-yousef:
Pull in the changes from trappy and wa.


qais-yousef:
Thanks for the reviews. Waiting for some tests to finish and then will merge

PR #893, opened 2019-04-26
lisa.tests.base: Use LRU cache for TestBundle.trace

douglas-raillard-arm:
Avoid storing the parsed trace forever, to avoid consuming too much
memory in processes manipulating a large number of TestBundle. Most of
these kind of scripts will only process a bundle once and then go to the
next one anyway.

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -475,8 +475,11 @@ class RTATestBundle(TestBundle, metaclass=RTATestBundleMeta):
 
         return (rta_start, rta_stop)
 
+    # Use LRU cache instead of memoized, to avoid caching the trace forever, in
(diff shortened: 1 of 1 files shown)


lisa/tests/base.py
```diff
@@ -475,8 +475,11 @@ def trace_window(self, trace):
 
         return (rta_start, rta_stop)
 
+    # Use LRU cache instead of memoized, to avoid caching the trace forever, in
+    # case the thread is manipulating a large number of TestBundles without
+    # deleting them.
     @property
-    @memoized
+    @functools.lru_cache(maxsize=50, typed=True)
```
valschneider: We could probably reduce that further to something like 20/30. You rarely re-execute things on more than that.

PR #894, opened 2019-04-26
doc: Add lisa man page

douglas-raillard-arm:
Available both in the HTML doc and as a man page

```
$ source init_env
$ cd doc && ./pedantic_build
$ xdg-open _build/html/index.html
$ man lisa
```


valschneider:
Travis complains with this:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

I can't say if this change made the test duration much worse or if we've been very close to 10 mins lately...

douglas-raillard-arm:
The doc building was probably stalled indeed, I need to update the PR with the fix

douglas-raillard-arm:
@valschneider Also added env var doc, with new env var:
```
export _DOC_LISA_DEVMODE="By default use internal libraries"
export LISA_DEVMODE=${LISA_DEVMODE:-1}
```
That documentation ends up in a table at the end of the man page, and is also shown on `source init_env`

PR #895, opened 2019-05-02
lisa: Use inspect.getdoc() instead of textwrap.dedent

douglas-raillard-arm:
inspect.getdoc()/cleandoc() are better suited to handle docstrings than
textwrap.dedent(), since they handle properly lack of indentation on the
first line.

### DIFF
--- a/lisa/conf.py
+++ b/lisa/conf.py
@@ -1148,7 +1148,7 @@ class ConfigurableMeta(abc.ABCMeta):
         # Update the docstring by using the configuration help
         docstring = inspect.getdoc(new_cls)
         if docstring:
-            new_cls.__doc__ = textwrap.dedent(docstring).format(
+            new_cls.__doc__ = docstring.format(
                 configurable_params=new_cls._get_rst_param_doc()
             )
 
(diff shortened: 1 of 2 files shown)


lisa/conf.py
```diff
@@ -1148,7 +1148,7 @@ def __new__(metacls, name, bases, dct, **kwargs):
         # Update the docstring by using the configuration help
         docstring = inspect.getdoc(new_cls)
         if docstring:
-            new_cls.__doc__ = textwrap.dedent(docstring).format(
```
douglas-raillard-arm: forgot to remove this one last time I introduced `getdoc()` 2 lines above

lisa/tests/base.py
```diff
@@ -638,7 +638,8 @@ def wrapper(self, *args,
             )
 
             #pylint: disable=no-member
```
valschneider: That warning was for accessing `wrapper.__doc__` which is now one line below. I guess you could remove it, at the time I added it feel to make it feel like we're progressing towards a pylint-checkable codebase...

douglas-raillard-arm: right, if we want to aim at that, we should add a travis check to make sure that new patches don't introduce pylint warnings, and we can fix the existing warnings a bit at a time during these 3 minutes before lunch break when it feels nothing big should be started :-)

PR #897, opened 2019-05-02
tools: lisa-test: Use --share '*Target'

douglas-raillard-arm:
This will share the same lisa.target.Target for all iterations, if more
than one iteration is requested (with -n). That avoids reconnecting to
the target at every iteration, which is likely to be the expected
behaviour for usual use cases.

More advanced users will use exekall directly instead of lisa-test if
sharing the Target between iterations is not desired.


douglas-raillard-arm:
@valschneider  documented this new `lisa-test` behaviour in `man lisa`

PR #898, opened 2019-05-03
doc: source init_env for ReadTheDocs setup

douglas-raillard-arm:
source init_env from conf.py, since it is the only place we can hook
into to execute code for RTD. This allows getting all environment
variable defined as on a regular setup, which is needed to generate
parts of the documentation.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -18,6 +18,8 @@ import re
 import subprocess
 import sys
 import unittest
+import textwrap
+import json
 
 from docutils import nodes
 from sphinx.util.docfields import TypedField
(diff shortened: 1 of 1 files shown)


valschneider:
Don't we need some more tweaking for $LISA_HOME? I get
```
  File "/home/docs/.pyenv/versions/3.7.1/lib/python3.7/subprocess.py", line 481, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'exekall run $LISA_HOME/lisa/tests --rst-list --inject-empty-target-conf' returned non-zero exit status 2.
```
at https://readthedocs.org/projects/my-lisa/builds/9008031/

PR #902, opened 2019-05-03
unchecked @requires_events(check=False)

douglas-raillard-arm:
Allows using requires_events() decorator without enabling the runtime check.
This fixes the infinite loop with the `trace` property and `trace_window()` method exposed in https://github.com/ARM-software/lisa/pull/829


douglas-raillard-arm:
@valschneider updated with a decorator that guards a property against recursion. When that happens, it raises `AttributeError`.

qperret:
> LGTM! @qperret, could you quickly confirm this solves your issue?

I'm late to the party, but FWIW: yes, this cures the issue I was seeing.

PR #903, opened 2019-05-03
[RFC] ipynb: Add tests/noise_analysis.ipynb

douglas-raillard-arm:
This notebook can be used to get an idea of what are the tasks that
will interfere the most with the synthetic tests. It reloads data from
exekall's ValueDB, which can be retrieved from its artifact dir or using
bisector report -oexport-db.


valschneider:
I think this all looks fine. One question perhaps is: do we want to concentrate a `TestMetric` consuming APIs somewhere in `lisa/` to make it more widely available? That can probably be done later tho, and we could just merge this..

douglas-raillard-arm:
It really depends on what you want to do, and most test metrics consumer will probably focus on a single TestBundle, which means the only API they will deal with is exekall's public API:
https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/exekall/public_api.html

PR #904, opened 2019-05-08
AggregatedResultBundle

douglas-raillard-arm:
Introduce AggregatedResultBundle to be used in load_tracking Invariance aggregated tests and #829

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -107,9 +107,18 @@ class ResultBundleBase:
         return self.result is Result.PASSED
 
     def __str__(self):
-        return self.result.name + ': ' + ', '.join(
-                '{}={}'.format(key, val)
-                for key, val in self.metrics.items())
+
+        def format_val(val):
+            # Handle recursive mappings, like metrics of AggregatedResultBundle
+            if isinstance(val, Mapping):
+                return '{' + ', '.join(
+                    '{}={}'.format(key, format_val(val))
+                    for key, val in val.items()
+                ) + '}'
+            else:
+                return str(val)
+
+        return self.result.name + ': ' + format_val(self.metrics)
 
     def add_metric(self, name, data, units=None):
         """
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -35,7 +35,10 @@ from lisa.analysis.tasks import TasksAnalysis
 from lisa.trace import Trace, requires_events
 from lisa.wlgen.rta import RTA
 
-from lisa.utils import Serializable, memoized, ArtifactPath, non_recursive_property
+from lisa.utils import (
+    Serializable, memoized, ArtifactPath, non_recursive_property,
+    LayeredMapping
+)
 from lisa.trace import FtraceCollector, FtraceConf
 
 class TestMetric:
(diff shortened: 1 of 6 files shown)


douglas-raillard-arm:
@ionela-voinescu Following our discussion:
* Updated AggregatedResultBundle to optionally force the `Result` to use, to allow using different result aggregation functions.
* Made use of it in `Invariance.test_freq_invariance`, which is now properly handling `Result.UNDECIDED`. It now returns an `AggregatedResultBundle`, itself wrapping one `AggregatedResultBundle` for each CPU. The inner per-CPU `AggregatedResultBundle` are regular ones, but the top-level one only requires one inner bundle to pass, instead of all of them.

lisa/tests/base.py
```diff
@@ -131,28 +166,94 @@ def from_bool(cls, cond, *args, **kwargs):
...
+            return combinator(
+                res_bundle.result is result
+                for res_bundle in self.result_bundles
+            )
+
+        if predicate(all, Result.UNDECIDED):
+            return Result.UNDECIDED
+        elif predicate(any, Result.FAILED):
+            return Result.FAILED
+        elif predicate(any, Result.PASSED):
```
ionela-voinescu: Shouldn't this be:
`elif predicate(all, Result.PASSED):`?

It makes more sense to me as:
```
if predicate(all, Result.PASSED):
    return Result.PASSED
 elif predicate(any, Result.FAILED):
    return Result.FAILED
 else:
    return Result.UNDECIDED
```

douglas-raillard-arm: The basic semantic of UNDECIDED is that it should be ignored. That means that a mix of PASSED and UNDECIDED should be interpreted as PASSED. The only time UNDECIDED "wins" is when there are only UNDECIDED, so there is no relevant data to be based on. 

In practice, we might want to be more cautious for some tests and treat a mix of passed and undecided as undecided. That said, the more results you aggregate, the higher the probability of having at least one decided. We don't want to end up with a test that returns "undecided" all the time.  This is one of the main argument against such aggregated tests, unless they are just heuristics on top of results that can be visualised in a different way (like test_freq_invariance and test_cpu_invariance).  If they can be visualised separately, we can keep an eye on the rate of undecided and make sure it stays within reasonable bounds.

We can introduce a parameter to the constructor to switch to one semantic or to another, but I'd rather avoid introducing too many tunables only supposed to enable writing tests that are to some degree broken by design.

ionela-voinescu: Okay, fair enough. For the moment my purpose was just to make sure that the logic in this PR is not broken. But it is intentional so it sounds good to me.

lisa/tests/scheduler/load_tracking.py
```diff
@@ -587,35 +570,49 @@ def test_freq_invariance(self) -> ResultBundle:
...
-        for cpu, submetric in metrics.items():
-            res.add_metric(cpu, submetric)
-
-        return res
+        group_result_bundles = [
+            make_group_bundle(cpu, item_group)
+            for cpu, item_group in groupby(self.invariance_items, key=lambda x: x.cpu)
+        ]
 
+        # The combination differs from the AggregatedResultBundle default one:
```
valschneider: I know this isn't part of this change, but I didn't realise/forgot this test behaved this way. Why don't we use the default combination method?

douglas-raillard-arm: The logic of the test itself makes it pass as long as at least one CPU is frequency invariant. If the invariance is broken on one CPU but not the other one, it's probably a cpu invariance issue. I say "probably" because at the end of the day, there is no way to tell apart a freq invariance issue from a cpu invariance issue with what we can observe. Both tests are just heuristics to guide the debugging, but ultimately cannot be relied upon. The only one true source of information are the (cpu, freq) tests, which should be looked at one by one if something goes wrong to figure out what is broken.

douglas-raillard-arm:
note to self: ready to be merged once the sched_android tests are validated using `AggregatedResultBundle`

PR #906, opened 2019-05-09
lisa.exekall_customize: Error for empty TestBundle.ftrace_conf['events']

douglas-raillard-arm:
There is no case where an FtraceCollector is requested to exekall
by a TestBundle subclass without providing any events to collect.
Therefore, it currently indicates a design issue, that would otherwise
trigger a cryptic error in trace-cmd itself.

Such design issue can arise for aggregation+item class pairs, where
exekall only knows about the aggregation class, but this class is an
empty shell and does not provide a useful FtraceConf. In that case, the
link must be manually made by adding ftrace_conf class attribute equal
to the one from the item class.

Catch the issue when we can display relevant context and raise an
exception.


douglas-raillard-arm:
@qperret This should catch the `ftrace_conf` issue you had yesterday

PR #907, opened 2019-05-09
lisa.analysis.task: Add TaskAnalysis.cpus_of_tasks() 

douglas-raillard-arm:
Allow getting the CPUs on which the given tasks executed. This is useful
to convert the tasks names of the rtapp profile into list of CPUs to
feed to plotting functions that need to know what CPU to look at. Since
the trace is usually already cropped around the execution of rtapp,
this allows spotting the relevant pieces of data.

### DIFF
--- a/lisa/analysis/tasks.py
+++ b/lisa/analysis/tasks.py
@@ -122,6 +122,22 @@ class TasksAnalysis(TraceAnalysisBase):
     def __init__(self, trace):
         super(TasksAnalysis, self).__init__(trace)
 
+    @requires_events('sched_switch')
+    def cpus_of_tasks(self, tasks):
+        """
+        Return the list of CPUs where the ``tasks`` executed.
+
+        :param tasks: Task names or PIDs to look for.
+        :type tasks: list(int or str)
+        """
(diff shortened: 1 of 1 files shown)


lisa/trace.py
```diff
@@ -588,6 +588,19 @@ def show(self):
             return webbrowser.open(self.ftrace.trace_path)
         self.get_logger().warning('No trace data available')
 
+    def cpus_of_tasks(self, tasks):
+        """
+        Return the list of CPUs where the ``tasks`` executed.
+
+        :param tasks: Task names or PIDs to look for.
+        :type tasks: int or str
```
valschneider: That should be a list, no?
```suggestion
        :type tasks: list(int or str)
```

douglas-raillard-arm:
@valschneider Maybe that method should live in the TaskAnlaysis instead, isn't it ?

valschneider:
> @valschneider Maybe that method should live in the TaskAnlaysis instead, isn't it ?

Right, we already have the PID stuff in `Trace` but this should go in `TaskAnalysis`.

PR #915, opened 2019-05-13
kernelshark-like activation plot

douglas-raillard-arm:
With `overlay=True`, it will adapt to data already plotted on the given ``axis``, with 1/4 of the height and alpha=0.5 . This allows quick overlaying over an existing plot, to see how rtapp activations are impacting whatever signal is being looked at:
foo

### DIFF
--- a/lisa/analysis/tasks.py
+++ b/lisa/analysis/tasks.py
@@ -455,6 +455,51 @@ class TasksAnalysis(TraceAnalysisBase):
 
         return res_df[:count]
 
+    @df_task_states.used_events
+    def df_task_activation(self, task, cpu=None, active_value=1, sleep_value=0):
+        """
+        DataFrame of a task's active time on a given CPU
+
+        :param task: the task to report activations of
+        :type task: int or str
+
+        :param cpu: the CPUs to look at. If ``None``, all CPUs will be used.
+        :type task: int or None
+
+        :param active_value: the value to use in the series when task is
+            active.
+        :type task: float
+
+        :param active_value: the value to use in the series when task is
+            sleeping.
+        :type task: float
+
+        :returns: a :class:`pandas.DataFrame` with:
+
+          * A timestamp as index
+          * A ``active`` column, containing ``active_value`` when the task is
+            not sleeping, ``sleep_value`` otherwise.
+        """
+
+        df = self.df_task_states(task)
+
+        def f(state):
+            if state == TaskState.TASK_ACTIVE:
+                return active_value
+            elif state == TaskState.TASK_INTERRUPTIBLE:
+                return sleep_value
+
+        if cpu is not None:
+            df = df[df['cpu'] == cpu]
+
+        curr_state = df.curr_state
+        active_series = curr_state[
+            (curr_state == TaskState.TASK_ACTIVE) |
+            (curr_state == TaskState.TASK_INTERRUPTIBLE)
+        ].map(f)
+
+        return active_series.to_frame('active')
+
 ###############################################################################
 # Plotting Methods
 ###############################################################################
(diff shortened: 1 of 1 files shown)


derkling:
I see you attempt to not clutter the data by using an alpha channel blending, but I still find this format a bit confusing. I would probably prefer a standalone plot right below the plot of the signals which share the same x axis. You can still get it if you don't pass in an existing axis, but that will be more work for the user.

douglas-raillard-arm:
> I would probably prefer a standalone plot right below the plot
`overlay=False` (which is the default) and plot on another axis/figure, that's completely opt-in. At least in my use-case, there is almost no value in having the plot if that needs to be in a separate one, I already know what my activations will look like. What I need is to be able to see the precise ordering of activations with utilization samples and frequency changes, which is very hard to do if not on the same figure. This seems to follow what `LoadTrackingAnalysis.plot_task_signals()` does, at it always uses `alpha=0.4` (may be worth making that optional now that I think about it).

Ideally, I would need a vertical bar following my cursor or at least coordinates, but last time I tried, `%matplotlib notebook` was not working on jupyter lab because it seems to rely on some javascript magic that works differently now.

The "y autoscale" feature is mostly there to be able to see what is going on in just a few seconds, to make sure one is looking at the right piece of the trace, without having to create new axes and add some boilerplate for this quick check.

> which share the same x axis. You can still get it if you don't pass in an existing axis,

Not sure to get it: this function will only ever plot the task activations, so even if we call `setup_plot(nrows=2)` when the user does not give an axis, we will end up with:
* one subplot with task activations
* one empty subplot, because we have nothing else to show on this one

> but that will be more work for the user.
This seems quite straightforward:
```
fig, axes = TraceAnalysisBase.setup_plot(nrows=2)                
trace.analysis.tasks.plot_task_activation('foo', axis=axes[0])
trace.analysis.frequency.plot_cpu_frequencies(cpu=0, axis=axes[1])
```

douglas-raillard-arm:
EDIT: looks like it still possible to get the matplotlib widget with jupyter lab, I'll create another PR to enable it.

lisa/analysis/tasks.py
```diff
@@ -423,6 +423,51 @@ def df_task_total_residency(self, task):
...
+            elif state == TaskState.TASK_INTERRUPTIBLE:
+                return sleep_value
+
+        if cpu is not None:
+            df = df[df['cpu'] == cpu]
+
+        curr_state = df.curr_state
+        active_series = curr_state[
+              (curr_state == TaskState.TASK_ACTIVE)
+            | (curr_state == TaskState.TASK_INTERRUPTIBLE)
```
valschneider: Nit: I'd rather have the operator on an end of line.

lisa/analysis/tasks.py
```diff
@@ -632,4 +677,50 @@ def plot_tasks_forks_heatmap(self, xbins=100, colormap=None, filepath=None):
...
+
+        :param alpha: transparency level of the plot.
+        :type task: float
+
+        :param overlay: If ``True``, assumes that ``axis`` already contains a
+            plot, so it will adjust automatically the height and transparency of
+            the plot to blend with existing data.
+        :type task: bool
+
+        :param kwargs: Keyword arguments passed to :meth:`df_task_activation`.
```
valschneider: Nit: I've used `:Keyword arguments:` elsewhere, but I think some `kwargs` have slipped here and there anyway...

douglas-raillard-arm: yes, I've never used that definition list, I'll batch convert all the `:param kwargs:` in a separate PR.

douglas-raillard-arm:
@valschneider updated with your comments.
I also re-purposed kwargs for `do_plot`, to align with #916

PR #917, opened 2019-05-14
Rework Analysis plot infrastructure

douglas-raillard-arm:
* Allow plot functions to return an HTML and reStructuredText output with `output="html"` and `output="rst"`. The output does not rely on any external resource (base64 image in data URL).
* Introduce `lisa-plot` command that allows generating analysis plots from a trace.dat file.

Each plot gets a short description (first line of docstring), a link to RTD doc and the value of relevant parameters.

Sample HTML of a plot method: https://pste.eu/p/QT8z.html


lisa/analysis/frequency.py
```diff
@@ -286,8 +286,9 @@ def mhz(x, pos):
         pl.savefig(figname, bbox_inches='tight')
 
 
+    @TraceAnalysisBase.plot_method(in_report=False)
     @requires_events('cpu_frequency')
-    def plot_cpu_frequencies(self, cpu, average=True, **kwargs):
+    def plot_cpu_frequencies(self, cpu, axis, local_fig, average=True):
```
valschneider: If I got this right, we're exposing `axis` and `local_fig` for the `plot_method` decorator, but it won't actually be required for the end user (otherwise it doesn't make sense to me)?

I'm not too hot on that change. At least with the `plotter()` way, we hide those inside the method and it's not exposed in the public signature. Even if we then perform surgery on that signature, the code still becomes more complicated to read.

Also, have both this way and the `return self.do_plot()` way adds to that complexity. I'd like to keep a single way to do plots.

douglas-raillard-arm: > If I got this right, we're exposing axis and local_fig for the plot_method decorator, but it won't actually be required for the end use

It's more constrained than that: `local_fig` cannot be user-specified. What you see now as the method is what used to be the plotter() function. There is nothing more or less exposed to the user than before.

The idea is that 99% of the plot functions only specify a basic plotter and call do_plot() without any other parameter than kwargs. With that decorator, it becomes possible to implement the function directly as a plotter, so the decorator takes care of calling do_plot().

This removes some bloat in the code and allows surgery on the signature+docstring, which ultimately gives a better documentation than the endless `.. sealso: do_plot()` sprinkled in all the docstrings. It should also be noted that although `do_plot()` is purely an implementation detail, it still ends up right in front of the user in the doc, which is not great.

#947 also makes use of that to provide a very easy path from notebook custom code to upstream analysis code without the need of deep refactoring.

> At least with the plotter() way, we hide those inside the method and it's not exposed in the public signature

That is still the case with this decorator. `local_fig` is not exposed to the user. `axis` is, the same way it used to be.

> Also, have both this way and the return self.do_plot() way adds to that complexity

I would favour making basic plot as simple as possible, and keep the advanced APIs for the few plots that actually need it. It's much easier to refactor the code if there is only a handful of odd cases, with 90% of the functions being programmatically taken care of by the decorator.

douglas-raillard-arm: With the decorator, the rendered documentation looks like that:

image

PR #921, opened 2019-05-15
Make RTATestBundle.trace useable for by analysis methods

douglas-raillard-arm:
Load all the events that are useful to analysis classes from the trace, so that they can be manipulated without any manual settings.

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -605,13 +605,25 @@ class RTATestBundle(TestBundle, metaclass=RTATestBundleMeta):
         """
         :returns: a :class:`lisa.trace.TraceView`
 
+        All events specified in ``ftrace_conf`` are parsed from the trace,
+        so it is suitable for direct use in methods.
+
         Having the trace as a property lets us defer the loading of the actual
         trace to when it is first used. Also, this prevents it from being
         serialized when calling :meth:`lisa.utils.Serializable.to_path` and
         allows updating the underlying path before it is actually loaded to
         match a different folder structure.
         """
-        trace = Trace(self.trace_path, self.plat_info, events=self.ftrace_conf["events"])
+        return self.get_trace(events=self.ftrace_conf["events"])
+
+    def get_trace(self, **kwargs):
+        """
+        :returns: a :class:`lisa.trace.TraceView` cropped to fit the ``rt-app``
+            tasks.
+
+        :Keyword arguments: forwarded to :class:`lisa.trace.Trace`.
+        """
+        trace = Trace(self.trace_path, self.plat_info, **kwargs)
         return trace.get_view(self.trace_window(trace))
 
     @property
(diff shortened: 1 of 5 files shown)


lisa/tests/base.py
```diff
@@ -495,7 +496,13 @@ def trace(self):
         match a different folder structure.
         """
         path = os.path.join(self.res_dir, self.TRACE_PATH)
-        trace = Trace(path, self.plat_info, events=self.ftrace_conf["events"])
+        # Merge all the events that we expect to see, plus all the ones that
+        # can be useful to any analysis method
+        events = sorted(
```
valschneider: I'm not sure I'm okay with having this in the `trace` property. IIUC, this means that when running test methods, we're going to load every single trace event that LISA can use + the ones from the bundle's conf. That's a lot of unused data being loaded.

Besides, unless you modify something (IIRC exekall can merge ftrace confs for you?), you won't get all of those events in the trace. What's your use-case here?

douglas-raillard-arm: Reloading serialized TestBundle in a notebook currently does not let you plot anything, even if the right events have been collected. For example, if you want to get the frequency graph from a TestBundle generated during the integration, you cannot, although the events have been collected (extra events in ftrace-conf key). 

> That's a lot of unused data being loaded
If the events are present, one could assume it's because the user wants to make use of it. By default they won't be there so it will not make a big difference. We also limit to 30 memoized traces, so it should not grow out of control too quickly. If that's still an issue, we can introduce a new `full_trace` attribute that loads all the "useable" events instead of modifying `trace`.

valschneider: Right, I had a chat with @derkling last week and he faced a similar issue. In his case it was fixable by calling `Trace(<bundle's trace.dat>, events=events)`, which is not ideal but usable at least.

What I'm not fond of is loading all of the extra events (specified in`ftrace-conf`) when running tests, because it is useless then and we know it. In that regards I'd be more in favour of a `full_trace` attribute but it stills feels icky.

douglas-raillard-arm: Yes, maybe we could expose a `get_trace()` method that passes `kwargs` to `Trace.__init__`. In addition to that, we could auto-load all events that can be used by analysis classes when `events` list is empty. This would give:
```
tb = TestBundle(...)
trace = tb.get_trace()
trace.analysis.foo.plot_bar(...)
```

That said, the real fix should be to make Trappy able to lazily load event's dataframes. That can only be done efficiently if the initial parsing gives an intermediate format that allows fast loading of individual events. Until then, we will face that issue again and again.

> `full_trace` attribute but it stills feels icky

It definitely is, that `trace` and `full_trace` split is really artificial and bakes a workaround into a core API, which is a quite nasty thing to do.

douglas-raillard-arm:
Updated with:
* Loading a `Trace` without specifying events will try to load all events used in analysis classes.
* `RTATestBundle.get_trace()` allows getting a trace with custom parameters passed to `Trace()` constructor.

PR #922, opened 2019-05-16
wlgen: sysbench: Expose --num-threads

qperret:
It is possible to specify the number of threads of sysbench using a
command line option. As this is often useful, expose it as an argument
of the run method(), like --max-time for example.

### DIFF
--- a/lisa/wlgen/sysbench.py
+++ b/lisa/wlgen/sysbench.py
@@ -61,7 +61,8 @@ class Sysbench(Workload):
 
 
     def run(self, cpus=None, cgroup=None, background=False, as_root=False,
-            test="cpu", max_duration_s=None, max_requests=None, **kwargs):
+            test="cpu", max_duration_s=None, max_requests=None,
+            num_threads=None, **kwargs):
         """
         Execute the workload on the configured target.
 
@@ -86,6 +87,9 @@ class Sysbench(Workload):
         :param: max_requests: Maximum number of event requests
         :type max_requests: int
 
+        :param: num_threads: Number of threads
+        :type num_threads: int
+
         :Keyword arguments: Additionnal arguments required by the specific
           test (run ``sysbench --test=<test> help``). Due to Python limitations,
           use ``_`` instead of ``-`` for argument naming.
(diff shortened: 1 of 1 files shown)


lisa/wlgen/sysbench.py
```diff
@@ -100,6 +104,9 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False,
         if max_requests is not None:
             command = "{} --max-requests={}".format(command, max_requests)
 
+        if num_threads is not None:
+            command = "{} --num-threads={}".format(command, num_threads)
```
douglas-raillard-arm: That may be the right time to clean up the surrounding code: command lines need to be built using `shlex.quote()`:
```
arg_list = ['foo', 'bar']
# modifications of arg_list ...
command  = ' '.join(shlex.quote(arg) for arg in arg_list)
```
https://docs.python.org/3/library/shlex.html#shlex.quote

Adding some helpers in lisa/utils.py may be worth it, since devlib always requires a pre-built command line to be used by a shell.

note: In usual code, that last step is avoided since that's the default input format of all `subprocess` functions for various reasons:
* you never know what shell is going to be used, what globbing it will recognize and so on.
* it's as bad as building SQL queries from untrusted data without escaping. Even when not being attacked, this can lead to subtle API breakages when the issue needs to be fixed for functional reasons (devlib root password is now escaped, but used to not be, which means that people that used to pre-escape the password to work around devlib like `r'\*foo' will now need to use '*foo').
https://docs.python.org/3/library/subprocess.html#security-considerations

qperret: Can we just keep things consistent with the existing code rather than refactoring the world again ? I'm honestly not too keen on doing this clean-up. These parameters are integers so there is no issue there. And if they're not, the user is a moron and there is nothing we can do about it, his system deserves to break.

douglas-raillard-arm: The current policy so far has been to cleanup/fix things as we modify them, so at one point we reach a clean state ... I certainly don't want any proliferation of code that follows documented bad practices, that gets copy pasted pretty much everywhere.

That is especially true for such a small functions, where the refactoring has zero impact on any other code, so it's cheap both to do it and review (you don't need to understand anything else about LISA to do it, other than Python APIs that are expected to be used).

qperret: I'm not even convinced this really is a clean-up. The reasons to use shlex don't apply here IMO, so we're just making the code more complex than necessary IMO.

qperret:
Actually let's forget about this. Using `**kwargs` should work just fine so exposing the parameter doesn't buy us that much. So let's not waste time arguing.

douglas-raillard-arm:
Cleanup implemented here in ~5min
https://github.com/ARM-software/lisa/pull/923

PR #925, opened 2019-05-24
tests: load_tracking: Decorate test methods with check_noisy_tasks()

valschneider:
Load tracking tests rely on having a quiet system, so it makes for
them to check for noisy tasks.

I wanted to validate this on an integration run, but I can't get exekall to re-run on the computed values:
```
bisector report TC2.report.yml.gz -oexport-db=VALUE_DB.pickle.xz
exekall run /data/work/lisa/lisa/tests/ --load-db VALUE_DB.pickle.xz --load-type '*TestBundle' -s '*util_task*'
[EXEKALL][2019-05-24 16:55:10,455] INFO  Nothing to do, check --help while passing some python sources to get the full help.
```

I also trying without any `select` statement and got the same output :(

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -362,6 +362,7 @@ class InvarianceItem(LoadTrackingBase):
         return bundle
 
     @_test_signal.used_events
+    @RTATestBundle.check_noisy_tasks(noise_threshold_pct=1)
     def test_task_util_avg(self, allowed_error_pct=15) -> ResultBundle:
         """
         Test that the mean of the util_avg signal matched the expected value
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
FTR as we discussed offline, `check-test-fix.py` is the scrip to use [1]. The reason is that the DB embedded in the report is minimal and contains only `ResultBundle` for size and loading speed reasons, so passing that to `--load-db` will only show tests pieces that can be computed from a `ResultBundle` (i.e. nothing since it is the "terminal" object in our tests "pipelines").

Note: it is possible to instruct bisector to not prune the DB that are embedded in the report using the `prune-db: false` (YAML) or `-oprune-db=no` (CLI) option so that your procedure works, but that will not scale to hundreds of iterations [2].

[1] https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/workflows/automated_testing.html#fixing-regressions
[2] https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/bisector/man/man.html#bisector-step-help

PR #928, opened 2019-06-03
lisa.trace: Add DmesgCollector 

douglas-raillard-arm:
Make sure we install a know dmesg binary.


lisa/trace.py
```diff
@@ -1530,4 +1530,11 @@ def from_user_conf(cls, target, base_conf=None, user_conf=None, merged_src='merg
         )
         return cls.from_conf(target, conf)
 
+
+class DmesgCollector(devlib.DmesgCollector):
+    def __init__(self, target, *args, **kwargs):
```
valschneider: Would we want to expose the devlib parameters here, like we've done for the `FtraceCollector`?

douglas-raillard-arm: The reason it's done for `FtraceCollector` is because it is tied to a configuration class. Since it's not the case for DmesgCollector so far, it's not really necessary. I should add docstring though.

PR #934, opened 2019-06-04
Vagrantfile: run the Travis tests after setup

douglas-raillard-arm:
This allows using `vagrant up` command to validate a LISA change, including full installation from a bare Ubuntu system.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -22,7 +22,7 @@ Vagrant.configure(2) do |config|
 
     cd /home/vagrant/lisa
     # Install required packages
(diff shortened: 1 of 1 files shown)


valschneider:
I can't really remember why I bothered with the vagrant triggers thing. I think having a trigger on "up" would mean running the tests whenever we restart the VM, but perhaps just running them on VM creating is [good enough](good enough).

douglas-raillard-arm:
Having it run once only seems a reasonable way of validating the installation before it is used the first time. Running it everytime the VM is booted seems unnecessary and would not trigger when the user pulls a new version anyway.

PR #936, opened 2019-06-05
lisa-install: Fix typo in install-nbextensions

balsini:
An extra line-return escape character was preventing lisa-install to
complete.

Signed-off-by: Alessio Balsini <balsini@android.com>


douglas-raillard-arm:
Ah, we really need Travis back on, sorry for the breakage :(

PR #944, opened 2019-06-07
lisa: platinfo: Add kernel/cmdline key

douglas-raillard-arm:
Record the kernel's cmdline so we have it in the logs and it becomes
available to other helpers to know what to expect from dmesg output.

### DIFF
--- a/lisa/platforms/platinfo.py
+++ b/lisa/platforms/platinfo.py
@@ -58,6 +58,7 @@ class PlatformInfo(MultiSrcConf, HideExekallID):
         LevelKeyDesc('kernel', 'Kernel-related information', (
             KeyDesc('version', '', [KernelVersion]),
             KernelConfigKeyDesc('config', '', [TypedKernelConfig]),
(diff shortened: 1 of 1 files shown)


valschneider:
Hm actually, would it make sense to create a `KernelCommandLine` class for this?
Similar to the `KernelConfig` one, we could more easily search for specific options.

Dumb example, but say I want to know if `acpi=force` was passed, it'd be nice if I could just search for an "acpi" key and find its value. What do you think? Maybe something just based on `shlex.split()`?

douglas-raillard-arm:
I was thinking about that, we can probably make a subclass of str with extra methods to lookup options. `shlex.split` should work, although it probably parses a format more complex than the kernel CLI. Do you know of any formalized document about the kernel CLI format ?

valschneider:
I can only remember https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html, dunno if there's another source.

douglas-raillard-arm:
let's close the PR for now until we actually directly need that information. At this point we should have more information on how we want to handle it.

PR #948, opened 2019-06-11
lisa-buildroot-update-kernel-config: Fix call from arbitrary folder

douglas-raillard-arm:
cd to the kernel tree before calling merge_config.sh


douglas-raillard-arm:
@deggeman Can you confirm this solves the issue you had with `lisa-buildroot-update-kernel-config` yesterday ?

PR #950, opened 2019-06-12
Allow plotting utilest signals

douglas-raillard-arm:
* Add may_use_events() decorator to supplement requires_events() and requires_one_event_of()
* Add `LoadTrackingAnalysis.df_(cpus|task)_signal` to replace `df_(cpus|task)_signals()`
* Allow plotting `util_est_enqueued` signals in `LoadTrackingAnalysis.plot_(cpus|task)_signals`
* Allow selecting signals to plot in `LoadTrackingAnalysis.plot_cpus_signals` to mirror the `task` variant of the method

EDIT:
* Add `utils.deprecate()` decorator and treat deprecation as errors in `lisa`  and warnings in `__main__`

### DIFF
--- a/lisa/analysis/load_tracking.py
+++ b/lisa/analysis/load_tracking.py
@@ -65,7 +65,7 @@ class LoadTrackingAnalysis(TraceAnalysisBase):
     def _df_uniformized_signal(self, event):
         df = self.trace.df_events(event)
 
-        df = df.rename(columns=self._columns_renaming(event))
+        df = df.rename(columns=self._columns_renaming(event), copy=True)
 
         if event == 'sched_load_se':
             df = df[df.path == "(null)"]
@@ -74,7 +74,7 @@ class LoadTrackingAnalysis(TraceAnalysisBase):
             df = df[df.path == "/"]
 
         to_drop = self._columns_to_drop(event)
-        df = df[[col for col in df.columns if col not in to_drop]]
+        df.drop(columns=to_drop, inplace=True)
 
         return df
 
--- a/lisa/analysis/load_tracking.py
+++ b/lisa/analysis/load_tracking.py
@@ -48,6 +48,12 @@ class LoadTrackingAnalysis(TraceAnalysisBase):
                 "load_avg" : "load"
             }
 
+        if event == 'sched_util_est_task':
+            return {
+                'est_enqueued': 'util_est_enqueued',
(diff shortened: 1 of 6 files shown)


lisa/analysis/load_tracking.py
```diff
@@ -109,13 +123,26 @@ def df_tasks_signals(self):
 
           * A ``util`` column (the average utilization of a task at time t)
           * A ``load`` column (the average load of a task at time t)
+          * A ``util_est_enqueued`` column (the utilest of a task at time t)
```
derkling: Since we already use shorter names (e.g. `util` instead of `util_avg`) perhaps we can also `s/util_est_enqueued/enqueued/`

... and add also a column for `ewma` reporesenting `util_est_ewma`?

lisa/analysis/load_tracking.py
```diff
@@ -211,8 +214,8 @@ def plotter(axes, local_fig):
                 axis.set_title('CPU{}'.format(cpu))
                 df = cpus_df[cpus_df["__cpu"] == cpu]
 
-                df[['util']].plot(ax=axis, drawstyle='steps-post', alpha=0.4)
-                df[['load']].plot(ax=axis, drawstyle='steps-post', alpha=0.4)
+                for signal in signals:
+                    df[[signal]].plot(ax=axis, drawstyle='steps-post', alpha=0.4)
```
derkling: isn't the above just:
```
df[[signals]].plot(ax=axis, drawstyle='steps-post', alpha=0.4)
```

i.e. `signals` passed to `df` directly

lisa/analysis/load_tracking.py
douglas-raillard-arm: I'm not quite sure about the `enqueued` rename, as it does not even reflect anymore the kind of thing the signal is (utilization). Maybe `enqueued_util` ?

I'll add `util_est_ewma` as well

lisa/analysis/load_tracking.py
douglas-raillard-arm: I can give it a try

lisa/analysis/load_tracking.py
douglas-raillard-arm: Actually there is nothing more to do to support `util_est_ewma`, since it's just a field of the util est event, so it works out of the box.

PR #951, opened 2019-06-12
misc fixes

douglas-raillard-arm:
* Fix events checker decorator unwrapping optimization
* Fix the bbox of save_plot()
* Add update_wrapper_doc() to factor out the docstring fixups wanted when writing "invasive" decorators

### DIFF
--- a/lisa/analysis/base.py
+++ b/lisa/analysis/base.py
@@ -131,7 +131,7 @@ class AnalysisHelpers(Loggable):
             guessed_format = mime_type.split('/')[1].split('.', 1)[-1].split('+')[0]
             img_format = img_format or guessed_format
 
-        figure.savefig(filepath, format=img_format)
+        figure.savefig(filepath, format=img_format, bbox_inches='tight')
 
     def save_plot(self, figure, filepath=None, img_format=None):
         """
--- a/lisa/utils.py
+++ b/lisa/utils.py
@@ -824,4 +824,90 @@ class LayeredMapping(MutableMapping):
             top=copy.copy(self.top),
         )
 
+
+def update_wrapper_doc(func, added_by=None, description=None, remove_params=None):
+    """
+    Equivalent to :func:`functools.wraps` that updates the signature by taking
+    into account the wrapper's extra parameters and the given description. Note
+    that var positional and var keyword arguments (``*args`` and ``**kwargs``)
+    are *not* added to the wrapped function's signature, since they are usually
+    used to transparently forward arguments from the decorator to the decorated
+    function.
+
+    :param added_by: Add some kind of reference to give a sense of where the
+        new behaviour of the wraps function comes from.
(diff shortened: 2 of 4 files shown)


lisa/analysis/base.py
```diff
@@ -131,7 +131,7 @@ def _save_plot(self, figure, default_dir, filepath=None, img_format=None, wrappe
             guessed_format = mime_type.split('/')[1].split('.', 1)[-1].split('+')[0]
             img_format = img_format or guessed_format
 
-        figure.savefig(filepath, format=img_format)
+        figure.savefig(filepath, format=img_format, bbox_inches='tight')
```
valschneider: Doesn't that cause issues with the resulting image file ? IIRC a handful of plot functions had this, but it caused more harm than good (usually stripped part of the title and such).

lisa/utils.py
```diff
@@ -817,4 +817,81 @@ def __copy__(self):
...
+
+        added_params = [
+            desc
+            for name, desc in wrapper_sig.parameters.items()
+            if (
+                name not in f_sig.parameters.keys()
+                # These kinds are usually present in the wrapper to forward
+                # to the wrapped function, so they are not interesting here
+                and not desc.kind in (
+                    inspect.Parameter.VAR_KEYWORD,
```
valschneider: IMO there should be a Big Fat Warning somewhere, at the very least in the doc if not at runtime when we ignore parameters.

I guess the reason you've not gone for a runtime warning is because you have kwargs/pos args in your use of it in the analysis code. If that's the case then I'd be ok with "just" big fat doc warning.

lisa/analysis/base.py
douglas-raillard-arm: On all the plots I tried to save so far, the title was always cropped without that option. I can make it the default only, so that plot functions needing something else can override it ?

lisa/utils.py
douglas-raillard-arm: This is just the list of parameters added by the decorator that we want to see on the function once decorated. The parameters we are collecting here come from `wrapper_sig`, and we ignore the ones 1) already defined in f() 2) *args and **kwargs since we don't want the decorator's kwargs to be added to the wrapped function signature (that would be wrong, the decorated function has no such parameter unless it already had initially, in which case they are preserved).

```
def decorator(f):
	@update_wrapper_doc(f)
	def wrapper(*args, foo=3, **kwargs):
		if foo == 42:
			return 'bar'
		else:
			return f(*args, **kwargs)
	return wrapper

@decorator
def f(y):
	pass

# The signature of f() should be "y, foo=3" and must not include
# "*args, **kwargs", since these are not parameters of the wrapped function.
```

valschneider: I'm fine with the behaviour, I'd just like to have it explicitly called out to avoid any confusion.

PR #952, opened 2019-06-12
utils: Make memoize() use a weakref for the first parameter

douglas-raillard-arm:
When used on methods, this avoids keeping a reference over the instances
alive forever even when nothing else apart from the cache needs it
anymore.


valschneider:
(Don't panic, quickly testing something with Travis)

PR #955, opened 2019-06-14
workload.run(): add timeout to avoid broken workloads

balsini:


lisa/wlgen/workload.py
```diff
@@ -131,13 +131,17 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False):
...
+                logger.info("Execution complete")
 
-            logfile = os.path.join(self.res_dir, 'output.log')
-            logger.debug('Saving stdout to %s...', logfile)
+                logfile = os.path.join(self.res_dir, 'output.log')
+                logger.debug('Saving stdout to %s...', logfile)
+
+                with open(logfile, 'w') as ofile:
+                    ofile.write(self.output)
+            except Exception as e:
```
valschneider: Generic `Exception` catching isn't really nice, but besides that I think we don't really need to catch it here. Just let it fire so it's obvious to the user that something went wrong, and let the user implement an specific exception handling if it makes sense for his/her use case.

balsini: Makes sense. I didn't want to expose the exception to the user, but actually, it may be triggered anyway in case of connection timeouts, so it wouldn't make any difference.
This behavior should be documented. :)

valschneider: You can add something like this to the docstring of the function:
```
        :raises: devlib.exception.TimeoutError if the specified ``timeout`` is hit.
```
Or something along those lines...

douglas-raillard-arm: nit: ``` :exc:`devlib.exception.TimeoutError` ```

balsini: Ouch, I didn't see your `:exc:`, I went for `:raise devlib.exception.TimeoutError: When the specified ``timeout`` is hit.`

valschneider: IIRC it's implied by the `raise` context: https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/misc_utilities.html?highlight=from_path#lisa.utils.Serializable.from_path

PR #957, opened 2019-06-14
doc: Add utils.show_doc()

douglas-raillard-arm:
Calling utils.show_doc() from a Jupyter notebook opens an IFrame with
online LISA documentation.

EDIT: ReadTheDocs exposes the intersphinx inventory, so no need to build anything locally.

image

### DIFF
--- a/lisa/utils.py
+++ b/lisa/utils.py
@@ -37,11 +37,16 @@ import threading
 import itertools
 import weakref
 from weakref import WeakKeyDictionary
+import urllib.parse
 import warnings
 import textwrap
 
 import ruamel.yaml
 from ruamel.yaml import YAML
+import sphobjinv
+
+import webbrowser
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
Not sure if the best behavior is an IFrame inside the notebook, or just opening a browser window. The IFrame does not have any history facility, so the browser experience might be better ...

valschneider:
I'd got for a browser window to follow `Trace.show()`: https://github.com/arm-software/lisa/blob//lisa/trace.py#L576

douglas-raillard-arm:
Default is now opening a browser, but `iframe=True` can also be used if someones fancies showing off our doc using notebook wizardry (s/notebook/jupyter lab/ for really cool 21st century guys using ugly iframe from the 90's).

PR #960, opened 2019-06-19
Tracepoints module

qais-yousef:
Add some infrastructure to allow building, installing, loading and unload a module and add a new sched_tp module which would be required to run Lisa tests against mainline kernel when the new bare tracepoints (ie: don't have TRACE_EVENTs on top/associated with them) land there.

### DIFF
--- a/doc/setup.rst
+++ b/doc/setup.rst
@@ -91,3 +91,82 @@ again is needed.
   ``lisa-install`` was executed. If a modification is detected, it will ask the
   user to run ``lisa-install`` again, since a dependency might have been added,
   or a version requirement might have been updated.
+
+External dependencies
+=====================
+
+Kernel modules
+++++++++++++++
+
+The following modules are required to run Lisa tests against some kernels.
+
+sched_tp
+--------
+
+From Linux v5.3, sched_load_cfs_rq and sched_load_se tracepoints are present in
+mainline as bare tracepoints without any events in tracefs associated with
+them.
+
+To help expose these tracepoints (and any additional one we might require in
+the future) as trace events, an external module is required and is provided
+under the name of sched_tp in $LISA_HOME/tools/kmodules/sched_tp
+
+Building a module
++++++++++++++++++
+
+The process is standard Linux external module build step. Helper scripts are
+provides too.
+
+Build
+-----
+
+.. code-block:: sh
+
+  $LISA_HOME/tools/kmodules/build_module path/to/kenrel path/to/kmodule [path/to/install/modules]
+
+This will build the module against the provided kernel tree and install it in
+``path/to/install/module`` if provided otherwise install it in
+``$LISA_HOME/tools/kmodules``.
+
+Clean
+-----
+
+.. code-block:: sh
+
+  $LISA_HOME/tools/kmodules/clean_module path/to/kenrel path/to/kmodule
+
+Highly recommended to clean when switching kernel trees to avoid unintentional
+breakage for using stale binaries.
+
+Pushing the module into the target
+----------------------------------
+
+You need to push the module into your rootfs either by installing it directly
+there or use commands like ``scp`` to copy it into your device.
+
+.. code-block:: sh
+
+  scp -r $LISA_HOME/tools/kmoudles/lib username@ip:/
+
+Loading the module
+------------------
+
+On the target run:
+
+.. code-block:: sh
+
+  modprobe sched_tp
+
+Integrating the module in your kernel tree
+++++++++++++++++++++++++++++++++++++++++++
+
+If you're rebuilding your kernel tree anyway, it might be easier to integrate
+the module into your kernel tree as a built-in module so that it's always
+present.
+
+Integrate using provided patch
+------------------------------
+
+.. code-block:: sh
+
+  cd path/to/kernel && git am path/to/patch
--- a/tools/kmodules/build_module
+++ b/tools/kmodules/build_module
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -eu
+
+KERNEL_SRC=`realpath "$1"`
+MODULE_SRC=`realpath "$2"`
+INSTALL_MOD_PATH=${3:-`dirname "$0"`}
+
+export INSTALL_MOD_PATH=`realpath "$INSTALL_MOD_PATH"`
+export ARCH=${ARCH:-arm64}
+
+case "$ARCH" in
+
+	arm64)
+		export CROSS_COMPILE=${CROSS_COMPILE:-aarch64-linux-gnu-}
+		;;
+	arm)
+		export CROSS_COMPILE=${CROSS_COMPILE:-arm-linux-gnueabi-}
+		;;
+	x86|x86_64)
+		;;
+	*)
+		echo "ERROR: Unknown ARCH=$ARCH"
+		exit 1
+		;;
+esac
+
+make -C "$KERNEL_SRC" M="$MODULE_SRC" modules
+make -C "$KERNEL_SRC" M="$MODULE_SRC" modules_install
--- a/tools/kmodules/clean_module
+++ b/tools/kmodules/clean_module
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -eu
+
+KERNEL_SRC=`realpath "$1"`
+MODULE_SRC=`realpath "$2"`
+
+export ARCH=${ARCH:-arm64}
+
+case "$ARCH" in
+
+	arm64)
+		export CROSS_COMPILE=${CROSS_COMPILE:-aarch64-linux-gnu-}
+		;;
+	arm)
+		export CROSS_COMPILE=${CROSS_COMPILE:-arm-linux-gnueabi-}
+		;;
+	x86|x86_64)
+		;;
+	*)
+		echo "ERROR: Unknown ARCH=$ARCH"
+		exit 1
+		;;
+esac
+
+make -C "$KERNEL_SRC" M="$MODULE_SRC" clean
--- a/tools/kmodules/sched_tp/sched_events.h
+++ b/tools/kmodules/sched_tp/sched_events.h
@@ -1,0 +1,134 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+#undef TRACE_SYSTEM
+#define TRACE_SYSTEM sched
+
+#if !defined(_SCHED_EVENTS_H) || defined(TRACE_HEADER_MULTI_READ)
+#define _SCHED_EVENTS_H
+
+#define PATH_SIZE		64
+#define SPAN_SIZE		NR_CPUS/4
+
+#include <linux/tracepoint.h>
+
+TRACE_EVENT(sched_load_cfs_rq,
+
+	TP_PROTO(int cpu, char *path, const struct sched_avg *avg),
+
+	TP_ARGS(cpu, path, avg),
+
+	TP_STRUCT__entry(
+		__field(	int,		cpu			)
+		__array(	char,		path,	PATH_SIZE	)
+		__field(	unsigned long,	load			)
+		__field(	unsigned long,	rbl_load		)
+		__field(	unsigned long,	util			)
+	),
+
+	TP_fast_assign(
+		__entry->cpu		= cpu;
+		strlcpy(__entry->path, path, PATH_SIZE);
+		__entry->load		= avg->load_avg;
+		__entry->rbl_load	= avg->runnable_load_avg;
+		__entry->util		= avg->util_avg;
+	),
+
+	TP_printk("cpu=%d path=%s load=%lu rbl_load=%lu util=%lu",
+		  __entry->cpu, __entry->path, __entry->load,
+		  __entry->rbl_load,__entry->util)
+);
+
+DECLARE_EVENT_CLASS(sched_pelt_rq_template,
+
+	TP_PROTO(int cpu, const struct sched_avg *avg),
+
+	TP_ARGS(cpu, avg),
+
+	TP_STRUCT__entry(
+		__field(	int,		cpu			)
+		__field(	unsigned long,	load			)
+		__field(	unsigned long,	rbl_load		)
+		__field(	unsigned long,	util			)
+	),
+
+	TP_fast_assign(
+		__entry->cpu		= cpu;
+		__entry->load		= avg->load_avg;
+		__entry->rbl_load	= avg->runnable_load_avg;
+		__entry->util		= avg->util_avg;
+	),
+
+	TP_printk("cpu=%d load=%lu rbl_load=%lu util=%lu",
+		  __entry->cpu, __entry->load,
+		  __entry->rbl_load,__entry->util)
+);
+
+DEFINE_EVENT(sched_pelt_rq_template, sched_pelt_rt,
+	TP_PROTO(int cpu, const struct sched_avg *avg),
+	TP_ARGS(cpu, avg));
+
+DEFINE_EVENT(sched_pelt_rq_template, sched_pelt_dl,
+	TP_PROTO(int cpu, const struct sched_avg *avg),
+	TP_ARGS(cpu, avg));
+
+DEFINE_EVENT(sched_pelt_rq_template, sched_pelt_irq,
+	TP_PROTO(int cpu, const struct sched_avg *avg),
+	TP_ARGS(cpu, avg));
+
+TRACE_EVENT(sched_load_se,
+
+	TP_PROTO(int cpu, char *path, char *comm, int pid, const struct sched_avg *avg),
+
+	TP_ARGS(cpu, path, comm, pid, avg),
+
+	TP_STRUCT__entry(
+		__field(	int,		cpu			)
+		__array(	char,		path,	PATH_SIZE	)
+		__array(	char,		comm,	TASK_COMM_LEN	)
+		__field(	int,		pid			)
+		__field(	unsigned long,	load			)
+		__field(	unsigned long,	rbl_load		)
+		__field(	unsigned long,	util			)
+	),
+
+	TP_fast_assign(
+		__entry->cpu		= cpu;
+		strlcpy(__entry->path, path, PATH_SIZE);
+		strlcpy(__entry->comm, comm, TASK_COMM_LEN);
+		__entry->pid		= pid;
+		__entry->load		= avg->load_avg;
+		__entry->rbl_load	= avg->runnable_load_avg;
+		__entry->util		= avg->util_avg;
+	),
+
+	TP_printk("cpu=%d path=%s comm=%s pid=%d load=%lu rbl_load=%lu util=%lu",
+		  __entry->cpu, __entry->path, __entry->comm, __entry->pid,
+		  __entry->load, __entry->rbl_load,__entry->util)
+);
+
+TRACE_EVENT(sched_overutilized,
+
+	TP_PROTO(int overutilized, char *span),
+
+	TP_ARGS(overutilized, span),
+
+	TP_STRUCT__entry(
+		__field(	int,		overutilized		)
+		__array(	char,		span,	SPAN_SIZE	)
+	),
+
+	TP_fast_assign(
+		__entry->overutilized	= overutilized;
+		strlcpy(__entry->span, span, SPAN_SIZE);
+	),
+
+	TP_printk("overutilized=%d span=0x%s",
+		  __entry->overutilized, __entry->span)
+);
+
+#endif /* _SCHED_EVENTS_H */
+
+/* This part must be outside protection */
+#undef TRACE_INCLUDE_PATH
+#define TRACE_INCLUDE_PATH .
+#define TRACE_INCLUDE_FILE sched_events
+#include <trace/define_trace.h>
--- a/tools/kmodules/sched_tp/sched_tp.c
+++ b/tools/kmodules/sched_tp/sched_tp.c
@@ -1,0 +1,132 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+#include <linux/module.h>
+
+#include <linux/sched.h>
+#include <trace/events/sched.h>
+
+#define CREATE_TRACE_POINTS
+#include "sched_events.h"
+
+static inline struct cfs_rq *get_group_cfs_rq(struct sched_entity *se)
+{
+#ifdef CONFIG_FAIR_GROUP_SCHED
+	return se->my_q;
+#else
+	return NULL;
+#endif
+}
+
+static void sched_pelt_cfs(void *data, struct cfs_rq *cfs_rq)
+{
+	if (trace_sched_load_cfs_rq_enabled()) {
+		const struct sched_avg *avg;
+		char path[PATH_SIZE];
+		int cpu;
+
+		avg = sched_trace_cfs_rq_avg(cfs_rq);
+		sched_trace_cfs_rq_path(cfs_rq, path, PATH_SIZE);
+		cpu = sched_trace_cfs_rq_cpu(cfs_rq);
+
+		trace_sched_load_cfs_rq(cpu, path, avg);
+	}
+}
+
+static void sched_pelt_rt(void *data, struct rq *rq)
+{
+	if (trace_sched_pelt_rt_enabled()) {
+		const struct sched_avg *avg = sched_trace_rq_avg_rt(rq);
+		int cpu = sched_trace_rq_cpu(rq);
+
+		if (!avg)
+			return;
+
+		trace_sched_pelt_rt(cpu, avg);
+	}
+}
+
+static void sched_pelt_dl(void *data, struct rq *rq)
+{
+	if (trace_sched_pelt_dl_enabled()) {
+		const struct sched_avg *avg = sched_trace_rq_avg_dl(rq);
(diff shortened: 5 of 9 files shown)


lisa/target.py
```diff
@@ -197,13 +198,14 @@ class used under the hood, and avoids messing up with ``devlib``
 
     CONF_CLASS = TargetConf
     INIT_KWARGS_KEY_MAP = {
+        'kernel_src': ['kernel_src'],
```
douglas-raillard-arm: That's not needed in basic case but will be needed with the `kernel` sublevel (just like the below devlib's `excluded-modules` case).

lisa/target.py
```diff
@@ -276,6 +276,8 @@ def __init__(self, kind, name='<noname>', tools=[], res_dir=None,
 
         logger.info('Effective platform information:\n{}'.format(self.plat_info))
 
+        self._init_extend_sched_events()
```
douglas-raillard-arm: We probably don't want to call that unconditionally, as just connecting to a target should work everywhere, including places where we cannot create/insert modules (lack of toolchain, lack of permissions, connecting to a target which is a container etc). 

My only thought for now is adding an `__init__` parameter (plus a `modules` config key under `kernel` sublevel) that would contain a list of kmodules to build/load. This way the user can choose which modules are needed, including custom ones, so we can add more of these if we want. 

@valschneider any thoughts on that ?

lisa/target.py
```diff
@@ -728,6 +730,30 @@ def remove_kmodule(self, kmodule):
...
+        sched_load_avg_task = os.path.join(sched_events_path, 'sched_load_avg_task')
+        sched_load_se = os.path.join(sched_events_path, 'sched_load_se')
+
+        has_sched_load_avg_task = self.target.directory_exists(sched_load_avg_task)
+        has_sched_load_se = self.target.directory_exists(sched_load_se)
+
+        self.need_to_load_sched_tp = not (has_sched_load_avg_task or has_sched_load_se)
+
+        if self.need_to_load_sched_tp:
+            self.build_kmodule('sched_tp')
```
douglas-raillard-arm: Would be worth adding a helper that does all 3 steps in one go, and that `rmmod` first in case the module is already loaded.

lisa/target.py
```diff
@@ -728,6 +730,30 @@ def remove_kmodule(self, kmodule):
...
+        On newer kernels (5.3 and above) PELT signals are present as bare
+        tracepoints and we need to load a module to convert them to proper
+        TRACE_EVENTs.
+
+        This function automatically handles this.
+        """
+
+        sched_events_path = '/sys/kernel/debug/tracing/events/sched/'
+
+        sched_load_avg_task = os.path.join(sched_events_path, 'sched_load_avg_task')
```
derkling: Since you use `os.path.join` here, shouldn't we do the same for `sched_events_path`?
Moreover, we should use devlib's `target.os.path.join` since these are target specific paths.

derkling: Moreover, the base should be probably collected from the target via `target.list_file_systems()` and looking for an `FstabEntry` with `fs_type='tracefs'`

derkling: Something along these lines should make it:
```python
for entry in target.list_file_systems():
    if entry.fs_type != 'tracefs':
        continue
target.path.join(entry.mount_point, 'events', 'sched')
```

lisa/target.py
```diff
@@ -197,13 +198,14 @@ class used under the hood, and avoids messing up with ``devlib``
...
+        'kernel_src': ['kernel_src'],
         'devlib_excluded_modules': ['devlib', 'excluded-modules']
     }
 
     def __init__(self, kind, name='<noname>', tools=[], res_dir=None,
         plat_info=None, workdir=None, device=None, host=None, port=None,
-        username='root', password=None, keyfile=None, devlib_platform=None,
-        devlib_excluded_modules=[]
+        username='root', password=None, keyfile=None, kernel_src=None,
+        devlib_platform=None, devlib_excluded_modules=[]
```
derkling: Why not just add it at the end?

tools/kmodules/build_module
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -eu
+
+MODULE=$(dirname $0)/$1
```
qperret: I'm thinking, we could pass the path to the module source as a parameter of this script, and provide a way for users to override it (by specifying something `target.conf` for example). We might need several versions of this module in case the in-kernel ABI changes (because of upstream progress, or because of device-specific changes) and keep backward compatibility.

tools/kmodules/sched_tp/sched_events.h
```diff
@@ -0,0 +1,134 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+#undef TRACE_SYSTEM
+#define TRACE_SYSTEM sched
+
+#if !defined(_SCHED_EVENTS_H) || defined(TRACE_HEADER_MULTI_READ)
+#define _SCHED_EVENTS_H
+
+#define PATH_SIZE		128
+#define SPAN_SIZE		128/4	/* assuming a max of 128 cpu system! */
```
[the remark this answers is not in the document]
qais-yousef: I thought about this but then one can override this value from command line if I remember correctly. It's not really a lot of memory. 128 bitmap is 2 words on 64bits systems and 4 on 32bits one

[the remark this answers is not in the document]
qais-yousef: Okay sod that I should have looked at the code again I already forgot it :D

Maybe I can make the events dynamic if the impact of this is really serious. The overutilized event is low traffic in general so I don't see that wasting too much space?

tools/kmodules/sched_tp/sched_events.h
```diff
@@ -0,0 +1,134 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+#undef TRACE_SYSTEM
+#define TRACE_SYSTEM sched
+
+#if !defined(_SCHED_EVENTS_H) || defined(TRACE_HEADER_MULTI_READ)
+#define _SCHED_EVENTS_H
+
+#define PATH_SIZE		128
```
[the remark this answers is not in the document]
qais-yousef: What value do you propose? I'm not really sure if it's worth the fuss but it's an easy to modify tunable

tools/kmodules/sched_tp/sched_tp.c
```diff
@@ -0,0 +1,132 @@
...
+#ifdef CONFIG_FAIR_GROUP_SCHED
+       return se->my_q;
+#else
+       return NULL;
+#endif
+}
+
+static void sched_pelt_cfs(void *data, struct cfs_rq *cfs_rq)
+{
+	if (trace_sched_load_cfs_rq_enabled()) {
```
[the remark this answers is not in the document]
qais-yousef: I don't mind but I kept it like that so one can insert their own code around this more easily if they too (in case anyone felt extending what they want to do when this tracepoint is triggered)

lisa/target.py
```diff
@@ -664,6 +665,68 @@ def install_tools(self, tools):
...
+        Compile kernel module that we want to load at the target at some point
+
+        :param kmodule: The name of the module to build
+        :type kmodule: str
+        """
+        logger = self.get_logger()
+
+        cmd = '{}/tools/kmodules/build_module'.format(LISA_HOME)
+
+        logger.info('Building kernel module [{}] using [{}] kernel tree'.format(kmodule, self._kernel_src))
```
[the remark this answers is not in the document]
qais-yousef: Is this a pure cosmetical change or is there a functional difference too?

tools/kmodules/clean_module
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -eu
+
+MODULE=$(dirname $0)/$1
+KERNEL_SRC=$2
+
+cd $MODULE
+KERNEL_SRC=$KERNEL_SRC make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- clean &> /dev/null
```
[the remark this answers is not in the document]
qais-yousef: Yeah it's on my todo list to extend this to other ARCHs, but as initial step I just wanted to keep things simple. I'd be happy to get this work done before this gets merged if you want, or do it as a separate change on top (which what I'd prefer if it's not a big issue)

lisa/target.py
qais-yousef: The code was using mixed stuff to be honest and I got lost in that. I'll fix it.

lisa/target.py
qais-yousef: To preserve the hierarchy, so it's at the end of the top level

lisa/target.py
```diff
@@ -119,6 +119,7 @@ class TargetConf(SimpleMultiSrcConf, HideExekallID):
         KeyDesc('port', 'SSH or ADB server port', [int, None]),
         KeyDesc('device', 'ADB device. Takes precedence over "host"', [str, None]),
         KeyDesc('keyfile', 'SSH private key file', [str, None]),
+        KeyDesc('kernel_src', 'Kernel source tree to build module', [str, None]),
```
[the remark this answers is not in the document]
qais-yousef: Okay will do

lisa/target.py
qais-yousef: :+1:

lisa/target.py
```diff
@@ -664,6 +665,68 @@ def install_tools(self, tools):
...
+        :param kmodule: The name of the module to build
+        :type kmodule: str
+        """
+        logger = self.get_logger()
+
+        cmd = '{}/tools/kmodules/build_module'.format(LISA_HOME)
+
+        logger.info('Building kernel module [{}] using [{}] kernel tree'.format(kmodule, self._kernel_src))
+
+        failed = subprocess.call([cmd, kmodule, self._kernel_src])
```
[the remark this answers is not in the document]
qais-yousef: I don't get you. I did test this to verify that we detect we get the right return code if we fail, which in this case we just log and carry on.

Are you saying better to use check_call() to enforce raising an error?

lisa/target.py
```diff
@@ -664,6 +665,68 @@ def install_tools(self, tools):
...
+    def install_kmodule(self, kmodule):
+        """
+        Install kernel modules on the target
+
+        :param kmodule: The name of the module to install
+        :type kmodule: str
+        """
+        logger = self.get_logger()
+
+        src = '{}/tools/kmodules/{}/{}.ko'.format(LISA_HOME, kmodule, kmodule)
```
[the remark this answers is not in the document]
qais-yousef: I copied that from install_tools, it was confusing the mixed usage all around

lisa/target.py
```diff
@@ -664,6 +665,68 @@ def install_tools(self, tools):
...
+        dst = '{}/kmodules/{}.ko'.format(self.working_directory, kmodule)
```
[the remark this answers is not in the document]
qais-yousef: :+1:

lisa/target.py
```diff
@@ -664,6 +665,68 @@ def install_tools(self, tools):
...
+
+        logger.info('Installing kernel module {}'.format(dst))
+
+        self.target.execute('mkdir -p {}'.format(self.path.dirname(dst)))
```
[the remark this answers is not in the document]
qais-yousef: :+1:

tools/kmodules/build_module
qais-yousef: Okay. How about default to our known location unless the user specifically wants to override?

lisa/target.py
qais-yousef: I tried to do the module insertion in FtraceCollect.__init__() but I had issues with that, let me try again more seriously this time

lisa/target.py
qais-yousef: I was hoping I'd get a nice pointer on how to clean up nicely so that we can't have a left over module loaded when we finish..

_init_extend_sched_events() is a wrapper around these, why create another wrapper?

lisa/target.py
```diff
@@ -659,6 +667,92 @@ def install_tools(self, tools):
...
+
+        sched_events_path = '/sys/kernel/debug/tracing/events/sched/'
+
+        sched_load_avg_task = os.path.join(sched_events_path, 'sched_load_avg_task')
+        sched_load_se = os.path.join(sched_events_path, 'sched_load_se')
+
+        has_sched_load_avg_task = self.target.directory_exists(sched_load_avg_task)
+        has_sched_load_se = self.target.directory_exists(sched_load_se)
+
+        self.need_to_load_sched_tp = not (has_sched_load_avg_task or has_sched_load_se)
```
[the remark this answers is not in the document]
qais-yousef: Fair enough

qais-yousef:
I have updated the tree taken all requested most changes into account. If it wasn't done it's either omitted by mistake or I had a good reason not to do that way. Either case please ping me if I seem to have 'ignored' your review comment.

I have changed the approach now so that we take a list of modules to be compiled and loaded as suggested by Douglas and we assume they are a requirement to run the test. This simplifies the logic and keeps things generic, but the user have to know whether they need the module for their kernel.. Maybe we can improve on this by taking a condition under which to load the module if we deem  it's necessary.

Lightly tested against Peter's latest queue.

Pointers to the right place in documentation to explain the module support would be appreciated.

Thanks for all the time taken by the reviewers to look and comment at this!

lisa/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: Yes, I'd rather have this fail if there is any issue with the module in one way or another. Since that feature has to stay optional in any case, if it breaks with an exception, the user has the choice of not trying to use that module to "fix" the issue. The caller of that code can therefore expect this function to always do its job if it returns (return != raise), which is the usual way of proceeding in Python.

lisa/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: Right, install tools should be fixed as well, I'll do that in a separate PR.

lisa/target.py
douglas-raillard-arm: I had the same thought over the weekend: we can let FtraceCollector decide if that modules is needed depending on the events that are asked for. That would allow connecting to the target without uploading the module when there won't be any consumer for it.

lisa/target.py
douglas-raillard-arm: > clean up nicely so that we can't have a left over module loaded when we finish.

There is no cleanup code for Target. We can turn `Target` into a context manager and add some cleanup steps if it's needed, but the code should never rely on the target being cleaned up when re-connecting, since that cleanup may not happen for various reasons.

> _init_extend_sched_events() is a wrapper around these, why create another wrapper?

Seems solved by the generic `_init_kmodules()` in the new version of the PR, which can be used for any module and not just sched stuff.

tools/kmodules/sched_tp/sched_events.h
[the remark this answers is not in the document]
douglas-raillard-arm: Random thought: If these can be changed with a `-DCONFIG_NR_CPUS=42` when compiling the module, it's quite easy to fill-in the right value directly from LISA. It can default to a reasonable number for non-LISA users.

lisa/target.py
```diff
@@ -664,6 +665,68 @@ def install_tools(self, tools):
...
             self._installed_tools.add(tool)
 
+    def build_kmodule(self, kmodule):
+        """
+        Compile kernel module that we want to load at the target at some point
+
+        :param kmodule: The name of the module to build
+        :type kmodule: str
+        """
+        logger = self.get_logger()
```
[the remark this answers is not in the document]
douglas-raillard-arm: `get_logger()` has to exist as a classmethod so it can be used from classmethods, but we could have a property on top when called in regular methods if we want.

lisa/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: If possible I'd rather have one kind of interpolation throughout LISA, and the logging's interpolation benefits are quite small for our use case (lazy `__str__` and string interpolation). It also allows easier manipulation of the message in general when debugging, as it's a plain string.

> Is this a pure cosmetical change or is there a functional difference too?

The only technical difference is the one that I mentioned, but Patrick was more referring to it as a cosmetic change.

douglas-raillard-arm:
> Pointers to the right place in documentation to explain the module support would be appreciated.

A good starting point is `Target`'s docstring, and if you want to give less "factual" information and explain some use cases, doc/target.rst sounds like a good place for your prose.

qais-yousef:
Ionela was suggesting to extend the description to provide a path to a prebuilt module which might make CI integration easier.

I am not strongly against the idea but neither in favor of it either. A simple clear and well defined rule of what modules will be supported will keep the code and the external support simple. Allowing exceptions will end up with some sort of spaghetti at some point. Mind you, spaghetti is delicious, but not when it's made of code :-)

I think if someone thinks it's fine to build their on module, then they're fine to put it on their file system and auto load it and keep Lisa out of it.

The use case I see for this module support is: I run and test several trees but Lisa needs this module to run the test, so let Lisa auto manage building and loading the modules as I switch trees/branches transparently. Well pseudo transparently as you have to update the yml config if you switch trees. Which is still simpler than maintaining the module on all of your trees/branches.

lisa/target.py
```diff
@@ -659,6 +661,66 @@ def install_tools(self, tools):
...
+        :param kernel_src: The path to the kernel source tree to use to build the module
+        :type kernel_src: str
+
+        :param kmodule_path: The path to the module to build
+        :type kmodule_path: str
+        """
+        self.get_logger().info('Building kernel module [%s] using [%s] kernel tree', kmodule_path, kernel_src)
+
+        cmd = os.path.join(LISA_HOME, 'tools/kmodules/clean_module')
+        failed = subprocess.call([cmd, kernel_src, kmodule_path])
```
douglas-raillard-arm: `subprocess.check_call` provides this check already, so this:
```
         failed = subprocess.call([cmd, kernel_src, kmodule_path]) 
        if failed:
            raise RuntimeError('Failed to clean kernel module: %s', kmodule_path) 
```
can be simplified into this:
```
         subprocess.check_call([cmd, kernel_src, kmodule_path]) 
```

lisa/target.py
```diff
@@ -659,6 +661,66 @@ def install_tools(self, tools):
...
+
+        :param kmodule_path: Path to the module to to be unloaded
+        :type kmodule_path: str
+        """
+        kmodule_path = os.path.join(kmodule_path, '*.ko')
+
+        temp = glob.glob(kmodule_path)
+        kmodule_names = []
+
+        for kname in temp:
```
douglas-raillard-arm: nit: `temp` is not really necessary as this is already readable:
```
for kname in glob.glob(kmodule_path)
```

tools/kmodules/build_module
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -eu
+
+KERNEL_SRC=`realpath $1`
+MODULE_SRC=`realpath $2`
```
douglas-raillard-arm: $1 and $2 need quoting
```
KERNEL_SRC=$(realpath "$1")
MODULE_SRC=$(realpath "$2")
```

tools/kmodules/build_module
```diff
@@ -0,0 +1,13 @@
...
+
+export ARCH=${ARCH:-arm64}
+
+if [ "$ARCH" != "`uname -m`" ]; then
+	export CROSS_COMPILE=${CROSS_COMPILE:-aarch64-linux-gnu-}
+fi
+
+make -C $KERNEL_SRC M=$MODULE_SRC modules &> /dev/null
```
douglas-raillard-arm: all variable expansions need quoting

lisa/target.py
```diff
@@ -554,6 +567,31 @@ def _init_target(self, kind, name, workdir, device, host,
...
+        :type name: str
+
+        :param kernel_modules: List of kernel modules paths to build and load
+        :type kernel_modules: StrList
+        """
+        if not kernel_modules:
+            return
+
+        if not kernel_src:
+            raise RuntimeError("Can't build modules without a kernel source tree")
```
douglas-raillard-arm: May be more appropriate to raise a ValueError, since it boils that to the user incorrectly giving a `None` value (explicitly or implicitly) for the use case. RuntimeError is more the "last resort" exception for things that don't fit in any other category:
https://docs.python.org/3/library/exceptions.html#ValueError

lisa/target.py
```diff
@@ -659,6 +703,66 @@ def install_tools(self, tools):
             self.target.install(tool)
             self._installed_tools.add(tool)
 
+    def build_kmodule(self, kernel_src, kmodule_path):
```
valschneider: I think I missed that one back when I first reviewed this PR, apologies for that.

I'm really not fond of having building logic within LISA. We used to have some bits of code that would try to build a kernel and install it on a target (while taking into account different setups like tftp and whatnot), and we agreed that this was outside of the scope of LISA.

ATM, to get the load tracking tests going, you need to have the proper "proprietary" TP patches cherry-picked to your kernel, and have that kernel running on the target. I'm fine with having logic to load the modules, but not to build them - that should be left up to the user (just like building the kernel is left to him/her).

This sort of matches the logic we have for tools ATM. You can specify that you want to load `trace-cmd`, but we don't build it from source every time we do that. So here I would suggest taking as input a *compiled* module path to load it, but leave the building outside.

It also makes sense for CI flows, because it builds the whole thing anyway (kernel + modules), so it would just be plumbing of artifacts we already generate. This would avoid duplicating code that takes care of the building.

qais-yousef: I can appreciate that point of view but I think it's either Lisa do full management of the module or the user. Splitting it here and there is a bad idea and will open a worse can of worms.

As a middle ground I'd be happy to provide a patch instead + a script to apply it into a tree then tell the user to rebuild their kernel with this change before running Lisa. The module is tiny and we can enforce it to always be builtin and that will be that least effort solution if the user has to build the module anyway. There's little or no gray areas here and since the code is completely isolated it won't conflict with anything and should be guaranteed to apply cleanly except to small conflicts in Makefile maybe.

The advantage of current solution is that it is generic to serve other purposes and Douglas showed some interest in that already, but I don't mind as we choose something that is consistent. ie: either user has full responsibility or Lisa, but not divide it.

qais-yousef:
As agreed in our discussion offline. The responsibility of managing the module is delegated to the user now and we only provide the source code and a patch and some instructions for the user to integrate this into their workflow in appropriate way.

The only thing that I didn't figure out yet, and orthogonal to this change, is how do we point the user in an easy and clear way that they might need to look in the new tools/kmodule/README.md?

One potential way is to print a helpful error message in Ftracecollector() if one of the events we know is only avaialble by sched_tp is required but missing on the target.

douglas-raillard-arm:
> The only thing that I didn't figure out yet, and orthogonal to this change, is how do we point the user in an easy and clear way that they might need to look in the new tools/kmodule/README.md?

Since our doc is now 100% reStructuredText and hosted on ReadTheDocs, I would suggest moving this readme in doc/setup.rst [1] since it's a step that is common to all workflows.

note: when converting to reST, you will need to build the doc to make sure it works:
```
cd doc && ./pedantic_build.sh
```
If the script returns non-zero, look for the "warnings to fix" section in the output.

[1] https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/setup.html

qais-yousef:
Are there any more comments about this PR?

doc/setup.rst
```diff
@@ -91,3 +91,70 @@ again is needed.
...
+Building a module
++++++++++++++++++
+
+The process is standard Linux external module build step. Helper scripts are
+provides too.
+
+Build
+-----
+
+`$LISA_HOME/tools/kmodules/build_module path/to/kenrel path/to/kmodule [path/to/install/modules]`
```
douglas-raillard-arm: would be better as a code block than relying on the default role:
```

.. code-block:: sh

    $LISA_HOME/tools/kmodules/build_module path/to/kenrel path/to/kmodule [path/to/install/modules]
```

doc/setup.rst
```diff
@@ -91,3 +91,70 @@ again is needed.
...
+
+This will build the module against the provided kernel tree and install it in
+path/to/install/module if provided otherwise install it in
```
douglas-raillard-arm: Could use literal text:
```
``path/to/install/module``
```

doc/setup.rst
```diff
@@ -91,3 +91,70 @@ again is needed.
...
+$LISA_HOME/tools/kmodules.
```
douglas-raillard-arm: same here

doc/setup.rst
```diff
@@ -91,3 +91,70 @@ again is needed.
...
+
+Clean
+-----
+
+`$LISA_HOME/tools/kmodules/clean_module path/to/kenrel path/to/kmodule`
```
douglas-raillard-arm: ditto about code-block

doc/setup.rst
```diff
@@ -91,3 +91,70 @@ again is needed.
...
+Pushing the module into the target
+----------------------------------
+
+You need to push the module into your rootfs either by installing it directly
+there or `scp -r $LISA_HOME/tools/kmoudles/lib username@ip:/` into your device.
+
+Loading the module
+------------------
+
+On the taraget run:
```
douglas-raillard-arm: s/taraget/target/

PR #969, opened 2019-07-03
Update PATH management

douglas-raillard-arm:
* cleanup ANDROID_HOME path-related bits
* give priority to binaries shipped in LISA over the ones from the distribution

### DIFF
--- a/shell/lisa_shell
+++ b/shell/lisa_shell
@@ -23,6 +23,9 @@
 # Set environment variables
 ################################################################################
 
+export _DOC_ANDROID_HOME="Base folder of Android SDK installation"
+export ANDROID_HOME="$LISA_HOME/tools/android-sdk-linux"
+
 export _DOC_LISA_HOME="Base directory of LISA environment"
 
 export _DOC_LISA_PYTHON="Python binary to use. This allows users to install
@@ -42,7 +45,9 @@ source "$LISA_HOME/shell/lisa_colors"
 export _DOC_LISA_HOST_ABI="Add some shell utilities to the PATH, with lower
 priority than system's one in case the user needs a different version of them"
 export LISA_HOST_ABI=${LISA_HOST_ABI:-x86_64}
-export PATH=$PATH:$LISA_HOME/shell/:$LISA_HOME/tools/$LISA_HOST_ABI:$LISA_HOME/tools
+
+_LISA_PATH="$LISA_HOME/shell/:$LISA_HOME/tools/$LISA_HOST_ABI:$LISA_HOME/tools/:$ANDROID_HOME/tools/:$ANDROID_HOME/platform-tools/"
+export PATH=$PATH:$_LISA_PATH
 
 export _DOC_LISA_CONF="Default configuration file for exekall"
 export LISA_CONF=${LISA_CONF:-$LISA_HOME/target_conf.yml}
@@ -55,6 +60,10 @@ export EXEKALL_ARTIFACT_ROOT=${EXEKALL_ARTIFACT_ROOT:-$LISA_RESULT_ROOT}
 
 # Add our man pages
 export MANPATH="$MANPATH:$LISA_HOME/doc/"
+
+# make sure it is unset if the folder does not exist, as some other tools will
+# not handle that well
+[[ ! -d "$ANDROID_HOME" ]] && unset ANDROID_HOME
 
 ################################################################################
 # Helpers
@@ -381,7 +390,7 @@ function lisa-wltest-series {
     export WA_USER_DIRECTORY="$LISA_HOME/tools/wa_user_directory"
 
     # Check that the environment is properly configured
-    if [[ -z "$ANDROID_HOME" ]]; then
+    if [[ ! -d "$ANDROID_HOME" ]]; then
 cat <<EOF
 
     ERROR: No Android SDK found!
--- a/shell/lisa_shell
+++ b/shell/lisa_shell
@@ -46,8 +46,15 @@ export _DOC_LISA_HOST_ABI="Add some shell utilities to the PATH, with lower
 priority than system's one in case the user needs a different version of them"
 export LISA_HOST_ABI=${LISA_HOST_ABI:-x86_64}
 
+export _DOC_LISA_USE_SYSTEM_BIN="Use the system binaries if 1, will use the ones shipped with LISA if 0"
+export LISA_USE_SYSTEM_BIN=${LISA_USE_SYSTEM_BIN:-0}
+
 _LISA_PATH="$LISA_HOME/shell/:$LISA_HOME/tools/$LISA_HOST_ABI:$LISA_HOME/tools/:$ANDROID_HOME/tools/:$ANDROID_HOME/platform-tools/"
-export PATH=$_LISA_PATH:$PATH
+if [[ "$LISA_USE_SYSTEM_BIN" == 1 ]]; then
+    export PATH=$PATH:$_LISA_PATH
+else
+    export PATH=$_LISA_PATH:$PATH
+fi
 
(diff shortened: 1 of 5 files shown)


shell/lisa_shell
```diff
@@ -42,7 +45,9 @@ source "$LISA_HOME/shell/lisa_colors"
 export _DOC_LISA_HOST_ABI="Add some shell utilities to the PATH, with lower
 priority than system's one in case the user needs a different version of them"
 export LISA_HOST_ABI=${LISA_HOST_ABI:-x86_64}
-export PATH=$PATH:$LISA_HOME/shell/:$LISA_HOME/tools/$LISA_HOST_ABI:$LISA_HOME/tools
+
+_LISA_PATH="$LISA_HOME/shell/:$LISA_HOME/tools/$LISA_HOST_ABI:$LISA_HOME/tools/:$ANDROID_HOME/tools/:$ANDROID_HOME/platform-tools/"
+export PATH=$_LISA_PATH:$PATH
```
valschneider: I'm really not fond of overriding the user choices. One of the requirements for our install of ADB/fastboot was that the user's one would be used in priority (@derkling argued for it), and I think the same applies for trace-cmd.

Yes, if your trace-cmd is busted (e.g. the one shipped by most distros), you'll get a segfault. But if you have a different trace-cmd already set up on your system (maybe you're working on a fancy custom kernel and you built your own trace-cmd to make it work with your hacks, or you just need a newer trace-cmd in general), then you would expect that to be used.

It's an "ease of first time use" vs "don't mess up existing user's setup" balancing, and I heavily lean towards the second one.

douglas-raillard-arm: Giving priority to LISA means you will just have to delete the binary in LISA's folder if you want to use the systems' one. This is easy to do, can be supported and won't break randomly. Deleting a file that is part of a package is a great way to break your system and is definitely *not* something that any package manager even remotely supports.

> ease of first time use
It's not about first time use. It's about use, even after a 1000 times. The problem does not go away at the second use, and fixing currently requires a patch to LISA that will need to evolve as things we add to PATH change. OTH, "fixing" the priority if you don't want LISA to mess up with your system binary is straightforward: delete some LISA binaries.

To break even, it would be easy to add a config option to let the user choose which PATH order they want. Default would still be to make lisa commands work. `source init_env` is supposed to create an environment in which all the `lisa-` commands we provide work. Since LISA is mostly distro-agnostic, that means we have to ship versions known to work for some tools. The user can open a separate shell if they need their distro's trace-cmd.

Now if we were to make things "perfectly", we would provide a Debian package, a Ubuntu package, an Archlinux package and maybe Fedora. It would contain a build of trace-cmd and would explicitly conflict/replace the systems' trace-cmd. The package manager would then be happy and the user will be aware of that, and end up with only one version of the tool. We are not shipping such packages.

qais-yousef: I think both points are valid. But general expectations is that system's binary take a priority, and silent override will catch some people out. At the same time if a broken trace-cmd is installed (like what happened to me when I first started using Lisa), it'd confuse the hell out of the user and make them silently drop Lisa.

As a middle ground, one can prompt the user to a series of questions when they do `. init_env` so that they can choose to use a prebuilt one or use their own. Or shout loudly it was overriden and give an option to set an ENV_VARIABLE to provide a path to trace-cmd they want to use etc. I think defaulting to a stable setup is a good idea given that we make it obvious how the user can take back control if they want to. Nice out of box experience is good to help wider adoption IMHO.

douglas-raillard-arm: As it is `source init_env` is supposed to work in non-interactive environment, but we could provide an interactive one if we want to start asking for user input like that.

Ultimately, people will install kernelshark since it's in install_ubuntu_base.sh, which provides a trace-cmd binary as well (that's not optional, installing kernelshark on Ubuntu implies installing trace-cmd). "taking back control" by asking affect people to delete binaries installed in system-wide location and managed by the distro package manager is *not* a valid approach. If that step is necessary to get LISA working, LISA is either broken or not supporting the distro. "unsupporting" distro just because of that when we have working binaries that we already ship would make no sense.

The only real choice we have here are:
* give priority to LISA binary.
* rename LISA binaries to something like lisa-trace-cmd, so both the distro one and the LISA one can coexist in lisa-shell.
* use a shim for trace-cmd that selects the one with the highest version number.
* use a shim for binaries we ship that will look at some env var to know what binary to use.

I'll update the PR with the last proposition implemented, so people can selectively choose what binary to use for each tool we ship if necessary.

douglas-raillard-arm:
Updated with a new mechanism to allow user override for tools we ship:
```
export LISA_BIN_TRACE_CMD=/usr/bin/trace-cmd
source init_env
# Calling trace-cmd will execute /usr/bin/trace-cmd instead of the one provided by LISA
```

tools/dispatch/__init__.py
```diff
@@ -0,0 +1 @@
+../dispatch.sh
```
valschneider: Does it make any sense to have user overrides for things like `__init__.py` or `extract_platinfo.py`?

Also, IIUC, adding new tools will require adding dispatch entries - arguably that can be scripted, but it just feels like unnecessary trouble.

What if we go for a e.g. `LISA_USER_BIN` env var, and then we can have something like:
```bash
_LISA_PATH="$LISA_HOME/shell/:$LISA_HOME/tools/$LISA_HOST_ABI:$LISA_HOME/tools/:$ANDROID_HOME/tools/:$ANDROID_HOME/platform-tools/"

if [[ -n $LISER_USER_BIN ]]; then
        # User bins take priority
        export PATH=$PATH:$_LISA_PATH
else
        # LISA bins take priority
        export PATH=$_LISA_PATH:$PATH
fi
```

That way, first-time users get the right setup, and existing users can just add a single env var to their .bashrc and keep on using their hodgepodge of binaries.

douglas-raillard-arm: I haven't added any symlinks for programs we ship, so `extract_platinfo.py` is not part of them, for the reason you state.

> Also, IIUC, adding new tools will require adding dispatch entries - arguably that can be scripted, but it just feels like unnecessary trouble.

They will need to be symlinked in the dispatch folder indeed. One good side is that we know exactly the list of things that we add in the PATH ...

> What if we go for a e.g. LISA_USER_BIN env var, and then we can have something like:

That's what I proposed at some point, however that has also its flaws: people might want the adb coming from the android SDK, and the trace-cmd of their distro at the same time. Settings like that conflict with each other and can only be solved by a per-binary decision, which is what this commit brings in a compact way.

douglas-raillard-arm: following offline discussion, we can adopt that solution now and if we need more flexibility in the future, we can always add that dispatch implementation then.

shell/lisa_shell
```diff
@@ -42,7 +45,16 @@ source "$LISA_HOME/shell/lisa_colors"
 export _DOC_LISA_HOST_ABI="Add some shell utilities to the PATH, with lower
 priority than system's one in case the user needs a different version of them"
 export LISA_HOST_ABI=${LISA_HOST_ABI:-x86_64}
-export PATH=$PATH:$LISA_HOME/shell/:$LISA_HOME/tools/$LISA_HOST_ABI:$LISA_HOME/tools
+
+export _DOC_LISA_USE_SYSTEM_BIN="Use the system binaries if 1, will use the ones shipped with LISA if 0"
+export LISA_USE_SYSTEM_BIN=0
```
valschneider: Shouldn't that be something like
```suggestion
export LISA_USE_SYSTEM_BIN=${LISA_USE_SYSTEM_BIN:-0}
```
?

douglas-raillard-arm: indeed, thanks

PR #980, opened 2019-07-10
Add example plots to documentation

douglas-raillard-arm:
A built documentation with the example plots can be found at:
https://htmlpreview.github.io/?https://raw.githubusercontent.com/douglas-raillard-arm/lisa/_pr228_build/doc/_build/html/trace_analysis.html

### DIFF
--- a/doc/plot_conf.yml
+++ b/doc/plot_conf.yml
@@ -1,0 +1,117 @@
+# 
+# Documentation's example plot settings
+#
+# A "default" key allows setting default values for each entry. All the
+# settings defined there can be overriden for specific methods.
+#
+# Each entry needs the following keys:
+#   trace: a trace *object* (use !call or !var tags provided by lisa.utils.Serializable)
+#   kwargs [optional]: keyword arguments to pass to the plot function. This may
+#       be mandatory for some plot functions that need a task or cpu name for
+#       example. Bear in mind that YAML is "typed", so 0 is not the same as "0"
+#   hide [optional]: Do not produce an example plot, in case no trace with
+#   necessary events is available for example.
+
+
+doc-plot-conf:
+    plots:
+        # Default values for the other entries
+        default:
+            trace: &trace0 !call:lisa.trace.Trace
+                trace_path: !env:interpolate $LISA_HOME/doc/traces/trace.dat
+                plat_info: !call:lisa.platforms.platinfo.PlatformInfo.from_yaml_map
+                    path: !env:interpolate $LISA_HOME/doc/traces/plat_info.yml
+
+        TasksAnalysis.plot_task_residency:
+            kwargs:
+                task: &task big_0
+
+        TasksAnalysis.plot_task_total_residency:
+            kwargs:
+                task: *task
+
+        TasksAnalysis.plot_task_activation:
+            kwargs:
+                task: *task
+
+        LoadTrackingAnalysis.plot_task_signals: 
+            kwargs:
+                task: *task
+
+        LoadTrackingAnalysis.plot_task_required_capacity: 
+            kwargs:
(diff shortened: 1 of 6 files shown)


lisa/doc/helpers.py
```diff
@@ -187,10 +193,56 @@ def autodoc_process_analysis_events(app, what, name, obj, options, lines):
...
+
+    print('Generating plot for {}'.format(obj.__qualname__))
+    rst_figure = TraceAnalysisBase.call_on_trace(obj, trace, {
+        'output': 'rst',
+        'always_save': False,
+        # avoid memory leaks
+        'interactive': False,
+        **kwargs
+    })
+    rst_figure = '\n\nExample plot:\n{}'.format(rst_figure)
```
valschneider:
```suggestion
    rst_figure = '\n:Example plot:\n\n{}'.format(rst_figure)
```
This should give us a header like we have for paremeters / required trace events (right now it's just plain black text on white background).

doc/plot_conf.yml
```diff
@@ -0,0 +1,117 @@
...
+        # Default values for the other entries
+        default:
+            trace: &trace0 !call:lisa.trace.Trace
+                trace_path: !env:interpolate $LISA_HOME/doc/traces/trace.dat
+                plat_info: !call:lisa.platforms.platinfo.PlatformInfo.from_yaml_map
+                    path: !env:interpolate $LISA_HOME/doc/traces/plat_info.yml
+
+        TasksAnalysis.plot_task_residency:
+            kwargs:
+                task: &task big_0
```
valschneider: Not a YAML guru - is there a clean way to declare those handles outside of a plot entry? IMO it would read better (maybe something like a "plot_parameters" entry). If that's too convoluted then that's fine, I'm not too attached to the idea.

doc/plot_conf.yml
```diff
@@ -0,0 +1,117 @@
...
+#   hide [optional]: Do not produce an example plot, in case no trace with
+#   necessary events is available for example.
+
+
+doc-plot-conf:
+    plots:
+        # Default values for the other entries
+        default:
+            trace: &trace0 !call:lisa.trace.Trace
+                trace_path: !env:interpolate $LISA_HOME/doc/traces/trace.dat
```
valschneider: That trace is pretty big (~7MB). The trace we use for Travis (`tests/assets/trace.txt`) is ~.5MB. It would be nice to consolidate these into a single trace, and if possible a not-too-big one. Although in the end both the `tests` and `doc` directories are outside of the `lisa` module directory, so it probably doesn't matter *too* much.

douglas-raillard-arm: I already cropped the trace to only contain the interesting (rt-app) section. The size comes from the fact lots of events were collected, so we can plot lots of things. As a compensation, I removed another unused trace that was ~90MB without any interesting events beyond sched_switch ( #981 )

tests/assets/trace.txt only contains sched_switches and wakeups, and 3 instances of sched_overutilized, so it cannot be used for plotting anything interesting.

doc/plot_conf.yml
douglas-raillard-arm: It can be defined outside, but I'd rather not pollute the file with extra keys that don't have direct influence (merely adding something in "plot_parameters" will not trigger anything).
Maybe we could establish a convention across LISA conf files and "reserve" a top-level `scratch` key if we want to do this kind of things.

doc/plot_conf.yml
valschneider: I also forgot that some tests are hardcoded to work with the test trace (e.g. overutilized duration), so it's probably best to keep test and doc traces separate anyway.

PR #984, opened 2019-07-22
rta: Don't force running rta as root for calibration

JaviMerino:
You don't need to run it as root. I have tested it on an unrooted phone.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -345,7 +345,7 @@ class RTA(Workload):
                                  profile={'task1': calib_task},
                                  calibration="CPU{}".format(cpu),
                                  res_dir=res_dir)
-            rta.run(as_root=True)
+            rta.run()
 
             for line in rta.output.split('\n'):
                 pload_match = re.search(pload_regexp, line)


derkling:
Mmm... AFAIR we use to run it as root to ensure we can run it as an RT task:
   https://github.com/ARM-software/lisa/blob/legacy/libs/wlgen/wlgen/rta.py#L129

That's the reason why we also compute `max_rtprio`, which is now a not user variable.

That looks like a nonintentional change since the "big switch", maybe @valschneider and/or
@douglas-raillard-arm know why? Running calibration as an RT task was intended to avoid interferences and get more consistent results.

douglas-raillard-arm:
I'll leave that to @valschneider as he refactored wlgen, but that seems not intentional indeed. We should also probably use the freezer as well to make sure nothing disturbs the calibration loop.

JaviMerino:
Ok, but now that the calibration can be done in non-rooted devices, it would be great if the RT task and the freezer was only done if `target.is_rooted`.

valschneider:
Right so that change was unintentional. I did run some side-by-side comparison and didn't notice differences in calibration values back at the time, so hopefully we haven't been too far-off.

IMO using the freezer only makes sense if we're not using the RT class, so we could go with:
- If we are rooted, go with RT class for the calibration task
- If we aren't rooted, go for default (CFS) class + freezer (and warn about !rooted status).

We already have something similar for cpufreq - if we don't have cpufreq available, which is the cases in most VMs, we just warn about it and keep going with the calibration.

douglas-raillard-arm:
We should probably expose in the target configuration if the target user has sudo rights or not, but I'll need to dive again in devlib to see how exactly that should be done. Last time I checked the rooted/unrooted statuses management were very confusing.

Once that info is exposed, various pieces of code could bail-out and warn if the target user is not sudo user, like the freezer.

douglas-raillard-arm:
> If we aren't rooted, go for default (CFS) class + freezer (and warn about !rooted status).

Will we be able to freeze everything without being root ?

JaviMerino:
> We should probably expose in the target configuration if the target user has sudo rights or not, but I'll need to dive again in devlib to see how exactly that should be done. Last time I checked the rooted/unrooted statuses management were very confusing.

devlib already exposes it in its Target class: it's got an `is_rooted` property. Can't you use that?

douglas-raillard-arm:
Looks like `is_rooted` should do the job, looks like my issue was that the host connection does not try to check if the username is already "root".

valschneider:
> > If we aren't rooted, go for default (CFS) class + freezer (and warn about !rooted status).
> 
> Will we be able to freeze everything without being root ?

Good point, I'm not sure. We already peacefully continue when we can't freeze because the freezer cgroup isn't there, so maybe we could do that there. I sense a "do this as root or skip it" building block....

douglas-raillard-arm:
Here is the PR that restores RT and userspace freezing when running calibration, and warns but proceeds when ran as non-sudo user:
https://github.com/ARM-software/lisa/pull/994

@JaviMerino If that fits the bill, you can close this PR

PR #985, opened 2019-07-22
Cleanup EM loading

douglas-raillard-arm:
Add `LinuxEnergyModel(EnergyModel)` and `LegacyEnergyModel(EnergyModel)` sublcasses of EnergyModel, so that each of them can implement their own `from_target()` and `check_from_target()` methods. 

`EnergyModel.from_target()` autodetects the appropriate subclass to use, so it's completely transparent.

This split can allow separate implementation of other methods as well, notably the ones dealing with idle states, allowing for potential speed-up in LinuxEnergyModel since this class registers a 0 cost for idle states/

### DIFF
--- a/lisa/energy_model.py
+++ b/lisa/energy_model.py
@@ -923,74 +923,6 @@ class EnergyModel(Serializable, Loggable):
                    freq_domains=perf_domains)
 
     @classmethod
-    def from_sysfsEM_target(cls, target,
-            directory='/sys/devices/system/cpu/energy_model'):
-        """
-        Create an EnergyModel by reading a target filesystem on a device with
-        the new Simplified Energy Model present in sysfs.
-
-        The patches exposing the Energy Model in sysfs have been abandonned
-        and this way of loading it is now deprecated.
-
-        :param target: Devlib target object to read filesystem from. Must have
-                       cpufreq and cpuidle modules enabled.
-        :returns: Constructed EnergyModel object based on the parameters
-                  reported by the target.
-        """
-        warnings.warn('The Energy Model in sysfs is DEPRECATED. Please use debugfs instead.',
-                      DeprecationWarning)
-
-        if 'cpuidle' not in target.modules:
-            raise TargetStableError('Requires cpuidle devlib module. Please ensure '
-                               '"cpuidle" is listed in your target/test modules')
-
-        # Simplified EM on-disk format (for each frequency domain):
-        #    /sys/devices/system/cpu/energy_model/<frequency_domain>/..
-        #        ../capacity
-        #           contains a space-separated list of capacities in increasing order
-        #        ../cpus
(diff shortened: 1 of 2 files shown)


lisa/energy_model.py
```diff
@@ -341,6 +341,12 @@ class EnergyModel(Serializable, Loggable):
         distinct estimation for each item in the series.
     """
 
+    _AUTODETECT_ORDER = None
+    """
+    Order in which subclasses are tried when auto-detecting the kind of energy
+    model to load from a target.
+    """
```
qperret: Do we care about the order ? We really don't expect having kernels with multiple EMs, so we should really only match one. But no big deal.

lisa/energy_model.py
```diff
@@ -758,107 +764,163 @@ def get_optimal_placements(self, capacities, capacity_margin_pct=0):
...
+    @classmethod
+    def from_target(cls, target):
+        """
+        Create an instance of (a subclass of) :class:``EnergyModel`` by reading a target filesystem.
 
-        ret = set()
+        If present, load an EM provided via dt using from_sd_target, since these
+        devices make the full EM available via the sched domain in sysfs. If
+        there is no EM at this location, attempt to load the simplified EM
+        made available via dedicated sysfs files.
```
qperret: I see this was an existing comment, but it is broken no ? We seem to look at the simplified EM first (according to `_AUTODETECT_ORDER`), and _then_ to the full/legacy EM. Or maybe I got that wrong ?

Also, I think we should say the simple EM is exposed in _debugfs_, and the legacy EM in _procfs_, rather than sysfs.

lisa/energy_model.py
```diff
@@ -920,6 +919,13 @@ def f(input_row):
 
         return inputs.apply(f, axis=1)
 
+    @classmethod
+    def _get_idle_states_name(cls, target, cpu):
+        if 'cpuidle' in target.modules:
+            return [s.name for s in target.cpuidle.get_states(cpu)]
+        else:
+            return ['placeholder-idle-state']
```
qperret: :heart:

lisa/energy_model.py
douglas-raillard-arm: It's mostly to get reproducible results and to speed up the common case, otherwise the order will be random, which is rarely a good idea. Other than that, I agree, the probe functions should only return true for the kernel they handle and False for the others

lisa/energy_model.py
douglas-raillard-arm: Yes, I replaced it by a .. seealso pointing at the others from_target()

douglas-raillard-arm:
Updated with:
* PR comments
* s/check_from_target/probe_target/ to make it more explicit, and s/_AUTODETECT_ORDER/_PROBE_ORDER/

PR #994, opened 2019-07-25
improve calibration reproducability: freeze userspace and run as RT

douglas-raillard-arm:


lisa/target.py
```diff
@@ -668,27 +668,32 @@ def freeze_userspace(self):
         logger = self.get_logger()
         if 'cgroups' not in self.target.modules:
             raise RuntimeError(
-                'Failed to freeze userspace. Ensure "cgroups" module is listed '
-                'among modules in target/test configuration')
+                'Failed to freeze userspace. Ensure "cgroups" devlib module is loaded.')
 
         controllers = [s.name for s in self.target.cgroups.list_subsystems()]
         if 'freezer' not in controllers:
             logger.warning('No freezer cgroup controller on target. '
```
valschneider: I know it's slightly ridiculous, but the discrepancy between the warning messages drives me nuts and I can't unsee it. Could you align them?

lisa/wlgen/rta.py
```diff
@@ -329,23 +329,39 @@ def _calibrate(cls, target, res_dir):
...
+        if target.is_rooted:
+            max_rtprio = int(target.execute('ulimit -Hr').split('\r')[0])
+            logger.debug('Max RT prio: %d', max_rtprio)
 
-        if max_rtprio > 10:
-            max_rtprio = 10
+            if max_rtprio > 10:
+                max_rtprio = 10
+
+            sched = {
```
valschneider: Urgh, the docstring is out of date - I killed the `sched` dict and split it into `sched_policy` and `priority` parameters.

lisa/wlgen/rta.py
```diff
@@ -329,23 +329,39 @@ def _calibrate(cls, target, res_dir):
...
+                'policy': 'FIFO',
+                'prio': max_rtprio,
+            }
+        else:
+            logger.warning('Will use CFS scheduler class instead of RT since the target is not rooted')
```
valschneider: Nit: unlikely to diverge but hey :man_shrugging: 
```suggestion
            logger.warning('Will use default scheduler class instead of RT since the target is not rooted')
```

lisa/target.py
douglas-raillard-arm: me too actually, let's fully give into our OCD :)

lisa/wlgen/rta.py
douglas-raillard-arm: I added a commit to fix the docstring

lisa/target.py
```diff
@@ -667,28 +667,31 @@ def freeze_userspace(self):
...
+            @contextlib.contextmanager
+            def cm():
+                logger.info('Freezing all tasks except: %s', ','.join(exclude))
+                try:
+                    yield self.target.cgroups.freeze(exclude)
+                finally:
+                    logger.info('Un-freezing userspace tasks')
+                    self.target.cgroups.freeze(thaw=True)
 
+        with cm() as x:
```
valschneider: Following the context manager flow hurt my brain a bit but I think I got it now...

douglas-raillard-arm: Context-manager composition 101:
```
@contextmanager
def cm1():
    yield

@contextmanager
def cm2():
    yield

@contextmanager
def cm3():
    with cm1:
        with cm2:
           yield
```

At the end, it's not much different than chaining function calls, with `return f()` replaced by `with f(): yield`

PR #999, opened 2019-07-26
Rework task selection

### LINKED ISSUE #926: Task names and PID mapping
The current identification of task names and mapping of names to PID is rather simplistic since it assumes that a task changes its name at most one time after being created. Thus, we assume the name for a give PID is the very last `comm` value we get from a `sched_switch` event.

Reality is that a given task can change its name multiple times or that the same PID can be re-used for different tasks.

A better solution would be to consider `sched_wakeup_new` events to identify when a new task is created and to ensure the correct name is assigned to each PID. In case more than two names are used for the same task we should also warn the user at the very least.

Other details have been discussed in #880

douglas-raillard-arm:
Fix #926

Make the task selection/resolution methods more consistent and strict when used with task name or task PID. This assumes that the simple comm/PID interface is only meant for non-ambigous cases.

Introduce a new way of selecting tasks by using `TaskID(pid, comm)` named tuples. It allows disambiguation in tricky cases. This include:
* when just specifying a comm or a PID is ambiguous, you can specify both to remove ambiguity
* when a specific PID is of interest regardless its task comm (or the opposit), the irrelevant field can be set to ``None``. This gives an explicit way of avoiding the exception that would be raised when specifying an PID or comm alone (i.e. not in a tuple).

### DIFF
--- a/lisa/analysis/latency.py
+++ b/lisa/analysis/latency.py
@@ -208,7 +208,7 @@ class LatencyAnalysis(TraceAnalysisBase):
                 df.plot(ax=axis, style='+', label="Preemption")
 
 
-        axis.set_title("Latencies of task \"{}\"".format(task))
+        axis.set_title('Latencies of task "{}"'.format(task))
         axis.set_ylabel("Latency (s)")
         axis.legend()
         axis.set_xlim(self.trace.start, self.trace.end)
@@ -278,7 +278,7 @@ class LatencyAnalysis(TraceAnalysisBase):
         axis.axvspan(0, threshold_s, facecolor=self.LATENCY_THRESHOLD_ZONE_COLOR,
                      alpha=0.5, label="{}ms threshold zone".format(threshold_ms));
 
-        axis.set_title("Latencies CDF of task \"{}\"".format(task))
+        axis.set_title('Latencies CDF of task "{}"'.format(task))
         axis.set_xlabel("Latency (s)")
         axis.set_ylabel("Latencies below the x value (%)")
         axis.legend()
@@ -311,7 +311,7 @@ class LatencyAnalysis(TraceAnalysisBase):
         axis.axvspan(0, threshold_s, facecolor=self.LATENCY_THRESHOLD_ZONE_COLOR, alpha=0.5,
                      label="{}ms threshold zone".format(threshold_ms));
 
-        axis.set_title("Latencies histogram of task \"{}\"".format(task))
+        axis.set_title('Latencies histogram of task "{}"'.format(task))
         axis.set_xlabel("Latency (s)")
         axis.legend()
 
@@ -363,7 +363,7 @@ class LatencyAnalysis(TraceAnalysisBase):
         if self.trace.has_events(plot_overutilized.used_events):
             plot_overutilized(axis=axis)
 
-        axis.set_title("Activation intervals of task \"{}\"".format(task))
+        axis.set_title('Activation intervals of task "{}"'.format(task))
 
         axis.set_xlim(self.trace.start, self.trace.end)
 
@@ -384,7 +384,7 @@ class LatencyAnalysis(TraceAnalysisBase):
         if self.trace.has_events(plot_overutilized.used_events):
             plot_overutilized(axis=axis)
 
-        axis.set_title("Per-activation runtimes of task \"{}\"".format(task))
+        axis.set_title('Per-activation runtimes of task "{}"'.format(task))
 
         axis.set_xlim(self.trace.start, self.trace.end)
 
--- a/lisa/analysis/latency.py
+++ b/lisa/analysis/latency.py
@@ -45,7 +45,7 @@ class LatencyAnalysis(TraceAnalysisBase):
         DataFrame of a task's wakeup latencies
 
         :param task: The task's name or PID
-        :type task: int or str
+        :type task: int or str or tuple(int, str)
 
         :returns: a :class:`pandas.DataFrame` with:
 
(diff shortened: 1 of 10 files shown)


lisa/analysis/latency.py
```diff
@@ -45,7 +45,7 @@ def df_latency_wakeup(self, task):
         DataFrame of a task's wakeup latencies
 
         :param task: The task's name or PID
-        :type task: int or str
+        :type task: int or str or tuple(int, str)
```
valschneider: Even though it's supposed to be equivalent, wouldn't it better to point to `TaskID` ?

lisa/trace.py
```diff
@@ -34,15 +32,43 @@
...
 
 from lisa.utils import Loggable, HideExekallID, memoized, deduplicate, deprecate
 from lisa.platforms.platinfo import PlatformInfo
 from lisa.conf import SimpleMultiSrcConf, KeyDesc, TopLevelKeyDesc, StrList, Configurable
-import devlib
-from devlib.target import KernelVersion
 
 NON_IDLE_STATE = -1
 
+class TaskID(namedtuple('TaskID', ('pid', 'comm'), module=__name__)):
```
valschneider: `module` was introduced in Python 3.6 and thus fails with Travis.

lisa/analysis/load_tracking.py
```diff
@@ -230,7 +228,9 @@ def df_top_big_tasks(self, util_threshold, min_samples=100):
         samples = samples.sort_values(ascending=False)
 
         top_df = pd.DataFrame(samples).rename(columns={"util" : "samples"})
-        top_df["comm"] = top_df.index.map(self.trace.get_task_by_pid)
+        def get_name(pid):
+            return self.trace.get_task_pid_names(pid)[-1]
```
valschneider: `get_task_pid()` (which should be used by most analysis functions ATM) returns `pid_list[0]` AKA the first PID.
Depending on what you're looking at, either makes sense, but it might better to keep that change for a separate PR.

lisa/trace.py
```diff
@@ -537,44 +600,106 @@ def get_task_by_pid(self, pid):
...
+        """
+        Helper that resolves a task PID or name to a :class:`TaskID`.
+
+        :param task: Either the task name, the task PID, or a tuple ``(pid, comm)``
+        :type task: int or str or tuple(int, str)
+
+        :param update: If a partially-filled :class:`TaskID` is passed (one of
+            the fields set to ``None``), returns a complete :class`TaskID`
+            instead of leaving the ``None`` fields.
+        :type update: bool
```
valschneider: Maybe add something along the lines of
```
:raises ValueError: if more than one task matches - see `get_task_ids()` for a non exception-raising version
```

lisa/analysis/latency.py
douglas-raillard-arm: The intention was to let people pass actual tuples, not necessarily `TaskID` (although they can). Since that is an opaque value, it the only thing that matters is that `get_task_id(s)` knows how to handle it, and then always return a `TaskID` so the rest of the code only needs to know how to manipulate `TaskID`.

lisa/trace.py
douglas-raillard-arm: Now that I think about it, `module` is useless here since we have a proper class declaration anyway

lisa/analysis/load_tracking.py
douglas-raillard-arm: Looks like you confused `get_task_by_pid()` with`get_task_pid()` :)

Regarding `get_task_pid()`, it actually returns the last PID of the first task that ran with the given name,  because of `self._tasks_by_pid = (df.drop_duplicates(subset=pid_key, keep='last')`
So if you had:
```
PID NAME
1 sh
2 sh
2 mycommand
2 mycommand_renamed_itself
3 sh
3 mycommand2
3 mycommand_renamed_itself
```
You would get:
```
# API deprecated by this PR
get_task_pid("mycommand_renamed_itself") == 2 # interesting but irrelevant in our case
get_task_by_pid(2) == "mycommand_renamed_itself" # this one is relevant for this comment

# New API
get_task_pid_names(2)[-1] == "mycommand_renamed_itself"
``

So here we can see that `get_task_by_pid(pid)` can be correctly replaced by `get_task_pid_names(pid)[-1]`.

Comment (valschneider) [reply lisa/analysis/load_tracking.py]: > Looks like you confused `get_task_by_pid()` with`get_task_pid()` :)
> 

No, but I did forget about the `drop_duplicates(..., keep='last)`. It's been there for some time, but I swear I've seen the warning fire at least once in `get_task_pid()`. I don't really know how to make sense of that with the `drop_duplicates()`.

Comment (douglas-raillard-arm) [reply lisa/analysis/load_tracking.py]: Ah ok, I was under that impression since the hunk you commented on does not contain any reference to `get_task_pid()`.

In the example I provided, that should give the warning you talk about:
```
get_task_pid("mycommand_renamed_itself")
```
The `drop_duplicates(pid, keep='last')` will turn the name table into:
```
1 sh
2 mycommand_renamed_itself
3 mycommand_renamed_itself
```
So the `pid_list` in `get_task_pid("mycommand_renamed_itself")` will be set to `[2, 3]`, which will give the warning and return `[2,3][0] == 2`. This choice of returning `2` seems completely arbitrary, as there is nothing more interesting in PID 2 than in 3 (they are different tasks, probably sharing a common goal and even code, but instantiated at two unrelated point in time). That really should raise an exception, which it will do after this pull request (`get_task_pid()` has been re-implemented in terms of `get_task_id()`, which is strict by default).

PR #1002, opened 2019-07-30
Fix load_tracking InvarianceItem workload

douglas-raillard-arm:
* Fix wlgen run parameter handling
* Make sure the workload can execute at the target OPP

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -420,7 +420,7 @@ class Phase(Loggable):
     :type period_ms: int
 
     :param duty_cycle_pct: the generated load in [%].
-    :type duty_cycle_pct: int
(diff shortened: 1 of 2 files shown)


lisa/wlgen/rta.py
```diff
@@ -492,7 +492,9 @@ def get_rtapp_repr(self, task_name):
                              running_time, sleep_time)
 
             phase['loop'] = cloops
-            phase['run'] = running_time
+            # rtapp fails to handle floating values correctly
+            # https://github.com/scheduler-tools/rt-app/issues/82
+            phase['run'] = int(running_time)
```
valschneider: Might want to do that for `sleep_time` and `running_time` above directly, so that we always log int values as well.

PR #1003, opened 2019-07-30
lisa.tests.cpufreq.sanity: Use separate res_dir for each sysbench inv…

douglas-raillard-arm:
…ocation

Avoid overwriting output.log files when using a single res_dir. Fix that by
making a separate res_dir for each invocations.

### DIFF
--- a/lisa/tests/cpufreq/sanity.py
+++ b/lisa/tests/cpufreq/sanity.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import os
 
 from lisa.tests.base import Result, ResultBundle, TestBundle
 from lisa.wlgen.sysbench import Sysbench
@@ -47,7 +48,6 @@ class UserspaceSanity(TestBundle):
         governor
         """
         cpu_work = {}
-        sysbench = Sysbench(target, "sysbench", res_dir)
 
         with target.cpufreq.use_governor("userspace"):
             for domain in target.cpufreq.iter_domains():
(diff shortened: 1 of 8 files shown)


lisa/tests/cpufreq/sanity.py
```diff
@@ -14,51 +14,81 @@
...
+        self.cpu = cpu
+        self.freq = freq
+        self.work = work
+
+    @classmethod
+    def _from_target(cls, target, res_dir, cpu, freq):
+        sysbench = Sysbench(target, "sysbench", res_dir)
+        work = None
+
+        with target.cpufreq.use_governor("userspace"):
```
valschneider: We used to have one governor change per freq domain, now it's one per frequency - that's going to considerably slow down the execution of this test. We could either move the governor (and frequency?) change to `UserspaceSanity`, or have both options and add a boolean parameter in to `UserspaceSanityItem._from_target()` to control whether the governor switch is done by the caller.

douglas-raillard-arm: It does double the execution time on Pixel 3. Just moving the change to the other level is not really a good idea, as it would prevent correct operations when creating an item for a given CPU directly, which is the whole point of this PR (would need some annotations for that, I'll add them after #1018 is merged).

A third option that does not require making the API more clunky is to just wrap the target with a cpufreq function not switching a governor to the same.

douglas-raillard-arm:
Updated with a `switch_governor` flag to disable switching inside the item class.

PR #1005, opened 2019-07-30
lisa.target: Add Target.from_custom_cli()

douglas-raillard-arm:
Allow registering extra command line parameters, allowing to easily
create a script that takes some domain-specific parameters, to share a
reproducer for example.

### DIFF
--- a/lisa/target.py
+++ b/lisa/target.py
@@ -350,7 +350,18 @@ class Target(Loggable, HideExekallID, Configurable):
         return cls.from_conf(conf=conf, plat_info=plat_info)
 
     @classmethod
-    def from_cli(cls, argv=None) -> 'Target':
+    def from_cli(cls, *args, **kwargs) -> 'Target':
+        """
+        Same as :meth:`from_custom_cli` without the custom parameters
+        capabilities.
+
+        :return: A connected :class:`Target`
+        """
+        args, target = cls.from_custom_cli(*args, **kwargs)
+        return target
+
+    @classmethod
+    def from_custom_cli(cls, argv=None, params=None) -> 'Target':
         """
         Create a Target from command line arguments.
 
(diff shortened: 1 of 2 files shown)


lisa/target.py
```diff
@@ -350,17 +350,34 @@ def from_one_conf(cls, path):
         return cls.from_conf(conf=conf, plat_info=plat_info)
 
     @classmethod
-    def from_cli(cls, argv=None) -> 'Target':
```
valschneider: What about just having a single `from_cli()` ? It's *slightly* (only slightly, I promise) ugly but if we pass `params=None` then we could just return `target`, otherwise we'd return `args, target`.

lisa/target.py
```diff
@@ -350,17 +350,34 @@ def from_one_conf(cls, path):
...
-        to be confusing (help message woes, argument clashes...), so for now
-        this should only be used in scripts that only expect Target args.
+        :param params: Dictionary of custom parameters to add to the parser. It
+            is in the form of
+            ``{param_name: {dict of ArgumentParser.add_argument() options}}``.
+        :type params: dict(str, dict)
+
+        :return: A tuple ``(args, target)``
+
+        .. note:: This method should not be relied upon to implemnt long-term
```
valschneider:
```suggestion
        .. note:: This method should not be relied upon to implement long-term
```

lisa/target.py
```diff
@@ -350,17 +350,34 @@ def from_one_conf(cls, path):
...
+            scripts, it's more designed for quick scripting.
```
valschneider: also, with this change, is it really still the case?

lisa/target.py
```diff
@@ -416,31 +428,62 @@ def from_cli(cls, argv=None) -> 'Target':
...
+        })
+
+        # Some sanity check to get better error messages
+        if not target_conf:
+            parser.error('--conf with target configuration or any of the connection options is required')
+
         if args.kind == 'android':
-            if not (args.host or args.device):
+            if ('host' not in target_conf) and ('device' not in target_conf):
+                breakpoint()
```
valschneider: Stray debug code?

lisa/target.py
douglas-raillard-arm: I'd rather avoid having this kind of ad-hoc polymorphism the wrong way around, it can quickly leads to very strange errors that are very hard to debug. The input data should drive the path taken by the control flow, but not the code structure itself.

lisa/target.py
douglas-raillard-arm: indeed

lisa/target.py
douglas-raillard-arm: Before removing that comment, I'd provide an API that works as:
```
parser = LisaArgumentParser(...)
# Target would implement `get_argparse_arguments()` and would build instances with `from_argparse_args()` or something similar
parser.register_class(Target)
...
args = parser.parse_args()
...
target = args.get_instance(Target)
```

This way, the user has full control of the parser, can set the description, the formatter etc. We can also add more "hooks" like this one for `FtraceCollector` for example. They would cooperate for the `--conf` parameter and otherwise add their stuff to the parser.

The issue with that API is that it takes 3 lines for what we currently can do in 1, but that is a much saner way to build a real script out of LISA lego blocks

PR #1011, opened 2019-08-05
lisa.tests.cpufreq.schedutil: Add LargeStepUp test for ramp boost fea…

douglas-raillard-arm:
…ture

Provide a test for the ramp boost feature as posted on LKML.
This test will mostly catch cases where schedutil is not called as often as
expected, since it reads most of the signals from the kernel instead of guessing
them.

### DIFF
--- a/lisa/tests/staging/schedutil.py
+++ b/lisa/tests/staging/schedutil.py
@@ -1,0 +1,273 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2019, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from math import ceil
+import os
+
+import pandas as pd
+
+from lisa.wlgen.rta import Ramp
+from lisa.tests.base import TestBundle, ResultBundle, Result, RTATestBundle
+from lisa.target import Target
+from lisa.trace import requires_events, FtraceCollector, FtraceConf
+from lisa.datautils import df_merge, series_mean
+from lisa.utils import ArtifactPath
+
+from lisa.analysis.frequency import FrequencyAnalysis
+from lisa.analysis.load_tracking import LoadTrackingAnalysis
+from lisa.analysis.rta import RTAEventsAnalysis
+
+
+class RampBoostTestBase(RTATestBundle):
+    """
+    Test schedutil's ramp boost feature.
+    """
+
+    def __init__(self, res_dir, plat_info, cpu):
+        super().__init__(res_dir, plat_info)
+        self.cpu = cpu
+
+    @requires_events('cpu_idle', 'cpu_frequency', 'sched_wakeup')
+    def estimate_nrg(self):
+        return self.plat_info['nrg-model'].estimate_from_trace(self.trace).sum(axis=1)
+
+    def get_avg_slack(self, only_negative=False):
+        analysis = self.trace.analysis.rta
+        def get_slack(task):
+            series = analysis.df_rtapp_stats(task)['slack']
+            if only_negative:
+                series = series[series < 0]
+
+            if series.empty:
+                return 0
+            else:
+                # average negative slack across all activations
+                return series.mean()
+
+        return {
+            task: get_slack(task)
+            for task in self.trace.analysis.rta.rtapp_tasks
+        }
+
+    @LoadTrackingAnalysis.df_cpus_signal.used_events
+    @requires_events('schedutil_em')
+    def df_ramp_boost(self):
+        """
+        Return a dataframe with schedutil-related signals, sampled at the
+        frequency decisions timestamps for the CPU this bundle was executed on.
+
+        .. note:: The computed columns only take into account the CPU the test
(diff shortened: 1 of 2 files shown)


lisa/tests/staging/schedutil.py
```diff
@@ -0,0 +1,271 @@
...
+
+from lisa.analysis.rta import PerfAnalysis
+from lisa.analysis.frequency import FrequencyAnalysis
+from lisa.analysis.load_tracking import LoadTrackingAnalysis
+
+
+class RampBoostTestBase(RTATestBundle):
+    """
+    Test schedutil's ramp boost feature.
+
```
valschneider:
```suggestion
```

lisa/tests/staging/schedutil.py
```diff
@@ -0,0 +1,271 @@
...
+    def __init__(self, res_dir, plat_info, cpu):
+        super().__init__(res_dir, plat_info)
+        self.cpu = cpu
+
+    @requires_events('cpu_idle', 'cpu_frequency', 'sched_wakeup')
+    def estimate_nrg(self):
+        return self.plat_info['nrg-model'].estimate_from_trace(self.trace).sum(axis=1)
+
+    def get_avg_slack(self, only_negative=False):
+        pa = PerfAnalysis.from_dir(self.res_dir)
```
valschneider: I'm late to the party but this'll have to be updated to the new ftrace-based analysis :(

lisa/tests/staging/schedutil.py
```diff
@@ -0,0 +1,271 @@
...
+        )
+        df['boost_points'] = boost_points
+
+        df['expected_cost_margin'] = (df['util'] - df['util_est_enqueued']).where(
+            cond=boost_points,
+            other=0,
+        )
+
+        # cost_margin values range from 0 to 1024
+        ENERGY_SCALE = 1024
```
valschneider: Is that meant to be the same scale as `SCHED_CAPACITY_SCALE`, or is just a [happy little accident](https://www.youtube.com/watch?v=_Tq5vXk0wTk)?

lisa/tests/staging/schedutil.py
```diff
@@ -0,0 +1,271 @@
...
+        return axis
+
+    @requires_events('sched_switch')
+    def trace_window(self, trace):
+        """
+        Skip the first phase in the trace, since it is heavily influenced by
+        what was immediately preceding it.
+        """
+        profile = self.rtapp_profile
+        phase_duration = max(
```
valschneider: This can probably be moved over to the new phase windows API. However, seeing as you've already tested the thing, I wouldn't mind if this was merged in staging as-is. We can then do a batch convert (misfit needs that treatment too).

lisa/tests/staging/schedutil.py
```diff
@@ -0,0 +1,271 @@
...
+        assert not (df['expected_cost_margin'] < 0).any()
+        assert not (df['cost_margin'] < 0).any()
+
+        # "rect" method is accurate here since the signal is really following
+        # "post" steps
+        expected_boost_nrg = series_mean(df['expected_cost_margin'])
+        actual_boost_nrg = series_mean(df['cost_margin'])
+
+        # Check that the total amount of boost is close to expectations
+        lower = max(0, expected_boost_nrg - nrg_threshold_pct)
```
valschneider: Shouldn't that `nrg_threshold_pct` be multiplied to something?

lisa/tests/staging/schedutil.py
```diff
@@ -0,0 +1,271 @@
...
+        good_shape_nr = (df['cost_margin'] <= df['expected_cost_margin']).sum()
+
+        df_len = len(df)
+        bad_shape_nr = df_len - good_shape_nr
+        bad_shape_pct = bad_shape_nr / df_len * 100
+
+        # Tolerate a few bad samples that added too much boost
+        passed_shape = bad_shape_pct < bad_samples_threshold_pct
+
+        passed = passed_overhead and passed_shape
```
valschneider: Would it make sense to split these into two separate tests? We'd get finer grained results.

lisa/tests/staging/schedutil.py
```diff
@@ -0,0 +1,271 @@
...
+        plat_info = target.plat_info
+
+        # Use a big CPU by default to allow maximum range of utilization
+        cpu = cpu if cpu is not None else plat_info["capacity-classes"][-1][0]
+
+        rtapp_profile = cls.get_rtapp_profile(plat_info, cpu, nr_steps)
+
+        # Ensure accurate duty cycle and idle state misprediction on some
+        # boards. This helps having predictable execution.
+        with target.disable_idle_states():
```
valschneider: We keep carrying over an explanation as to why we do this... Perhaps we should stick it in a load-tracking specific "I care about signals so give me a white room` context manager.

lisa/tests/staging/schedutil.py
douglas-raillard-arm: That's the scale used to represent "per-1024-age" of costs as defined in:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/energy_model.h?h=v5.3#n17

That's not homogeneous to a capacity

lisa/tests/staging/schedutil.py
douglas-raillard-arm: Yes, we will go over the tests and update them with the new API once merged

lisa/tests/staging/schedutil.py
douglas-raillard-arm: I'm not sure if that would really help, all these checks are just checking the overall shape of the signal. If I could simulate the signals with 100% accuracy, they would all go away and replaced by a Series equality check.

lisa/tests/staging/schedutil.py
douglas-raillard-arm: Good cleanup idea

lisa/tests/staging/schedutil.py
douglas-raillard-arm: No, these series are already in percents

PR #1012, opened 2019-08-05
lisa.target: Avoid freezing the main process on LocalLinuxTarget

douglas-raillard-arm:
This allows using the freezing feature on the host, while still being able to
unfreeze at the end. Output usually freezes as well if it comes in a virtual
terminal, but otherwise, things do work.


douglas-raillard-arm:
Yes, maybe we should add some platform info keys to let the user customize that depending on the system being used, with some defaults depending on the target type ...

PR #1013, opened 2019-08-05
Prepare lisa for PyPI submission

douglas-raillard-arm:
Make sure `lisa` python package is self-contained, so that it will be installable from PyPI in the future.

TODO (in another PR):
* remove last dependency over bart
* update tools/{exekall,bisector}/setup.py so they can be uploaded as well
* Allow giving a package name instead of a filename to exekall, so the user does not need to reference site-package directory directly


shell/lisa_shell
```diff
@@ -42,8 +42,8 @@ export LISA_PRESERVE_SHELL=${LISA_PRESERVE_SHELL:-0}
 # Setup colors
 source "$LISA_HOME/shell/lisa_colors"
 
-export _DOC_LISA_HOST_ABI="Add some shell utilities to the PATH based on the host ABI. Priority is determined by LISA_USE_SYSTEM_BIN"
-export LISA_HOST_ABI=${LISA_HOST_ABI:-x86_64}
+export _DOC_LISA_HOST_ABI="Add some shell utilities to the PATH, based on the host ABI. Priority is determined by LISA_USE_SYSTEM_BIN""
+export LISA_HOST_ABI=${LISA_HOST_ABI:-$(uname -m | sed 's/aarch64/arm64/')}
```
valschneider: I was going to argue it might be simpler to keep things un-renamed (and have the binary directories straight up match `uname -m`), but from looking at https://github.com/ARM-software/devlib/blob/master/devlib/utils/misc.py#L55 I guess we do want to do it.

douglas-raillard-arm: The issue is that the Linux world for some reason decided to settle on the `arm64` name, which does not match anything defined by the ARM architecture, since it's usually used to refer to ARMv8-A, although ARMv8-A defines both aarch32 and aarch64. Since `arm64` is used almost everywhere, including Debian packaging and so on, it seems better to align with that convention.

valschneider: And we have to align with devlib anyway, so there isn't really a choice...

PR #1017, opened 2019-08-08
Add pylint conf and fix issues

douglas-raillard-arm:


douglas-raillard-arm:
I updated the PR with a comment where some linters are disabled. Otherwise, the diff can be generated using that commands, and the default may change over time, so ultimately what matters is the actual config we use (default conf is sensible but still pretty arbitrary at the end):
```
pylint --generate-rcfile > default
diff -u default .pylintrc
```

PR #1018, opened 2019-08-08
Remove `from_target` boilerplate definitions

douglas-raillard-arm:
Add `TestBundleMeta` metaclass to generate the shim automatically, by moving the annotations of `_from_target` to a shim `from_target`

### DIFF
--- a/lisa/tests/scheduler/misfit.py
+++ b/lisa/tests/scheduler/misfit.py
@@ -66,16 +66,6 @@ class MisfitMigrationBase(RTATestBundle):
         HZ = plat_info['kernel']['config']['CONFIG_HZ']
         return ((HZ * plat_info['cpus-count']) // 10) * (1. / HZ)
 
-    @classmethod
-    def from_target(cls, target:Target, res_dir:ArtifactPath=None, ftrace_coll:FtraceCollector=None) -> 'MisfitMigrationBase':
-        """
-        Factory method to create a bundle using a live target
-
-        This will execute the rt-app workload described in
-        :meth:`~lisa.tests.base.RTATestBundle.get_rtapp_profile`
-        """
-        return super().from_target(target, res_dir, ftrace_coll=ftrace_coll)
-
 class StaggeredFinishes(MisfitMigrationBase):
     """
     One 100% task per CPU, with staggered completion times.
--- a/lisa/tests/staging/sched_android.py
+++ b/lisa/tests/staging/sched_android.py
@@ -60,16 +60,7 @@ class SchedTuneItemBase(RTATestBundle):
     @classmethod
     # Not annotated, to prevent exekall from picking it up. See
     # SchedTuneBase.from_target
-    def from_target(cls, target, boost, prefer_idle, res_dir=None, ftrace_coll=None):
-        """
-        .. warning:: `res_dir` is at the end of the parameter list, unlike most
-            other `from_target` where it is the second one.
-        """
-        return super().from_target(target, res_dir, boost=boost,
-                prefer_idle=prefer_idle, ftrace_coll=ftrace_coll)
-
-    @classmethod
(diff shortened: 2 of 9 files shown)


douglas-raillard-arm:
Needs  #1020

lisa/tests/scheduler/misfit.py
```diff
@@ -66,16 +66,6 @@ def _get_max_lb_interval(cls, plat_info):
         HZ = plat_info['kernel']['config']['CONFIG_HZ']
         return ((HZ * plat_info['cpus-count']) // 10) * (1. / HZ)
 
-    @classmethod
-    def from_target(cls, target:Target, res_dir:ArtifactPath=None, ftrace_coll:FtraceCollector=None) -> 'MisfitMigrationBase':
```
valschneider: How does that work? How do we end up with something annotated as returning a misfit-typed bundle?

douglas-raillard-arm: `TestBundleMeta` documents what happens:
1) If `_from_target` is defined in a class but `from_target` is not, a dummy `from_target` is created, just calling `super().from_target()`
2) The annotations of `_from_target` are moved to the dummy `from_target`. 

This means that having an annotated `_from_target` is now enough, there is no need for the boilerplate anymore. Since annotations are removed from `_from_target`, it will not be picked up by exekall. 

All in all, that is almost completely transparent to the user, no "weird" behavior should happen. It's still observable, since definitions of `from_target` will now pop, and the name of the method will reflect that it was redefined in the subclass (which is exactly what we want here).

douglas-raillard-arm: Note that in Python 3.6, all the uses of metaclasses we have in `lisa` will be convertible to `__init_subclass__`:

https://docs.python.org/3/reference/datamodel.html#customizing-class-creation

valschneider: That's my point, there is no `_from_target()` in `misfit.py`. Does `RTATestBundle`'s new `from_target()` suffice? It's only annotated as returning a `RTATestBundle`, although I must admit I'm a bit confused since it looks like for misfit we only had something annotated as returning a `MisfitMigrationBase` while the only concrete implementation was `StaggeredFinishes`

douglas-raillard-arm: Ah sorry, missed your point. `RTATestBundle.from_target()` is enough indeed, as misfit was not adding extra parameters. `exekall` is aware of the subclass hierarchy and uses it to create its expressions. What is detected as factory methods (classmethod with return annotation == the class it is defined in) have a "special" treatment: `exekall` considers them as if they were defined on each subclass, returning an instance of that subclass. As long as all the intermediate classes are abstract or don't have any test methods (returning `ResultBundle`), they will not be picked up.

valschneider: Okay, that clears up some of my confusion, thanks.

lisa/tests/staging/sched_android.py
```diff
@@ -91,15 +82,11 @@ def __init__(self, res_dir, plat_info, test_bundles):
         self.test_bundles = test_bundles
 
     @classmethod
-    def from_target(cls, target:Target, res_dir:ArtifactPath=None,
+    def _from_target(cls, target:Target, res_dir:ArtifactPath=None,
             ftrace_coll:FtraceCollector=None) -> 'SchedTuneBase':
         """
         Creates a SchedTuneBase bundle from the target.
```
valschneider: We probably want to keep the warning from `from_target()`

douglas-raillard-arm: Actually the warning does not apply anymore, since the function actually exposed is  `RTATestBundle.from_target`. That said, the signature of the stub is actually wrong, since it copied it from `_from_target`, whereas in fact the actual method is `RTATestBundle.from_target`. I'll revisit that and push to the PR again once it's fixed.

PR #1020, opened 2019-08-09
lisa.tests.base: Call _from_target() using only keyword arguments.

douglas-raillard-arm:
This gives more flexibility in the order of parameters in implementations of
_from_target(), which is needed when some parameters could have a default value.


douglas-raillard-arm:
https://github.com/ARM-software/lisa/pull/1018 now takes care of that

PR #1035, opened 2019-08-20
Fix CPUMigrationBase

douglas-raillard-arm:
* Fix the CPU rq util expected mean
* cleanup CPUMigrationBase
* Generalize to N-tasks on N-CPUs

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+import abc
 import copy
 import os
 from collections import OrderedDict
@@ -908,6 +909,13 @@ class CPUMigrationBase(LoadTrackingBase):
     period, so it's important to set it to a known validate value in that class.
     """
 
+    @abc.abstractmethod
+    def get_nr_required_cpu(cls, plat_info):
+        """
+        The number of CPUs of same capacity involved in the test
+        """
+        pass
+
     @classmethod
     def _run_rtapp(cls, target, res_dir, profile, ftrace_coll, cgroup=None):
         # Just do some validation on the profile
@@ -919,20 +927,16 @@ class CPUMigrationBase(LoadTrackingBase):
 
         super()._run_rtapp(target, res_dir, profile, ftrace_coll, cgroup)
 
-    def __init__(self, res_dir, plat_info):
-        super().__init__(res_dir, plat_info)
-
-        self.cpus = set()
-
-        self.reference_task = list(self.rtapp_profile.values())[0]
-        self.nr_phases = len(self.reference_task.phases)
-
-        for task in self.rtapp_profile.values():
-            for phase in task.phases:
-                self.cpus.update(phase.cpus)
-
-        self.phases_durations = [phase.duration_s
-                                 for phase in self.reference_task.phases]
+    @property
+    def cpus(self):
+        """
+        All CPUs used by RTapp workload.
+        """
+        return set(itertools.chain.from_iterable(
+            phase.cpus
+            for task in self.rtapp_profile.values()
+            for phase in task.phases
+        ))
 
     @classmethod
     def check_from_target(cls, target):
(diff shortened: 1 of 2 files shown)


lisa/tests/scheduler/load_tracking.py
```diff
@@ -902,6 +902,12 @@ class CPUMigrationBase(LoadTrackingBase):
     The duration of a single phase
     """
 
+    TASK_PERIOD_MS = 16
```
valschneider: Don't we already inherit that from `RTATestBundle` ?

lisa/tests/scheduler/load_tracking.py
```diff
@@ -1087,51 +1088,60 @@ def get_rtapp_profile(cls, plat_info):
...
-            # A task that will migrate from CPU B to CPU A
-            profile["migrating1"] += Periodic(
-                duty_cycle_pct=20, duration_s=cls.PHASE_DURATION_S,
-                period_ms=cls.TASK_PERIOD_MS, cpus=[cpus[1 - i]])
+        # Define one task per CPU, and create all the possible migrations by
+        # shuffling around these tasks
+        for cpus_combi in itertools.permutations(cpus, r=len(cpus)):
+            for i, cpu in enumerate(cpus_combi):
+                profile[make_name(i)] += Periodic(
+                    duty_cycle_pct=20, duration_s=cls.PHASE_DURATION_S,
```
valschneider: As discussed offline, this should be changed scale on the number of requested tasks. Since we're still only using the 1 and 2 tasks definitions, that can be done in a later change.

lisa/tests/scheduler/load_tracking.py
douglas-raillard-arm: We do, but here that value is necessary for correctness. If we were to change the one from RTATestBundle, we really need this one to stay the same, they are not defined with the same constraints.

lisa/tests/scheduler/load_tracking.py
douglas-raillard-arm: Done in the new version of the PR

lisa/tests/scheduler/load_tracking.py
```diff
@@ -971,13 +977,19 @@ def get_trace_cpu_util(self):
...
-                cpu_util[cpu][phase] = series_integrate(util) / (phase_duration)
+                # The runqueue util signal's average does not match the duty
+                # cycle of the task, since it "decays instantly" at next task
+                # wakeup, but stays at its previous value when the task sleeps.
+                # This means that rq PELT signal average is higher than the
+                # idealized PELT signal. Using trapz integration allows to
+                # lower the contribution of the sleep-time util, since it links
+                # with a straight line the point when task goes to sleep with
+                # the wakeup util point.
+                cpu_util[cpu][phase] = series_mean(util, method='trapz')
```
qais-yousef: FWIW,  that's what BART was doing

https://pythonhosted.org/bart-py/bart.common.Utils.html

And I still noticed the approximation errors but I think for the purposes of all of our tests it's good enough. I did document them in an old PR with that noisy tasks if anyone cares to look at it in the future

douglas-raillard-arm: Yes, that was the default behavior of bart. In this case, we don't face an approximation error, as stated in the comment introduced by this PR. We use trapz to workaround the fact that the real mean of the signal we look at is not the mean of the utilization of the task, regardless of how many sample points we would take on that signal. It's a bit brittle and the error is not really controlled, but there is no better/more robust alternative ATM ...

FYI `lisa.datautils.series_integrate()` uses the rectangle method by default since that is the correct thing to do to integrate almost all signals we manipulate, with the exception of the power probe measurements. All trace signals we have are 100% accurate, since we get an event as soon as their value is updated (i.e. watching the value of the signal every picosecond would not expose more information than we already have from the trace).

qais-yousef: Hmm okay. I didn't dig into the details with the newer code. But I think the error that I encountered in bart was due to the linear approximation of the decay - which is exponential in reality. The scheduler doesn't record any samples during the decay. So we end up with a slightly higher util_mean than the real value. The significance of this error will depend on the duty cycle and the ratio of run/sleep. We didn't do much about at the end because it didn't turn out to be a real problem and the noisy tasks were the real culprit. Then buildroot solved the noise issue - mostly at least.

Anyway, I have nothing against the new change - just trying to add some references/docs for the future.

douglas-raillard-arm: Yes, that's exactly what is now described in the comment, and the reason why I forced the task period to a known value in that class (changing it would break that hack)

qais-yousef: Okay. I didn't parse the comment that way tbh. But as long as we are talking about the same thing it's all good :)

PR #1036, opened 2019-08-20
Wlgen uclamp support

Jaime-VH:
This patch adds the ability to set clamp values for RTATask objects. Additional parameters are added to the constructors of RTATask classes which allow one to provide functions as arguments which are used to compute the desired clamp value for a given phase of the task. Example usage is provided in [this notebook](https://gist.github.com/Jaime-VH/c70acebef1977ee1c2e1c9a1f2256695)

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -432,7 +432,7 @@ class Phase(Loggable):
     :type barrier_after: str
     """
 
-    def __init__(self, duration_s, period_ms, duty_cycle_pct, cpus=None, barrier_after=None):
+    def __init__(self, duration_s, period_ms, duty_cycle_pct, cpus=None, barrier_after=None, util_min=None, util_max=None):
         if barrier_after and duty_cycle_pct != 100:
             # This could be implemented but currently don't foresee any use.
             raise ValueError('Barriers only supported when duty_cycle_pct=100')
@@ -442,6 +442,8 @@ class Phase(Loggable):
         self.duty_cycle_pct = duty_cycle_pct
         self.cpus = cpus
         self.barrier_after = barrier_after
+        self.util_min = util_min
+        self.util_max = util_max
 
     def get_rtapp_repr(self, task_name):
         """
@@ -500,6 +502,16 @@ class Phase(Loggable):
 
         if self.cpus is not None:
             phase['cpus'] = self.cpus
+            
+
+        if self.util_min is not None:
+            phase['util_min'] = self.util_min
+            logger.info(' | util_min %7d', self.util_min)
+
+        
+        if self.util_max is not None:
+            phase['util_max'] = self.util_max
+            logger.info(' | util_max %7d', self.util_max)
 
         return phase
 
@@ -573,11 +585,21 @@ class Ramp(RTATask):
     :param cpus: the list of CPUs on which task can run.
                 .. note:: if not specified, it can run on all CPUs
     :type cpus: list(int)
+    
+    :param util_mins: when called with the index and load of a phase, 
+                       sets the util_min value for that phase.
+    :type util_mins: callable
+    
+    :param util_maxs: when called with the index and load of a phase, 
+                       sets the util_max value for that phase.
+    :type util_maxs: callable
     """
 
     def __init__(self, start_pct=0, end_pct=100, delta_pct=10, time_s=1,
                  period_ms=100, delay_s=0, loops=1, sched_policy=None,
-                 priority=None, cpus=None):
+                 priority=None, cpus=None, 
+                 util_mins = lambda idx, load: 0,
+                 util_maxs = lambda idx, load: 1024):
         super(Ramp, self).__init__(delay_s, loops, sched_policy, priority)
 
         if not (0 <= start_pct <= 100 and 0 <= end_pct <= 100):
@@ -592,12 +614,38 @@ class Ramp(RTATask):
         # clip the last step
         steps[-1] = end_pct
 
+        # Clamp values for previous phases.
+        # These values are remembered so the clamp values are not set on the start
+        #   of every phase
+        # The initial values should probably be set on the first phase,
+        #   this protects against retaining a clamp from a previous loop.
+        util_min = None
+        util_max = None
+
         phases = []
-        for load in steps:
+        for idx, load in enumerate(steps):
+            # If clamp values are unchanged, don't bother setting them again
+            new_util_min = util_mins(idx, load)
+            if new_util_min == util_min:
+                new_util_min = None
+            new_util_max = util_maxs(idx, load)
+            if new_util_max == util_max:
+                new_util_max = None
+                
             if load == 0:
-                phase = Phase(time_s, 0, 0, cpus)
+                phase = Phase(time_s, 0, 0, cpus, 
+                              util_min=new_util_min, 
+                              util_max=new_util_max)
             else:
-                phase = Phase(time_s, period_ms, load, cpus)
+                phase = Phase(time_s, period_ms, load, cpus, 
+                              util_min=new_util_min, 
+                              util_max=new_util_max)
+                
+            if new_util_min != None:
+                util_min = new_util_min
+            if new_util_max != None:
+                util_max = new_util_max
+        
             phases.append(phase)
 
         self.phases = phases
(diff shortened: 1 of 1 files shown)


lisa/wlgen/rta.py
```diff
@@ -573,11 +581,21 @@ class Ramp(RTATask):
...
+    :param clamp_maxs: when called with the index and load of a phase, 
+                       sets the clamp_max value for that phase.
+    :type clamp_maxs: callable
     """
 
     def __init__(self, start_pct=0, end_pct=100, delta_pct=10, time_s=1,
                  period_ms=100, delay_s=0, loops=1, sched_policy=None,
-                 priority=None, cpus=None):
+                 priority=None, cpus=None, 
+                 clamp_mins = lambda idx, load: 0,
```
douglas-raillard-arm: nit: no spaces around `=`

lisa/wlgen/rta.py
```diff
@@ -592,12 +610,38 @@ def __init__(self, start_pct=0, end_pct=100, delta_pct=10, time_s=1,
         # clip the last step
         steps[-1] = end_pct
 
+        # Clamp values for previous phases.
+        # These values are remembered so the clamp values are not set on the start
+        #   of every phase
```
douglas-raillard-arm: nit: you can leave just one space between # and the first word

lisa/wlgen/rta.py
```diff
@@ -573,11 +581,21 @@ class Ramp(RTATask):
     :param cpus: the list of CPUs on which task can run.
                 .. note:: if not specified, it can run on all CPUs
     :type cpus: list(int)
+    
+    :param clamp_mins: when called with the index and load of a phase,
```
douglas-raillard-arm: Looks like it could be fused in a single `get_clamps` parameter, returning a tuple of (min, max) so we need to introduce only one parameter. `get_` prefix shows that it's a function rather than just an iterable that is expected.

lisa/wlgen/rta.py
```diff
@@ -592,12 +610,38 @@ def __init__(self, start_pct=0, end_pct=100, delta_pct=10, time_s=1,
...
+                phase = Phase(time_s, 0, 0, cpus, 
+                              clamp_min=new_clamp_min, 
+                              clamp_max=new_clamp_max)
             else:
-                phase = Phase(time_s, period_ms, load, cpus)
+                phase = Phase(time_s, period_ms, load, cpus, 
+                              clamp_min=new_clamp_min, 
+                              clamp_max=new_clamp_max)
+                
+            if new_clamp_min != None:
```
douglas-raillard-arm: Use `new_clamp_min is not None` for comparison with None. None is a singleton so it will always work, and using `is/is not` is recommended by PEP8 in these cases (mostly for speed purpose and to avoid broken equality implementations I assume):
https://www.python.org/dev/peps/pep-0008/#programming-recommendations

lisa/wlgen/rta.py
```diff
@@ -677,11 +741,34 @@ def __init__(self, start_pct=100, end_pct=0, time_s=1, period_ms=100,
...
+        clamp_max = None
+
         phases = []
-        for load in [start_pct, end_pct]:
+        for idx, load in enumerate([start_pct, end_pct]):
             if load == 0:
                 continue
+                
+            # If clamp values are unchanged, don't bother setting them again
+            new_clamp_min = clamp_mins(idx, load)
```
douglas-raillard-arm: Since the exact same pattern came up twice, you could factor it out in a function like this one:
```
import itertools

def replace_consecutive_values(l, placeholder=None):
    """
    Replace consecutive runs of the same value with ``placeholder``.
    """
    if len(l) < 2:
        return list(l)
    prev_val = l[0]
    deduplicated = []
    for val in l[1:]:
        if val == prev_val:
            dedup_val = None
        else:
            dedup_val = val
            prev_val = val

        deduplicated.append(dedup_val)
    return deduplicated

def compute_clamps(steps, get_clamps):
    if not get_clamps:
        get_clamps = lambda idx, load: (None, None)

    # Use starmap instead of map to unpack the tuple from enumerate
    clamps = starmap(get_clamps, enumerate(steps))
    # Unzip the clamp tuples into two separate iterables
    both_clamps = zip(*clamps)
    # Remove consecutive runs of the same clamp value
    both_clamps = map(replace_consecutive_values, both_clamps)
    # Zip back into a single list of tuples
    return list(zip(*both_clamps))
```
And then use it as:
```
# get_clamps is either None or a function returning (clamp_min, clamp_max)
clamps = compute_clamps(steps, get_clamps)
for load, (min_clamp, max_clamp) in zip(steps, clamps):
    ...
```

Note: `replace_consecutive_values` may benefit from being moved to `lisa/utils.py`, as it is generally useful (there are some other itertools-based helpers there).

lisa/wlgen/rta.py
```diff
@@ -495,7 +495,7 @@ def get_rtapp_repr(self, task_name):
                              period, self.duty_cycle_pct)
             logger.info(' |  run_time %6d [us], sleep_time %6d [us]',
                              running_time, sleep_time)
-
+
```
douglas-raillard-arm: nit: trailing whitespace

lisa/wlgen/rta.py
Jaime-VH: These comments have now been reformatted.

lisa/wlgen/rta.py
Jaime-VH: The new commit fuses the parameters and renames it accordingly.

lisa/wlgen/rta.py
Jaime-VH: This change has now been made.

lisa/wlgen/rta.py
```diff
@@ -573,11 +584,16 @@ class Ramp(RTATask):
...
+    :param get_utils: when called with the index and load of a phase, 
+                      returns (util_min, util_max)
+    :type get_utils: callable
     """
 
     def __init__(self, start_pct=0, end_pct=100, delta_pct=10, time_s=1,
                  period_ms=100, delay_s=0, loops=1, sched_policy=None,
-                 priority=None, cpus=None):
+                 priority=None, cpus=None, 
+                 get_utils=lambda idx, load: (0, 1024)):
```
valschneider: Also nit: given that we don't know the count of phases when this gets called, does it make sense to keep the `idx` parameter?

I was originally thinking that we could even drop the callbacks and let the user specify the clamps manually (e.g. create the `RTATask` object, iterate over `.phases` and tweak accordingly - the good thing is you have the total count of phases here), but @douglas-raillard-arm made a point that the callback simplifies quite a few scenarios.

Also, seeing as you're duplicating this lambda in a few places, it could be worth having it defined as an actual method somewhere. The character count won't get better (it might get worse), but at least the default will live in a centralised place.

lisa/wlgen/rta.py
```diff
@@ -677,11 +732,33 @@ def __init__(self, start_pct=100, end_pct=0, time_s=1, period_ms=100,
...
                 continue
+                
+            # If clamp values are unchanged, don't bother setting them again
+            new_util_min, new_util_max = get_utils(idx, load)
+            if new_util_min == util_min:
+                new_util_min = None
+            if new_util_max == util_max:
+                new_util_max = None
+                
             phase = Phase(time_s, period_ms, load, cpus)
```
valschneider: Isn't this missing the actual setting of the clamp values?

valschneider:
Just had another thought - in a way those uclamp values are quite similar to the cpu affinity we can set for `RTATask`. In that case, wouldn't it make more sense to use task composition to assemble phases of different clamp values? That's how we generally build phases with different CPU affinities.

I'd see something like this:

```
common_params = {
    "duty_cycle_pct" : 40,
    "period_ms" : 16
}

task = Periodic(**common_params, durations_s=0.1, uclamps=(200, 1024)) + \
    Periodic(**common_params, durations_s=0.2, uclamps=(300, 850)) + \
    Periodic(**common_params, duration_s=0.15, uclamps=(400, 600))
```

The good thing with this is that you (the user of this API) don't have to get your hands in `Phase`. @derkling what do you think?

douglas-raillard-arm:
general comment: it would be good to have `uclamp` in the name of things we set (parameters etc), since there are many utilization, and even more since what we call `duty_cycle` is actually a request for average utilization nowadays.

qais-yousef:
What are the current sticking point for this to get merged?

Jaime-VH:
> What are the current sticking point for this to get merged?
It was decided that this API is too complex and should be changed to something more in line with the existing usage of wlgen. 
The new API will have a single clamp value for each RTATask object rather than a callback function used to determine the clamp value for each phase of the task. Fine grained control over the clamps can then be achieved using task composition instead, which is more natural and avoids exposing the implementation detail of rtatask phases.

Jaime-VH:
This PR is closed. Refer to #1146 instead.

PR #1037, opened 2019-08-20
lisa.wlgen.rta: Fix race condition in RTA workload execution

douglas-raillard-arm:
The rtapp JSON config file is pushed to the target in `RTA.__init__`, but is only
used when RTA.run() is called. Since the name is not unique, something else
could have modified the file in between, leading to the wrong workload to be
executed.

Fix that by prepending a UUID to the remote filename, so that there is a
one-to-one mapping between remote conf files and RTA() instances.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -22,6 +22,7 @@ import re
 import sys
 from collections import OrderedDict
 import copy
+import uuid
 
 from lisa.wlgen.workload import Workload
 from lisa.utils import Loggable, ArtifactPath, TASK_COMM_MAX_LEN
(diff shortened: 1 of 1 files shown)


lisa/wlgen/rta.py
```diff
@@ -63,13 +65,14 @@ def _early_init(self, target, name, res_dir, json_file):
             json_file = '{}.json'.format(self.name)
 
         self.local_json = ArtifactPath.join(self.res_dir, json_file)
-        self.remote_json = self.target.path.join(self.run_dir, json_file)
+        remote_json_file = '{}_{}'.format(uuid.uuid4(), json_file)
```
valschneider: Might want to swap the order? If we end up with several in a directory they'll be sorted together.

douglas-raillard-arm: If we swap the order we will get `foo.json_XXXXXX`, and since the file name can come from the user I figured it would be better to avoid assumptions on the (multiple) extensions the file could have.
Once https://github.com/ARM-software/lisa/pull/958 is merged we could actually drop that commit and only keep the `shlex.quote()` one, since the folder containing it will be unique.

douglas-raillard-arm:
Closing as https://github.com/ARM-software/lisa/pull/958 is now merged and is a more general fix than this one

PR #1041, opened 2019-08-20
lisa.tests.base: Add SimpleRTATestBundle

douglas-raillard-arm:
Having RTATestBundle._from_target() implemented in RTATestBundle breaks the
generality of RTATestBundle.from_target(), as it gains parameters specific to
RTATestBundle._from_target that will be expected to be available on
implementations of _from_target in all subclasses.

Splitting that feature in SimpleRTATestBundle allows clearing the way to
TestBundle.from_target(), which does not add unnecessary constraints on the
expected signature of downstream _from_target.

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -1102,8 +1102,14 @@ class RTATestBundle(FtraceTestBundle):
         dmesg_coll.get_trace(dmesg_path)
         return trace_path
 
+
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
Abandoned as we decided to force subclass's _from_target() to implement at least all the parameters of their mother class's _from_target()

PR #1053, opened 2019-08-28
lisa/wa_results_collector: Fix parse_traces=True

valschneider:
Daniel pointed out that passing `parse_traces=True` to
WaResultsCollector breaks horribly on a sad piece of code.

We're attempting to create a PlatformInfo to give to Trace, but we're
only filling it with the kernel version which mostly useless. Just skip
that step, and only pass a non-None platinfo if the user specified one.

Ideally we should collect one platform_info per wa directory, but that
would mean adding a LISA-specific step to lisa-wltest which I think we
all would like to avoid.

### DIFF
--- a/lisa/wa_results_collector.py
+++ b/lisa/wa_results_collector.py
@@ -38,7 +38,6 @@ from trappy.utils import handle_duplicate_index
 
 from IPython.display import display
 
-from lisa.platforms.platinfo import PlatformInfo
 from lisa.trace import Trace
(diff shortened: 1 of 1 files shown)


valschneider:
> LGTM. That was much needed indeed, thanks for cleaning that up. I guess it's another example of something that could use some CI testing, but that's probably for later.

Not sure how easy it would be to run WA workloads on the Travis slave, but yes it's something we'd want.

PR #1055, opened 2019-09-02
lisa.wlgen.workload.Workload background execution revamp

valschneider:


lisa/wlgen/workload.py
```diff
@@ -133,16 +133,19 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False, timeout=N
...
         if background:
-            target.background(_command, as_root=as_root)
-        else:
-            self.output = target.execute(_command, as_root=as_root, timeout=timeout)
-            logger.info("Execution complete")
+            return target.background(_command, as_root=as_root)
 
-            logfile = ArtifactPath.join(self.res_dir, 'output.log')
-            logger.debug('Saving stdout to %s...', logfile)
+        self.output = target.execute(_command, as_root=as_root, timeout=timeout)
```
douglas-raillard-arm: Do we need `output` attribute outside of that function ? If possible, it would be better to only keep the "static configuration" in the state of the object and just let people look at the return of `run()` if they need it, rather than introducing extra mutable state that can be looked at when it's garbage (i.e. None).

lisa/wlgen/workload.py
```diff
@@ -133,16 +133,19 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False, timeout=N
 
         logger.info("Execution start: %s", _command)
 
+        # Just return the Popen handle
         if background:
-            target.background(_command, as_root=as_root)
-        else:
-            self.output = target.execute(_command, as_root=as_root, timeout=timeout)
-            logger.info("Execution complete")
+            return target.background(_command, as_root=as_root)
```
douglas-raillard-arm: I'd rather have `run(background=True)` return None as before with a DeprecationWarning and introduce a new `run_background()` function that does the right thing. Ultimately, we can kill the `background` parameter of `run()`, which will mean:
* `run()` is meant to be the foreground run, which is what most people will need
* `run_background()` for background runs, returning the handle, and maybe some background-specific options.

If that makes sense from code-sharing point of view, we can have an internal `_run()` that returns one or the other, but IMO it's better to keep background-related options/return value separated.

lisa/wlgen/workload.py
```diff
@@ -111,6 +111,37 @@ def wipe_run_dir(self):
...
+        """
+        command = self.command
+
+        if not command:
+            raise RuntimeError("Workload does not specify any command to execute")
+
+        if cpus:
+            taskset_bin = self.target.which('taskset')
+            if not taskset_bin:
+                raise RuntimeError("Could not find 'taskset' executable on the target")
```
douglas-raillard-arm: since this method is the one deciding which binaries are going to get used, it would make sense to `self.target.install()` rather than just complaining to the caller that the binary is not there.

lisa/wlgen/workload.py
```diff
@@ -141,7 +139,42 @@ def wrap_command(self, cpus=None, cgroup=None):
...
+
+        .. note:: Popen handles can be used as context managers, so you could
+          (and probably should) use this like so::
+
+            with run_background() as handle:
+                do_stuff_with(handle)
+                handle.communicate()
+        """
+        logger = self.get_logger()
+        if not self.command:
```
douglas-raillard-arm: can this happen ? Maybe it would be better to catch that in `__init__` rather than here.

lisa/wlgen/workload.py
```diff
@@ -170,17 +203,13 @@ def run(self, cpus=None, cgroup=None, background=False, as_root=False, timeout=N
         command = self.wrap_command(cpus, cgroup)
 
         logger.info("Execution start: %s", command)
+        self.output = target.execute(command, as_root=as_root, timeout=timeout)
```
douglas-raillard-arm: Wouldn't it make more sense for `run()` to return the output, rather than storing it as an attribute ? Avoiding statefull APIs usually reduce the confusion. With state like that, it's not always obvious if a given workload is suitable or not: does it have an output yet or do we need to run it first ? Then the client code will add tests to check if that attribute is None or not, and `__init__` will initialize that to `None`.  We can avoid all this confusion and boilerplate by just having objects with a single responsibility: representing the workload, and another one representing the result of an execution. These two things are separate, and if merged together usually give rise to half of the methods checking some attributes for `None` and raising `RuntimeError` at the beginning of the method (like we used to have in `Trace` for example).

lisa/wlgen/workload.py
valschneider: Hmm actually it's already installed at `__init__()` time since `taskset` is in the list of `required_tools` for the base `Workload` class. We've always had this safety check, I suppose I could remove it?

lisa/wlgen/workload.py
valschneider: Some workloads (e.g. sysbench) can have their command defined at `run()` time, which now that I look at it is pretty broken for splitting execution methods.

AFAICT I had done that to not have to recreate several `Sysbench` instances for slightly different invocations. I actually wrote a piece about that in the doc:

```
    ``__init__`` is there to initialize a given workload, and :meth:`run`                            
    can be called on it several times, with varying arguments.                                       
    As much work as possible should be delegated to :meth:`run`, so that                             
    different flavours of the same workload can be run without the hassle of                         
    creating a superfluous amount of new instances. However, when persistent                         
    data is involved (e.g. the workload depends on a file), then this data                           
    should be exposed as an ``__init__`` parameter.  
```

I need to have a think about this.

lisa/wlgen/workload.py
valschneider: I think it makes sense, I just need to have a look around and see how many users of this API there is.

lisa/wlgen/workload.py
douglas-raillard-arm: The issue with that approach is that if we really do that, we will end up with everything as `run()` parameters and nothing in `__init__`. Anything in between will just be an arbitrary draw of whoever introduced the class, without any guarantee that the split will be sensible for the next user.
Functional programming comes to the rescue and can save us the `hassle of creating a superfluous amount of new instances` ! We put everything back in `__init__` and use currying:
```
import functools
class Workload:
     @classmethod
     def partial(cls, *args, **kwargs):
          return functools.partial(cls, *args, **kwargs)

class MyWorkload(Workload):
     def __init__(self, foo, bar):
         ....

make = MyWorkload.partial(foo=3)
make(bar=4).run()
make(bar=5).run() # No hassle !
```

To finish the work, we also need to have a way of amending the configuration after the workload has been created, for example for the RTA workloads (`run_rtapp()` may need to force some settings, e.g. disabling mlockall for CFS tasks when target is not rooted). This means pushing things to the target has to happen later than `__init__`. A nice place to store that is in `Workload.__enter__`, since it will be run right before we call `run()`. We can also have a "manual" `deploy()` method if somebody does not use the CM (or we can even do that automatically in `run()` if it has not been done already).

douglas-raillard-arm: Note that with a bit of metaprogramming we can even make the currying automatic, i.e.:
```
wload = Workload(foo=1)
wload(bar=3).run()
wload(bar=4).run()
# Raise an exception explicitly saying that "bar" was not given a value
wload.run()
```

douglas-raillard-arm:
Closing as the issue has been solved

PR #1059, opened 2019-09-02
lisa.tests.base: Add DmesgTestBundle

douglas-raillard-arm:
DmesgTestBundle allows writing tests based on dmesg. It comes with a generic
test that checks that no message more critical than "warn" appeared during the
execution of the test.

Signed-off-by: Douglas RAILLARD <douglas.raillard@arm.com>

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -33,6 +33,8 @@ from collections import OrderedDict, ChainMap
 from collections.abc import Mapping
 from inspect import signature
 
+from devlib.trace.dmesg import KernelLogEntry
+
 from lisa.analysis.tasks import TasksAnalysis
 from lisa.trace import Trace, requires_events, TaskID
 from lisa.wlgen.rta import RTA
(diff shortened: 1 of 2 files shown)


lisa/tests/base.py
```diff
@@ -835,19 +837,74 @@ def get_trace(self, **kwargs):
...
-class RTATestBundle(FtraceTestBundle):
+class DmesgTestBundle(TestBundle):
     """
-    Abstract Base Class for :class:`lisa.wlgen.rta.RTA`-powered TestBundles
-
-    .. seealso: :class:`lisa.tests.base.FtraceTestBundleMeta` for default
-        ``ftrace_conf`` content.
+    Abstract Base Class for TestBundles based on dmesg output.
     """
-
```
valschneider: Spurious change? I think this newline made sense.

lisa/tests/base.py
```diff
@@ -835,19 +837,74 @@ def get_trace(self, **kwargs):
...
+
+        :param facility: Only select entries emitted by the given dmesg
+            facility like `kern`. Note that not all versions of `dmesg` are
+            able to print it, so specifying it may lead to no entry being
+            inspected at all. If ``None``, the facility is ignored.
+        type facility: str or None
+        """
+        levels = DmesgCollector.LOG_LEVELS
+        # Consider as an issue all levels more critical than `level`
+        issue_levels = levels[:levels.index(level) + 1]
```
valschneider: This only makes sense if message priority decreases with the level, is that the case?

lisa/tests/base.py
douglas-raillard-arm: It's actually not a change per say as I'm adding new code, not modifying it (don't be confused by diff, there are gazillions of equivalent patches, and diff just picks one that is usually reasonable, but not always).
I've added the newline in the new class.

lisa/tests/base.py
douglas-raillard-arm: Yes, LOG_LEVELS is explicitly defined with the right order:
```
    # taken from "dmesg --help"
    # This list needs to be ordered by priority
    LOG_LEVELS = [
        "emerg",        # system is unusable
        "alert",        # action must be taken immediately
        "crit",         # critical conditions
        "err",          # error conditions
        "warn",         # warning conditions
        "notice",       # normal but significant condition
        "info",         # informational
        "debug",        # debug-level messages
    ]
```

PR #1063, opened 2019-09-05
load_tracking: rewrite get_task_duty_cycle_pct

ionela-voinescu:
Rewrite get_task_duty_cycle_pct using task analysis methods. It seems
that the Bart method for obtaining a tasks's run time on a CPU
sometimes fails to provide a correct value.

Given that we're using task analysis methods on all other tests with this
function being the only one remaining using a Bart (SchedAssert) method,
rewrite this function to use a more reliable method to obtain a task's
run time.

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -25,7 +25,6 @@ from statistics import mean
 import matplotlib.pyplot as plt
 
 from bart.sched import pelt
-from bart.sched.SchedAssert import SchedAssert
 
 from trappy.stats.Topology import Topology
 
@@ -121,11 +120,11 @@ class LoadTrackingHelpers:
 
     @classmethod
     def get_task_duty_cycle_pct(cls, trace, task_name, cpu):
-        window = cls.get_task_window(trace, task_name, cpu)
 
-        top = Topology()
-        top.add_to_level('cpu', [[cpu]])
-        return SchedAssert(trace._ftrace, top, execname=task_name).getDutyCycle(window)
+        df = trace.analysis.tasks.df_task_total_residency(task_name)
+        run_time = df['runtime'][cpu]
+
+        return (run_time * 100) / trace.time_range
 
     @staticmethod
     def get_task_window(trace, task_name, cpu=None):


lisa/tests/scheduler/load_tracking.py
```diff
@@ -121,11 +121,11 @@ def get_max_capa_cpu(cls, plat_info):
...
     def get_task_duty_cycle_pct(cls, trace, task_name, cpu):
-        window = cls.get_task_window(trace, task_name, cpu)
 
-        top = Topology()
-        top.add_to_level('cpu', [[cpu]])
-        return SchedAssert(trace._ftrace, top, execname=task_name).getDutyCycle(window)
+        df = trace.analysis.tasks.df_task_total_residency(task_name)
+        run_time = df.iloc[cpu].values[0]
+
+        return (run_time * 100)/trace.time_range
```
douglas-raillard-arm: nit: spaces around `/`

lisa/tests/scheduler/load_tracking.py
```diff
@@ -121,11 +121,11 @@ def get_max_capa_cpu(cls, plat_info):
 
     @classmethod
     def get_task_duty_cycle_pct(cls, trace, task_name, cpu):
-        window = cls.get_task_window(trace, task_name, cpu)
 
-        top = Topology()
-        top.add_to_level('cpu', [[cpu]])
-        return SchedAssert(trace._ftrace, top, execname=task_name).getDutyCycle(window)
+        df = trace.analysis.tasks.df_task_total_residency(task_name)
+        run_time = df.iloc[cpu].values[0]
```
douglas-raillard-arm: Untested: wouldn't it be more robust with ?
```
df[cpu]['runtime']
```
`iloc` will take the n'th row, but according to the doc the CPU is the index. We could have:
```
CPU XXX
0       B
4       B
```
Which would give:
```
# Looking for CPU4 using `iloc` means accessing the 2nd row
df.iloc[1]['XXX'] == A
# This is the expected behavior, where the CPU ID is used as the index value, rather than as a row number.
df[4]['XXX'] == B
```
Columns (`runtime`) need to be accessed by names rather than making a Series from the row and then accessing the nth item. We make no guarantee on the order of columns, we might add new ones, and it makes the code much more readable.

douglas-raillard-arm:
This line can also be removed now:
```
from bart.sched.SchedAssert import SchedAssert
```

lisa/tests/scheduler/load_tracking.py
ionela-voinescu: Right! I meant to use loc intead of iloc - I always get them mixed up. But I think you meant to say I should use: ``` run_time = df[cpu]['runtime'] ``` It id definitely a better solution. I'll fix it! Thanks!

douglas-raillard-arm:
LGTM, waiting for @valschneider ack

ionela-voinescu:
Got the ack from @valschneider in person.

PR #1066, opened 2019-09-05
Allow whitelist for test_dmesg()

douglas-raillard-arm:


douglas-raillard-arm:
@valschneider This could become the idiomatic pattern for adding test-specific parameters that need to be set by a regular user when running the tests (as opposed as hacking with `exekall run --param` or `--sweep`).

douglas-raillard-arm:
@valschneider This PR can be considered "finished" now and is not RFC anymore.
The commits reflect the "history of the idea", going from the manual implementation based on explicit classes and methods to a completely equivalent and automatic way of doing it.

This reduces the code clutter and allows straightforward use of any key from any configuration class without having to add extra boilerplate first.

lisa/tests/base.py
```diff
@@ -873,7 +889,8 @@ def dmesg_entries(self):
                 if line.strip()
             ]
 
-    def test_dmesg(self, level='warn', facility=None) -> ResultBundle:
+    def test_dmesg(self, level='warn', facility=None, pattern_whitelist:DmesgTestConf.PatternWhitelist=None) -> ResultBundle:
```
valschneider: Nit: `pattern_ignore` would be more explicit. I was a bit confused at first (i.e. does having a whitelist mean we'll only consider messages matching the pattern, or will we let messages matching the pattern "pass"?).

lisa/conf.py
```diff
@@ -415,6 +447,9 @@ class MultiSrcConfMeta(abc.ABCMeta):
     template with the ``{generated_help}`` placeholder replaced by a snippet of
     ResStructuredText containing the list of allowed keys.
 
+    It will also create the types specified using ``newtype`` in the
+    :class:`KeyDesc`, along with a getter to expose it to ``exekall``.
```
valschneider: Might want to document here (or somewhere else - anywhere as long as it's documented!) that it's exposed as a camelcased attribute.

lisa/tests/base.py
douglas-raillard-arm: I'll go for `ignore_patterns` so it shows that it's a list (or at least an iterable).

lisa/conf.py
douglas-raillard-arm: It was shortly documented in KeyDesc docstring, but is now more explicit.

lisa/tests/base.py
```diff
@@ -886,16 +902,32 @@ def test_dmesg(self, level='warn', facility=None) -> ResultBundle:
...
+        logger = self.get_logger()
+
+        if ignored_patterns:
+            logger.info('Will ignore patterns in dmesg output: {}'.format(ignored_patterns))
+            regex_whitelist = [
+                re.compile(pattern)
+                for pattern in ignored_patterns
+            ]
+        else:
+            regex_whitelist = []
```
valschneider: Nit: might want to rename that to `regex_ignore` for consistency. Otherwise LGTM.

PR #1072, opened 2019-09-06
Update subtrees

douglas-raillard-arm:


douglas-raillard-arm:
> Do you know why github's complaining about merge conflicts?

It complains if two PRs are conflicting, which is the case since they would touch the same places in the same files. Maybe yours has one less or one more change factored in, so they are not exactly the same.

douglas-raillard-arm:
Stalled until https://github.com/ARM-software/devlib/issues/428 gets fixed

douglas-raillard-arm:
https://github.com/ARM-software/devlib/issues/428 now fixed and PR updated

valschneider:
Github's still reporting conflicts, did we mess up our subtrees in some way? :/

PR #1074, opened 2019-09-13
devlib usage update to use new features

douglas-raillard-arm:
Couple of things to update in LISA to make use of new devlib features:

* Remove some duplication in dmesg output parsing
* Root adb upon connection before loading the devlib modules. Modules that need root access will work even if the device is not already rooted at the time the target connects.

Depends on https://github.com/ARM-software/lisa/pull/1072

### DIFF
--- a/lisa/target.py
+++ b/lisa/target.py
@@ -115,7 +115,7 @@ class TargetConf(SimpleMultiSrcConf, HideExekallID):
         KeyDesc('kind', 'Target kind. Can be "linux" (ssh) or "android" (adb)', [str]),
 
         KeyDesc('host', 'Hostname or IP address of the host', [str, None]),
-        KeyDesc('username', 'SSH username', [str, None]),
+        KeyDesc('username', 'SSH username. On ADB connections, "root" username will root adb upon target connection', [str, None]),
         PasswordKeyDesc('password', 'SSH password', [str, None]),
         KeyDesc('port', 'SSH or ADB server port', [int, None]),
         KeyDesc('device', 'ADB device. Takes precedence over "host"', [str, None]),
@@ -516,6 +516,7 @@ class Target(Loggable, HideExekallID, ExekallTaggable, Configurable):
                 device = 'DEFAULT'
 
             conn_settings['device'] = device
+            conn_settings['adb_as_root'] = (username == 'root')
 
         elif kind == 'linux':
             logger.debug('Setting up Linux target...')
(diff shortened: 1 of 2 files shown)


lisa/target.py
```diff
@@ -115,7 +115,7 @@ class TargetConf(SimpleMultiSrcConf, HideExekallID):
         KeyDesc('kind', 'Target kind. Can be "linux" (ssh) or "android" (adb)', [str]),
 
         KeyDesc('host', 'Hostname or IP address of the host', [str, None]),
-        KeyDesc('username', 'SSH username', [str, None]),
+        KeyDesc('username', 'SSH username. On ADB connections, "root" username will root adb upon target connection', [str, None]),
```
valschneider: I'm really not keen on the double meaning of this key. Could we have an ADB specific key (like `device`) instead?

lisa/target.py
```diff
@@ -516,6 +516,7 @@ def _init_target(self, kind, name, workdir, device, host,
                 device = 'DEFAULT'
 
             conn_settings['device'] = device
+            conn_settings['adb_as_root'] = (username == 'root')
```
valschneider: Sadly this doesn't seem to cope with devices you can't root:

```
File "/data/work/lisa/lisa/target.py", line 266, in __init__
    wait_boot_timeout=wait_boot_timeout,
  File "/data/work/lisa/lisa/target.py", line 595, in _init_target
    target.connect(check_boot_completed=wait_boot, timeout=wait_boot_timeout)
  File "/data/work/lisa/external/devlib/devlib/target.py", line 1212, in connect
    super(AndroidTarget, self).connect(timeout=timeout, check_boot_completed=check_boot_completed)
  File "/data/work/lisa/external/devlib/devlib/target.py", line 263, in connect
    self._connections[tid] = self.get_connection(timeout=timeout)
  File "/data/work/lisa/external/devlib/devlib/target.py", line 283, in get_connection
    return self.conn_cls(timeout=timeout, **self.connection_settings)  # pylint: disable=not-callable
  File "/data/work/lisa/external/devlib/devlib/utils/android.py", line 272, in __init__
    self.adb_root(enable=True)
  File "/data/work/lisa/external/devlib/devlib/utils/android.py", line 334, in adb_root
    raise TargetStableError(output)
devlib.exception.TargetStableError: adbd cannot run as root in production builds
```
Dunno if we want to handle that properly or not. If the action to root depends on an extra conf key, maybe not.

lisa/target.py
douglas-raillard-arm: It's not actually double meaning, since requesting "root" will actually log you as root on the device, just like for SSH. If possible, I'd rather avoid proliferation of mutually-exclusive or context specific keys at the top-level, especially if their meaning 100% overlap. The only real reason `device` exists is that `host` can also be used for ADB, so we needed another one.

lisa/target.py
douglas-raillard-arm: You will need to set a different username if your device cannot be rooted. LISA defaults to `Target.__init__(username="root")`. We already used to run `target.adb_root(enable=True)` unconditionally for android targets, so currently working setups will keep working (which was probably not the case with yours)

PR #1077, opened 2019-09-17
lisa.utils.deprecate: Allow deprecating a parameter

douglas-raillard-arm:
Deprecated parameters will emit a warning when used, and the docstring will be
updated in right spot at the end of the relevant :param: block.


valschneider:
It will create several "param" sections in the output, but that's fine by me, what matters is getting the warning:
image

PR #1079, opened 2019-09-18
lisa.wlgen.workload: Make Workload a context manager 

douglas-raillard-arm:


lisa/tests/base.py
```diff
@@ -1272,14 +1277,15 @@ def run_rtapp(cls, target, res_dir, profile=None, ftrace_coll=None, cg_cfg=None)
...
+        wload_cm = wload if wipe_run_dir else nullcontext(wload)
 
         # Pre-hit the calibration information, in case this is a lazy value.
         # This avoids polluting the trace and the dmesg output with the
         # calibration tasks. Since we know that rt-app will always need it for
         # anything useful, it's reasonable to do it here.
         target.plat_info['rtapp']['calib']
 
-        with dmesg_coll, ftrace_coll, target.freeze_userspace():
+        with wload_cm, dmesg_coll, ftrace_coll, target.freeze_userspace():
```
valschneider: We need to make sure we don't add an earlier context manager that tries to do something in the run dir, but AFAICT that should always be contained in the `run()` method. Other than this slippery slope, LGTM

douglas-raillard-arm: Yes, the current code has no issue with the cleanup. Future code will fail quite obviously so we should be fine

PR #1080, opened 2019-09-18
lisa/tests: misfit: Fix rt-app naming issue

valschneider:
Rt-app task naming was changed with the introduction of the fork
functionality, and this was reflected in

  5b066eef0578 ("lisa.tests.base: Introduce RTATestBundle.rtapp_tasks property")

The misfit test case uses the name of the task to figure out on which
CPU it is initially pinned, which is somewhat icky (and now broken with
the task naming change). Look at the full fledged profile to figure this
out instead.

Briefly tested on a H960 with the latest integration branch:
```
[EXEKALL][2019-09-18 12:53:26,537] INFO  Artifacts dir: /data/work/lisa/results/20190918_12:49:12_1a273d9cd97c42fab8b7195bee0b317f
[EXEKALL][2019-09-18 12:53:26,537] INFO  Result summary:
StaggeredFinishes[board=hikey960]:test_dmesg        FAILED: {dmesg output=
crit: [481.849941] hisi_thermal fff30000.tsensor: sensor <1> THERMAL ALARM: 65370 > 65000
crit: [483.044989] hisi_thermal fff30000.tsensor: sensor <1> THERMAL ALARM stopped: 60860 < 65000}

StaggeredFinishes[board=hikey960]:test_preempt_time PASSED: ...
StaggeredFinishes[board=hikey960]:test_throughput   PASSED: ...
```

### DIFF
--- a/lisa/tests/scheduler/misfit.py
+++ b/lisa/tests/scheduler/misfit.py
@@ -118,9 +118,15 @@ class StaggeredFinishes(MisfitMigrationBase):
 
         sdf = sdf[self.trace.start + self.IDLING_DELAY_S * 0.9:]
 
-        for task in self.rtapp_tasks:
-            task_cpu = int(task.strip("{}_".format(self.task_prefix)))
-            task_start = sdf[(sdf.next_comm == task) & (sdf["__cpu"] == task_cpu)].index[0]
+        # Find out when all tasks started executing on their designated CPU
+        for task, profile in self.rtapp_profile.items():
+            task_cpu = profile.phases[0].cpus[0]
+
+            names = self.rtapp_tasks_map[task]
+            assert len(names) == 1
+            name = names[0]
+
+            task_start = sdf[(sdf.next_comm == name) & (sdf["__cpu"] == task_cpu)].index[0]
             last_start = max(last_start, task_start)
 
         self.start_time = last_start


lisa/tests/scheduler/misfit.py
```diff
@@ -118,9 +118,15 @@ def __init__(self, res_dir, plat_info):
...
-            task_start = sdf[(sdf.next_comm == task) & (sdf["__cpu"] == task_cpu)].index[0]
+        # Find out when all tasks started executing on their designated CPU
+        for task, profile in self.rtapp_profile.items():
+            task_cpu = profile.phases[0].cpus[0]
+
+            names = self.rtapp_tasks_map[task]
+            assert len(names) == 1
+            name = names[0]
+
+            task_start = sdf[(sdf.next_comm == name) & (sdf["__cpu"] == task_cpu)].index[0]
```
douglas-raillard-arm: Could we do something like that ?
```
df = df_filter_task_ids(sdf, TaskID(None, name), comm_col='next_comm')
task_start = df[df["__cpu"] == task_cpu].index[0]
```

That's a bit more verbose but I'd rather maintain the logic of task filtering in one place if possible, and ease the transition to using TaskID as much as possible, so that it becomes the common currency.

valschneider: I'm not convinced it is adding much value here seeing as those are names of tasks we entirely control.

I also filter on `prev_comm` and `next_comm` values in `test_preempt_time()`. This *could* be converted to using `TaskID` filters, but would require one call to `df_filter_task_ids` per column filtering rather than a trivial (for both reader and writers) dataframe filter.

IMO `TaskID` has value in analysis methods since it makes them more robust to non-synthetic traces (i.e. analysing "real" workloads), but less so for synthetic tests.

douglas-raillard-arm: True, that is definitely needed for "random" traces, but we control what executes at the same time as us. The only potential issue is if the freezer is disabled (non-root) and another rt-app task lingers around for some reason, but we can probably ignore that case

douglas-raillard-arm: The other reason I had in mind would be if we start having a `task_id` column or something like that, completely standardized between dataframes. The filtering function could just be updated, otherwise we will have to convert the world to make that happen.

valschneider: > The other reason I had in mind would be if we start having a `task_id` column or something like that, completely standardized between dataframes. 

Mmm interesting idea, it would make chaining filters easier. Dunno if that can be done as a dynamic attribute (i.e. not actually grow the size of the df).

douglas-raillard-arm: Does pandas support dynamic columns like that ? OTH we could have a column containing Python objects (i.e. pointers to TaskID tuples). Filtering on these should be fast if we make sure we don't use duplicated TaskIDs (need to check if pandas support filtering by object identity, i.e. pointer comparison). It should even be faster than current string comparison of `comm`.

This would also allow dropping the comm and PID columns in some cases (with a helper to re-add them for debugging), which should reduce the current memory footprint.

douglas-raillard-arm: Another advantage of abstracting these kind of common filters is that we may be able to hand over some of them to the future trace parser, for even better performances (not sure exactly what it would like yet, but if we want to do that, we need an abstraction).

PR #1082, opened 2019-09-18
lisa.tests.hotplug: Add number of missing online CPUs metric

douglas-raillard-arm:
This metric is redundant with the list of CPUs that are online, but can be
easily plotted, unlike the list of CPUs.

### DIFF
--- a/lisa/tests/hotplug/torture.py
+++ b/lisa/tests/hotplug/torture.py
@@ -204,8 +204,9 @@ class HotplugBase(TestBundle):
         Test that all CPUs came back online after the hotplug operations
         """
         res = ResultBundle.from_bool(self.hotpluggable_cpus == self.live_cpus)
-        res.add_metric("hotpluggable CPUs", self.hotpluggable_cpus)
-        res.add_metric("Online CPUs", self.live_cpus)
+        dead_cpus = sorted(set(self.hotpluggable_cpus) - set(self.live_cpus))
+        res.add_metric("dead CPUs", dead_cpus)
+        res.add_metric("number of dead CPUs", len(dead_cpus))
         return res
 
 class HotplugTorture(HotplugBase):


lisa/tests/hotplug/torture.py
```diff
@@ -206,6 +206,7 @@ def test_cpus_alive(self) -> ResultBundle:
         res = ResultBundle.from_bool(self.hotpluggable_cpus == self.live_cpus)
         res.add_metric("hotpluggable CPUs", self.hotpluggable_cpus)
         res.add_metric("Online CPUs", self.live_cpus)
+        res.add_metric("Number of missing online CPUs", len(self.hotpluggable_cpus) - len(self.live_cpus))
```
valschneider: I don't necessarily object, but if some plotting tool needs this data, couldn't it build it itself from the other two metrics?

Otherwise we could get rid of the other two metrics and only have a "dead CPUs" metric which tell us everything we need: we expect those CPUs to be online but they aren't. Less noise, and we can plot it :D

douglas-raillard-arm: > I don't necessarily object, but if some plotting tool needs this data, couldn't it build it itself from the other two metrics?

So far, we managed to make the rest of the world 100% test agnostic. `bisector` knows nothing about the test cases, `exekall` knows nothing about the test cases, `grafana` knows nothing about them either. I'd rather keep the situation as it is and centralize all the code that deals with test semantic in the test itself or its class, so I don't end up spending days pushing related commits to tens of repositories to update lists of foos, mapping from bars to human readable foobars and creating virtual metrics out of baz :-)

douglas-raillard-arm: > which tell us everything we need: we expect those CPUs to be online but they aren't

Do you mean a list of CPUs ? A list can unfortunately not be plotted by things outside our notebook world. If we talk about a number of CPUs, we clearly miss the detail of what CPU failed to come online, so we can't remove that.

valschneider: Right, fair points (on both the agnostism and plotting lists).

Then I'd suggesting killing the current metrics, and only have: "dead CPUs" and "number of dead CPUs".

douglas-raillard-arm:
@valschneider updated with the metrics you suggested

PR #1083, opened 2019-09-18
lisa.wlgen.rta: Disable mlockall ("lock_pages") when running as non-root

douglas-raillard-arm:
mlockall requires privileges to succeed, so disable it when running as a regular
user.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -185,6 +185,11 @@ class RTA(Workload):
             'duration': -1 if not max_duration_s else max_duration_s,
             'calibration': calibration,
             'logdir': self.run_dir,
+            # TODO: this can only be enabled when rt-app is running as root.
+            # unfortunately, that's currently decided when calling
+            # run(as_root=True), at which point we already generated and pushed
+            # the JSON
+            'lock_pages': False,
         }
 
         if max_duration_s:


douglas-raillard-arm:
On hold, since it looks like an rtapp bug (mlockall should be disabled for CFS tasks).

douglas-raillard-arm:
This is an rt-app issue introduced in:
395ec42c28c1e7dbcd08904903c01936ef7f7e27 rt-app: add per phase set priority event

rt-app more recent that this commit will just apply the lock_page settings, regardless of the sched policy. Since it's enabled by default, rtapp will fail when running as non-privileged user.

lisa/wlgen/rta.py
```diff
@@ -180,11 +180,18 @@ def by_profile(cls, target, name, profile, res_dir=None, default_policy=None,
 
         calibration = self._process_calibration(calibration)
 
+        if target.is_rooted:
+            lock_pages = True
+        else:
+            lock_pages = False
+            logger.debug('rtapp global option "lock_pages" disabled when non-root')
```
derkling: Since this could affect results, would not be better to warn here?

douglas-raillard-arm: I'm not sure if warning just because somebody runs as non-root is a good thing to do. So far rt-app was not using mlockall for any of the CFS stuff and we did not have lots of issues. That said, if it really makes a noticeable difference on some setups, it would make sense to actually warn.

douglas-raillard-arm:
Updated with lock_pages unconditionally disabled

PR #1086, opened 2019-09-19
lisa.conf: Restrict valid key names

douglas-raillard-arm:
Avoid fancy names that would make it impossible to build new things out of the
names, such as key paths etc.

### DIFF
--- a/lisa/conf.py
+++ b/lisa/conf.py
@@ -24,6 +24,7 @@ import inspect
 import itertools
 import textwrap
 import logging
+import re
 
 from lisa.utils import (
     Serializable, Loggable, get_nested_key, set_nested_key, get_call_site,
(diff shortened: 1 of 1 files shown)


lisa/conf.py
```diff
@@ -56,7 +57,14 @@ class KeyDescBase(abc.ABC):
     to sanitize user input and generate help snippets used in various places.
     """
     INDENTATION = 4 * ' '
+    _VALID_NAME_PATTERN = r'^[a-zA-Z0-9-]+$'
```
valschneider: No underscores? I guess that's fine since we go with dash instead.

douglas-raillard-arm: yes, dashes are in use and I'd rather avoid having a mix of both, and the tentation for somebody to play with `setattr()` (if that somebody is motivated enough, they will do like in argparse).

PR #1087, opened 2019-09-19
Update tracepoint module and patch to fix some kbuild errors

qais-yousef:
kernel/sched/sched_tp.c: In function 'sched_pelt_se':
>> kernel/sched/sched_tp.c:77:22: error: 'struct sched_entity' has no member named 'cfs_rq'; did you mean 'on_rq'?
      void *cfs_rq = se->cfs_rq;
                         ^~~~~~
                         on_rq

### DIFF
--- a/tools/kmodules/sched_tp/sched_tp.c
+++ b/tools/kmodules/sched_tp/sched_tp.c
@@ -16,6 +16,15 @@ static inline struct cfs_rq *get_group_cfs_rq(struct sched_entity *se)
 #endif
 }
 
+static inline struct cfs_rq *get_se_cfs_rq(struct sched_entity *se)
+{
+#ifdef CONFIG_FAIR_GROUP_SCHED
+	return se->cfs_rq;
+#else
+	return NULL;
+#endif
+}
+
 static void sched_pelt_cfs(void *data, struct cfs_rq *cfs_rq)
 {
 	if (trace_sched_load_cfs_rq_enabled()) {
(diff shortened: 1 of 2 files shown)


tools/kmodules/sched_tp/sched_tp.c
```diff
@@ -16,6 +16,15 @@ static inline struct cfs_rq *get_group_cfs_rq(struct sched_entity *se)
 #endif
 }
 
+static inline struct cfs_rq *get_se_cfs_rq(struct sched_entity *se)
+{
+#ifdef CONFIG_FAIR_GROUP_SCHED
+	return se->cfs_rq;
+#else
+	return NULL;
```
valschneider: Or you could return the root CFS rq of the task's CPU rq (see `fair.c::cfs_rq_of()`):
```
        struct task_struct *p = task_of(se);                                                         
        struct rq *rq = task_rq(p);                                                                  
                                                                                                     
        return &rq->cfs;    
```

douglas-raillard-arm: I would go with the one that either works with our tests, or fail in the most obvious way if making it work is not possible

qais-yousef: The function is called get_se_cfs_rq(), so if we don't return the se->cfs_rq we returning the wrong thing, no? This function is called in sched_pelt_se() trace event

qais-yousef: @douglas-raillard-arm agreed. There's no functional changes here and the kmodule stayed faithful to whatever we had before moving to tracepoints. So unless this breaks something, I'd rather not do any improvements. Especially I have to maintain multiple versions of this now.

valschneider: > if we don't return the se->cfs_rq we returning the wrong thing, no?

Well the root CFS rq *is* the entity's CFS rq it is enqueued on, so from a naming perspective that's still be correct. It might still be useful to get the CPU the thing's enqueued on for !group scheduling configs, but that was just a drive-by comment anyway so I won't fight for it :)

qais-yousef: I will certainly consider your comment next time I look at this in detail :) At the moment this PR is to sync Lisa with changes I made to our integration branch a while back.

PR #1091, opened 2019-09-21
Add key documentation to YAML conf output

douglas-raillard-arm:
TODO: cli tool/online doc generating default conf for all SimpleMultiSrcConf subclasses in one file.

### DIFF
--- a/lisa/conf.py
+++ b/lisa/conf.py
@@ -30,6 +30,8 @@ from lisa.utils import (
     is_running_sphinx, get_cls_name, HideExekallID,
 )
 
+from ruamel.yaml.comments import CommentedMap
+
 class DeferredValue:
     """
     Wrapper similar to functools.partial.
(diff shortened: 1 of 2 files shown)


lisa/conf.py
```diff
@@ -1235,6 +1243,62 @@ def from_map(cls, *args, **kwargs):
...
+            ``add_placeholder`` is used.
+        :type placeholder: object
+        """
+
+        def format_comment(key_desc):
+            comment = key_desc.help
+            if not comment:
+                return comment
+            else:
+                return comment[0].upper() + comment[1:]
```
valschneider: Is that just a capitalization of the first letter? If so, wouldn't `return comment.title()` suffice?

douglas-raillard-arm: title() makes all first letters uppercase, not just for the first word:
```
>>> 'foo bar'.title()
'Foo Bar'
```

valschneider: Aaah. What about `.capitalize()`? :D

douglas-raillard-arm: Good point, the docstring of `capitalize()` is a kind of tautology ...

PR #1093, opened 2019-09-23
Test util est

douglas-raillard-arm:
Respin of https://github.com/ARM-software/lisa/pull/1085
Changes compared to initial PR are left as extra commits on top.

TODO: decide on how kernel version should be checked.

### DIFF
--- a/lisa/tests/staging/scheduler/util_tracking.py
+++ b/lisa/tests/staging/scheduler/util_tracking.py
@@ -225,7 +225,7 @@ class Convergence(UtilTrackingBase):
                     failures.append(phase.start)
                     continue
 
-        bundle = ResultBundle.from_bool(len(failure_reasons)==0)
+        bundle = ResultBundle.from_bool(failure_reasons)
         bundle.add_metric("fast ramp", self.fast_ramp)
         bundle.add_metric("phases stats", metrics)
         if not failure_reasons:
@@ -328,12 +328,10 @@ class Convergence(UtilTrackingBase):
                     failures.append(activation)
                     continue
 
-        bundle = ResultBundle.from_bool(len(failure_reasons)==0)
+        if failure_reasons:
+            self._plot_signals(test_task, 'activations', failures)
+
+        bundle = ResultBundle.from_bool(not failure_reasons)
         bundle.add_metric("signals", metrics)
-        if not failure_reasons:
-            return bundle
-
-        self._plot_signals(test_task, 'activations', failures)
         bundle.add_metric("failure reasons", failure_reasons)
-
         return bundle
--- a/lisa/tests/staging/scheduler/util_tracking.py
+++ b/lisa/tests/staging/scheduler/util_tracking.py
@@ -1,0 +1,339 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2019, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+from collections import OrderedDict, namedtuple
+import itertools
+from statistics import mean
+
+import matplotlib.pyplot as plt
+import pylab as pl
+
+from devlib.target import KernelVersion
+
+from lisa.tests.base import (TestMetric, Result, ResultBundle, TestBundle,
+                             RTATestBundle, CannotCreateError)
+from lisa.target import Target
+from lisa.utils import ArtifactPath, memoized
+from lisa.wlgen.rta import Periodic, Ramp
+from lisa.trace import FtraceCollector, requires_events
+from lisa.analysis.rta import RTAEventsAnalysis
+from lisa.analysis.tasks import TaskState, TasksAnalysis
+from lisa.datautils import series_integrate
+
+from lisa.tests.scheduler.load_tracking import LoadTrackingHelpers
+
+class UtilTrackingBase(RTATestBundle, LoadTrackingHelpers):
+    """
+    Base class for shared functionality of utilization tracking tests
+    """
+
+    @classmethod
+    def _from_target(cls, target:Target, *,
+                     res_dir:ArtifactPath=None,
+                     ftrace_coll:FtraceCollector=None) -> 'UtilTrackingBase':
+        plat_info = target.plat_info
+        rtapp_profile = cls.get_rtapp_profile(plat_info)
+
+
+        # After a bit of experimenting, it turns out that on some platforms
+        # misprediction of the idle time (which leads to a shallow idle state,
+        # a wakeup and another idle nap) can mess up the duty cycle of the
+        # rt-app task we're running. In our case, a 50% duty cycle, 16ms period
+        # task would always be active for 8ms, but it would sometimes sleep for
+        # only 5 or 6 ms.
+        # This is fine to do this here, as we only care about the proper
+        # behaviour of the signal on running/not-running tasks.
+        with target.disable_idle_states():
+            with target.cpufreq.use_governor('performance'):
+                cls.run_rtapp(target, res_dir, rtapp_profile, ftrace_coll)
+
+        return cls(res_dir, plat_info)
+
+class Convergence(UtilTrackingBase):
+    """
+    Basic checks for estimated utilization signals
+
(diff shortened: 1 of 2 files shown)


lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+    def _plot_signals(self, task, test, failures):
+        signals = ['util', 'util_est_enqueued', 'util_est_ewma']
+        ax = self.trace.analysis.load_tracking.plot_task_signals(task, signals=signals)
+        ax = self.trace.analysis.rta.plot_phases(task, axis=ax);
+        for start in failures:
+            ax.axvline(start, alpha=0.5, color='r', axis=ax)
+        filepath = os.path.join(self.res_dir, 'util_est_{}.png'.format(test))
+        self.trace.analysis.rta.save_plot(ax.figure, filepath=filepath)
+
+    @requires_events('sched_util_est_task', 'sched_load_se')
```
valschneider: I was going to say that should be `df_tasks_signal.used_events` but it's a bit trickier than this... Ohwell.

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+    @RTAEventsAnalysis.task_phase_windows.used_events
+    @RTATestBundle.check_noisy_tasks(noise_threshold_pct=1)
+    def test_areas(self) -> ResultBundle:
+        """
+        Test signals are properly "dominated".
+
+        The integral of `util_est_enqueued` is expected to be always not
+        smaller than that of `util_avg`, since this last is subject to decays
```
valschneider: I have only one negation unit in my sentence reading pipeline, what about `expected to be always greater than that of...`

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+        while the first not.
+
+        On fast-ramp systems, the `util_est_ewma` signal is never smaller then
```
valschneider:
```suggestion
        On fast-ramp systems, the `util_est_ewma` signal is always greater than
```

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+                  ensure `util_avg` saturation.
+        - DOWN: periodic ramp-down task, to slowly decay `util_avg`
+        - UP: periodic ramp-up task, to slowly increase `util_avg`
+
+        """
+        failure_reasons = {}
+        metrics = {}
+
+        # We have only two task: the main 'rt-app' task and our 'test_task'
+        test_task = self.trace.analysis.rta.rtapp_tasks[-1]
```
valschneider: That's sorted on `TaskID`s, so we should search for actual task we want (by `comm`)

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+        failures = []
+        for phase in self.trace.analysis.rta.task_phase_windows(test_task):
+            phase_df = ue_df[phase.start:phase.end]
+            area_enqueued = series_integrate(phase_df.util_est_enqueued)
+            area_ewma = series_integrate(phase_df.util_est_ewma)
+
+            phase_df = ua_df[phase.start:phase.end]
+            area_util = series_integrate(phase_df.util)
+
+            metrics[phase.id] = self.PhaseStats(phase.start, phase.end,
```
valschneider: I think you commented on this already, but this should be "properly" named. Also, I never really thought about it - do we want to let `metrics` contain non-standard types? We probably don't for ease of plotting/human-parsing

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+                failure_reasons[phase_name] = 'Enqueued smaller then Util Average'
+                failures.append(phase.start)
+                continue
+
+            # Running on FastRamp kernels:
+            if self.fast_ramp:
+
+                # STABLE, DOWN and UP:
+                if area_ewma < area_enqueued:
+                    failure_reasons[phase_name] = 'NO_FAST_RAMP: EWMA smaller then Enqueued'
```
valschneider: Those failure strings are inverted, no? This should be `FAST_RAMP:`

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+            else:
+
+                # STABLE: ewma ramping up
+                if phase.id == 0 and area_ewma > area_enqueued:
+                    failure_reasons[phase_name] = 'FAST_RAMP(STABLE): EWMA bigger then Enqueued'
+                    failures.append(phase.start)
+                    continue
+
+                # DOWN: ewma ramping down
+                if 0 < phase.id < 5 and area_ewma < area_enqueued:
```
valschneider: Fiddling with phase ids is finicky, we should do something like allow specifying a friendly name to an `RTATask` and have it copied in its phases. That way we could do something like
`if rtapp_task.phases[phase.id].name == "ramp_down" and ...`

That way we can change the rtapp profile without having to figure out how many phases it generates. It's not a problem here because we pin the task on a big CPU, but if we change that in the future we'll need to adapt the profile with the capacity landscape, which might result in a dynamically varying number of phases anyway...

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+                    failures.append(phase.start)
+                    continue
+
+                # UP: ewma ramping up
+                if phase.id > 4 and area_ewma > area_enqueued:
+                    failure_reasons[phase_name] = 'FAST_RAMP(UP): EWMA bigger then Enqueued'
+                    failures.append(phase.start)
+                    continue
+
+        bundle = ResultBundle.from_bool(len(failure_reasons)==0)
```
valschneider:
```suggestion
        bundle = ResultBundle.from_bool(not failures)
```

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+        bundle.add_metric("fast ramp", self.fast_ramp)
+        bundle.add_metric("phases stats", metrics)
+        if not failure_reasons:
+            return bundle
+
+        # Plot signals to support debugging analysis
+        self._plot_signals(test_task, 'areas', failures)
```
valschneider: IMO that should be done regardless of the test result - we do plot a few things unconditionally in `load_tracking`.

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+
+    @requires_events('sched_util_est_task')
+    @TasksAnalysis.df_task_states.used_events
+    @RTATestBundle.check_noisy_tasks(noise_threshold_pct=1)
+    def test_activations(self) -> ResultBundle:
+        """
+        Test signals are properly "aggregated" at enqueue/dequeue time.
+
+        On fast-ramp systems, the `util_est_enqueud` is expected to be always
+        not bigger then `util_est_ewma`.
```
valschneider:
```suggestion
        smaller than `util_est_ewma`.
```

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+        # Define a time interval to correlate relative trace events.
+        def restrict(df, time, delta=1e-3):
+            return df[time-delta:time+delta]
+
+        failures = []
+        for idx, activation in enumerate(activations):
+            avg, enq, ewma = restrict(df, activation)[[
+                'util_avg', 'util_est_enqueued', 'util_est_ewma']].iloc[-1]
+
+            metrics[idx+1] = self.ActivationSignals(activation, avg, enq, ewma)
```
valschneider: Ditto on the metrics

lisa/tests/staging/scheduler/util_tracking.py
```diff
@@ -0,0 +1,338 @@
...
+
+    ActivationSignals = namedtuple("ActivationSignals", [
+        'time', 'util_avg', 'util_est_enqueued', 'util_est_ewma'])
+
+    @classmethod
+    def get_rtapp_profile(cls, plat_info):
+        big_cpu = plat_info["capacity-classes"][-1][0]
+
+        task = (
+            # Big task: 15 activations at full utilization
```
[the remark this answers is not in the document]
douglas-raillard-arm: Maybe that is because of the util ramp up time ? "full utilization" could mean "once the signal has settled"

lisa/tests/staging/scheduler/util_tracking.py
douglas-raillard-arm: Are you handling that as part of the wlgen update you are baking atm ?

valschneider: Not yet, but can/should do.

PR #1097, opened 2019-09-23
lisa.wlgen.rta: Restrict valid rt-app names

douglas-raillard-arm:
Since rt-app itself generates suffixes based on dashes and numbers, only
allow simple names: chars (upper and lower case), numbers and
underscore.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -46,6 +46,8 @@ class RTA(Workload):
 
     sched_policies = ['OTHER', 'FIFO', 'RR', 'DEADLINE']
 
+    ALLOWED_TASK_NAME_REGEX = r'^[a-zA-Z0-9_]+$'
+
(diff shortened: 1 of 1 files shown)


lisa/wlgen/rta.py
```diff
@@ -46,6 +46,8 @@ class RTA(Workload):
 
     sched_policies = ['OTHER', 'FIFO', 'RR', 'DEADLINE']
 
+    ALLOWED_TASK_NAME_REGEX = r'^[a-zA-Z0-9_]+$'
```
valschneider: Why not dashes? Is it for easier regexp'ing due to e.g. forking? Just curious.

douglas-raillard-arm: Dashes followed by numbers can introduce ambiguity since rt-app can use one or more `-XX` prefixes. If rt-app decides to add further suffixes, it will likely be based on even more dashes, so I'd rather avoid them altogether at our level

PR #1115, opened 2019-10-03
lisa.analysis.tasks: Add 'cpu' column to df_task_activations()

douglas-raillard-arm:
Allow easily knowing where the task was running.

### DIFF
--- a/lisa/analysis/tasks.py
+++ b/lisa/analysis/tasks.py
@@ -519,19 +519,15 @@ class TasksAnalysis(TraceAnalysisBase):
         def f(state):
             if state == TaskState.TASK_ACTIVE:
                 return active_value
-            elif state == TaskState.TASK_INTERRUPTIBLE:
+            else:
                 return sleep_value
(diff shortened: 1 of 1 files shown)


lisa/analysis/tasks.py
```diff
@@ -524,13 +525,13 @@ def f(state):
         if cpu is not None:
             df = df[df['cpu'] == cpu]
 
-        curr_state = df.curr_state
-        active_series = curr_state[
+        curr_state = df['curr_state']
+        df['active'] = curr_state[
```
valschneider: The change itself looks fine, but I wonder why we only look at `TASK_ACTIVE` and `TASK_INTERRUPTIBLE` ? Why not do `TASK_ACTIVE` means task is active, all other states means not active ("sleeping")?

douglas-raillard-arm: I'm not quite sure, if you positively say so then we can make this change as well.

PR #1123, opened 2019-10-07
Fix trace window

douglas-raillard-arm:


lisa/tests/base.py
```diff
@@ -989,8 +990,22 @@ def trace_window(self, trace):
...
+            start_time = row['Time']
+            task = TaskID(comm=comm, pid=pid)
+            start_swdf = df_filter_task_ids(swdf, [task], pid_col='next_pid', comm_col='next_comm')
+            return start_swdf[start_swdf.index < start_time].index[-1]
+
+        # Find when the first rtapp phase starts, and take the associated
+        # sched_switch that is immediately preceding
+        rta_start = min(
+            get_first_switch(swdf, idx, row)
+            for idx, row in trace.analysis.rta.df_rtapp_phases_start().iterrows()
```
valschneider: `iterrows()` should be avoided as much as possible - it's a performance killer (and isn't even guaranteed to yield every row exactly once...). Now those are rtapp phases so we shouldn't have a gazillion of them, but perhaps this would be slightly better (will require tuning the inline function parameters):

```
df_phase_switch =  trace.analysis.rta.df_rtapp_phases_start().apply(get_first_switch())
rta_start = df_phase_switch.min
```

Maybe you can `timeit.timeit` and see if it's really noticeable.

douglas-raillard-arm: `timeit.timeit` wont show anything since my df has less than 10 rows, but I'll make the change if that works (also simplifies the code)

PR #1126, opened 2019-10-07
lisa.wlgen.rta: Strengthen sanity check on calibration values

douglas-raillard-arm:
* Check against the fastest CPU in each capacity class. This CPU is the
  one that has been the less impacted by external disturbances, that are
  usually random.

* Check that the rt-app calibration values are inversely proportional to
  the CPU capacities. When the scaling factor diverges too much for some
  CPUs, raise an exception. This indicates that rt-app will be unable to
  provide predictable utilizations across different CPUs, which is the
  underlying assumption of all wlgen-based code in LISA.

* Use the CPU capacities from plat_info rather than trying to fetch them
  directly using devlib. This follows the practice in LISA, allowing the
  user to provide the values in a configuration file in order to run as
  non-root (it's also faster and uses values printed in the log, for a
  consistent debugging experience).

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -23,9 +23,11 @@ import sys
 from collections import OrderedDict
 from shlex import quote
 import copy
+import itertools
 
 from lisa.wlgen.workload import Workload
 from lisa.utils import Loggable, ArtifactPath, TASK_COMM_MAX_LEN, groupby, nullcontext
+from lisa.pelt import PELT_SCALE
 
 class RTA(Workload):
     """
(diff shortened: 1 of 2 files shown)


lisa/wlgen/rta.py
```diff
@@ -408,15 +410,16 @@ def _calibrate(cls, target, res_dir):
 
         logger.info('Target RT-App calibration: %s', pload)
 
-        if 'sched' not in target.modules:
-            return pload
```
valschneider: The return should remain, you'd just have to turn the condition in `if 'cpu-capacities' not in plat_info`

lisa/wlgen/rta.py
```diff
@@ -408,15 +410,16 @@ def _calibrate(cls, target, res_dir):
...
-        cpu_capacities = target.sched.get_capacities()
+        cpu_capacities = plat_info['cpu-capacities']
 
-        # Find the max pload per capacity level
+        # Find the min pload per capacity level, i.e. the fastest detected CPU.
+        # It is more likely to represent the right pload, as it has suffered
+        # from less IRQ slowdown or similar disturbances that might be random.
         capa_pload = {
-            capacity: max(pload[cpu] for cpu, capa in cpu_caps)
+            capacity: min(pload[cpu] for cpu, capa in cpu_caps)
```
valschneider: The reason for using max watch to catch e.g. a rogue big CPU that happened to have a bigger pload than a LITTLE. So you might want to keep it there (or do both).

lisa/wlgen/rta.py
```diff
@@ -430,8 +433,45 @@ def _calibrate(cls, target, res_dir):
...
+        # the kernel uses
+        true_capacities = cls.get_cpu_capacity_from_calibrations(pload)
+        capa_factors_pct = {
+            cpu: true_capacities[cpu] / cpu_capacities[cpu] * 100
+            for cpu in cpu_capacities.keys()
+        }
+        dispersion_pct = max(abs(100 - factor) for factor in capa_factors_pct.values())
+
+        if dispersion_pct > 2:
+            logger.warning('The calibration values are not inversely proportional to the CPU capacities, the duty cycles will up to {:.2f}% off on some CPUs: {}'.format(dispersion_pct, capa_factors_pct))
```
valschneider: I think you're missing a word in the sentence - also, could we split the line somehow?

lisa/wlgen/rta.py
douglas-raillard-arm: problem with max() is that we select the pload that is the less relevant for computing the capacity. Background noise will always lead to higher pload, so the minimum one is the most representative.

I'll add another check to find the condition you mentioned instead.

lisa/wlgen/rta.py
douglas-raillard-arm: I can't find any missing word, where do you fail parsing the sentence ?
Splitting lines used to emit user-visible errors is usually a bad idea, for the reason mentioned in the kernel coding style:

> However, never break user-visible strings such as printk messages, because that breaks the ability to grep for them.
https://www.kernel.org/doc/html/v4.10/process/coding-style.html

PR #1131, opened 2019-10-10
Helper to get idle CPUs at a given timestamp

valschneider:
Wrote that thing to test out some wakeup change, it's proven useful.

### DIFF
--- a/lisa/analysis/idle.py
+++ b/lisa/analysis/idle.py
@@ -51,22 +51,22 @@ class IdleAnalysis(TraceAnalysisBase):
         :param cpu: CPU ID
         :type cpu: int
 
-        :returns: A :class:`pandas.Series` that equals 1 at timestamps where the
-          CPU is reported to be non-idle, 0 otherwise
+        :returns: A :class:`pandas.Series` that equals True at timestamps where the
+          CPU is reported to be non-idle, False otherwise
         """
         idle_df = self.trace.df_events('cpu_idle')
         cpu_df = idle_df[idle_df.cpu_id == cpu]
 
         cpu_active = cpu_df.state.apply(
-            lambda s: 1 if s == -1 else 0
+            lambda s: True if s == -1 else False
         )
 
         start_time = self.trace.start
 
         if cpu_active.empty:
-            cpu_active = pd.Series([0], index=[start_time])
+            cpu_active = pd.Series([False], index=[start_time])
         elif cpu_active.index[0] != start_time:
-            entry_0 = pd.Series(cpu_active.iloc[0] ^ 1, index=[start_time])
+            entry_0 = pd.Series(not cpu_active.iloc[0], index=[start_time])
             cpu_active = pd.concat([entry_0, cpu_active])
 
         # Fix sequences of wakeup/sleep events reported with the same index
@@ -123,8 +123,8 @@ class IdleAnalysis(TraceAnalysisBase):
         sr = pd.Series()
         for cpu in cpus:
             cpu_sr = self.signal_cpu_active(cpu)
-            cpu_sr = cpu_sr[cpu_sr == 1]
-            cpu_sr = cpu_sr.replace(1, cpu)
+            cpu_sr = cpu_sr[cpu_sr]
+            cpu_sr = cpu_sr.replace(True, cpu)
             sr = sr.append(cpu_sr)
 
         return pd.DataFrame({'cpu': sr}).sort_index()
--- a/lisa/analysis/idle.py
+++ b/lisa/analysis/idle.py
@@ -25,6 +25,7 @@ from trappy.utils import handle_duplicate_index
 from lisa.utils import memoized
 from lisa.datautils import series_integrate
 from lisa.analysis.base import TraceAnalysisBase
+from lisa.analysis.tasks import TaskState, TasksAnalysis
 from lisa.trace import requires_events
 
 
@@ -71,6 +72,50 @@ class IdleAnalysis(TraceAnalysisBase):
 
         # Fix sequences of wakeup/sleep events reported with the same index
         return handle_duplicate_index(cpu_active)
+
+    @memoized
+    @signal_cpu_active.used_events
+    @TasksAnalysis.df_tasks_states.used_events
+    def signal_cpu_sched_active(self, cpu):
+        """
+        Build a square wave representing the active (i.e. non-idle) CPU time
+        from the scheduler's point of view.
+
+        :param cpu: CPU ID
+        :type cpu: int
+
+        :returns: A :class:`pandas.Series` that equals True at timestamps where the
+          CPU is reported to be non-idle, False otherwise
+        """
+        # Find the earliest time at which the scheduler placed a task on
+        # this CPU's rq. The scheduler won't consider this CPU idle at this
+        # time. A proper rq depth signal would make this more accurate.
+        signal_active = self.signal_cpu_active(cpu).copy()
+        df_states = self.trace.analysis.tasks.df_tasks_states()
+
+        df_states = df_states[df_states.target_cpu == cpu]
+        sched_active = pd.Series(index=df_states.index, data=1)
+
+        # If there's an enqueue when the CPU is deemed inactive, we want to
+        # "move" the active signal earlier in the df.
+        # To do that, first set the signal to active whenever there's such a
+        # wakeup, then clear up redundant values.
+        #
+        # XXX: bitwise 'or' has surprising results:
+        # s1 = pd.Series(index=[1.25], data=[True])
+        # s2 = pd.Series(index=[1.05], data=[True])
+        # print(s1 | s2)
+        # 1.05    False
+        # 1.25     True
+        def orify(a, b):
+            return a or b
+
+        signal_active = signal_active.combine(sched_active, orify, fill_value=False)
+
+        # Delete successive duplicate values
+        signal_active = signal_active[signal_active.shift() != signal_active].dropna()
+
+        return signal_active
 
     @signal_cpu_active.used_events
     def signal_cluster_active(self, cluster):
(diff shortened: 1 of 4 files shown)


lisa/analysis/idle.py
```diff
@@ -236,6 +236,27 @@ def df_cluster_idle_state_residency(self, cluster):
...
+        """
+        idles = []
+        for cpu in cpus:
+            df_active = trace.analysis.idle.signal_cpu_active(cpu)
+            before, after = df_float_index_neighbours(df_active, timestamp)
+
+            # If the CPU went idle before T, it was idle at T
+            # If the CPU left idle after T, it was idle at T
+            # Explicitely check value in case it's None
+            if before == False or after == True:
```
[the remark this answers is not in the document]
valschneider: Well the good thing is we get away with a single check. The way I see it a separate `None` check would be
```
if after is not None:
    idle |= after
if before is not None:
    idle |= not before

if idle:
    idles.append(cpu)
```

Because both `after` or `before` can be `None`, it's somewhat annoying to check that in a single if. I think the explicit check is fine (and it's commented!), but if you're adamant about it I'll split it out.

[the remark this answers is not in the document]
douglas-raillard-arm: The issue is that the only type known for sure here is `None`. However, it seems that `True == 1.0` and `False == 0.0` as well, so Python seems to diverges from its usual strong typing behavior and allows comparing any numeric type (bool is a subclass of int). Given that, your code should be robust against type change.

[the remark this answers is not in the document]
valschneider: Completely forgot about the types in those columns, you're right.

valschneider:
Got some more changes after playing around with the thing I was trying to test. I dropped the neighbour helper and just went with helpers that give the prev/next index. Also added some unrelated changes for stuff I happened to stare at.

lisa/trace.py
```diff
@@ -152,6 +152,27 @@ def add_events_deltas(self, df, col_name='delta', inplace=True):
 
         return df
 
+    def df_discretize_series(self, series, time_delta):
```
douglas-raillard-arm: Could we move that to datautils ? It does not really have anything to do with the trace itself since it could just work from series.index[0] to series.index[-1]. 

Also that's not really a discretization operation, but some kind of local integration (don't really know how to name that).

lisa/datautils.py
```diff
@@ -504,6 +504,55 @@ def df_window(df, window, method='inclusive', clip_window=False):
     """
     return _data_window(df, window, method, clip_window)
 
+def _data_prev_index(data, ref):
```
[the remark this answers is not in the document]
valschneider: That'd be better aye

lisa/analysis/idle.py
```diff
@@ -236,6 +236,32 @@ def df_cluster_idle_state_residency(self, cluster):
...
         return idle_time_df
 
+    @requires_events('cpu_idle')
+    def idle_cpus_at(self, timestamp):
+        """
+        :returns: list of CPUs that were idle at the given ``timestamp``
+        """
+        idles = []
+        for cpu in range(self.trace.cpus_count):
+            df_active = trace.analysis.idle.signal_cpu_active(cpu)
```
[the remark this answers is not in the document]
valschneider: AFAICT this is the equivalent of only using `df_prev_index()`.

It might not be super likely, but you could have started the trace with a certain CPU busy, and it only goes in idle much later in the trace. That means you have a chunk of your trace where there is no previous index for any requested timestamp, but you can still figure out if the CPU was idle by looking at the next index (which is what I do).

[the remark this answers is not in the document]
douglas-raillard-arm: yes, that issue should be solved at the "signal" level rather than bloating the rest of LISA with partial workarounds. Even when considering the next index, the issue will still creep up if you use a small window on the trace. Once the signal fixup is in place, all these issues will disappear, so I'd rather keep the code clean so we don't end up reimplementing/simplifying dozens of helpers just because we fixed the root issue.

lisa/trace.py
douglas-raillard-arm: > And yes, naming is hard :(

How about `series_fixed_window_integrate()` ? That's actually what it does.

valschneider:
Moved the discretize/bucketer/integral helper to datautils, and applied @douglas-raillard-arm's logic to the idle CPU search. I'm not a big fan of purposedly not doing a better implementation by handling requests before a CPU's first `cpu_idle` event, but so be it.

lisa/analysis/idle.py
```diff
@@ -51,22 +51,22 @@ def signal_cpu_active(self, cpu):
...
         )
 
         start_time = self.trace.start
 
         if cpu_active.empty:
-            cpu_active = pd.Series([0], index=[start_time])
+            cpu_active = pd.Series([False], index=[start_time])
         elif cpu_active.index[0] != start_time:
-            entry_0 = pd.Series(cpu_active.iloc[0] ^ 1, index=[start_time])
+            entry_0 = pd.Series(cpu_active.iloc[0] ^ True, index=[start_time])
```
douglas-raillard-arm: Wouldn't that be expressed more simply as ?: 
```
not cpu_active.iloc[0]
```
rather than:
```
cpu_active.iloc[0] ^ True
```

douglas-raillard-arm:
This PR might provide an alternative to `series_fixed_window_integrate`

https://github.com/ARM-software/lisa/pull/1167
```
# Rolling mean over 1s window, output indexes centered on the window
series_rolling_apply(series, lambda s: series_mean(s), 1, center=True)
```

The output of that rolling sum should carry more information than a fixed window integration, since the local extrema should match the local extrema of the original series. It has no risk of dampening them because of bad alignment of the window with the data.

In other words, `x.shift().rolling(f) == x.rolling(f).shift()` (with `shift()` shifting a constant amount of time and not a number of samples). That (very desirable) property does not hold for a non-rolling window filtering.

douglas-raillard-arm:
>  I'm not a big fan of purposedly not doing a better implementation by handling requests before a CPU's first cpu_idle event, but so be it.

We will get there without spreading this clutter, I promise ;) . I'll make a PR soon to introduce something like that

valschneider:
@douglas-raillard-arm I've split out the fixed window out of this PR, I'm going to test out the rolling window separately. In the meantime I've just refactored the datautils helpers - there's a single entrypoint now that takes a `prev` boolean to decide whether to go for the previous or next index.

lisa/datautils.py
```diff
@@ -504,6 +504,40 @@ def df_window(df, window, method='inclusive', clip_window=False):
...
+
+def df_adjacent_index(df, ref, prev=True):
+    """
+    :param ref: The index value to look around
+
+    :param prev: Whether to find the previous index (``True``)
+      or the next index (``False``)
+    :type prev: bool
+
+    :returns: The index value directly preceding ``ref`` if ``prev``,
```
douglas-raillard-arm: That's not strictly the case.: if df[ref] exists, it will be returned instead of the prev/next value.

Maybe something like:
```
The row at index directly preceding ``ref`` if ``prev`` if (``ref`` is not present in the index)
```

lisa/datautils.py
```diff
@@ -504,6 +504,40 @@ def df_window(df, window, method='inclusive', clip_window=False):
...
+    elif ref < data.index[-1]:
+        res = data.index.get_loc(ref, "bfill")
+
+    return res
+
+def df_adjacent_index(df, ref, prev=True):
+    """
+    :param ref: The index value to look around
+
+    :param prev: Whether to find the previous index (``True``)
```
douglas-raillard-arm: Could we use `method="prev"` and `method="next"` instead of a boolean, and remove the default value ?
That avoids putting some arbitrary emphasis on one or the other, and makes the client code clear as to which behavior is in use. Since the name of the function is neutral (as it should), there is no reason for it to privilege prev over next.

lisa/analysis/idle.py
```diff
@@ -72,6 +73,50 @@ def signal_cpu_active(self, cpu):
...
+        :param cpu: CPU ID
+        :type cpu: int
+
+        :returns: A :class:`pandas.Series` that equals True at timestamps where the
+          CPU is reported to be non-idle, False otherwise
+        """
+        # Find the earliest time at which the scheduler placed a task on
+        # this CPU's rq. The scheduler won't consider this CPU idle at this
+        # time. A proper rq depth signal would make this more accurate.
+        signal_active = self.signal_cpu_active(cpu).copy()
```
douglas-raillard-arm: is copy() necessary ? We already do a `dropna(inplace=False)`

lisa/analysis/idle.py
```diff
@@ -72,6 +73,50 @@ def signal_cpu_active(self, cpu):
...
+        # XXX: bitwise 'or' has surprising results:
+        # s1 = pd.Series(index=[1.25], data=[True])
+        # s2 = pd.Series(index=[1.05], data=[True])
+        # print(s1 | s2)
+        # 1.05    False
+        # 1.25     True
+        def orify(a, b):
+            return a or b
+
+        signal_active = signal_active.combine(sched_active, orify, fill_value=False)
```
douglas-raillard-arm: might be easier to add all the columns into a dataframe, and then just return a final column ? That ensures indexes are aligned and is generally easier to printf-debug

lisa/analysis/idle.py
```diff
@@ -236,6 +281,37 @@ def df_cluster_idle_state_residency(self, cluster):
...
+        for cpu in range(self.trace.cpus_count):
+            series_active = active_signal_fn(cpu)
+
+            # Get very first event before 'timestamp'
+            #
+            # XXX: NaN if 'timestamp' is before the first event in the series.
+            # This could be fixed by checking the first event after 'timestamp':
+            # if the CPU goes out of idle later, it was idle during 'timestamp'
+            # and vice-versa.
+            is_active = series_active.reindex([timestamp], method='ffill').iloc[0]
```
douglas-raillard-arm: Can df_adjacent_index() be used here ?

lisa/analysis/idle.py
```diff
@@ -72,6 +73,50 @@ def signal_cpu_active(self, cpu):
         # Fix sequences of wakeup/sleep events reported with the same index
         return handle_duplicate_index(cpu_active)
 
+    @memoized
+    @signal_cpu_active.used_events
+    @TasksAnalysis.df_tasks_states.used_events
+    def signal_cpu_sched_active(self, cpu):
```
douglas-raillard-arm: Do we want to create a new function or just add an `as_sched=False` parameter to the existing `signal_cpu_active()` ?

lisa/analysis/idle.py
```diff
@@ -236,6 +281,37 @@ def df_cluster_idle_state_residency(self, cluster):
...
+            # if the CPU goes out of idle later, it was idle during 'timestamp'
+            # and vice-versa.
+            is_active = series_active.reindex([timestamp], method='ffill').iloc[0]
+            if not is_active:
+                idles.append(cpu)
+
+        return sorted(idles)
+
+    @signal_cpu_active.used_events
+    def idle_cpus_at(self, timestamp):
```
douglas-raillard-arm: Same about `as_sched=True`, this could save on docstring copy-pasting.

douglas-raillard-arm:
Accidentally merged, let's finish the review before merging

lisa/datautils.py
valschneider: I went with that, then backtracked because it was a boolean information. Let's backtrack again!

lisa/analysis/idle.py
valschneider: Originally I used the both prev/next values, but you suggested using only prev with a reindex which I've just copied here(see previous comments).

douglas-raillard-arm: Right, the original intent of the comment was to avoid dealing with both prev and next. Now that we only care about one of them, either that oneliner or the df_adjacent_index() can be used. df_adjacent_index() seems better since we still have it, as it avoids generating a full intermediate df.

lisa/analysis/idle.py
valschneider: Oh, I just remembered why I did this: they have different events requirements. One only needs `cpu_idle`, the other needs `cpu_idle`, `sched_switch` and `sched_wakeup`. I don't expect anyone to trace `cpu_idle` without tracing `sched_switch`, but it's easy to forget `sched_wakeup`, so I kept these separate.

douglas-raillard-arm: +1, it's better to avoid function that seemingly depend on everything, or optionally depend on everything and then do manual validation. Even when using optional events, everything ends up being traced and event list from `lisa-test --list` would quickly become meaningless.

PR #1133, opened 2019-10-11
lisa.utils: Add Loggable.logger property

douglas-raillard-arm:
Allows abbreviating self.get_logger() into self.logger .

In classmethods, cls.get_logger() is still needed, since having a "class
property" would require a metaclass. That would bring more potential (trivial)
metaclass conflicts that could confuse people writing new classes so it has not
been done.


douglas-raillard-arm:
Abandoning since it does not play well with inheritance

PR #1134, opened 2019-10-11
lisa.wlgen.rta: Speed up calibration

douglas-raillard-arm:
Disable idle state once before calibrating all CPUs, rather than for each CPU
everytime.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -402,7 +402,7 @@ class RTA(Workload):
                                  calibration="CPU{}".format(cpu),
                                  res_dir=res_dir)
 
-            with rta, target.disable_idle_states(), target.freeze_userspace():
+            with rta, target.freeze_userspace():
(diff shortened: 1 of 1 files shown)


lisa/wlgen/rta.py
```diff
@@ -402,7 +402,7 @@ def _calibrate(cls, target, res_dir):
                                  calibration="CPU{}".format(cpu),
                                  res_dir=res_dir)
 
-            with rta, target.disable_idle_states(), target.freeze_userspace():
+            with rta, target.freeze_userspace():
```
valschneider: Could we give the freezer the same treatment?

douglas-raillard-arm: Since we need to interact with the board between each run, I'd rather avoid upsetting the userspace. Maybe things would just work, but I'd rather avoid freezing things more than really necessary (not a very hard opinion though).

PR #1137, opened 2019-10-15
lisa.trace: Add Trace.from_target() factory classmethod

douglas-raillard-arm:
Allows creating a Trace object right from a Target, without an intermediate
FtraceCollector.

### DIFF
--- a/lisa/trace.py
+++ b/lisa/trace.py
@@ -30,6 +30,8 @@ import logging
 import webbrowser
 import inspect
 import shlex
+import contextlib
+import tempfile
 from functools import reduce, wraps
 from collections.abc import Sequence
 from collections import namedtuple
@@ -41,7 +43,7 @@ import trappy
 import devlib
 from devlib.target import KernelVersion
 
-from lisa.utils import Loggable, HideExekallID, memoized, deduplicate, deprecate
+from lisa.utils import Loggable, HideExekallID, memoized, deduplicate, deprecate, nullcontext
 from lisa.platforms.platinfo import PlatformInfo
 from lisa.conf import SimpleMultiSrcConf, KeyDesc, TopLevelKeyDesc, StrList, Configurable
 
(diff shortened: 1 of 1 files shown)


lisa/trace.py
```diff
@@ -352,6 +354,69 @@ def ftrace(self):
...
+
+            def __getattr__(self, attr):
+                return getattr(self.base_trace, attr)
+
+        proxy = TraceProxy()
+
+        with ftrace_coll:
+            yield proxy
+
+        with tempfile.NamedTemporaryFile(suffix='.dat', delete=delete_trace) as _file:
```
valschneider: In `ipynb/examples/typical_experiment.ipynb` we do:
```python
trace_path = os.path.join(wload.res_dir, "trace.dat")
with ftrace_coll:
    wload.run()
ftrace_coll.get_trace(trace_path)
```

So what about letting us specify a result directory? If it's present, save the file there, otherwise go with a tempfile that gets deleted.

douglas-raillard-arm:
@valschneider Updated with a `filepath` parameter for persistence. When not specified, a temp file is created and removed right after being parsed

lisa/trace.py
```diff
@@ -352,6 +354,80 @@ def ftrace(self):
...
+            def get_view(self, *args, **kwargs):
+                return self.base_trace.get_view(*args, **kwargs)
+
+            def __getattr__(self, attr):
+                return getattr(self.base_trace, attr)
+
+        proxy = TraceProxy()
+
+        with ftrace_coll:
+            yield proxy
```
valschneider: IIUC when we yield this it's actually "empty", but once we go through the equivalent of `__exit__()` we populate its `base_trace` which makes it work.

It seems obvious, but maybe we should add a warning in the docstring that the yielded object is only really usable *outside* of the `with` statement?

douglas-raillard-arm: True, or maybe I should raise an explicit exception if it used inside the `with` statement.

douglas-raillard-arm:
@valschneider Updated with warning comment in the docstring example and the following exception when used inside the `with` block:
```
RuntimeError: The trace instance can only be used outside its "with" statement.
```

PR #1146, opened 2019-10-18
Wlgen uclamp support (revision)

JaimeVHArm:
This pull request adds wlgen support for uclamp.
A pull request was previously opened for this feature in #1036, but this PR is a rewrite with little code in common. 
This PR uses a single uclamp_min/uclamp_max pair for each RTATask object. Complex combinations of clamp values are to be constructed using RTATask composition. This is unlike the previous PR where a callback applied to each phase of an RTATask is intended to allow for the construction of these behaviours.


lisa/wlgen/rta.py
```diff
@@ -553,7 +553,8 @@ class Phase(Loggable):
     :type barrier_after: str
     """
 
-    def __init__(self, duration_s, period_ms, duty_cycle_pct, cpus=None, barrier_after=None):
+    def __init__(self, duration_s, period_ms, duty_cycle_pct, cpus=None, barrier_after=None,
+                 uclamp_min=None, uclamp_max=None):
```
valschneider: While packing them into a single tuple might make it look tidier, I think we want to keep this interface. AFAICT there is room for more uclamp settings, so this is a more future-proof API (that or creating a uclamp class to pass in as parameter...).

lisa/wlgen/rta.py
```diff
@@ -717,10 +728,15 @@ def __init__(self, start_pct=0, end_pct=100, delta_pct=10, time_s=1,
...
             if load == 0:
-                phase = Phase(time_s, 0, 0, cpus)
+                phase = Phase(time_s, 0, 0, cpus, uclamp_min=uclamp_min,
+                              uclamp_max=uclamp_max)
             else:
-                phase = Phase(time_s, period_ms, load, cpus)
+                phase = Phase(time_s, period_ms, load, cpus, 
+                              uclamp_min=uclamp_min, uclamp_max=uclamp_max)
             phases.append(phase)
+            # It is only necessary to set clamp values on the first phase.
```
valschneider: We don't do this for the `cpus` parameter, which AFAICT have a similar behaviour (CPU affinity carries over phases if not specified otherwise). This should be an optimization on rt-app's side I think, so maybe we can just repeatedly specify uclamp values on our side.

lisa/wlgen/rta.py
```diff
@@ -806,8 +824,12 @@ def __init__(self, start_pct=100, end_pct=0, time_s=1, period_ms=100,
         for load in [start_pct, end_pct]:
             if load == 0:
                 continue
-            phase = Phase(time_s, period_ms, load, cpus)
+            phase = Phase(time_s, period_ms, load, cpus, uclamp_min=uclamp_min,
+                          uclamp_max=uclamp_max)
             phases.append(phase)
+            # It is only necessary to set clamp values on the first phase.
```
valschneider: Ditto on uclamp values.

lisa/wlgen/rta.py
JaimeVHArm: OK, this is no longer done.

PR #1150, opened 2019-10-21
Change plot defaults: always_save=False interactive=is_running_ipython()

douglas-raillard-arm:


valschneider:
This will need to update the tests that generate plots to keep saving them as artefacts. Otherwise LGTM.

douglas-raillard-arm:
I just checked and they actually all use an explicit `filepath` so they won't be affected. The other calls to plot methods are just intermediate, to add plots to an existing axis (and the last one saves to the file).

PR #1154, opened 2019-10-22
Load devlib modules on-demand

douglas-raillard-arm:
Load devlib modules on demand rather than loading them all in advance. This speeds up target connection time.

Depends on https://github.com/ARM-software/lisa/pull/1153


douglas-raillard-arm:
@JaviMerino This should save you from providing and excluded module list

valschneider:
Gave it a spin on our dual Xeon E5 system without any `excluded-modules` - I usually exclude everything because it's horrendously slow.

It does seem to be working fine, but is still taking about 2 minutes to setup the target, mostly because we want to fill information in the `plat_info`: frequencies, search for an energy model:

```
2019-10-23 14:50:01,377 INFO    : lisa.target.Target : Target configuration:
[...
2019-10-23 14:50:12,072 INFO    : lisa.target.Target : Loading target devlib module cpufreq
2019-10-23 14:51:37,255 INFO    : lisa.target.Target : Loading target devlib module sched
2019-10-23 14:52:05,435 INFO    : sched        : Scheduler sched_domain procfs entries found
2019-10-23 14:52:05,436 INFO    : sched        : Detected kernel compiled with SCHED_DEBUG=n
2019-10-23 14:52:05,437 INFO    : sched        : CPU capacity sysfs entries not found
2019-10-23 14:52:32,871 INFO    : lisa.target.Target : Effective platform information:
[...]
```

I think this mostly stems from too many sysfs files to read (It's dual socket, 40 CPUs total with SMT).
While this is fine to merge, we may want to rethink our `plat_info` strategy. Maybe we could move as much stuff as possible to be `DeferredValue`?

douglas-raillard-arm:
I'd rather avoid moving too much to deferred values, since that makes the log essentially useless. Right now, looking at the log gives the basic information that people will be interested in. If that disappears, debugging and usability will take a big hit. 

However, we could make devlib cpufreq module lazier so that it only loads what it's asked for. Maybe we don't use all the stuff it parses by default when loading ?

JaviMerino:
Cool, this has reduced my list of excluded modules to just ['cpufreq', 'sched']. Thanks!

PR #1162, opened 2019-10-29
lisa.wlgen.rta: Add RTA.resolve_trace_task_names()

douglas-raillard-arm:
Allow resolving RTA profile names to task names as found in a trace. To avoid
filtering the trace again and again, cache the result in RTA.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -24,6 +24,7 @@ from collections import OrderedDict
 from shlex import quote
 import copy
 import itertools
+import weakref
 
 from lisa.wlgen.workload import Workload
 from lisa.utils import Loggable, ArtifactPath, TASK_COMM_MAX_LEN, groupby, nullcontext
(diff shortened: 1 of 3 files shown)


douglas-raillard-arm:
@qais-yousef This should give what you were trying to do a few days ago:
```
 RTA.resolve_trace_task_names(trace, names)
```

lisa/wlgen/rta.py
```diff
@@ -534,6 +535,72 @@ def get_cpu_calibrations(cls, target, res_dir=None):
...
+            name: comms
+            for name, comms in task_map.items()
+            if name in names
+        }
+
+    def get_trace_task_names(self, trace):
+        """
+        Get the name used in the given trace for this task.
+        """
+        return self.resolve_trace_task_names(trace, [self.name])[self.name]
```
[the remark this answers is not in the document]
douglas-raillard-arm: the problem is that the task name is the key of the rta profile dict, and AFAIK it's not reflected anywhere else. Either we use `RTA.resolve_trace_task_names(profile.keys())` or we embed the name in the task objects (which you apparently plan to make optional).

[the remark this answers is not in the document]
valschneider: The task names are saved in `_late_init()`. They're either passed as parameter (for the alternative constructors we've already gutted the file, so it's a slight optimization not to re-read it), or read from the file when e.g. the regular constructor is used.

They show up in `self.tasks`, which is why I was suggesting to use this.

lisa/wlgen/rta.py
```diff
@@ -534,6 +535,72 @@ def get_cpu_calibrations(cls, target, res_dir=None):
...
+        # Only return what we were asked for, so the client code does not
+        # accidentally starts depending on whatever was requested in earlier
+        # calls
+        return {
+            name: comms
+            for name, comms in task_map.items()
+            if name in names
+        }
+
+    def get_trace_task_names(self, trace):
```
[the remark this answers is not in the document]
douglas-raillard-arm: not really:
1) using `memoized` would create a regular reference over the task, keeping it alive forever
2) we only want to memoize on the `trace`, and then be able to update the mapping in place.

douglas-raillard-arm:
Updated to return `TaskID` rather than plain names. This is generally more useful as it contains more information (the PID as well), and all the analysis APIs will play nicely with that.

lisa/wlgen/rta.py
```diff
@@ -534,6 +535,72 @@ def get_cpu_calibrations(cls, target, res_dir=None):
...
+        missing = sorted(prefix for prefix, task_ids in task_map.items() if not task_ids)
+        if missing:
+            raise RuntimeError("Missing tasks matching the following rt-app profile names: {}"
+                                .format(', '.join(missing)))
+        return task_map
+
+    # Mapping of Trace objects to their task map.
+    # We don't want to keep traces alive just for this cache, so we use weak
+    # references for the keys.
+    _traces_task_map = weakref.WeakKeyDictionary()
```
qais-yousef: It feels weird to read this between function definitions. Or my eyes just got tricked and it isn't? Shouldn't it be moved into init()?

douglas-raillard-arm: no, that's a class attribute

qais-yousef: Defining it between function still is weird. Define it at the top just after declaring the class where all the other attributes are makes more sense, no?

douglas-raillard-arm: The main reason I put it there is because it's right above the only method that uses it, so one does not have to scroll endlessly to see what goes together. It's not something public-facing in any case, so it's not a big deal if people don't notice it at first sight

qais-yousef: This is deviating from best practices in general IMO. But it's not a big deal if you really prefer it this way. FWIW people don't rely on scrolling when inspecting code only - but could expect all class attributes to be grouped together at the top.

douglas-raillard-arm: My point is that grouping by itself is useless, what matters is what grouping brings. It's useful to group public attributes if someone wants to discover what the class has to offer. Since that attribute is definitely not part of that list, it sounds more useful to have it close to where it's actually used. This attribute is actually an internal persistent state of `resolve_trace_task_names()` and nothing else. If I could, I would declare it inside the function but Python does not have `static` local variables as in C.

PR #1171, opened 2019-11-01
lisa.trace: Add FtraceCollector(trace_clock='global', saved_cmdlines_…

douglas-raillard-arm:
…nr=4096)

Allow changing the number of entries in the cmdlines/PID table created by the
kernel while tracing. This is needed on systems generating a lot of
tasks (Android), to avoid having placeholder task names such as "<...>".

Also allow changing the trace clock source, and default to the "global"
monotonic shared clock so that timestamp between CPUs are consistent.

Depends on new devlib features:
* https://github.com/ARM-software/devlib/pull/434
* https://github.com/ARM-software/devlib/pull/435

Function tracer can be used with trappy PR:
https://github.com/ARM-software/trappy/pull/299

Note that the functions are left as pointers rather than names, but that's the only way to get both entry and exit events. Dataframe sanitization will resolve the symbol names when parsed kallsyms from plat info is available. After sanitization, the dataframe looks like:
```
         __comm    __pid  __cpu  __line                  func  depth            func_name
Time                                                                                     
0.000000  <...>  2094921      2       0  18446744072324090592      0  select_task_rq_fair
0.000118  <...>  2094921      2       4  18446744072324090592      0  select_task_rq_fair
0.000318  <...>  2094920      7       8  18446744072324090592      0  select_task_rq_fair
0.001215  <...>  2094920      7      12  18446744072324090592      0  select_task_rq_fair
0.001521  <...>  2094918      0      18  18446744072324090592      0  select_task_rq_fair
...         ...      ...    ...     ...                   ...    ...                  ...
0.076765  <...>  2094947      4    1566  18446744072324090592      0  select_task_rq_fair
0.076771  <...>  2094947      4    1568  18446744072324090592      0  select_task_rq_fair
0.076892  <...>  2094948      6    1573  18446744072324090592      0  select_task_rq_fair
0.077026  <...>  2094948      6    1576  18446744072324090592      0  select_task_rq_fair
0.079242  <...>  2094948      6    1580  18446744072324090592      0  select_task_rq_fair
```


lisa/platforms/platinfo.py
```diff
@@ -108,6 +115,8 @@ def add_target_src(self, target, rta_calib_res_dir, src='target', **kwargs):
         if target.is_module_available('sched'):
             info['cpu-capacities'] = target.sched.get_capacities(default=1024)
 
+        info['kernel']['symbols-address'] = self._read_kallsyms(target)
```
valschneider: Should this be a deferred value?

lisa/trace.py
```diff
@@ -1512,22 +1547,68 @@ class FtraceCollector(devlib.FtraceCollector, Loggable, Configurable):
 
     CONF_CLASS = FtraceConf
 
-    def __init__(self, target, events=None, functions=None, buffer_size=10240, autoreport=False, **kwargs):
+    def __init__(self, target, events=None, functions=None, buffer_size=10240, autoreport=False, trace_clock=None, saved_cmdlines_nr=4096, tracer=None, **kwargs):
         events = events or []
         functions = functions or []
-        kwargs.update(dict(
+        trace_clock = trace_clock or 'global'
```
valschneider: Do we have an idea of the overhead/limitations of using the global clock?

lisa/platforms/platinfo.py
douglas-raillard-arm: there isn't really a good reason for a deferred value here, that's fast to get. Deferred values have a number of drawbacks, and that piece of info will be necessary if we want to plot anything offline.

lisa/trace.py
douglas-raillard-arm: it might be slower. That said on ARM arch, the system clock is a architecturally a system-wide clock so I expect that using `global` will make no difference compared to `local`. On x86, that might change. Trappy now expects monotonic timestamps anyway, so using `local` clock sounds like a pretty bad idea.

PR #1172, opened 2019-11-02
pyupgrade cleanup

douglas-raillard-arm:


doc/conf.py
```diff
@@ -46,16 +45,19 @@
     try:
         m = obj.__module__
         obj.__module__ = 'unittest' if m == 'unittest.case' else m
-    except Exception: pass
+    except Exception:
+        pass
 
 # This is a hack to prevent :ivar: docs from attempting to create a reference
 # Credit goes to https://stackoverflow.com/a/41184353/5096023
+
```
valschneider: Is this expected/intended?

lisa/analysis/__init__.py
```diff
@@ -21,11 +21,14 @@
 # Import all the submodules before they are asked for by user code, since
 # we need to create the *Analysis classes in order for them to be
 # registered against TraceAnalysisBase
+
```
valschneider: Ditto on the double newlines

doc/conf.py
douglas-raillard-arm: Seems like it treats a comment as a declaration. That's not really intended, but I'm not sure if that's really harmful either. That will make happy these kind of tools, so if we agree to proceed with that, diffs will be restricted to modified code in the future.

lisa/analysis/tasks.py
```diff
@@ -46,37 +47,38 @@ def __or__(self, other):
...
     """
     Represents the task state as visible in sched_switch
 
     * Values are extracted from include/linux/sched.h
     * Chars are extracted from fs/proc/array.c:get_task_state()
     """
-    #pylint-suppress: bad-whitespace
-    TASK_RUNNING         = 0x0000, "R", "Running"
-    TASK_INTERRUPTIBLE   = 0x0001, "S", "Sleeping"
+    # pylint-suppress: bad-whitespace
```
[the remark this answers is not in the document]
douglas-raillard-arm: indeed, it would be better to be able to preserve that, but I don't think we can disable that linter without loosing a lot of actually useful cleanup :/

PR #1174, opened 2019-11-04
lisa.conf: Fix DeferredValue docstring

douglas-raillard-arm:
Replace half-written sentence by an actual description.

### DIFF
--- a/lisa/conf.py
+++ b/lisa/conf.py
@@ -35,10 +35,16 @@ from ruamel.yaml.comments import CommentedMap
 
 class DeferredValue:
     """
-    Wrapper similar to functools.partial.
+    Wrapper similar to :func:`functools.partial` allowing to defer computation
+    of the value until the key is actually used.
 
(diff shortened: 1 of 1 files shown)


lisa/conf.py
```diff
@@ -35,10 +35,13 @@
...
-    Wrapper similar to functools.partial.
+    Wrapper similar to :func:`functools.partial` allowing to defer computation
+    of the value until the key is actually used.
 
-    I is a nown class to detect and to derive from to create different
-    categories of deferred values.
+    Once computed, the deferred value is replaced by the value that was
+    computed. This is useful for values that are very costly to compute, but
+    should be used with care as it means it will usually not be available in
+    the offline :class:`lisa.platforms.platinfo.PlatformInfo` instances.
```
valschneider: LGTM, might just want to append some extra justification at the end, e.g.:
```
(and thus not universally available for offline analysis).
```

PR #1183, opened 2019-11-06
lisa.analysis.tasks: Get rid of non-rolling windowing

valschneider:
Depends on https://github.com/ARM-software/lisa/pull/1167

### DIFF
--- a/lisa/analysis/tasks.py
+++ b/lisa/analysis/tasks.py
@@ -23,7 +23,7 @@ import pandas as pd
 
 from lisa.analysis.base import TraceAnalysisBase
 from lisa.utils import memoized
-from lisa.datautils import df_filter_task_ids
+from lisa.datautils import df_filter_task_ids, series_rolling_apply
 from lisa.trace import requires_events
 
 class StateInt(int):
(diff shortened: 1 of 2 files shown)


lisa/analysis/tasks.py
```diff
@@ -671,10 +656,10 @@ def plot_tasks_wakeups(self, target_cpus=None, time_delta=0.01, axis=None,
         if target_cpus:
             df = df[df.target_cpu.isin(target_cpus)]
 
-        df = self._df_discretize_series(df["target_cpu"], time_delta, "Wakeup count")
-
-        df.plot(ax=axis, legend=False)
+        series = series_rolling_apply(df["target_cpu"], lambda x: x.count(),
```
douglas-raillard-arm: `lambda x: x.count()` could be updated with `series_mean` so that the values that the magnitude of the output does not depend on `time_delta`, and is actually in unit of `x per seconds`.

lisa/analysis/tasks.py
```diff
@@ -726,13 +710,13 @@ def plot_tasks_forks(self, target_cpus=None, time_delta=0.01, axis=None, local_f
         if target_cpus:
             df = df[df.target_cpu.isin(target_cpus)]
 
-        df = self._df_discretize_series(df["target_cpu"], time_delta, "Forks count")
-        df.plot(ax=axis, legend=False)
+        series = series_rolling_apply(df["target_cpu"], lambda x: x.count(),
```
douglas-raillard-arm: ditto

lisa/analysis/tasks.py
valschneider: I think you're right, however `series_mean` gives me funny results (because it's average the contents of the series rather than the count). I think what I want here is

```
lambda x: x.count() / (x.max() - x.min())
```

douglas-raillard-arm: right, `x` must be a series of bool in order to use `series_mean()`, in which case the result the same as your formula.

lisa/analysis/tasks.py
```diff
@@ -671,10 +656,10 @@ def plot_tasks_wakeups(self, target_cpus=None, time_delta=0.01, axis=None,
         if target_cpus:
             df = df[df.target_cpu.isin(target_cpus)]
 
-        df = self._df_discretize_series(df["target_cpu"], time_delta, "Wakeup count")
-
-        df.plot(ax=axis, legend=False)
+        series = series_rolling_apply(df["target_cpu"], lambda x: x.count() / (x.max() - x.min()),
```
douglas-raillard-arm: s/x.max() - x.min()/x.index[-1] - x.index[0]/

Since `x` could contain only one item, you'll need to handle division by 0. A dropna() before plotting should be enough to avoid having blanks segments in the graph.

PR #1185, opened 2019-11-06
[RFC] Use DataFrame.query() in df_filter_tasks()

douglas-raillard-arm:
Allow using the query language of pandas for filtering dataframes. Since it can be slower for small input and lead to MemoryError or ValueError for very large queries, abstract over it to allow both regular indexing and query building.

TODO: If that is deemed to be an interesting feature, the query building blocks might be turned into classes with overloaded operators, so that it's possible to write code once and then choose the method of evaluation separately.


douglas-raillard-arm:
Closing as DataFrame.query() is unreliable due to some limitations in numexpr parser.

PR #1188, opened 2019-11-07
Helper to get idle CPUs at a given timestamp

valschneider:
https://github.com/ARM-software/lisa/pull/1131 back from the dead.

### DIFF
--- a/lisa/analysis/idle.py
+++ b/lisa/analysis/idle.py
@@ -51,22 +51,22 @@ class IdleAnalysis(TraceAnalysisBase):
         :param cpu: CPU ID
         :type cpu: int
 
-        :returns: A :class:`pandas.Series` that equals 1 at timestamps where the
-          CPU is reported to be non-idle, 0 otherwise
+        :returns: A :class:`pandas.Series` that equals True at timestamps where the
+          CPU is reported to be non-idle, False otherwise
         """
         idle_df = self.trace.df_events('cpu_idle')
         cpu_df = idle_df[idle_df.cpu_id == cpu]
 
         cpu_active = cpu_df.state.apply(
-            lambda s: 1 if s == -1 else 0
+            lambda s: True if s == -1 else False
         )
 
         start_time = self.trace.start
 
         if cpu_active.empty:
-            cpu_active = pd.Series([0], index=[start_time])
+            cpu_active = pd.Series([False], index=[start_time])
         elif cpu_active.index[0] != start_time:
-            entry_0 = pd.Series(cpu_active.iloc[0] ^ 1, index=[start_time])
+            entry_0 = pd.Series(not cpu_active.iloc[0], index=[start_time])
             cpu_active = pd.concat([entry_0, cpu_active])
 
         # Fix sequences of wakeup/sleep events reported with the same index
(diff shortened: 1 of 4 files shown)


lisa/analysis/idle.py
```diff
@@ -72,6 +73,50 @@ def signal_cpu_active(self, cpu):
...
+        :param cpu: CPU ID
+        :type cpu: int
+
+        :returns: A :class:`pandas.Series` that equals True at timestamps where the
+          CPU is reported to be non-idle, False otherwise
+        """
+        # Find the earliest time at which the scheduler placed a task on
+        # this CPU's rq. The scheduler won't consider this CPU idle at this
+        # time. A proper rq depth signal would make this more accurate.
+        signal_active = self.signal_cpu_active(cpu).copy()
```
douglas-raillard-arm: is copy() necessary ? We already do a dropna(inplace=False)

lisa/analysis/idle.py
```diff
@@ -72,6 +73,50 @@ def signal_cpu_active(self, cpu):
...
+        # XXX: bitwise 'or' has surprising results:
+        # s1 = pd.Series(index=[1.25], data=[True])
+        # s2 = pd.Series(index=[1.05], data=[True])
+        # print(s1 | s2)
+        # 1.05    False
+        # 1.25     True
+        def orify(a, b):
+            return a or b
+
+        signal_active = signal_active.combine(sched_active, orify, fill_value=False)
```
douglas-raillard-arm: might be easier to add all the columns into a dataframe, and then just return a final column ? That ensures indexes are aligned and is generally easier to printf-debug

valschneider: I don't see how to make this simpler using dataframes since I just want to 'or' two signals. Is there something specific you were thinking of?

douglas-raillard-arm: Actually adding the series as a column would give a column of mostly NaN so it's not what we want.

If I understand correctly the current code, it's equivalent to a reindex. The use of `X | 1` is a way of getting `X` on indices of the 1's series and NaN the rest of the time, and then we drop the NaN.

```
df_states = self.trace.analysis.tasks.df_tasks_states()
df_states = df_states[df_states.target_cpu == cpu] 
cpu_active = self.signal_cpu_active(cpu)

active = cpu_active.reindex(df_states.index, method='ffill')
return series_deduplicate(active, consecutives=True, keep='first')
```

douglas-raillard-arm: EDIT: I misunderstood the actual meaning of `fill_value`, i.e. it's replacing NaN in the inputs to the function, not a blanket replace in the output when any of the input is NaN.

lisa/analysis/idle.py
```diff
@@ -72,6 +73,50 @@ def signal_cpu_active(self, cpu):
         # Fix sequences of wakeup/sleep events reported with the same index
         return handle_duplicate_index(cpu_active)
 
+    @memoized
```
douglas-raillard-arm: not sure if we want to memoize things that return series or dataframe as it leads to high memory usage, even if the user of the trace will not look at that df anymore.
https://github.com/ARM-software/lisa/pull/1235 will bring a better way of doing that, avoiding excessive memory usage and swapping the df to disk if that's faster than recomputing it.

douglas-raillard-arm:

PR #1195, opened 2019-11-19
Update CPU capacities while running rt-app

DouglasRaillard:
Allow rt-app to run with actual CPU invariant utilization.

TODO: check the CPUMigration test debug plot to make sure there is no change in util value when migrating the same task.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -528,7 +528,7 @@ class RTA(Workload):
             logger.warning('The calibration values are not inversely proportional to the CPU capacities, the duty cycles will be up to {:.2f}% off on some CPUs: {}'.format(dispersion_pct, capa_factors_pct))
 
         if dispersion_pct > 20:
-            raise CalibrationError('The calibration values are not inversely proportional to the CPU capacities. Either rt-app calibration failed, or the rt-app busy loops has a very different instruction mix compared to the workload used to establish the CPU capacities: {}'.format(capa_factors_pct))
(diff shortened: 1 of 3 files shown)


lisa/wlgen/rta.py
```diff
@@ -463,7 +500,7 @@ def _calibrate(cls, target, res_dir):
             logger.warning('The calibration values are not inversely proportional to the CPU capacities, the duty cycles will be up to {:.2f}% off on some CPUs: {}'.format(dispersion_pct, capa_factors_pct))
 
         if dispersion_pct > 20:
-            raise CalibrationError('The calibration values are not inversely proportional to the CPU capacities. Either rt-app calibration failed, or the rt-app busy loops has a very different instruction mix compared to the workload used to establish the CPU capacities: {}'.format(capa_factors_pct))
+            logger.waring('The calibration values are not inversely proportional to the CPU capacities. Either rt-app calibration failed, or the rt-app busy loops has a very different instruction mix compared to the workload used to establish the CPU capacities: {}'.format(capa_factors_pct))
```
valschneider:
```suggestion
            logger.warning('The calibration values are not inversely proportional to the CPU capacities. Either rt-app calibration failed, or the rt-app busy loops has a very different instruction mix compared to the workload used to establish the CPU capacities: {}'.format(capa_factors_pct))
```

That, or keep it as an exception?

douglas-raillard-arm: I've removed the exception since it's not a problem at all as long as we update the CPU capacities. For example on Pixel 3, the exception will always be raised unless you modify the capacities in the device tree. Since it will be done at runtime with the same effects, there is no reason to prevent running tests.

PR #1199, opened 2019-11-20
Use mainline Linux load tracking events names

douglas-raillard-arm:


douglas-raillard-arm:
Actually, the kernel does not have a `sched_pelt_se` event, it only has a `sched_pelt_se` tracepoint. That said, it's probably a better idea to give the same name to the trace point and associated event.

TODO: update tools/kmodules/ to emit the `sched_pelt_se` name rather than `sched_load_se`

lisa/analysis/load_tracking.py
```diff
@@ -61,7 +71,7 @@ def _columns_to_drop(cls, event):
         """
         The extra columns not shared between trace event versions
         """
-        if event in ['sched_load_cfs_rq', 'sched_load_se']:
+        if event in ['sched_load_cfs_rq', 'sched_load_se', 'sched_pelt_se']:
```
qais-yousef: Can't you use the new _SCHED_PELT_SE_NAMES instead here?

lisa/analysis/load_tracking.py
```diff
@@ -71,7 +81,7 @@ def _columns_to_drop(cls, event):
         """
         The extra columns not shared between trace event versions
         """
-        if event in ['sched_load_cfs_rq', 'sched_load_se', 'sched_pelt_se']:
+        if event in [*cls._SCHED_PELT_CFS_NAMES, 'sched_load_se', 'sched_pelt_se']:
```
qais-yousef: Same here. Why not juse _SCHED_PELT_CFS_NAMES?

lisa/analysis/load_tracking.py
douglas-raillard-arm: No, `sched_load_cfs_rq` is part of `_SCHED_PELT_CFS_NAMES`, so it seems to really depend on the specific event. If you have examples of them, we can sort out what fields they actually have.

lisa/analysis/load_tracking.py
douglas-raillard-arm: `_SCHED_PELT_CFS_NAMES` does not contain `sched_load_se`/`sched_pelt_se`

PR #1242, opened 2019-12-04
Added new tests for NUMA related behavior of scheduler

PeterPuhov-Linaro:
LISA can be used to test scheduler behavior for task placement on hardware with complex NUMA
topology. I already submitted Pull Request to devlib for NUMA introspection:
https://github.com/ARM-software/devlib/pull/441

This Pull Request has a couple of very basic tests to ensure that tasks with relatively low
utilization do not migrate to different cores.
I am planning to add NUMA related tests in my next Pull Request after some NUMA related changes in rt-app.

### DIFF
--- a/lisa/tests/staging/numa_behaviour.py
+++ b/lisa/tests/staging/numa_behaviour.py
@@ -1,0 +1,116 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2019, Linaro and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import os.path
+
+import pandas as pd
+
+from lisa.wlgen.rta import Periodic
+from lisa.tests.base import ResultBundle, RTATestBundle, TestMetric
+from lisa.utils import ArtifactPath
+from lisa.trace import requires_events, FtraceCollector
+from lisa.target import Target
+from lisa.datautils import df_filter_task_ids, df_deduplicate
+
+class NUMABehaviour(RTATestBundle):
+    """
+    Abstract class for NUMA related scheduler testing.
+    """
+    @classmethod
+    def check_from_target(cls, target):
+        if target.number_of_nodes < 2:
+            raise CannotCreateError(
+                "Target doesn't have at least two NUMA nodes")
+
+    @requires_events('sched_switch')
+    def _get_task_cpu_df(self, task_id):
+        """
+        Get a DataFrame for task mirgrations
+
+        Use the sched_switch trace event to find task mirgation from one CPU to another.
+
+        :returns: A Pandas DataFrame for the task, showing the
+                  CPU's that the task was migrated to
+        """        
+        df = self.trace.analysis.tasks.df_task_states(task_id)
+        cpu_df = df_deduplicate(df, cols=['cpu'], keep='first', consecutives=True)
+                
+        return cpu_df
+
+    @_get_task_cpu_df.used_events
+    def test_task_remains(self) -> ResultBundle:
+        """
+        Test that task remaines on the same core 
+        """
+        test_passed = True
+        metrics = {}
+        
+        for task_id in self.rtapp_task_ids:
+            cpu_df = self._get_task_cpu_df(task_id)
+            core_migrations = len(cpu_df.index)
+            metrics[task_id] = TestMetric(core_migrations)
+
+            # Ideally, task with 50% utilization
+            # should stay on the same core
+            if  core_migrations > 1:
+                test_passed = False
+
+        res = ResultBundle.from_bool(test_passed)
+        res.add_metric("Migrations", metrics)
+        
+        return res
+
+class NUMASmallTaskPlacement(NUMABehaviour):
+    """
+    A single task with 50% utilization
+    """
+
+    task_prefix = "tsk"
+
+    @classmethod
+    def get_rtapp_profile(cls, plat_info):
+        rtapp_profile = {}
+        rtapp_profile[cls.task_prefix] = Periodic(
+            duty_cycle_pct=50,
+            duration_s=30,
+            period_ms=cls.TASK_PERIOD_MS
+        )
+
+        return rtapp_profile
+    
+class NUMAMultipleTasksPlacement(NUMABehaviour):
+    """
+    Multiple tasks with 50% utilization
+    """
+    task_prefix = "tsk"
+    
+    @classmethod
+    def get_rtapp_profile(cls, plat_info):
+        # Four CPU's is enough to demonstrate task migration problem
+        cpu_count = min(4, plat_info["cpus-count"])        
+
+        rtapp_profile = {}
+        for cpu in range(cpu_count):            
+            rtapp_profile["{}{}".format(cls.task_prefix, cpu)] = Periodic(
+                duty_cycle_pct=50,
+                duration_s=30,
+                period_ms=cls.TASK_PERIOD_MS
+            )
+
+        return rtapp_profile    
+# vim :set tabstop=4 shiftwidth=4 textwidth=80 expandtab


lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+import os.path
+
+import pandas as pd
+
+from lisa.wlgen.rta import Periodic
+from lisa.tests.base import ResultBundle, RTATestBundle
+from lisa.utils import ArtifactPath
+from lisa.trace import requires_events
+from lisa.target import Target
+from lisa.trace import FtraceCollector
```
douglas-raillard-arm: Can be merged with earlier import:
```
from lisa.trace import FtraceCollector, requires_events
```

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+class NUMABehaviour(RTATestBundle):
+    """
+    Abstract class for NUMA related scheduler testing.
+
+    This class provides :meth:`test_task_placement` to validate the basic
+    behaviour of NUMA related scheduling decisions. 
+    """
+
+    @classmethod
+    def _from_target(cls, target: Target, *, res_dir: ArtifactPath = None, ftrace_coll: FtraceCollector = None) -> 'NUMABehaviour':
```
douglas-raillard-arm: AFAICS this definition of `_from_target` is redundant with what is inherited from `RTATestBundle` so removing should still work (let me know if it does not work so I can fix `RTATestBundle`).

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Use the sched_switch trace event to find task mirgation from one CPU to another.
+
+        :returns: A Pandas DataFrame for the task, showing the
+                  new CPU that the task was migrated to
+        """        
+        df = self.trace.df_events('sched_switch')[['next_pid', 'next_comm', '__cpu']]
+        task_id = self.trace.get_task_id(task, update=False)
+        _df = df_filter_task_ids(df, [task_id], pid_col='next_pid', comm_col='next_comm')
+        cpu_df = _df.drop_duplicates(subset='__cpu', keep='first', inplace=False)
```
douglas-raillard-arm: Might want to consider using the helper unifying deduplication recipes:
```
from lisa.datautils import df_deduplicate
cpu_df = df_deduplicate(cpu_df, cols=['__cpu'], keep='first', consecutives=False)
```

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+                
+        return cpu_df
+
+    @requires_events('sched_switch')
```
douglas-raillard-arm: The function actually getting the dataframes need the `requires_events` and other functions built on top of that can refer the first function. This would give:
```
     @requires_events('sched_switch') 
     def _get_task_cpu_df(self, task): 
          ....

    @_get_task_cpu_df.used_events
    def test_task_placement(self) -> ResultBundle: 
          ....
```
This way we only have to change the decoration of the class using the events directly if it starts using a new event. All the other ones built on top will follow the change automatically.

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_task_placement(self) -> ResultBundle:
+        """
+        Test that task remained on the same core 
+
+        """
+        test_passed = True
+        failed_tasks = []
+        
+        for task in self.rtapp_tasks:
```
douglas-raillard-arm: If you use `self.rtapp_task_ids` instead you can remove that line from `_get_task_cpu_df()` since `task` will already be a `TaskID`:
```
task_id = self.trace.get_task_id(task, update=False)
```

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+            if  core_migrations > 1:
+                test_passed = False
+                failed_tasks.append("{} core_migrations {}".format(task, core_migrations))
+
+        res = ResultBundle.from_bool(test_passed)
+        res.add_metric("Failed tasks", failed_tasks)
+        
+        return res
+
+class NUMA_SmallTaskPlacement(NUMABehaviour):
```
douglas-raillard-arm: nit: s/NUMA_SmallTaskPlacement/NUMASmallTaskPlacement/ to follow PEP8:
https://www.python.org/dev/peps/pep-0008/#class-names

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+        rtapp_profile = {}
+        rtapp_profile[cls.task_prefix] = Periodic(
+            duty_cycle_pct=50,
+            duration_s=30,
+            period_ms=cls.TASK_PERIOD_MS
+        )
+
+        return rtapp_profile
+    
+class NUMA_MultipleTasksPlacement(NUMABehaviour):
```
douglas-raillard-arm: s/NUMA_MultipleTasksPlacement/NUMAMultipleTasksPlacement/

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+    task_prefix = "ntask"
+    
+    @classmethod
+    def get_rtapp_profile(cls, plat_info):
+        # Four CPU's is enough to demonstrate task migration problem
+        cpu_count = min(4, plat_info["cpus-count"])        
+
+        rtapp_profile = {}
+        for cpu in range(cpu_count):            
+            rtapp_profile["{}_{}".format(cls.task_prefix, cpu)] = Periodic(
```
douglas-raillard-arm: FYI the task name is currently limited to 7 chars (kernel saved `comm` size minus some suffix added by rtapp itself). If you revisit the number of CPUs involved in the test and if it becomes > 10, you will end up with names like `ntask_10` which won't fit anymore. What we did in other tests is drop the `_` to avoid such issues.

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+        
+        for task in self.rtapp_tasks:
+            cpu_df = self._get_task_cpu_df(task)
+            core_migrations = len(cpu_df.index)
+            # Ideally, task with low utilization
+            # should stay on the same core and
+            # do not cross NUMA boundaries
+            if  core_migrations > 1:
+                test_passed = False
+                failed_tasks.append("{} core_migrations {}".format(task, core_migrations))
```
valschneider: This only adds metrics when there is a failure; it's good to unconditionally add metrics (to reassure ourselves that the test did the right thing when it says it passed).

Maybe you could something like
```
metrics = {}

...

metrics[task] =TestMetric(core_migrations)
...
res.add_metric("Migrations", metrics)
```

PeterPuhov-Linaro:
@douglas-raillard-arm and @valschneider , 
Thank you very much for your review!
I will make all recommended changes and
will update Pull Request shortly.

lisa/tests/scheduler/numa_behaviour.py
PeterPuhov-Linaro: It think there is a bug (typo) in:
def _data_deduplicate(data, keep, consecutives, cols, all_col):
     ...
     if not all_cols: 
https://github.com/ARM-software/lisa/blob/master/lisa/datautils.py#L839
Should I fix it in this PR or in separate one ?

douglas-raillard-arm: Typo indeed, I've fixed it in master.

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+        Use the sched_switch trace event to find task mirgation from one CPU to another.
+
+        :returns: A Pandas DataFrame for the task, showing the
+                  new CPU that the task was migrated to
+        """        
+        df = self.trace.df_events('sched_switch')[['next_pid', 'next_comm', '__cpu']]
+        task_id = self.trace.get_task_id(task, update=False)
+        _df = df_filter_task_ids(df, [task_id], pid_col='next_pid', comm_col='next_comm')
+        cpu_df = _df.drop_duplicates(subset='__cpu', keep='first', inplace=False)
+
```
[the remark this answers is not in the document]
douglas-raillard-arm: If we are interested in getting the events where the CPU is different from the previous one, we want to keep the last occurrence of  consecutive duplicates rather than the first, i.e. `shift=-1`.

This becomes clearer as:
```
# Only get occurrences where the CPU is different from the previous event
cpu_df = df_deduplicate(cpu_df, cols=['cpu'], consecutives=True, keep='last')
# Ignore the first event since we don't know if it was different from the previous one or not
cpu_df = cpu_df[1:]
```

douglas-raillard-arm:
@PeterPuhov-Linaro FYI that was merged recently:
https://github.com/ARM-software/lisa/pull/1239
(you can check the generated rtapp JSON configuration in the tests artifacts to see what are the effects for your profile).

This ensures that by the time the test workload defined by the rtapp profile starts, all tasks are placed on the CPU they are supposed to according to the `duty_cycle_pct` that was set, and their utilisation signals have converged to a stable value. This should make all tests robust against random-ish initial placement by default.

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+        test_passed = True
+        failed_tasks = []
+        
+        for task in self.rtapp_tasks:
+            cpu_df = self._get_task_cpu_df(task)
+            core_migrations = len(cpu_df.index)
+            # Ideally, task with low utilization
+            # should stay on the same core and
+            # do not cross NUMA boundaries
+            if  core_migrations > 1:
```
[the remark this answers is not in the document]
PeterPuhov-Linaro: I will change comment.
I already added 
number_of_nodes() and list_nodes_cpus()
to target.py from devlib and this PR already merged
I am planning to rename this particular test to test_task_remains()
as you suggested.
In my next PR I am planning to add test for NUMA boundaries.
Does it sounds reasonable?

PeterPuhov-Linaro:
Thank you,
I noticed that tasks have additional 2 sec. phase now.
What would be the proper way to trim traces for last 10 seconds
(Sorry, I am still ramping up)
I simply want to make sure that task with 50% utilization does not migrate between cores.
I am testing large systems (128 cores) and it should not be a reason for migration.

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,132 @@
...
+                test_passed = False
+                failed_tasks.append("{} core_migrations {}".format(task, core_migrations))
+
+        res = ResultBundle.from_bool(test_passed)
+        res.add_metric("Failed tasks", failed_tasks)
+        
+        return res
+
```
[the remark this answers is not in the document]
PeterPuhov-Linaro: devlib/target.py now has number_of_nodes() function
https://github.com/ARM-software/devlib/blob/master/devlib/target.py#L169
I am planning to use it

lisa/tests/scheduler/numa_behaviour.py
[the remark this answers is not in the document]
valschneider: > I already added number_of_nodes() and list_nodes_cpus()

Perfect, then we can also use that for  `check_from_target()`

> In my next PR I am planning to add test for NUMA boundaries. Does it sounds reasonable?

Sure does, but then I have to ask: how does this current test fit with NUMA topologies? I expect some level of leniency wrt migrating tasks within a same NUMA node (or more generally within the same LLC). Even with a quiet userspace I wouldn't be surprised to see the tasks move around a bit, so `core_migrations` wouldn't always be 1. Or is that actually the point - we want the tasks to get a CPU each and ensure they stay there, and if they don't well go and try to fix the scheduler?

douglas-raillard-arm:
>  What would be the proper way to trim traces for last 10 seconds

You actually don't need to do anything, since `RTATestBundle.trace` will already be trimed to what `RTATestBundle.trace_window()` [1] returns. By default, it goes from the beginning of the first phase in the profile to the end of the last phase. This removes the rt-app main thread setup part, and everything that LISA does before, and also everything that comes after rt-app is finished.

[1] https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/kernel_tests.html#lisa.tests.base.RTATestBundle.trace_window

Makes me realize that `RTATestBundle.trace` does not mention automatic trimming, I'll add that soon.

> (Sorry, I am still ramping up)

No worries, that's (at least) half of what reviews are for

lisa/tests/scheduler/numa_behaviour.py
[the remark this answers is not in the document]
PeterPuhov-Linaro: These tests will help to test/debug load balancing part of kernel scheduler
currently developed in KWG Linaro.
https://git.linaro.org/people/vincent.guittot/kernel.git

Tasks with medium utilization (less than 75%) should not migrate between cores if we have small number of tasks and large number of cores (in my case 128)
I don't see a reason for 4 tasks with 50% utilization to migrate between cores in this case.
But they do. And this test will help other folks to reproduce it. I am also planning to propose 
fix for kernel scheduler.
In addition, I am planning to add test with 100% utilization tasks. They will migrate, but migration should be within NUMA boundaries.
I am planning to add more tests in this area, this just the first once.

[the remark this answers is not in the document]
valschneider: Okay, makes sense, thanks for the explanation.

lisa/tests/scheduler/numa_behaviour.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Get a DataFrame for task mirgrations
+
+        Use the sched_switch trace event to find task mirgation from one CPU to another.
+
+        :returns: A Pandas DataFrame for the task, showing the
+                  CPU's that the task was migrated to
+        """        
+        df = self.trace.df_events('sched_switch')[['next_pid', 'next_comm', '__cpu']]
+        cpu_df = df_filter_task_ids(df, [task_id], pid_col='next_pid', comm_col='next_comm')
```
valschneider: IMO those first two lines should be reduced to:
```
df = self.trace.analysis.tasks.df_task_states(task_id)
```

This keeps the logic of which columns to get / which column to filter on in the above helper.
You can then call `df_deduplicate` with `cols=['cpu']`.

PeterPuhov-Linaro: Sure, will do

PeterPuhov-Linaro:
> LGTM once rebased to cleanup the PR's history and the file moved to `lisa/tests/staging/numa_behaviour.py` 
I am planning to move the test into staging directory and submit a new PR.
Is it reasonable?

valschneider:
> > LGTM once rebased to cleanup the PR's history and the file moved to `lisa/tests/staging/numa_behaviour.py`

> I am planning to move the test into staging directory and submit a new PR.
> Is it reasonable?

The usual github PR flow is to force-push to your branch, so you could do that here. This has the added benefit of keeping all of the review comments in one place.

douglas-raillard-arm:
Yeah, github is annoying because force-pushing nuke old versions of the commits, but opening multiple PR and spreading comments around are an even greater annoyance so force-pushing is the sanest thing to do. Maybe one day github will match Gerrit or gitlab and give inter-versions diff of the series ...

lisa/tests/staging/numa_behaviour.py
```diff
@@ -0,0 +1,120 @@
...
+from lisa.utils import ArtifactPath
+from lisa.trace import requires_events, FtraceCollector
+from lisa.target import Target
+from lisa.datautils import df_filter_task_ids, df_deduplicate
+
+class NUMABehaviour(RTATestBundle):
+    """
+    Abstract class for NUMA related scheduler testing.
+
+    This class provides :meth:`test_task_placement` to validate the basic
```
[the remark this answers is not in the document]
PeterPuhov-Linaro: I removed this part of docstring as you suggested

valschneider:
FYI I pulled the file locally to fix a small issue @douglas-raillard-arm noticed, and my pylint caught some other things that I fixed. See the following commits:

fa6fb9039 lisa.tests.staging.numa_behaviour: Clean/fix imports
4378b37d8 lisa.tests.staging.numa_behaviour: Docstring & whitespace cleanup
3a17d868e lisa.tests.staging.numa_behaviour: Fix analysis events decorator

PR #1243, opened 2019-12-04
RTATestBundle: crop the trace around rtapp execution

douglas-raillard-arm:
TODO: figure out if loosing the value of some signals is a real problem. Among them, the devlib_cpu_frequency should be lost.

Credits to @vdonnefort  for the idea :)

Depends on https://github.com/ARM-software/lisa/pull/1246

### DIFF
--- a/lisa/assets/__init__.py
+++ b/lisa/assets/__init__.py
@@ -17,6 +17,21 @@
 #
 import os
 import stat
+import platform
+
+def _get_abi():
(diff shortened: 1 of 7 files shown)


douglas-raillard-arm:
The issue is actually better solved by:
https://github.com/ARM-software/lisa/pull/1248

This will avoid parsing the trace one more time, and the potential issues related to missing frequency events.

PR #1250, opened 2019-12-05
Cleanup TargetScript

douglas-raillard-arm:


lisa/tests/hotplug/torture.py
```diff
@@ -182,14 +182,16 @@ def _from_target(cls, target: Target, *, res_dir: ArtifactPath = None, seed=None
...
-                script.wait()
-
-                target_alive = True
-                target.hotplug.online_all()
-            except TargetNotRespondingError:
-                target_alive = False
+        try:
+            with script.background(as_root=True) as bg:
+                while bg.poll() is None:
+                    if not script.target.check_responsive():
```
valschneider: IIRC this need the `explode=True` just like in the now-departed `TargetScript.wait()`, otherwise this doesn't timeout.

douglas-raillard-arm: I'm using the return value here rather than relying on exception raising so `explode=True` is not needed.

valschneider: Right you are

PR #1251, opened 2019-12-06
Fix misfit test issues

valschneider:
I've been hitting some nasty issues with the misfit tests since the introduction of the rt-app buffer phase. Some (most?) of it is due to shady things I wrote in that test, so I only really have myself to blame...

I tested this by keeping tasks pinned in the last phase (IOW forcing a failure scenario), and the detected big CPU idle times were exactly what I expected, so AFAICT this works fine.

### DIFF
--- a/lisa/tests/scheduler/misfit.py
+++ b/lisa/tests/scheduler/misfit.py
@@ -39,17 +39,9 @@ class MisfitMigrationBase(RTATestBundle):
     @classmethod
     def _has_asym_cpucapacity(cls, target):
         """
-        :returns: Whether the target has SD_ASYM_CPUCAPACITY set on any of its sd
+        :returns: Whether the target has asymmetric CPU capacities
         """
-        # Just try to find at least one instance of that flag
-        sd_info = target.sched.get_sd_info()
-
-        for cpu, domain_node in sd_info.cpus.items():
-            for domain in domain_node.domains.values():
-                if SchedDomainFlag.SD_ASYM_CPUCAPACITY in domain.flags:
-                    return True
-
-        return False
+        return len(set(target.plat_info["cpu-capacities"].values())) > 1
 
     @classmethod
     def _get_max_lb_interval(cls, plat_info):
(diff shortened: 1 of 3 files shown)


lisa/trace.py
```diff
@@ -303,7 +303,7 @@ def __init__(self,
                  trace_path,
                  plat_info=None,
                  events=None,
-                 normalize_time=True,
+                 normalize_time=False,
```
douglas-raillard-arm: Assuming we don't have traces with more than ~10 days of uptime that should be alright. After that, the nanosecond information will become noise.

valschneider: The thing I'm trying to "fix" is somewhat orthogonal to this; the thing is I can easily re-load a `TestBundle` and print some of its internal data to figure out what happened. The issue is that the internal `Trace` is loaded with the default `normalize_time` (thus `True`) and that isn't easily hijackable.

So I would print some df, get normalized timestamps, and wouldn't be able to compare this with kernelshark. Other than the long uptime issue, AFAIR the main reason for normalization was for "pretty plots". In my experience this has been more of an annoyance (my bread and butter being kernelshark), so I'd like to change that.

douglas-raillard-arm: We can change the default if we really don't expect anything more than 10 days uptime (and we could add a warning for non normalized time in that case), but it's pretty easy to get a trace with the parameter you want:
```
trace = test_bundle.get_trace(events=['foo', 'bar'], normalize_time=False)
```
`test_bundle.trace` is meant to be suited to test methods more than anything else, conflicting requirements can either be met with `get_trace()` or by introducing a new property (maybe a more friendly `test_bundle.debug_trace` that would parse all events, skip trace_window and provide non-normalized timestamps ?)

lisa/tests/scheduler/misfit.py
```diff
@@ -128,29 +129,20 @@ def duration(self):
...
+        phase_df.set_index("Time", inplace=True)
+        phase_df.sort_index(inplace=True)
 
-            return sdf[(sdf.next_comm == name) & (sdf["__cpu"] == task_cpu)].index[0]
-
-        return max(
-            get_start_time(sdf, task, profile)
-            for task, profile in self.rtapp_profile.items()
-        )
+        return phase_df[phase_df.phase == 1].index[-1]
```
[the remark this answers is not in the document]
douglas-raillard-arm: EDIT: there sadly is no satisfactory solution for now, we would need named phases, which are not there yet

PR #1262, opened 2019-12-19
tests: staging: sched_android: Ensure frequencies are sorted

qperret:
Cpufreq's scaling_available_frequencies sysfs file does not guarantee
any particular order for the frequencies. As such, on devices where
they're not sorted in increasing order, the schedtune frequency test
will fail to compute the right target frequency, leading to test
failures.

Fix this by sorting the frequencies when necessary in sched_android.py.

Signed-off-by: Quentin Perret <qperret@google.com>

### DIFF
--- a/lisa/tests/staging/sched_android.py
+++ b/lisa/tests/staging/sched_android.py
@@ -190,7 +190,7 @@ class SchedTuneFreqItem(SchedTuneItemBase):
         # into a real OPP
         boost = self.boost
         target_freq = min(max_freq, max_freq * boost / 80)
-        target_freq = list(filter(lambda f: f >= target_freq, freqs))[0]
+        target_freq = sorted(list(filter(lambda f: f >= target_freq, freqs)))[0]
 
         # Get the real average frequency
         avg_freq = self.trace.analysis.frequency.get_average_cpu_frequency(cpu)


lisa/tests/staging/sched_android.py
```diff
@@ -190,7 +190,7 @@ def test_stune_frequency(self, freq_margin_pct=10) -> ResultBundle:
         # into a real OPP
         boost = self.boost
         target_freq = min(max_freq, max_freq * boost / 80)
-        target_freq = list(filter(lambda f: f >= target_freq, freqs))[0]
+        target_freq = sorted(list(filter(lambda f: f >= target_freq, freqs)))[0]
```
valschneider: Methinks we should make sure `self.plat_info['freqs'][cpu]` is always sorted - either when we fetch them or in devlib directly.

Looking at this, I think `UserspaceSanity.test_performance_sanity()` suffers from the same problem: it does kinda expect we ran the workload on increasing frequencies.

douglas-raillard-arm: +1, let's sort self.plat_info['freqs'] or target.cpufreq.list_frequencies(cpu) in devlib:
https://github.com/ARM-software/lisa/blob/master/lisa/platforms/platinfo.py#L142

douglas-raillard-arm:
@qperret FYI that PR will ensure that provided list of CPUs and frequencies are sorted, so it will raise an exception if devlib/the user is not doing it:
https://github.com/ARM-software/lisa/pull/1272

As commented, it's probably better to sort the output of devlib functions so that everyone benefits from it rather than just fixing it in PlatformInfo.

douglas-raillard-arm:
Closing as it's now fixed in devlib cpufreq module (and the subtree in LISA is up to date). @qperret Feel free to reopen an issue if you still have related problems, but non-sorted freq will be caught at the PlatformInfo level so it won't be silent.

PR #1271, opened 2019-12-30
lisa.target: Fix enable/disable idle states

ambroise-arm:
Using target.disable_idle_states() only disables the idle states for the first CPU of each domain, not for all of them. I assume it is not the intended behavior.
This patch enables and disables the idle states for every CPUs, not only the first one of each domain.

### DIFF
--- a/lisa/target.py
+++ b/lisa/target.py
@@ -803,9 +803,6 @@ class Target(Loggable, HideExekallID, ExekallTaggable, Configurable):
         """
         Context manager that lets you disable all idle states
         """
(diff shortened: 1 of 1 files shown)


lisa/target.py
```diff
@@ -819,12 +816,14 @@ def disable_idle_states(self):
             def cm():
                 try:
                     for domain in self.target.cpufreq.iter_domains():
```
douglas-raillard-arm: Makes me realize that we don't need to mess with cpufreq domains at all now:
```
for cpu in range(self.plat_info['cpus-count']):
    ...
```

PR #1304, opened 2020-01-22
Update workload-automation and devlib to latest master

ionela-voinescu:
6f9856cf pcmark: Update popup dismissal to be case insensitive.
0f9331da fw/job: copy classifiers from the spec
659e6041 fw/exec: Add add_classifier() method
796f62d9 commands/process: partial results + write info
f60032a5 fw/target/manager: Update to use `module_name_set`
977ce499 utils/types: Add `module_name_set` type
a66251dd Antutu: Updating to work with major version 8.

git-subtree-dir: external/workload-automation
git-subtree-split: 6f9856cf2e9b431b6efd2d13fa109bd7d83ffb6c


douglas-raillard-arm:
@ionela-voinescu Can you reroll with devlib update as well while we are at it please ? (with lisa-update-subtrees)

ionela-voinescu:
@douglas-raillard-arm I've done that initially but I was only interested in wa :). Coming up!

douglas-raillard-arm:
lisa-update-subtrees also create the merge commit with some information on the remote and the branch used for the update, so it's better to use it in all cases (it's possible to comment out some subtrees in external/subtrees.conf if you really don't want to update one thing, but that never happened so far)

PR #1353, opened 2020-02-21
eas_behaviour: restrict bigs duty cycle

ionela-voinescu:
Restrict 'big' duty cycle to a value that ensures migration to
bigs while it does not go high enough to cause potential thermal or
power throttling.

### DIFF
--- a/lisa/tests/scheduler/eas_behaviour.py
+++ b/lisa/tests/scheduler/eas_behaviour.py
@@ -59,6 +59,37 @@ class EASBehaviour(RTATestBundle):
         return self.plat_info['nrg-model']
 
     @classmethod
+    def get_big_duty_cycle(cls, plat_info, load):
+        """
+        For load 0 or negative it returns the smallest duty-cycle percentage
+        guaranteeing the task will be placed on the biggest CPUs of the system.
+        This allows minimizing thermal issues while ensuring the expected
+        placement. Otherwise it return the duty-cycle percentage equivalent to
+        load on big CPUs.
+        """
+        # If a negative/zero load is passed we return the duty-cycle equivalent
+        # to the capacity of CPUs with the closest capacity to bigs
+        if load <= 0:
+            try:
+                bigs = plat_info["capacity-classes"][-2]
+            except IndexError:
+                bigs = plat_info["capacity-classes"][0]
+            load = 100
+        else:
+            bigs = plat_info["capacity-classes"][-1]
+
+        return cls.unscaled_utilization(plat_info, bigs[0], load)
+
+    @classmethod
+    def get_little_duty_cycle(cls, plat_info, load):
+        """
+        Returns the duty-cycle percentage equivalent to load on little CPUs
+        """
+        littles = plat_info["capacity-classes"][0]
+
+        return cls.unscaled_utilization(plat_info, littles[0], load)
+
+    @classmethod
     def check_from_target(cls, target):
         for domain in target.cpufreq.iter_domains():
             if "schedutil" not in target.cpufreq.list_governors(domain[0]):
@@ -404,8 +435,7 @@ class OneSmallTask(EASBehaviour):
 
     @classmethod
     def get_rtapp_profile(cls, plat_info):
-        littles = plat_info["capacity-classes"][0]
-        duty = cls.unscaled_utilization(plat_info, littles[0], 50)
+        duty = cls.get_little_duty_cycle(plat_info, 50)
 
         rtapp_profile = {}
         rtapp_profile[cls.task_name] = Periodic(
@@ -446,8 +476,7 @@ class ThreeSmallTasks(EASBehaviour):
 
     @classmethod
     def get_rtapp_profile(cls, plat_info):
-        littles = plat_info["capacity-classes"][0]
-        duty = cls.unscaled_utilization(plat_info, littles[0], 50)
+        duty = cls.get_little_duty_cycle(plat_info, 50)
 
         rtapp_profile = {}
         for i in range(3):
@@ -469,8 +498,7 @@ class TwoBigTasks(EASBehaviour):
 
     @classmethod
     def get_rtapp_profile(cls, plat_info):
-        bigs = plat_info["capacity-classes"][-1]
-        duty = cls.unscaled_utilization(plat_info, bigs[0], 80)
+        duty = cls.get_big_duty_cycle(plat_info, 0)
 
         rtapp_profile = {}
         for i in range(2):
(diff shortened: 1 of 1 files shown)


lisa/tests/scheduler/eas_behaviour.py
```diff
@@ -469,8 +469,11 @@ class TwoBigTasks(EASBehaviour):
 
     @classmethod
     def get_rtapp_profile(cls, plat_info):
-        bigs = plat_info["capacity-classes"][-1]
-        duty = cls.unscaled_utilization(plat_info, bigs[0], 80)
+        num_classes = len(plat_info["capacity-classes"])
```
douglas-raillard-arm: Wouldnt that be more straightforward than the modulo arithmetic ?
```
try:
     pre_bigs = plat_info["capacity-classes"][-2]
except IndexError:
     pre_bigs = plat_info["capacity-classes"][0]
```

lisa/tests/scheduler/eas_behaviour.py
```diff
@@ -493,11 +496,14 @@ class TwoBigThreeSmall(EASBehaviour):
 
     @classmethod
     def get_rtapp_profile(cls, plat_info):
+        # A duty cycle equivalent to the max capacity for all CPUs != big
```
douglas-raillard-arm: if that pattern is repeated, it needs to go into a method of EASBehaviour so all these tests share it.
Something like:
```
def get_big_duty_cycle(self)
     """
     Returns the smallest duty cycle percentage guaranteeing the task will be placed on the biggest CPUs of the system. This allows minimizing thermal issues while ensuring the expected placement.
     """
     ....
```

lisa/tests/scheduler/eas_behaviour.py
```diff
@@ -469,8 +469,11 @@ class TwoBigTasks(EASBehaviour):
 
     @classmethod
     def get_rtapp_profile(cls, plat_info):
-        bigs = plat_info["capacity-classes"][-1]
-        duty = cls.unscaled_utilization(plat_info, bigs[0], 80)
+        num_classes = len(plat_info["capacity-classes"])
+        # A duty cycle equivalent to the max capacity for all CPUs != big
+        # is enough to ensure migration to bigs
+        pre_bigs = plat_info["capacity-classes"][-2 % num_classes]
+        duty = cls.unscaled_utilization(plat_info, pre_bigs[0], 100)
```
douglas-raillard-arm: We should add some margin on top. Having the mean util_avg equal to pre-big capa means the util_avg will jump around that value. We don't want the task to migrate back and forth as the value evolves (unless migration is triggered based on util_est_enqueued ?)

ionela-voinescu: That is the reason I've put a duty-cycle equivalent to 100% pre-big capacity. As long as the jumping around is in within of the 20% capacity margin in the scheduler
> #define fits_capacity(cap, max)»((cap) * 1280 < (max) * 1024)
I don't believe it will ping pong back and forth. 

I would rather not add an extra hard-coded margin here unless we find we need it.

lisa/tests/scheduler/eas_behaviour.py
ionela-voinescu: I'd call it a judgement call - I happen to like the modulo arithmetic :).

douglas-raillard-arm: I'd still rather have something that is fundamentally a case disjunction expressed as a disjunction. Fundamentally, wrapping around hasn't got anything to do with what we try to achieve here, it just happens to work for lists of size one. If you feed a list of size zero to that, you end up with that:
```
ZeroDivisionError: integer division or modulo by zero
```
An `IndexError` or `KeyError` is what you would expect to be raised, and that would be fixed by addding an extra case to the existing disjunction. You won't be able to extend the modulo arithmetic version in a natural way.

Here is another interesting way of getting the values :) 
```
z = [0.0, -0.0, -0.0, 0.0, -0.0, 0.0, 0.0, 0.0, 0.0, -0.0, -0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -0.0, 0.0, -0.0, 0.0, 0.0, -0.0, -0.0, 8.117465778995262e-144, -1.3574776080583756e-141, -5.793790283191115e-140, 4.9840742267000315e-138, 9.483612326238007e-136, 7.314895422521054e-134, 1.6099195392731198e-132, -4.105478669405513e-130, -7.427221233524743e-128, -7.498702592300715e-126, -4.809848575740552e-124, -6.703810270441985e-123, 3.419583656069603e-120, 6.16987557559085e-118, 6.886498915099258e-116, 5.512904334664376e-114, 2.5991703324739765e-112, -9.129750425547004e-111, -4.048043258402688e-108, -5.951570641787871e-106, -6.174123639843241e-104, -4.71390995477634e-102, -2.0075522064015385e-100, 1.1739594818680068e-98, 3.977777625422032e-96, 5.637528271031509e-94, 5.7258028232725405e-92, 4.225801270898365e-90, 1.5637492339139832e-88, -1.5240372007239469e-86, -4.140309230461677e-84, -5.5260305859683e-82, -5.248444788185906e-80, -3.368003518906445e-78, -4.361671207044824e-77, 2.63656622930671e-74, 4.847720982842532e-72, 5.4252689001986225e-70, 4.102285350788563e-68, 1.2882210568762295e-66, -2.037514578659163e-64, -4.611843918300611e-62, -5.351598609850831e-60, -3.8513749336501494e-58, -6.300041250627269e-57, 2.941068446804099e-54, 5.171296173080001e-52, 4.832242159774863e-50, 1.8946601598668503e-48, -2.1911296414489678e-46, -5.100540781296532e-44, -4.854584519044269e-42, -1.3094108973687599e-40, 3.333316363415806e-38, 5.5777568241833485e-36, 3.2477626777116643e-34, -2.1497155288492977e-32, -5.73007790825629e-30, -3.355815585833561e-28, 3.1365136314711764e-26, 6.0081100004061e-24, 4.119324187259597e-23, -6.324191177489764e-20, -2.1793815219873502e-18, 6.898607088339474e-16, 6.458311574773866e-15, -8.178907293608016e-12, 8.212511881361966e-10, -4.4996266722290385e-08, 1.608581302976781e-06, -3.9673426678985286e-05, 0.0006854557579420314, -0.008235640617195734, 0.06701533015648299, -0.35163723949483455, 1.0939320596889983, -0.7322453637909365, -0.013090805886546167]

def get_index(l):
    x = len(l)
    if x < len(z):
        return sum((x**i) * a for i, a in enumerate(reversed(z)))
    else:
        return x - 1

pre_bigs = plat_info["capacity-classes"][get_index(plat_info["capacity-classes"])]
```

ionela-voinescu: Agreed and already changed! :)

lisa/tests/scheduler/eas_behaviour.py
douglas-raillard-arm: We can definitely get 10 to 20% of util variation because of CPU invariance not doing its job, plus the util swing that depends on the task period. We can expect ~7 or 8% (i.e. 80 or 100 PELT units) oscillation in util_avg as well. That means it's not very hard to have an overall 20% variation if the CPU capacities are not writeable, or if we end up increasing the period for better testing coverage.

lisa/tests/scheduler/eas_behaviour.py
douglas-raillard-arm: I got to do polynomial fitting on Monday morning thanks to you anyway :)

PR #1357, opened 2020-02-24
lisa.tests.scheduler.eas_behavior: Align get_big_duty_cycle() paramet…

douglas-raillard-arm:
…er with unscaled_utilization()

* Use utilization_pct instead of load
* Use utilization_pct=None as a default and marker value instead of negative
  values.

### DIFF
--- a/lisa/tests/scheduler/eas_behaviour.py
+++ b/lisa/tests/scheduler/eas_behaviour.py
@@ -59,35 +59,39 @@ class EASBehaviour(RTATestBundle):
         return self.plat_info['nrg-model']
 
     @classmethod
-    def get_big_duty_cycle(cls, plat_info, load):
+    def get_big_duty_cycle(cls, plat_info, utilization_pct=None):
         """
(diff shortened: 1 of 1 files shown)


lisa/tests/scheduler/eas_behaviour.py
```diff
@@ -59,35 +59,41 @@ def nrg_model(self):
...
+
+        :param utilization_pct: If ``None``, the duty cycle will be chosen so
+            that the task will not fit on the second to biggest CPUs in the
+            system, thereby forcing up-migration while minimizing the thermal
+            impact. If a number is passed, the value will be used as a
+            percentage of utilization of a big CPU, and will be converted to a
+            duty cycle.
+        :type utilization_pct: int or None
         """
         # If a negative/zero load is passed we return the duty-cycle equivalent
```
ionela-voinescu: You should remove this as now there is enough detail in the description! Thanks!

PR #1367, opened 2020-02-28
wa_results_collector: kernel identifier improvement

vdonnefort:
This PR allows to improve the kernel identification into wa_results_collector:

sha1 ref >> commit message >> sha1 >> kernel version

This allow to support kernel version without any sha1.

### DIFF
--- a/lisa/git.py
+++ b/lisa/git.py
@@ -67,6 +67,17 @@ def get_uncommited_patch(repo):
     """
     return git(repo, 'diff', 'HEAD')
 
+def get_commit_message(repo, ref='HEAD', format='%s'):
+    """
+    Get the reference commit message.
+
+    :returns: String containing the commit message.
+
+    :param format: Format string passed to ``git show --format``. Default to subject.
+    :type format: str
+    """
+    return git(repo, 'show', '--format=' + format, '-s', ref)
+
 def find_commits(repo, ref='HEAD', grep=None):
     """
     Find git commits.
--- a/lisa/wa_results_collector.py
+++ b/lisa/wa_results_collector.py
@@ -39,7 +39,7 @@ from trappy.utils import handle_duplicate_index
 from IPython.display import display
 
 from lisa.trace import Trace
-from lisa.git import find_shortest_symref
+from lisa.git import find_shortest_symref, get_commit_message
 from lisa.utils import Loggable, memoized
 from lisa.datautils import series_integrate, series_mean
 
@@ -169,7 +169,10 @@ class WaResultsCollector(Loggable):
                 try:
                     symref = find_shortest_symref(kernel_repo_path, sha1)
                 except ValueError:
-                    symref = sha1
+                    try:
+                        symref = get_commit_message(kernel_repo_path, sha1)
+                    except subprocess.CalledProcessError:
+                        symref = sha1
                 kernel_refs[sha1] = symref
             else:
                 kernel_refs[sha1] = sha1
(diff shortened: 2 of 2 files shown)


lisa/wa_results_collector.py
```diff
@@ -163,11 +163,16 @@ def __init__(self, base_dir=None, wa_dirs=".*", plat_info=None,
...
+            if sha1 is None:
+                continue
             if kernel_repo_path:
                 try:
                     symref = find_shortest_symref(kernel_repo_path, sha1)
                 except ValueError:
-                    symref = sha1
+                    try:
+                        symref = get_commit_message(kernel_repo_path, sha1)
+                    except:
```
douglas-raillard-arm: `except` needs an exception type to catch, otherwise it will catch things that you definitely don't want to catch: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

I assume that you are mostly interested in catching `subprocess.CalledProcessError`

lisa/wa_results_collector.py
```diff
@@ -571,18 +580,25 @@ def _get_extra_job_metrics(self, job_dir, workload):
         else:
             return pd.DataFrame()
 
+    @memoized
+    def _wa_get_kernel_version(self, wa_dir):
```
douglas-raillard-arm: WAResultCollector instances already has a `plat_info` attribute, so you should be able to use `self.plat_info['kernel']['version']` (it's already a `KernelVersion` object [1]). 

Since plat_info is currently optional, you can build it in `WAResultCollector.__init__` with something like that:
```
if plat_info is None:
        with open(os.path.join(wa_dir, '__meta', 'target_info.json')) as f:
            target_info = json.load(f)

        plat_info = PlatformInfo({
              'kernel': {
                     'version': KernelVersion(target_info['kernel_release']),
               },
        })
self.plat_info = plat_info
```     

[1] https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/target.html#lisa.platforms.platinfo.PlatformInfo

lisa/git.py
```diff
@@ -67,6 +67,17 @@ def get_uncommited_patch(repo):
...
+def get_commit_message(repo, ref='HEAD', format='%s'):
+    """
+    Get the reference commit message.
+
+    :returns: String containing the commit message.
+
+    :param format: Format string passed to ``git show --format``. Default to subject.
+    :type format: str
+    """
+    return git(repo, 'show', ref, '--format=' + format, '-s')
```
douglas-raillard-arm: maybe `ref` needs to be moved at the end, since the manpage indicates:
```
git show [<options>] [<object>...]
```
that may help avoiding conflicts between option names and branch names.

lisa/wa_results_collector.py
vdonnefort: Will do.

lisa/wa_results_collector.py
vdonnefort: That wouldn't completely work here: WAResultsCollector is created for several WA "kernels". The value needs to be per WA folder.

lisa/git.py
vdonnefort: Will do.

lisa/wa_results_collector.py
douglas-raillard-arm: interesting, that means that this `plat_info` parameter is somewhat useless. The strange thing is that it replaces a parameter that used to be something equivalent, so it's surprising that it does not make sense. I guess that will wait for a round of cleanup and documentation.

PR #1369, opened 2020-02-28
Support for new rt-app feature (nodes_membind)

PeterPuhov-Linaro:
Latest rt-app supports NUMA memory binding:
https://github.com/scheduler-tools/rt-app/blob/b3a5408c6d9f7eeb1b91d8286757cbd38fefa9cb/doc/tutorial.txt#L287

This will allow testing of kernel NUMA_BALANCING code.
The goal is to be able to bind memory allocations to a set of specific 
NUMA nodes and test that task placement decisions will be correct.

### DIFF
--- a/lisa/wlgen/rta.py
+++ b/lisa/wlgen/rta.py
@@ -116,7 +116,8 @@ class RTA(Workload):
         # Move configuration file to target
         self.target.push(self.local_json, self.remote_json)
 
-    def run(self, cpus=None, cgroup=None, background=False, as_root=False, update_cpu_capacities=None):
+    def run(self, cpus=None, cgroup=None, background=False, as_root=False, update_cpu_capacities=None,
+            numa_nodes_membind=None):
         logger = self.get_logger()
         plat_info = self.target.plat_info
         writeable_capacities = plat_info['cpu-capacities']['writeable']
@@ -159,7 +160,7 @@ class RTA(Workload):
             capa_cm = nullcontext()
 
         with capa_cm:
-            super().run(cpus, cgroup, background, as_root)
+            super().run(cpus, cgroup, background, as_root, numa_nodes_membind)
 
         if background:
             # TODO: handle background case
@@ -726,10 +727,16 @@ class Phase(Loggable):
                           when reaching the end of this phase. Currently only
                           supported when duty_cycle_pct=100
     :type barrier_after: str
+
+    :param numa_nodes_membind: the list of NUMA Nodes.
+        Task will only allocate memory from these nodes during this phase.
+        If unspecified, that phase will be allowed to allocate memory from any 
+        NUMA node, regardless of the previous phase settings.
+    :type numa_nodes_membind: list(int) or None
     """
 
     def __init__(self, duration_s, period_ms, duty_cycle_pct, cpus=None, barrier_after=None,
-                 uclamp_min=None, uclamp_max=None):
+                 uclamp_min=None, uclamp_max=None, numa_nodes_membind=None):
         if barrier_after and duty_cycle_pct != 100:
             # This could be implemented but currently don't foresee any use.
             raise ValueError('Barriers only supported when duty_cycle_pct=100')
@@ -741,6 +748,7 @@ class Phase(Loggable):
         self.barrier_after = barrier_after
         self.uclamp_min = uclamp_min
         self.uclamp_max = uclamp_max
+        self.numa_nodes_membind = numa_nodes_membind
 
     def get_rtapp_repr(self, task_name, plat_info):
         """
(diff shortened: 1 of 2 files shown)


tools/recipes/rt-app.recipe
```diff
@@ -9,11 +9,21 @@ build_jsonc() {
...
+}
+
 build_rtapp() {
     cd rt-app
     export ac_cv_lib_json_c_json_object_from_file=yes
+    export ac_cv_lib_numa_numa_available=yes
     ./autogen.sh
-    ./configure --host=$CONFIGURE_HOST LDFLAGS="--static -L$BUILD_DIR/json-c/" CFLAGS="-I$BUILD_DIR" --with-deadline
+    ./configure --host=$CONFIGURE_HOST LDFLAGS="--static -L$BUILD_DIR/json-c/ -L$BUILD_DIR/numactl/" CFLAGS="-I$BUILD_DIR" --with-deadline
+    make clean
```
douglas-raillard-arm: Is `make clean` actually needed ? The expected behavior of `lisa-build-asset` is to make each build in a clean work tree (either a clean clone or a copy of a clean clone).

If `make clean` is required for the build to succeed, that either indicates that :
*  the build relies on something that stays shared outside of `lisa-build-asset` control
* that `lisa-build-asset` has a bug somehow
* that rtapp repo contains files that need to be removed before building (unlikely)

lisa/wlgen/rta.py
```diff
@@ -726,10 +727,15 @@ class Phase(Loggable):
                           when reaching the end of this phase. Currently only
                           supported when duty_cycle_pct=100
     :type barrier_after: str
+
+    :param nodes_membind: the list of NUMA Nodes.
```
douglas-raillard-arm: Given the variety of origins of parameters to wlgen.rta constructors, it might be a good idea to prefix them so that someone not familiar with NUMA knows it's about NUMA (node being a very generic term). It would give something like `numa_nodes_membind`.

@valschneider Any thoughts on that ?

tools/recipes/rt-app.recipe
```diff
@@ -9,11 +9,21 @@ build_jsonc() {
     make
 }
 
+build_numactl() {
```
douglas-raillard-arm: The recipe looks good to me, but I'll need to get clearance from our legal department before merging the PR since we distribute prebuilt binaries of rtapp in LISA.

If you want, you can split the PR into two parts:
* build recipe update for numactl
* everything else

This way we can merge the Python side faster and avoid potential merge conflicts arising from stale PRs.

lisa/wlgen/rta.py
```diff
@@ -726,10 +727,15 @@ class Phase(Loggable):
...
+        Task will only allocate memory from these nodes during this phase.
+        If unspecified, that phase will inherit memory allocation policy from previous phase.
```
valschneider: To follow the spirit of what @douglas-raillard-arm did in ed748695503e ("lisa.wlgen.rta: Align cpus phase parameter with its documentation"), we should probably reset the restriction if unspecified in the rt-app description of a given phase.

The main argument back then was that this prevents composition from being deterministic: if you add two rt-app tasks together you don't want the first phase of the second task to be impacted by the last phase of the first task.

lisa/wlgen/rta.py
valschneider: In ~l33t~kernelspeak, node very much means NUMA node. I don't have any big preference towards having (or not having) the "numa" prefix.

douglas-raillard-arm: The thing is that it's not the kernel, it's LISA, and there are already other kind of nodes like this one:
https://github.com/ARM-software/lisa/blob/master/lisa/energy_model.py#L147

I'd rather have something relatively obvious than some undecipherable set of abbreviations. Python never had this sort of restriction, so there is no point in obfuscation :-)
http://port70.net/~nsz/c/c11/n1570.html#6.4.2.1p5

tools/recipes/rt-app.recipe
PeterPuhov-Linaro: I will remove make clean.

lisa/wlgen/rta.py
PeterPuhov-Linaro: I will rename it to numa_nodes_membind.
I tried to make it consistent with rt-app name.

tools/recipes/rt-app.recipe
PeterPuhov-Linaro: I will remove build recipe update from this PR 
and will create a separate PR for new rt-app and numactl.

lisa/wlgen/rta.py
PeterPuhov-Linaro: I will follow the same concept.

PR #1370, opened 2020-03-02
Wa results collector/parsing optimizations

vdonnefort:
This intends to drastically improve the parsing speed, especially with energy_measure data.

e.g. 50 PCMark runs on a Pixel4

Before: 20min
After: 1min

### DIFF
--- a/lisa/wa_results_collector.py
+++ b/lisa/wa_results_collector.py
@@ -366,11 +366,11 @@ class WaResultsCollector(Loggable):
             if extra_df.empty:
                 continue
 
-            extra_df.loc[:, 'workload'] = workload
-            extra_df.loc[:, 'iteration'] = iteration
-            extra_df.loc[:, 'id'] = job_id
-            extra_df.loc[:, 'tag'] = tag
-            extra_df.loc[:, 'test'] = test
+            extra_df['workload'] = workload
+            extra_df['iteration'] = iteration
+            extra_df['id'] = job_id
+            extra_df['tag'] = tag
+            extra_df['test'] = test
             # Collect all these DFs to merge them in one go at the end.
             extra_dfs.append(extra_df)
 
@@ -393,8 +393,8 @@ class WaResultsCollector(Loggable):
         # now, though - that refactoring would probably belong alongside a
         # refactoring to use WA's own API for reading output directories.
         df['_job_dir'] = df['id'].replace(job_dir_map)
-        df.loc[:, 'kernel_name'] = self._wa_get_kernel_name(wa_dir)
-        df.loc[:, 'kernel_sha1'] = self._wa_get_kernel_sha1(wa_dir)
+        df['kernel_name'] = self._wa_get_kernel_name(wa_dir)
+        df['kernel_sha1'] = self._wa_get_kernel_sha1(wa_dir)
 
         return df
 
@@ -520,8 +520,8 @@ class WaResultsCollector(Loggable):
             # 'device_power_sample'.
             device_name = artifact_name[len('energy_instrument_output') + 1:]
             name_extra = device_name or 'device'
-            df.loc[:, 'metric'] = '{}_power_sample'.format(name_extra)
-            df.loc[:, 'units'] = 'watts'
+            df['metric'] = '{}_power_sample'.format(name_extra)
+            df['units'] = 'watts'
 
         elif 'output_power' in df.columns and 'USB_power' in df.columns:
             # Looks like this is from a Monsoon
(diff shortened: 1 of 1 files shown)


lisa/wa_results_collector.py
```diff
@@ -492,6 +492,48 @@ def row_filter(r): return True
 
         return ret
 
+    def _read_energy_instrument_metrics(self, path):
+        """
+        Parse the WA energy_measurement output.
+        """
+        path_cache = path + ".parquet"
+
+        if not os.path.isfile(path_cache):
```
douglas-raillard-arm: Could swap both legs and remove the negation in the condition

lisa/wa_results_collector.py
```diff
@@ -492,6 +492,48 @@ def row_filter(r): return True
...
+            df.loc[:, 'units'] = 'watts'
+
+        elif 'output_power' in df.columns and 'USB_power' in df.columns:
+            # Looks like this is from a Monsoon
+            # For monsoon the USB and device power are collected
+            # together with the same timestamps, so we can just add them
+            # up.
+            power_samples = df['output_power'] + df['USB_power']
+            df = pd.DataFrame({'value': power_samples})
+            df.loc[:, 'metric'] = 'device_power_sample'
```
douglas-raillard-arm:
```
 df.loc[:, 'metric'] = ...
```
Can be turned into:
```
 df['metric'] = ...
```

lisa/wa_results_collector.py
```diff
@@ -492,6 +492,48 @@ def row_filter(r): return True
...
+    def _read_energy_instrument_metrics(self, path):
+        """
+        Parse the WA energy_measurement output.
+        """
+        path_cache = path + ".parquet"
+
+        if not os.path.isfile(path_cache):
+            try:
+                df = pd.read_csv(path)
+            except pandas.errors.ParserError as e:
```
douglas-raillard-arm: Given how the function is used, there seems to be no point in catching the exception since it is not handled here. By turning the exception into `None` and then checking for `None`, all information about the original issue is lost without any benefits.

lisa/wa_results_collector.py
```diff
@@ -540,41 +582,12 @@ def _get_extra_job_metrics(self, job_dir, workload):
                 continue
 
             if artifact_name.startswith('energy_instrument_output'):
-
-                try:
-                    df = pd.read_csv(path)
-                except pandas.errors.ParserError as e:
+                df = self._read_energy_instrument_metrics(path)
```
douglas-raillard-arm: following previous suggested change, this would turn into:
```
try:
    df = self._read_energy_instrument_metrics(path) 
except pandas.errors.ParserError as e:
     logger.info(" no data for %s", path)
     continue
else:
      .... # do something with df
```

lisa/wa_results_collector.py
```diff
@@ -492,6 +492,48 @@ def row_filter(r): return True
...
 
+    def _read_energy_instrument_metrics(self, path):
+        """
+        Parse the WA energy_measurement output.
+        """
+        path_cache = path + ".parquet"
+
+        if not os.path.isfile(path_cache):
+            try:
+                df = pd.read_csv(path)
```
douglas-raillard-arm: Could the CSV be modified in any way ?
I guess not, in which case it's fine to not checksum the CSV and just use parquet if it's there. Maybe add a comment stating that assumption

lisa/wa_results_collector.py
douglas-raillard-arm: rather than `os.path.isfile`, you can try to read the file and handled the exception. That's the preferred pattern as it avoids race condition and also duplication of assumptions (read_parquet is already aware of what it can handle)

vdonnefort: Done

lisa/wa_results_collector.py
vdonnefort: Done

lisa/wa_results_collector.py
vdonnefort: Done

lisa/wa_results_collector.py
vdonnefort: Done

lisa/wa_results_collector.py
vdonnefort: Done

lisa/wa_results_collector.py
vdonnefort: As this code has been moved, I changed it into the follow-up commit which fixes the `.loc` for all the found occurrences within this class.

PR #1372, opened 2020-03-02
WIP: lisa.pelt: Make kernel_util_mean() accurate for non-50% tasks

douglas-raillard-arm:
Attempt to compute the "true" utilisation mean even though the signal is not decayed as expected.

DO NOT MERGE (that code is currently broken)


douglas-raillard-arm:
Closing as this code is actually broken

PR #1384, opened 2020-03-11
init_env: add a post-merge githook to remind user of re-init init_env

qais-yousef:
post-merge githook will run after every git pull, and of course after
any merge. Remind the developer that init_env must be re-sourced afresh
to conitnue to get reliable lisa behavior.


init_env
```diff
@@ -62,6 +62,13 @@ else
 	echo "Please, source this configuration from a terminal running BASH."
 fi
 
+if [[ -d $LISA_HOME/.git/hooks ]]; then
```
douglas-raillard-arm: init_env is only the toplevel wrapper with shell-specific setup, everything else is done in shell/lisa_shell.
You can add a function there:
```
function _lisa-install-git-hooks {
     # works even with worktrees
     local hooks="$(git -C "$LISA_HOME" rev-parse --git-common-dir)/hooks"
     # Avoid overriding user hooks
     [[ ! -e "$hooks/post-merge" ]] && cp "$LISA_HOME/tools/post-merge" "$hooks/post-merge"
     [[ ! -e "$hooks/post-checkout" ]] && cp "$LISA_HOME/tools/post-checkout" "$hooks/post-checkout"
}
```
and call it at the end of `lisa-install`. You will also need to update:
https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/setup.html#updating

tools/post-merge
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+echo ""
+echo ""
+echo ""
+echo "#########################################################################"
```
douglas-raillard-arm: if you want you can source "$LISA_HOME/shell/lisa_colors" and use that to make the message red or something.

tools/post-merge
```diff
@@ -0,0 +1,17 @@
...
+echo "#                                                                       #"
+echo "# !!!WARNING!!!                                                         #"
+echo "#                                                                       #"
+echo "# Must execute:                                                         #"
+echo "#                                                                       #"
+echo "#    \`env -i bash\`                                                    #"
```
douglas-raillard-arm: s/bash/$(basename "$SHELL")/ so it stays accurate for zsh as well :)

init_env
qais-yousef: That's a lot better, yes! Thanks

tools/post-merge
qais-yousef: Can do :)

tools/post-merge
qais-yousef: Sure.

init_env
qais-yousef: `[[ ! -e "$hooks/post-merge" ]]` check won't be good enough. we will fail to update it after the first install by the way

shell/lisa_shell
```diff
@@ -219,6 +219,28 @@ function _lisa-install-system-check {
...
+
+     # Copy/Update lisa specific git-hook
+     cp "$LISA_HOME/tools/$hook.lisa" "$hooks_dir/$hook.lisa"
+
+     # Avoid overriding user hooks, but make sure lisa specific hook is
+     # called too
+     if [[ ! -e "$hooks_dir/$hook" ]]; then
+         cp "$LISA_HOME/tools/$hook" "$hooks_dir/$hook"
+     else
+         grep -qxF '"$0".lisa' "$hooks_dir/$hook" ||  echo '"$0".lisa' >> "$hooks_dir/$hook"
```
douglas-raillard-arm: I'm not sure there is any reliable way of making that work. If the hook ends with `exit` or something like that it will never be called.
Also, hooks can be symlinks, which means we have no idea of what we are modifying. This could be a script stored in user's home, in which case they would be pretty upset if lisa started adding random stuff to it. 

Ultimately, if someone dropped a hook in there, it's because they want that and if they want to call our things as well they can easily call it themselves in a way that makes sense. By default, no file with that name will exist since recent versions of git store the examples as `xxxx.sample` (rather than `xxxx` without the executable permission as it used to do), and most people don't have custom hooks anyway so the vast majority of users will get the feature.

qais-yousef: TBH I think this is too cautious. It'll help the majority of the users to not trip over an essential way of the workflow. Apps modify .bashrc all the time too to add and remove stuff, I don't see this being a problem here. It can be smarter, but I don't think it's worth the effort.

Keep in mind this only applies to their checkout lisa only. It's not a global git-hook for all projects.

By the way, if you ever use git review it automatically installs git-hooks on your behalf, which could run into the same issue.

I have no better alternative anyway.

douglas-raillard-arm: As mentioned, the alternative is to copy the file if the target does not exist, otherwise do nothing. The vast majority of people won't have any hook installed so it will work. People that have a hook installed will not want anyone to go and break that hook, so we don't.

> Apps modify .bashrc all the time too to add and remove stuff
That's exactly the kind of thing that makes me ditch crappy broken installers. You end up with hidden breakages waiting to explode, and you don't even have a guarantee it will work. I've spent enough time getting people to cleanup their .bashrc from PYTHONPATH setup done by lisa legacy after spending an afternoon debugging what's going on to not want that to history to be repeated :)

> By the way, if you ever use git review it automatically installs git-hooks on your behalf, which could run into the same issue.

Looking at the sources, it seems to be doing exactly what I'm describing, i.e. it won't touch anything if it detects a hook is already there:
https://opendev.org/opendev/git-review/src/branch/master/git_review/cmd.py#L1652
(the `os.access(hook_file, os.X_OK)` is there because old versions of git used to ship sample hooks with the normal filename, but without the executable bit. This means that if exec bit is missing, that is the default sample hook that can safely be overwritten).

qais-yousef: Hmm okay I think this makes more sense, you're right. Let me print a warning instead then.

shell/lisa_shell
```diff
@@ -219,6 +219,34 @@ function _lisa-install-system-check {
...
+     # Copy/Update lisa specific git-hook
+     cp "$LISA_HOME/tools/$hook.lisa" "$hooks_dir/$hook.lisa"
+
+     # Avoid overriding user hooks
+     if [[ ! -e "$hooks_dir/$hook" ]]; then
+         cp "$LISA_HOME/tools/$hook" "$hooks_dir/$hook"
+     else
+         installed=`grep -qxF '"$0".lisa' "$hooks_dir/$hook" || echo $?`
+
+         if [[ $installed -ne 0 ]]; then
```
douglas-raillard-arm: Can be simplified into:
```

     if [[ ! -e "$hooks_dir/$hook" ]]; then
         cp "$LISA_HOME/tools/$hook" "$hooks_dir/$hook"
     elif ! grep -qxF '"$0".lisa' "$hooks_dir/$hook"; then
         echo ""
         echo -ne "$LISASHELL_RED"
         echo "!!WARNING!! Couldn't copy git-hook: $LISA_HOME/tools/$hook"
         echo -ne "$LISASHELL_RESET"
    fi
```

qais-yousef: Does it make a difference to warrant the churn?

douglas-raillard-arm: There's no concerning churn, the PR is not merged yet. In any case, a PR update would be needed to make `installed` variable local, so we might as well eliminate it.

Also I've seen this pattern with multiplication of variables and tests in a lot of the existing code in LISA so I assume people don't know `if` takes a command as parameter, hence the comments. Next time there should not be a need for such comment hopefully.

qais-yousef: Okay I converted install to local. I'm aware of if can take a command but I don't like this style in general. Variables are there to help code more readable, and my preference is to keep it like this because you can still skim the code and understand what is happening without having to read the whole grep command.

PR #1417, opened 2020-05-19
Small wlgen cleanups

valschneider:


valschneider:
@PeterPuhov-Linaro Any chance you could double-check the `numa_nodes_membind` bit? In case you have some downstream `Workload` classes that actually use that `run()` parameter.

douglas-raillard-arm:
Actually, it seems the code is currently broken and the parameter `numa_node_membind` must be removed from `run()`, as it is passed to `super().run(timeout=numa_node_membind)` (as a positional parameter).

lisa/wlgen/rta.py
```diff
@@ -724,9 +724,17 @@ class Phase(Loggable):
                           supported when duty_cycle_pct=100
     :type barrier_after: str
 
+    :param uclamp_min: the task uclamp.min value to set for the task for the
```
douglas-raillard-arm, line 727: s/the task/the task's/
s/uclamp.min/``uclamp.min``/ since that is what we use in most docstrings to reflect either externally defined names or parameter names.

lisa/wlgen/rta.py
```diff
@@ -826,16 +834,20 @@ def get_rtapp_repr(self, task_name, plat_info):
...
+
         return phase
 
 class RTATask:
     """
     Base class for conveniently constructing params to :meth:`RTA.by_profile`
 
-    :param sched_policy: the scheduler policy for this task.
+    :param sched_policy: the scheduler policy for this task. Defaults to
+      ``SCHED_OTHER``, see ``man sched`` for information on scheduler policies.
```
douglas-raillard-arm, line 845: Maybe we can start using the manpage role ?
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-manpages_url

lisa/wlgen/rta.py
```diff
@@ -189,7 +189,7 @@ def _process_calibration(self, calibration):
         return calibration
 
     @classmethod
-    def by_profile(cls, target, name, profile, res_dir=None, default_policy=None,
+    def by_profile(cls, target, profile, name=None, res_dir=None, default_policy=None,
```
PeterPuhov-Linaro, line 193: Should we fix doc/workloads.rst :  "wload = RTA.by_profile(target, "4_50_tasks", profile)"
and also doc/transition_guide.rst: "wload = RTA.by_profile(te, "foo", profile, res_dir, calibration)"
The following example seems outdated as well:
https://github.com/valschneider/lisa/blob/wlgen/pre_pypi_cleanup/lisa/wlgen/rta.py#L227

PeterPuhov-Linaro:
I re-run all my tests and found no regression.
I had to update few places when workload created using RTA.by_profile
LGTM, thanks.

lisa/wlgen/rta.py
valschneider: Done!

lisa/wlgen/rta.py
valschneider: Yes indeed!

valschneider:
@PeterPuhov-Linaro Thanks a ton for having a look and giving this a shot!

douglas-raillard-arm:
@valschneider Can you include the string "BREAKING CHANGE" in the commit changing the by_profile parameter order, along with an example of old vs new recommended calls to it.

The list of commits with such marker ends up listed here, so if the commit is self-sufficient it acts as a poor man's migration guide with 0 maintenance cost :)
https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/transition_guide.html

PR #1446, opened 2020-06-29
Test util task migration v5

deggeman:
Solution for https://jira.arm.com/browse/PWRSOFT-2145

Last CI Tests: https://jenkins.oss.arm.com/view/Power/job/power/job/synthetic-tests/job/power-start-matrix-lisa-synthetics/381

- juno-r0:  OneTaskCPUMigration[board=juno-r0/00:02:F7:00:58:F3]:test_util_task_migration:  FAILED 11/50 (22.0%)
- juno-r2:  OneTaskCPUMigration[board=juno-r2/00:02:f7:00:67:99]:test_util_task_migration:  FAILED 7/50  (14.0%)
- hikey960: OneTaskCPUMigration[board=hikey960/00:0e:c6:b9:54:d7]:test_util_task_migration: passed 50/50 (100.0%)
- TC2:      OneTaskCPUMigration[board=tc2/00:02:F7:00:48:A4]:test_util_task_migration:      passed 50/50 (100.0%)
- db845c:   OneTaskCPUMigration[board=db845c/00:e0:4c:75:10:0f]:test_util_task_migration:   passed 50/50 (100.0%)

Commit 8cf1198ad718 "lisa.tests.scheduler.load_tracking: Fix remaining issue for CPUMigrationBase::test_util_task_migration" brings the error rate for juno-r0 and juno-r2 furher down to:

- juno-r0:  OneTaskCPUMigration[board=juno-r0/00:02:F7:00:58:F3]:test_util_task_migration:  FAILED 3/50 (6.0%)
- juno-r2:  OneTaskCPUMigration[board=juno-r2/00:02:f7:00:67:99]:test_util_task_migration:  FAILED 2/50  (4.0%)

### DIFF
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -994,7 +994,7 @@ class CPUMigrationBase(LoadTrackingBase):
             # Ignore the first quarter of the util signal of each phase, since
             # it's impacted by the phase change, and util can be affected
             # (rtapp does some bookkeeping at the beginning of phases)
-            start += duration / 4
+            # start += duration / 4
             phase_df = df_window(df, (start, end), method='pre', clip_window=True)
 
             for cpu in self.cpus:
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -29,14 +29,14 @@ from lisa.tests.base import (
 )
 from lisa.target import Target
 from lisa.utils import ArtifactPath, groupby, ExekallTaggable
-from lisa.datautils import series_mean, df_window, df_filter_task_ids, series_refit_index, df_split_signals, df_refit_index
+from lisa.datautils import series_mean, df_window, df_filter_task_ids, series_refit_index, df_split_signals, df_refit_index, df_extend_index
 from lisa.wlgen.rta import RTA, Periodic, RTATask
 from lisa.trace import FtraceCollector, requires_events, may_use_events, MissingTraceEventError
 from lisa.analysis.load_tracking import LoadTrackingAnalysis
 from lisa.analysis.tasks import TasksAnalysis
 from lisa.analysis.rta import RTAEventsAnalysis
 from lisa.analysis.frequency import FrequencyAnalysis
-from lisa.pelt import PELT_SCALE, simulate_pelt, pelt_settling_time, kernel_util_mean
+from lisa.pelt import PELT_SCALE, simulate_pelt, pelt_settling_time, kernel_util_mean, pelt_interpolate, simulate_pelt_clock
 
 UTIL_SCALE = PELT_SCALE
 
@@ -906,7 +906,6 @@ class CPUMigrationBase(LoadTrackingBase):
                 phase_df = df_window(df, window, clip_window=True)
 
                 for cpu in self.cpus:
-
                     if cpus_rel_freq is None:
                         rel_freq_mean = 1
                     else:
@@ -960,9 +959,35 @@ class CPUMigrationBase(LoadTrackingBase):
         migr_task = sorted(task for task in tasks if task.startswith('migr'))[0]
         migr_task = self.rtapp_task_ids_map[migr_task][0]
 
+        df_state = {}
+
+        for task in self.rtapp_task_ids_map.keys():
+            comm = self.rtapp_task_ids_map[task][0].comm
+            df_state[task] = self.trace.analysis.tasks.df_task_activation(comm)
+            df_state[task] = df_state[task][df_state[task]['active'] == 0]
+
+        cpu_capacities = self.plat_info['cpu-capacities']['rtapp']
         cpu_util = {}
+        cpu_freqs = self.plat_info['freqs']
+
+        try:
+            freq_df = self.trace.analysis.frequency.df_cpus_frequency()
+        except MissingTraceEventError:
+            raise RuntimeError('cpus_rel_freq = None')
+        else:
+            cpus_rel_freq = {
+                    # Frequency, normalized according to max frequency on that CPU
+                    cols['cpu']: df['frequency'] / max(cpu_freqs[cols['cpu']])
+                    for cols, df in df_split_signals(freq_df, ['cpu'])
+            }
+
         for row in self.trace.analysis.rta.df_phases(migr_task).itertuples():
             phase = row.phase
+
+            # TODO: remove that once we have named phases to skip the buffer phase.
+            if phase == 0:
+                continue
+
             duration = row.duration
             start = row.Index
             end = start + duration
@@ -973,7 +998,20 @@ class CPUMigrationBase(LoadTrackingBase):
             phase_df = df_window(df, (start, end), method='pre', clip_window=True)
 
             for cpu in self.cpus:
-                util = phase_df[phase_df['cpu'] == cpu]['util']
+
+                cpu_phase_df = phase_df[phase_df['cpu'] == cpu].copy()
+
+                df_state_clipped = pd.concat(df_window(df_state[task], (start, end), method='pre', clip_window=True)
+                                             for task, wlgen_task in self.rtapp_profile.items()
+                                             if cpu in wlgen_task.phases[phase-1].cpus)
+
+                df_state_clipped = df_state_clipped.sort_index()
+                df_util = df_extend_index(cpu_phase_df, df_state_clipped[[]])
+
+                cpu_capacity = cpu_capacities[cpu] * cpus_rel_freq[cpu]
+                clock = simulate_pelt_clock(cpu_capacity, df_util.index.to_series())
+
+                util = pelt_interpolate(df_util['util'], clock)
                 cpu_util.setdefault(cpu, {})[phase] = kernel_util_mean(util, plat_info=self.plat_info)
 
         return cpu_util
--- a/lisa/tests/scheduler/load_tracking.py
+++ b/lisa/tests/scheduler/load_tracking.py
@@ -957,11 +957,11 @@ class CPUMigrationBase(LoadTrackingBase):
         """
         df = self.trace.analysis.load_tracking.df_cpus_signal('util')
         tasks = self.rtapp_task_ids_map.keys()
-        task = sorted(task for task in tasks if task.startswith('migr'))[0]
-        task = self.rtapp_task_ids_map[task][0]
+        migr_task = sorted(task for task in tasks if task.startswith('migr'))[0]
+        migr_task = self.rtapp_task_ids_map[migr_task][0]
 
         cpu_util = {}
-        for row in self.trace.analysis.rta.df_phases(task).itertuples():
+        for row in self.trace.analysis.rta.df_phases(migr_task).itertuples():
             phase = row.phase
             duration = row.duration
             start = row.Index
--- a/lisa/datautils.py
+++ b/lisa/datautils.py
@@ -1848,4 +1848,29 @@ SignalDesc._SIGNALS_MAP = {
     for event, signal_descs in groupby(_SIGNALS, key=attrgetter('event'))
 }
 
+def _data_extend_index(data, extension):
+    """
+    ``data`` can either be a :class:`pandas.DataFrame` or :class:`pandas.Series`
+    """
+    return data.join(extension, how='outer')
+
+@SeriesAccessor.register_accessor
+def series_extend_index(series, extension):
+    """
+    Extending the index of a :class:`pandas.Series
+
+    :param: series: series to extend
+    :type series: :class:`pandas.Series`
+
+    :param extension: series holding the extension
+    :type extension: class:`pandas.Series`
+    """
+    return _data_extend_index(series, extension)
+
+@DataFrameAccessor.register_accessor
+def df_extend_index(df, extension):
+    """
+    Same as :func:`series_extend_index` but acting on a :class:`pandas.DataFrame`
+    """
+    return _data_extend_index(df, extension)
 # vim :set tabstop=4 shiftwidth=4 textwidth=80 expandtab
--- a/lisa/pelt.py
+++ b/lisa/pelt.py
@@ -20,7 +20,7 @@ import math
 import pandas as pd
 import numpy as np
 
-from lisa.datautils import series_envelope_mean
+from lisa.datautils import series_envelope_mean, df_add_delta, series_extend_index
 
 PELT_WINDOW = 1024 * 1024 * 1e-9
 """
@@ -265,5 +265,55 @@ def kernel_util_mean(util, plat_info):
         utilisation mean.
     """
     return series_envelope_mean(util)
+
+def pelt_interpolate(util, clock, interpolate_at=None):
+    """
+    Interpolate the utilization with an interpolate_at signal and
+    re-indexing on clock.
+
+    :param util: CPU utilization over time.
+    :type util: pandas.Series
+
+    :param clock: A series of timestamps providing the simulated PELT clock.
+    :type clock: pandas.Series
+
+    :param interpolate_at: A series of additional timestamps for which the
+        CPU utilization has to be calculated. It can be omitted in case
+        util already contains those extra timestamps.
+    :type interpolate_at: pandas.Series
+    """
+    if interpolate_at is not None:
+        util = series_extend_index(util, interpolate_at)
+
+    df_util = util.to_frame()
+    df_util = df_util.assign(new_index=clock.values)
+    df_util = df_util.set_index('new_index')
+
+    df_util = df_add_delta(df_util)
+    df_util['delta'] = df_util['delta'].shift()
+    df_util['prev_util'] = df_util['util'].shift()
+
+    df_util = df_util.dropna(subset=['delta'])
+
+    def compute_switch_phase_df(row):
+        # Applying the function on an empty dataframe will lead to being called
+        # with a Series, so the return value will not matter.
+        if row.empty:
+            return
+
+        timestamp = row.name
+        last_update = timestamp - row.delta
+        prev_util = row.prev_util
+        window_shrink = 1e3
+        activations = pd.Series([1, 1], index=[last_update, timestamp])
+        simulated_phase_df = simulate_pelt(activations, init=prev_util,
+                                           window=PELT_WINDOW/window_shrink,
+                                           half_life=PELT_HALF_LIFE*window_shrink)
+        return simulated_phase_df.iloc[-1]
+
+    switch_loc = df_util['util'].isnull()
+    df_util.loc[switch_loc, 'util'] = df_util.loc[switch_loc].apply(compute_switch_phase_df, axis='columns')
+
+    return df_util['util']
 
 # vim :set tabstop=4 shiftwidth=4 textwidth=80 expandtab
--- a/lisa/pelt.py
+++ b/lisa/pelt.py
@@ -316,4 +316,52 @@ def pelt_interpolate(util, clock, interpolate_at=None):
 
     return df_util['util']
 
+def simulate_pelt_clock(capacity, clock, scale=PELT_SCALE):
+    """
+    Simulate a PELT clock of an entity from the capacities of the CPU it's
+    residing on.
+    :param capacity: CPU capacity over time.
+    :type capacity: pandas.Series
+    :param clock: A series of timestamps at which the clock is to be observed.
+        The returned :class:`pandas.Series` will provide the simulated clock
+        values at these instants.
+    :type clock: pandas.Series
+    :param scale: Maximum value allowed for CPU capacity.
+    :type scale: float
+    """
+    # Ensures the clock's index is the same as the clock
+    clock = clock.copy(deep=False)
+    clock.index = clock
+    df = pd.DataFrame(
+        dict(
+            clock=clock,
+            capacity=capacity,
+        ),
+    )
+    # Remember which row is part of the user-provided clock
+    df['orig_clock'] = ~df['clock'].isna()
+    # Needed for "time" interpolation
+    df.index = pd.TimedeltaIndex(df.index, unit='s')
+    # Shift so that the capacity is aligned with the corresponding delta
+    df['capacity'] = df['capacity'].fillna(method='ffill').shift()
+    # Time flows linearly between 2 samples of the clock
+    df['clock'].interpolate(method='time', inplace=True)
+    # If there is an initial NaN in the clock or capacity, remove it since
+    # interpolate() cannot cope with that correctly even with
+    # limit_direction='both'
+    df.dropna(inplace=True)
+    df['delta'] = df['clock'].diff()
+    # Scale each delta independantly
+    df['delta'] *= df['capacity'] / scale
+    # Fill the NaN with the initial value for the cumsum() fold
+    df['delta'].iat[0] = df['clock'].iat[0]
+    # Reverse df_add_delta() now that we scaled each delta
+    df['new_clock'] = df['delta'].cumsum()
+    # Back to Float64Index
+    df.index = df.index.total_seconds()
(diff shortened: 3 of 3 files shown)


lisa/tests/scheduler/load_tracking.py
```diff
@@ -971,7 +979,44 @@ def get_trace_cpu_util(self):
...
 
             for cpu in self.cpus:
-                util = phase_df[phase_df['cpu'] == cpu]['util']
+
+                cpu_phase_df = phase_df[phase_df['cpu'] == cpu].copy()
+
+                for task in tasks:
+                    if self.rtapp_profile[task].phases[phase-1].cpus.count(cpu):
+                        df_state_clipped = df_window(df_state[task], (start, end), method='pre', clip_window=True)
+                        cpu_phase_df = cpu_phase_df.join(df_state_clipped[[]], how='outer')
```
douglas-raillard-arm, line 988: extending the index of a series/dataframe with another index or series should be factored out in lisa.datautils to preserve readability, especially if it's relying on a hack like this join()

lisa/tests/scheduler/load_tracking.py
```diff
@@ -971,7 +979,44 @@ def get_trace_cpu_util(self):
...
+
+                cpu_phase_df['cpu'].fillna(method='ffill', inplace=True)
+                cpu_phase_df = df_add_delta(cpu_phase_df)
+                cpu_phase_df['delta'] = cpu_phase_df['delta'].shift()
+                cpu_phase_df['prev_util'] = cpu_phase_df['util'].shift()
+
+                cpu_phase_df = cpu_phase_df[cpu_phase_df['cpu'].notna()]
+                cpu_phase_df = cpu_phase_df[cpu_phase_df['delta'].notna()]
```
douglas-raillard-arm, line 996:
```
                 cpu_phase_df = cpu_phase_df.dropna(subset=['cpu', 'delta'])
```

lisa/tests/scheduler/load_tracking.py
```diff
@@ -971,7 +979,44 @@ def get_trace_cpu_util(self):
...
+                    window_shrink = 1e3
+                    activations = pd.Series([1, 1], index=[last_update, timestamp])
+                    simulated_phase_df = simulate_pelt(activations, init=prev_util,
+                                                       window=PELT_WINDOW/window_shrink,
+                                                       half_life=PELT_HALF_LIFE*window_shrink)
+                    util = simulated_phase_df.iloc[-1]
+                    return util
+
+                switch_loc = cpu_phase_df['util'].isnull()
+                cpu_phase_df.loc[switch_loc, 'util'] = cpu_phase_df.apply(compute_switch_phase_df, axis='columns')
```
douglas-raillard-arm, line 1017: It looks like apply() should be called on `cpu_phase_df.loc[switch_loc]`, otherwise we do the (very costly) computation on the whole dataframe, and then throw away most of it.

lisa/tests/scheduler/load_tracking.py
```diff
@@ -975,7 +975,7 @@ def get_trace_cpu_util(self):
             # Ignore the first quarter of the util signal of each phase, since
             # it's impacted by the phase change, and util can be affected
             # (rtapp does some bookkeeping at the beginning of phases)
-            start += duration / 4
+            # start += duration / 4
```
douglas-raillard-arm, line 978: there were some issues (as per the comment) with the beginning, so it's quite likely that the problem now is just that we "cut" the signal in the middle of an activation, leading to having the 2nd half of an activation without its beginning. This would increase the average slightly.

If that's the problem, we can "cut" the signal at the first activation past the 4th of the duration. This way we will not have a half-activation.

lisa/tests/scheduler/load_tracking.py
```diff
@@ -958,6 +957,15 @@ def get_trace_cpu_util(self):
         migr_task = sorted(task for task in tasks if task.startswith('migr'))[0]
         migr_task = self.rtapp_task_ids_map[migr_task][0]
 
+        df_sw = self.trace.df_event('sched_switch')
+        df_sw = df_sw[df_sw['next_comm'].str.startswith('swapper')]
+
+        df_state = {}
+
+        for task in tasks:
```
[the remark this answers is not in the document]
deggeman: OK.

`-        for task in tasks:
+        for task in self.rtapp_task_ids_map.keys():
`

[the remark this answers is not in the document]
deggeman: OK.

```
-        for task in tasks:
+        for task in self.rtapp_task_ids_map.keys():
```

lisa/tests/scheduler/load_tracking.py
```diff
@@ -958,6 +957,15 @@ def get_trace_cpu_util(self):
...
+            comm = self.rtapp_task_ids_map[task][0].comm
+            df_state[task] = df_sw[(df_sw['prev_comm'] == comm) & (df_sw['prev_state'] == TaskState.TASK_INTERRUPTIBLE)]
```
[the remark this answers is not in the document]
deggeman: The only difference is that df_task_activation() also contains events when a test task is switched out for another test task, and not only for the idle task. I thought this will mess up the calculation in lisa.pelt.kernel_util_mean() but it looks like this isn't the case. Tested only on one particular run on Juno-r0 so far.

```
-        df_sw = self.trace.df_event('sched_switch')
-        df_sw = df_sw[df_sw['next_comm'].str.startswith('swapper')]
-
         df_state = {}
 
         for task in self.rtapp_task_ids_map.keys():
             comm = self.rtapp_task_ids_map[task][0].comm
-            df_state[task] = df_sw[(df_sw['prev_comm'] == comm) & (df_sw['prev_state'] == TaskState.TASK_INTERRUPTIBLE)]
+            df_state[task] = self.trace.analysis.tasks.df_task_activation(comm)
+            df_state[task] = df_state[task][df_state[task]['active'] == 0]
```

lisa/tests/scheduler/load_tracking.py
```diff
@@ -971,7 +979,44 @@ def get_trace_cpu_util(self):
             phase_df = df_window(df, (start, end), method='pre', clip_window=True)
 
             for cpu in self.cpus:
-                util = phase_df[phase_df['cpu'] == cpu]['util']
+
+                cpu_phase_df = phase_df[phase_df['cpu'] == cpu].copy()
+
+                for task in tasks:
```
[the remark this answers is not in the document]
deggeman: OK.

```
-                for task in tasks:
+                for task, wlgen_task in self.rtapp_profile.items():
```

lisa/tests/scheduler/load_tracking.py
```diff
@@ -971,7 +979,44 @@ def get_trace_cpu_util(self):
...
+                    if self.rtapp_profile[task].phases[phase-1].cpus.count(cpu):
```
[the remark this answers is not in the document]
deggeman: OK.

```
-                for task in tasks:
-                    if self.rtapp_profile[task].phases[phase-1].cpus.count(cpu):
+                for task, wlgen_task in self.rtapp_profile.items():
+                    if cpu in wlgen_task.phases[phase-1].cpus:
```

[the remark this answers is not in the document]
deggeman: Should I raise a runtime error in case phase = row.phase is not greater than 1?

```
         cpu_util = {}
         for row in self.trace.analysis.rta.df_phases(migr_task).itertuples():
             phase = row.phase
+
+            if phase < 1:
+                raise RuntimeError('phase has to be greater than 0')
+
             duration = row.duration
             start = row.Index
             end = start + duration
```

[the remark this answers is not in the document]
douglas-raillard-arm: Why do we need to look at the previous phase rather than the current one ?

lisa/tests/scheduler/load_tracking.py
deggeman: OK.

```
-                cpu_phase_df = cpu_phase_df[cpu_phase_df['cpu'].notna()]
-                cpu_phase_df = cpu_phase_df[cpu_phase_df['delta'].notna()]
+                cpu_phase_df = cpu_phase_df.dropna(subset=['cpu', 'delta'])
```

lisa/tests/scheduler/load_tracking.py
deggeman: OK.

```
-                cpu_phase_df.loc[switch_loc, 'util'] = cpu_phase_df.apply(compute_switch_phase_df, axis='columns')
+                cpu_phase_df.loc[switch_loc, 'util'] = cpu_phase_df.loc[switch_loc].apply(compute_switch_phase_df, axis='columns')
```

lisa/tests/scheduler/load_tracking.py
deggeman: OK, something like this?

```
diff --git a/lisa/datautils.py b/lisa/datautils.py
index a430bff53952..a47daf8794ee 100644
--- a/lisa/datautils.py
+++ b/lisa/datautils.py
@@ -401,6 +401,8 @@ from lisa.utils import TASK_COMM_MAX_LEN, groupby, deprecate
 
     return functools.reduce(merge, df_list)
 
+def df_extend_index(df1, df2):
+    return df1.join(df2[[]], how='outer')
 
 def _resolve_x(y, x):
     """
index 8787f2f06ae2..c94d159a69b0 100644
@@ -29,7 +29,7 @@ from lisa.tests.base import (
 )
 from lisa.target import Target
 from lisa.utils import ArtifactPath, groupby, ExekallTaggable
-from lisa.datautils import series_mean, df_window, df_filter_task_ids, series_refit_index, df_split_signals, df_refit_index, df_add_delta
+from lisa.datautils import series_mean, df_window, df_filter_task_ids, series_refit_index, df_split_signals, df_refit_index, df_add_delta, df_extend_index
 from lisa.wlgen.rta import RTA, Periodic, RTATask
 from lisa.trace import FtraceCollector, requires_events, may_use_events, MissingTraceEventError
 from lisa.analysis.load_tracking import LoadTrackingAnalysis
@@ -987,7 +987,7 @@ Time in seconds for util_avg to converge (i.e. ignored time)
                 for task, wlgen_task in self.rtapp_profile.items():
                     if cpu in wlgen_task.phases[phase-1].cpus:
                         df_state_clipped = df_window(df_state[task], (start, end), method='pre', clip_window=True)
-                        cpu_phase_df = cpu_phase_df.join(df_state_clipped[[]], how='outer')
+                        cpu_phase_df = df_extend_index(cpu_phase_df, df_state_clipped)
 
                 cpu_phase_df['cpu'].fillna(method='ffill', inplace=True)
                 cpu_phase_df = df_add_delta(cpu_phase_df)
```

deggeman:
Related to the pelt_interpolate() implementation:

Currently, I'm not parsing the clock, put cpu_capacity as a third parameter.  

```
diff --git a/lisa/pelt.py b/lisa/pelt.py
index 8a30146c077e..5cadf3eeb1a4 100644
--- a/lisa/pelt.py
+++ b/lisa/pelt.py
@@ -20,7 +20,7 @@ import math
 import pandas as pd
 import numpy as np
 
-from lisa.datautils import series_envelope_mean
+from lisa.datautils import series_envelope_mean, df_add_delta, df_extend_index
 
 PELT_WINDOW = 1024 * 1024 * 1e-9
 """
@@ -266,4 +266,34 @@ PELT_SCALE = 1024
     """
     return series_envelope_mean(util)
 
+def pelt_interpolate(util, interpolate_at, cpu_capacity):
+    util = df_extend_index(util, interpolate_at)
+
+    util['cpu'].fillna(method='ffill', inplace=True)
+    util = df_add_delta(util)
+    util['delta'] = util['delta'].shift()
+    util['prev_util'] = util['util'].shift()
+
+    util = util.dropna(subset=['cpu', 'delta'])
+
+    def compute_switch_phase_df(row):
+        timestamp = row.name
+        # Transform ftrace timestamp to PELT clock time.
+        # Only CPU capacity (and not frequency) is considered
+        # here. It has the bigger impact.
+        delta = row.delta * cpu_capacity/PELT_SCALE
+        last_update = timestamp - delta
+        prev_util = row.prev_util
+        window_shrink = 1e3
+        activations = pd.Series([1, 1], index=[last_update, timestamp])
+        simulated_phase_df = simulate_pelt(activations, init=prev_util,
+                                           window=PELT_WINDOW/window_shrink,
+                                           half_life=PELT_HALF_LIFE*window_shrink)
+        return simulated_phase_df.iloc[-1]
+
+    switch_loc = util['util'].isnull()
+    util.loc[switch_loc, 'util'] = util.loc[switch_loc].apply(compute_switch_phase_df, axis='columns')
+
+    return util['util']
+
 # vim :set tabstop=4 shiftwidth=4 textwidth=80 expandtab
index c94d159a69b0..5afaa3fc3c60 100644
@@ -29,14 +29,14 @@ from lisa.tests.base import (
 )
 from lisa.target import Target
 from lisa.utils import ArtifactPath, groupby, ExekallTaggable
-from lisa.datautils import series_mean, df_window, df_filter_task_ids, series_refit_index, df_split_signals, df_refit_index, df_add_delta, df_extend_index
+from lisa.datautils import series_mean, df_window, df_filter_task_ids, series_refit_index, df_split_signals, df_refit_index
 from lisa.wlgen.rta import RTA, Periodic, RTATask
 from lisa.trace import FtraceCollector, requires_events, may_use_events, MissingTraceEventError
 from lisa.analysis.load_tracking import LoadTrackingAnalysis
 from lisa.analysis.tasks import TaskState, TasksAnalysis
 from lisa.analysis.rta import RTAEventsAnalysis
 from lisa.analysis.frequency import FrequencyAnalysis
-from lisa.pelt import PELT_WINDOW, PELT_HALF_LIFE, PELT_SCALE, simulate_pelt, pelt_settling_time, kernel_util_mean
+from lisa.pelt import PELT_SCALE, simulate_pelt, pelt_settling_time, kernel_util_mean, pelt_interpolate
 
 UTIL_SCALE = PELT_SCALE
 
@@ -987,37 +987,9 @@ Time in seconds for util_avg to converge (i.e. ignored time)
                 for task, wlgen_task in self.rtapp_profile.items():
                     if cpu in wlgen_task.phases[phase-1].cpus:
                         df_state_clipped = df_window(df_state[task], (start, end), method='pre', clip_window=True)
-                        cpu_phase_df = df_extend_index(cpu_phase_df, df_state_clipped)
-
-                cpu_phase_df['cpu'].fillna(method='ffill', inplace=True)
-                cpu_phase_df = df_add_delta(cpu_phase_df)
-                cpu_phase_df['delta'] = cpu_phase_df['delta'].shift()
-                cpu_phase_df['prev_util'] = cpu_phase_df['util'].shift()
-
-                cpu_phase_df = cpu_phase_df.dropna(subset=['cpu', 'delta'])
 
                 cpu_capacity = self.plat_info['cpu-capacities']['rtapp'][cpu]
-
-                def compute_switch_phase_df(row):
-                    timestamp = row.name
-                    # Transform ftrace timestamp to PELT clock time.
-                    # Only CPU capacity (and not frequency) is considered
-                    # here. It has the bigger impact.
-                    delta = row.delta * cpu_capacity/PELT_SCALE
-                    last_update = timestamp - delta
-                    prev_util = row.prev_util
-                    window_shrink = 1e3
-                    activations = pd.Series([1, 1], index=[last_update, timestamp])
-                    simulated_phase_df = simulate_pelt(activations, init=prev_util,
-                                                       window=PELT_WINDOW/window_shrink,
-                                                       half_life=PELT_HALF_LIFE*window_shrink)
-                    util = simulated_phase_df.iloc[-1]
-                    return util
-
-                switch_loc = cpu_phase_df['util'].isnull()
-                cpu_phase_df.loc[switch_loc, 'util'] = cpu_phase_df.loc[switch_loc].apply(compute_switch_phase_df, axis='columns')
-
-                util = cpu_phase_df['util']
+                util = pelt_interpolate(cpu_phase_df, df_state_clipped, cpu_capacity)
                 cpu_util.setdefault(cpu, {})[phase] = kernel_util_mean(util, plat_info=self.plat_info)
 
         return cpu_util
```

lisa/tests/scheduler/load_tracking.py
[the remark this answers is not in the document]
deggeman: phase = row.phase gives me 1 or 2

wlgen_task.phases is a list with two members:

(Pdb) wlgen_task.phases[0].cpus
[0]
(Pdb) wlgen_task.phases[1].cpus
[3]
(Pdb) wlgen_task.phases[2].cpus
*** IndexError: list index out of range

So it's just to handle this transition from {1,2} to {0,1}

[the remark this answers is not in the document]
douglas-raillard-arm: ah right, phases are not named which leads to problems with the buffer phase. There has been plans for more than a year to rework wlgen to fix that but lack of manpower on that front has delayed it.

For now, just skip when `phase == 0` with that comment to make it easy to remove the hack when wlgen is fixed. Looking at the CPU of the previous phase is not what we want to do:
```
            # TODO: remove that once we have named phases to skip the buffer phase
```

lisa/tests/scheduler/load_tracking.py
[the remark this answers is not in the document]
douglas-raillard-arm: > also contains events when a test task is switched out for another test task, and not only for the idle task

It does, there is a `preempted_value` parameter you can set to control what value you will get in this case. By default, it is NaN.

The snippet be turned into this to make it more obvious what is loop-local state and what is the actual content of the df_state is:
```
def filter_state(task_id):
	df = self.trace.analysis.tasks.df_task_activation(task_id)
	return df[df['active'] == 0]

df_state[task] = {
	task: filter_state(task_ids[0])
	for task, task_ids in self.rtapp_task_ids_map.items()
}
```

Also as a general rule, if a TaskID is available, it should always be used instead of a part of it (comm or PID), with `df_filter_task_ids()` if necessary:
https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/misc_utilities.html#lisa.datautils.df_filter_task_ids

lisa/tests/scheduler/load_tracking.py
douglas-raillard-arm: It would be good to have a function that works both on Series and DataFrame, you can take example on `_data_window()`/`df_window()`/`series_window()` in `lisa.datautils` on how it's done and documented:
* make an internal `_data_extend_index()` that works on both Series and DataFrame
* make a public `df_extend_index()` and `series_extend_index()` decorated with `DataFrameAccessor.register_accessor`/`SeriesAccessor.register_accessor`
* Document one of the public function and just refer to it in the docstring of the other one.

DataFrame and Series method are always separated in the public API to be able to add parameters to deal with pandas type-specific quirks if necessary (they can be very nasty and only show up in some conditions not obvious at the time of writing).

`_data_extend_index()` should operate on pandas Index types and the original object reindexed with the new merged index with `data.reindex()`:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.reindex.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.reindex.html

douglas-raillard-arm:
> ```+def pelt_interpolate(util, interpolate_at, cpu_capacity):```

`util` should be a Series, not a dataframe, so the function becomes simpler and more generically applicable. It's the caller's responsibility to provide a pelt clock that goes with the entity represented by `util`.
For example, this function does not need to know what CPU the task is placed on, because that information feeds from the PELT clock, which is computed separately with something like that:
```
def simulate_pelt_clock(capacity, clock=None, scale=PELT_SCALE):
    """
    Simulate a PELT clock of an entity from the capacities of the CPU it's
    residing on.

    :param capacity: CPU capacity over time.
    :type capacity: pandas.Series

    :param clock: If provided, a series of timestamps at which the clock is to
        be observed. The returned :class:`pandas.Series` will provide the
        simulated clock values at these instants.
    :type clock: pandas.Series or None

    :param scale: Maximum value allowed for CPU capacity.
    :type scale: float
    """
    user_clock = clock is not None
    if user_clock:
        # Ensures the clock's index is the same as the clock
        clock = clock.copy(deep=False)
        clock.index = clock
    else:
        clock = capacity.index

    df = pd.DataFrame(
        dict(
            clock=clock,
            capacity=capacity,
        ),
    )
    # Needed for "time" interpolation
    df.index = pd.TimedeltaIndex(df.index, unit='s')

    # Time flows linearly between 2 samples of the clock
    df['clock'].interpolate(method='time', inplace=True)
    # Shift so that the capacity is aligned with the corresponding delta
    df['capacity'] = df['capacity'].fillna(method='ffill').shift()
    df['delta'] = df['clock'].diff()

    # Scale each delta independantly
    df['delta'] *= df['capacity'] / scale

    # Fill the NaN with the initial value for the cumsum() fold
    df['delta'].iat[0] = df['clock'].iloc[0]
    # Reverse df_add_delta() now that we scaled each delta
    df['new_clock'] = df['delta'].cumsum()

    # Back to Float64Index
    df.index = df.index.total_seconds()

    if user_clock:
        df = df.loc[clock]
    return df['new_clock']
```

lisa/tests/scheduler/load_tracking.py
deggeman: OK, will change this in the new version.

lisa/tests/scheduler/load_tracking.py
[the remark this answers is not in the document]
deggeman: I switched to use self.trace.analysis.tasks.df_task_activation in the new version.

lisa/tests/scheduler/load_tracking.py
deggeman: We ran the last 2 EAS mainline integrations with this patch. I haven't seen issues related to this patch in the other testcases which use this function. I kept the patch in the new version.

douglas-raillard-arm: >     Turns out that CPUMigrationBase.get_trace_cpu_util() advances 'start' of the window by 25% and CPUMigrationBase.get_expected_cpu_util() does not.

It looks like the other function is supposed to do but the code is also commented for unclear reasons, maybe by mistake. Have you tested removing the first 25% on both sides as well ?

lisa/tests/scheduler/load_tracking.py
```diff
@@ -973,7 +998,20 @@ def get_trace_cpu_util(self):
             phase_df = df_window(df, (start, end), method='pre', clip_window=True)
 
             for cpu in self.cpus:
-                util = phase_df[phase_df['cpu'] == cpu]['util']
+
+                cpu_phase_df = phase_df[phase_df['cpu'] == cpu].copy()
+
+                df_state_clipped = pd.concat(df_window(df_state[task], (start, end), method='pre', clip_window=True)
+                                             for task, wlgen_task in self.rtapp_profile.items()
+                                             if cpu in wlgen_task.phases[phase-1].cpus)
```
douglas-raillard-arm, line 1006: Why are we inspecting the CPUs of the previous phase ?

deggeman: phase = 1 and phase = 2 are the two testphases which correspond to wlgen_task.phases[0] and wlgen_task.phases[1].

phase = 0 is the buffer phase which is visible in testdata on some boards (sd845 ?)

I added this:

# TODO: remove that once we have named phases to skip the buffer phase.
if phase == 0:
    continue

to skip over the buffer phase.

douglas-raillard-arm: Right, we need to re-align the number we get from rt-app to look it up inside the profile indeed

lisa/tests/scheduler/load_tracking.py
deggeman: No I haven't. We could just go one way and see with the next integration.

Issue #1455, opened 2020-07-02
WIP: Tracking fixes for Exoplayer & Jankbench failures upstream.


credp:
DO NOT MERGE

Exoplayer and Jankbench workloads fail on a clean install with Ubuntu 20.04 and Python 3.8.

The exoplayer issue has been fixed upstream in devlib in https://github.com/ARM-software/devlib/pull/495
I sent a PR to wa for the other issue in https://github.com/ARM-software/workload-automation/pull/1116

Not sure how to track this kind of dependency, but I think we just do an external/ update once fixed. Devlib could happen now, I guess.

credp:
All changes upstream, PR ready.

PR #1457, opened 2020-07-03
Subtree update 121603072020

credp:
[WIP]

This updates the subtrees to latest devlib and wa as mentioned in https://github.com/ARM-software/lisa/issues/1455 now that the changes are in.

There are other changes as well, so I'd like to run some tests before merging.


credp:
OK, I triggered a smoke test on Jenkins and did not see any devlib problems in executing across TC2, Juno (both), H960 and db845. Results look normal.
I also ran this branch with wltests across exoplayer, jankbench and pcmark suites for just a single kernel and obtained results.

PR #1489, opened 2020-09-28
lisa.tests.scheduler.uclamp

vdonnefort:
This new test intends to verify basic support for UClamp. A series of 8 phases
sets a UClamp min value for a task whom duty cycle would create a lower
utilization. The test will make sure the actual capacity, allocated for this
task, matches the UClamp min value.

### DIFF
--- a/lisa/tests/staging/utilclamp.py
+++ b/lisa/tests/staging/utilclamp.py
@@ -1,0 +1,373 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2020, Arm Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import functools
+from operator import itemgetter
+
+import numpy as np
+import os
+import pandas as pd
+
+from lisa.analysis.frequency import FrequencyAnalysis
+from lisa.analysis.load_tracking import LoadTrackingAnalysis
+from lisa.datautils import df_add_delta, series_mean, df_window
+from lisa.pelt import PELT_SCALE
+from lisa.tests.base import ResultBundle, RTATestBundle, TestMetric
+from lisa.wlgen.rta import Periodic
+
+
+class UtilClamp(RTATestBundle):
+    """
+    Validate that UtilClamp min values are honoured properly by the kernel.
+
+    The test is split into 8 phases. For each phase, a UtilClamp value is set
+    for a task, whose duty cycle would generate a lower utilization. Then the
+    actual capacity, allocated to the task during its activation is checked.
+
+    The 8 phases UtilClamp values are picked to cover the entire SoC's CPU
+    scale. (usually between 0 and 1024)
+
+    .. code-block:: text
+
+                   |<-- band 0 -->|<-- band 1 -->|<-- band 2 -->|<-- ...
+      capacities:  0      |      128     |      256            512
+                          |              |
+      --------------------|--------------|-------------------------------
+      phase 1:       uclamp_val          |
+                                         |
+      -----------------------------------|-------------------------------
+      phase 2:                       uclamp_val
+         ...
+
+      phase 8:
+
+    """
+
+    NR_PHASES = 8
+    CAPACITY_MARGIN = 0.8  # kernel task placement a 80% capacity margin
+
+    @classmethod
+    def _collect_capacities(cls, plat_info):
+        """
+        Returns, for each CPU a mapping frequency / capacity:
+
+        dict(cpu, dict(freq, capacity))
+
+        where capacity = max_cpu_capacity * freq / max_cpu_frequency.
+        """
+
+        max_capacities = plat_info['cpu-capacities']['rtapp']
+
+        return {
+            cpu: {
+                freq: int(max_capacities[cpu] * freq / max(freqs))
+                for freq in freqs
+            }
+            for cpu, freqs in plat_info['freqs'].items()
+        }
+
+    @classmethod
+    def _collect_capacities_flatten(cls, plat_info):
+        capacities = [
+            capa
+            for freq_capas in cls._collect_capacities(plat_info).values()
+            for capa in freq_capas.values()
+        ]
+
+        # Remove the duplicates from the list
+        return sorted(set(capacities))
+
+    @classmethod
+    def _get_bands(cls, capacities):
+        bands = list(zip(capacities, capacities[1:]))
+
+        # Only keep a number of bands
+        nr_bands = cls.NR_PHASES
+        if len(bands) > nr_bands:
+            # Pick the bands covering the widest range of util, since they
+            # are easier to test
+            bands = sorted(
+                bands,
+                key=lambda band: band[1] - band[0],
+                reverse=True
+            )
+            bands = bands[:nr_bands]
+            bands = sorted(bands, key=itemgetter(0))
+
+        return bands
+
+    @classmethod
+    def _get_phases(cls, plat_info):
+        """
+        Returns a list of phases. Each phase being described by a tuple:
+
+          (uclamp_val, util)
+        """
+
+        capacities = cls._collect_capacities_flatten(plat_info)
+        bands = cls._get_bands(capacities)
+
+        def band_mid(band):
+            return int((band[1] + band[0]) / 2)
+
+        return [
+            (
+                band_mid(band),
+                band_mid(band) / 2
+            )
+            for band in bands
+        ]
+
+    @classmethod
+    def get_rtapp_profile(cls, plat_info):
+        periods = [
+            Periodic(
+                duty_cycle_pct=(util / PELT_SCALE) * 100,  # util to pct
+                duration_s=5,
+                period_ms=cls.TASK_PERIOD_MS,
+                uclamp_min=uclamp_val,
+                uclamp_max=uclamp_val
+            )
+            for uclamp_val, util in cls._get_phases(plat_info)
+        ]
+
+        return {'task': functools.reduce(lambda a, b: a + b, periods)}
+
+    def _get_trace_df(self):
+        task = self.rtapp_task_ids_map['task'][0]
+
+        df = self.trace.analysis.tasks.df_task_activation(task)
+        df = df[['active', 'cpu']]
+
+        df_freq = self.trace.analysis.frequency.df_cpus_frequency()
+        df_freq = df_freq[['cpu', 'frequency']]
+        df_freq = df_freq.pivot(index=None, columns='cpu', values='frequency')
+        df_freq.reset_index(inplace=True)
+        df_freq.set_index('Time', inplace=True)
+
+        df = df.merge(df_freq, how='outer', left_index=True, right_index=True)
+
+        # Ensures that frequency values are propogated through the entire
+        # DataFrame, as it is possible that no frequency event occur
+        # during a phase.
+        for cpu in self.plat_info['cpu-capacities']['rtapp']:
+            df[cpu].ffill(inplace=True)
+
+        return df
+
+    def _get_phases_df(self):
+        task = self.rtapp_task_ids_map['task'][0]
+
+        df = self.trace.analysis.rta.df_phases(task).copy()
+        df.reset_index(inplace=True)
+        df.rename(columns={'index': 'start'}, inplace=True)
+        df['end'] = df['start'].shift(-1)
+
+        phases = [(0, 0, 0)] + self._get_phases(self.plat_info)
+        df['uclamp_val'] = df.apply(
+                lambda x: phases[int(x.phase)][0],
+                axis=1)
+
+        # TODO: To remove once we have named phases.
+        df = df[df.phase > 0]
+
+        return df
+
+    def _for_each_phase(self, callback):
+        df_phases = self._get_phases_df()
+        df_trace = self._get_trace_df()
+
+        def parse_phase(phase):
+            start = phase['start']
+            end = phase['end']
+            df = df_trace
+
+            # During a phase change, rt-app will wakeup and then change
+            # UtilClamp value will be changed. We then need to wait for the
+            # second wakeup for the kernel to apply the most recently set
+            # UtilClamp value.
+            start = df[(df.index >= start) &
+                       (df['active'] == 1)].first_valid_index()
+
+            end = end if not np.isnan(end) else df.last_valid_index()
+
+            if (start >= end):
+                raise ValueError('Phase ends before it has even started')
+
+            df = df_trace[start:end].copy()
+
+            return callback(df, phase)
+
+        return df_phases.apply(parse_phase, axis=1)
+
+    def _plot_phases(self, test, failures, signal=None):
+        task = self.rtapp_task_ids_map['task'][0]
+        ax = self.trace.analysis.tasks.plot_task_activation(task,
+                                                            which_cpu=True)
+        ax = self.trace.analysis.rta.plot_phases(task, axis=ax)
+        for failure in failures:
+            ax.axvline(failure, alpha=0.5, color='r')
+        if signal is not None:
+            signal.plot(ax=ax.twinx(), drawstyle='steps-post')
+        filepath = os.path.join(self.res_dir, 'utilclamp_{}.png'.format(test))
+        self.trace.analysis.rta.save_plot(ax.figure, filepath=filepath)
+
+        return ax
+
+    @FrequencyAnalysis.df_cpus_frequency.used_events
+    @LoadTrackingAnalysis.df_tasks_signal.used_events
+    def test_placement(self) -> ResultBundle:
+        """
+        For each phase, checks if the task placement is compatible with
+        UtilClamp requirements. This is done by comparing the maximum capacity
+        of the CPU on which the task has been placed, with the UtilClamp
+        value.
+        """
+
+        metrics = {}
+        test_failures = []
+        capacity_margin = self.CAPACITY_MARGIN
+        cpu_max_capacities = self.plat_info['cpu-capacities']['rtapp']
+
+        def parse_phase(df, phase):
+            uclamp_val = phase['uclamp_val']
+            num_activations = df['active'][df['active'] == 1].count()
+            cpus = set(df.cpu.dropna().unique())
+            fitting_cpus = set([cpu for cpu, cap in cpu_max_capacities.items()
+                                if (cap == PELT_SCALE) or
+                                (cap * capacity_margin) > uclamp_val])
+
+            failures = df[(
+                df['active'] == 1) & (df['cpu'].isin(cpus - fitting_cpus))
+            ].index.tolist()
+            num_failures = len(failures)
+            test_failures.extend(failures)
+
+            phase_str = 'Phase-{}'.format(phase['phase'])
+            metrics[phase_str] = {
+                'uclamp-min': TestMetric(uclamp_val),
+                'cpu-placements': TestMetric(cpus),
+                'expected-cpus': TestMetric(fitting_cpus),
+                'bad-activations': TestMetric(
+                    num_failures * 100 / num_activations, "%"),
+            }
+
+            return cpus.issubset(fitting_cpus)
+
+        res = ResultBundle.from_bool(self._for_each_phase(parse_phase).all())
+        res.add_metric('Phases', metrics)
+
+        self._plot_phases('test_placement', test_failures)
+
+        return res
+
+    @FrequencyAnalysis.df_cpus_frequency.used_events
+    @LoadTrackingAnalysis.df_tasks_signal.used_events
+    def test_freq_selection(self) -> ResultBundle:
+        """
+        For each phase, checks if the task placement and frequency selection
+        is compatible with UtilClamp requirements. This is done by comparing
+        the current CPU capacity on which the task has been placed, with the
+        UtilClamp value.
+
+        The expected capacity is the schedutil projected frequency selection
+        for the given uclamp value.
+        """
+
+        metrics = {}
+        test_failures = []
+        capacity_dfs = []
+        # (
+        #    # schedutil factor that converts util to a frequency for a
+        #    # given CPU:
+        #    #
+        #    #   next_freq = max_freq * C * util / max_cap
+        #    #
+        #    #   where C = 1.25
+        #    schedutil_factor,
+        #
+        #    # list of frequencies available for a given CPU.
+        #    frequencies,
+        # )
+        cpu_frequencies = {
+            cpu: (
+                (max(capacities) * (1 / self.CAPACITY_MARGIN)) / max(capacities.values()),
+                sorted(capacities)
+            )
+            for cpu, capacities in
+            self._collect_capacities(self.plat_info).items()
+        }
+        cpu_capacities = self._collect_capacities(self.plat_info)
+
+        def schedutil_map_util_cap(cpu, util):
+            """
+            Returns, for a given util on a given CPU, the capacity that
+            schedutil would select.
+            """
+
+            schedutil_factor, frequencies = cpu_frequencies[cpu]
+            schedutil_freq = schedutil_factor * util
+
+            # Find the first available freq that meet the schedutil freq
+            # requirement.
+            for freq in frequencies:
+                if freq >= schedutil_freq:
+                    break
+
+            return cpu_capacities[cpu][freq]
+
+        def parse_phase(df, phase):
+            uclamp_val = phase['uclamp_val']
+            num_activations = df['active'][df['active'] == 1].count()
+            expected = schedutil_map_util_cap(df['cpu'].unique()[0],
+                                              uclamp_val)
+
+            # Activations numbering
+            df['activation'] = df['active'].cumsum()
+
+            # Only keep the activations
+            df.ffill(inplace=True)
+            df = df[df['active'] == 1]
+
+            # Actual capacity at which the task is running
+            for cpu, freq_to_capa in cpu_capacities.items():
+                df[cpu] = df[cpu].map(freq_to_capa)
+            df['capacity'] = df.apply(lambda line: line[line.cpu], axis=1)
+
+            failures = df[df['capacity'] != expected]
+            num_failures = failures['activation'].nunique()
+
+            test_failures.extend(failures.index.tolist())
+            capacity_dfs.append(df[['capacity']])
+
+            phase_str = 'Phase-{}'.format(phase['phase'])
+            metrics[phase_str] = {
+                'uclamp-min': TestMetric(uclamp_val),
+                'expected-capacity': TestMetric(expected),
+                'bad-activations': TestMetric(
+                    num_failures * 100 / num_activations, "%"),
+            }
+
+            return failures.empty
+
+        res = ResultBundle.from_bool(self._for_each_phase(parse_phase).all())
+        res.add_metric('Phases', metrics)
+
+        self._plot_phases('test_frequency', test_failures,
+                          pd.concat(capacity_dfs))
+
+        return res
--- a/doc/kernel_tests.rst
+++ b/doc/kernel_tests.rst
@@ -299,3 +299,6 @@ actively evaluated.
 
 .. automodule:: lisa.tests.staging.numa_behaviour
    :members:
+
+.. automodule:: lisa.tests.staging.utilclamp
+   :members:


valschneider (commented):
Have some comments on the pure Python side of things; I'll try to grok the test itself next week.

lisa/tests/scheduler/uclamp.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @classmethod
+    def _collect_capacities(cls, plat_info):
+        max_capacities = plat_info['cpu-capacities']['rtapp']
+
+        # returns, for each CPU a mapping frequency / capacity.
+        # cpu:
+        #   freq: capacity
+        #
+        # where capacity = max_cpu_capacity * freq / max_cpu_frequency.
```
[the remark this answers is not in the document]
vdonnefort: As the function is an internal one, I thought it wasn't interesting for a user to know about it, hence this choice. But if you think this should be expose, I can change that.

lisa/tests/scheduler/uclamp.py
```diff
@@ -0,0 +1,209 @@
...
+                Periodic(
+                    duty_cycle_pct=(util / 1024) * 100,  # util to pct
+                    duration_s=5,
+                    period_ms=10,
+                    uclamp_min=uclamp_val,
+                    uclamp_max=uclamp_val)
+                for uclamp_val, util, expected in cls._get_phases(plat_info)
+        ]
+
+        return {'task': functools.reduce(lambda a, b: a + b, periods)}
```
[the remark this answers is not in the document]
vdonnefort: Can't recall where was the problem, but this sadly can't work:
`Unsupported operand type for +: 'int' and 'Periodic'`

lisa/tests/scheduler/uclamp.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import functools
+import itertools
```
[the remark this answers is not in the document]
vdonnefort: That was to anticipate your proposal for `_collect_capacities_flatten` 🙂

lisa/tests/scheduler/uclamp.py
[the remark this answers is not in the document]
vdonnefort: Done.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+
+        def _expected_minmax(row):
+            uclamp_val = row['uclamp_val']
+            expected = 1024
+
+            for cap in max_capacities:
+                # kernel task placement uses 80% margin on the capacity
+                if (cap * 0.8) > uclamp_val:
+                    expected = cap
+                    break
```
douglas-raillard-arm, line 258: nit: `for` loops have an optional `break` clause that can be thought as paired with an `if` statement breaking the loop:
```
            for cap in max_capacities:
                # kernel task placement uses 80% margin on the capacity
                if (cap * 0.8) > uclamp_val:
                    expected = cap
                    break
            else:
                expected = 1024
```

It's equivalent to that pseudo-code (there is no binding for `cap`):
```
            if any((cap * 0.8) > uclamp_val for cap in max_capacities):
                expected = cap
            else:
                expected = 1024
```

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+    def test_freq_selection(self) -> ResultBundle:
+        """
+        For each phase, checks if the task placement and frequency selection
+        match UtilClamp requirements.
+
+        The expected capacity is :
+          - min: The first capacity below the UtilClamp value.
+          - max: The 4th capacity above the UtilClamp value. This arbitrary
+            value must be high enough to cover the migration and schedutil
+            margins.
```
douglas-raillard-arm, line 277: Same as for the other test, we don't want arbitrary margins. What is are the issues exactly when we don't have such margin ? We need to know precisely, and then take them into account (schedutil is fairly easy to simulate since we have access to util_est).

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+
+        df = self._get_analysis()
+        capacities = self._collect_capacities_flatten(self.plat_info)
+
+        def _expected_minmax(row):
```
douglas-raillard-arm, line 283: ditto, no need for `_` prefix for local functions

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+import itertools
+
+import pandas as pd
+import numpy as np
+from devlib.utils.misc import memoized
+
+from lisa.analysis.frequency import FrequencyAnalysis
+from lisa.analysis.load_tracking import LoadTrackingAnalysis
+from lisa.datautils import df_add_delta
+from lisa.datautils import series_mean
```
douglas-raillard-arm, line 29: can be fused into one:
```

from lisa.datautils import df_add_delta, series_mean
```

douglas-raillard-arm (changes requested):
To go with the round number of comments:

>  “This must be Thursday,' said Arthur to himself, sinking low over his beer. 'I never could get the hang of Thursdays.”
― Douglas Adams, The Hitchhiker's Guide to the Galaxy

lisa/tests/scheduler/uclamp.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @classmethod
+    def _collect_capacities_flatten(cls, plat_info):
+        capacities = []
+
+        for cpu, states in cls._collect_capacities(plat_info).items():
+            for freq, cap in states.items():
+                capacities.append(cap)
+
+        return sorted(list(dict.fromkeys(capacities)))
```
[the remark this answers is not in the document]
douglas-raillard-arm: @valschneider I left a comment on the new version, I think it's better to go with a plain comprehension here, since I don't think it really brings anything to stack "shorthands" for the sake of saving a couple of lines (it's not like the problem was something cleanly solvable by let's say in pseudo code `sorted(map(itertools.chain.from_iterable^5, deeply_nested))`)

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+from lisa.pelt import PELT_SCALE
+from lisa.tests.base import ResultBundle, RTATestBundle
+from lisa.wlgen.rta import Periodic
+
+
+class UtilClamp(RTATestBundle):
+    """
+    Validate that UtilClamp min values are honoured properly by the kernel.
+
+    The test is split into 8 phases. For each phase, a UtilClamp value is set
```
douglas-raillard-arm, line 39: Adding a new comment since @valschneider 's one was marked as resolved and apparently cannot be resurrected:

> That smells like annoying refactoring, but this feels like something that should be a parameter rather than a hardcoded value.
> Annoyingly this controls the rtapp profile, so this can't be a test parameter... Perhaps a class attribute for now?

We should be able to add an optional parameter to `_from_target()` method like `UserspaceSanity`:
```
    @classmethod
    def _from_target(cls, target: Target, *,
		res_dir: ArtifactPath = None,
		nr_phases=None,
    ) -> 'UtilClamp':
```

And in `__init__(..., nr_phases=None)`, we replace `None` by 8. Hardcoding 8 as default value in different paths leads to missed updates when the value needs to change, it's much better to use a standard `None` "marker" and deal with that in a single place like:
```
self.nr_phases = 8 if nr_phases is None else nr_phases 
```

with the appropriate docstring.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+                for cpu in range(plat_info['cpus-count'])
+        }
+
+    @classmethod
+    def _collect_capacities_flatten(cls, plat_info):
+        capacities = itertools.chain.from_iterable(
+                      list(v.values()) for v in
+                      cls._collect_capacities(plat_info).values())
+
+        return sorted(dict.fromkeys(capacities))
```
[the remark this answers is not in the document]
vdonnefort: This is to remove the duplicates from the list.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+        capacities = self._collect_capacities(self.plat_info)
+        for cpu in self.plat_info['nrg-model'].cpu_nodes:
+            cpu_id = cpu.cpus[0]
+            df[cpu_id] = df[cpu_id].map(capacities[cpu_id])
+
+        df['capacity'] = df.apply(lambda x: x[x.cpu], axis=1)
+        df['capacity_max'] = df.apply(
+                lambda x: max(capacities[x.cpu].values()), axis=1)
+
+        return [round(series_mean(df['capacity'], x=df['Time'])),
```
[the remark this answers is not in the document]
vdonnefort: IIRC, this has to be a list so that we can use it in `apply` with the option `result_type='expand'`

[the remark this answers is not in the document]
vdonnefort: Ahem ... yes, but I must admit that's completely empirical though.

With all the margins we have in the kernel, even with the very large band I used, we sometimes slightly exceed the maximum value. Rounding here prevent that problem, as the difference is always less than 1.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,296 @@
...
+                round(series_mean(df['capacity_max'], x=df['Time']))]
```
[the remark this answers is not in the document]
vdonnefort: This is to give the test a chance to succeed if the task needs to down migrate.

lisa/tests/staging/utilclamp.py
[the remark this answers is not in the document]
douglas-raillard-arm: ok in this case you can either just use `sorted(set(xs))` or `lisa.utils.deduplicate` if you need more control (or to keep the initial order of the values)

lisa/tests/staging/utilclamp.py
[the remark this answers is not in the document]
douglas-raillard-arm: Can you give a go at tuples ? In my experience, `pandas` treats both in the same way (which is most of the time a large annoyance but in this case might be useful)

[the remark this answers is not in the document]
douglas-raillard-arm: If we only expect a difference of one, we should check for that explicitly at the spot the value is used. `round()` is the last thing we want to use for that, as it will not behave like "integer arithmetic rounding":
```
>>> round(0.4)
0
>>> round(0.6)
1
```

lisa/tests/staging/utilclamp.py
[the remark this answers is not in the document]
douglas-raillard-arm: If we expect migrations to happen within a phase, we should check for it and ignore what happens before, averaging is not something we want to end up doing for that purpose. In the past 2 years we (painfully) scrubbed LISA to eliminate as much of that sort of margin as possible because they are:
* never computed from a knowledge model, it's always empirical
* from the above follows:
    * no-one can understand them, because usually no-one ever did, including the original test writer. You might think you know why you added it, but the reality is that you could be subject to an rt-app bug or some CPU-invariance issue and you would not realize, unless you are digging to check a specific hypothesis.
    * they simply fail randomly on some devices, because they amounted to essentially hardcoding some characteristics that should be derived from PlatformInfo
* hiding real issues (both in the kernel and in the test)

The last point can escalate to a much worse issue: a test appears to be simple, and check an "obvious" property of a kernel's signal. Turns out the signal does not actually have the property to start with, so the test should actually be expected to fail, but it passed because margins were added to "accommodate" the situation (test_util_migration is a prime example of how a "high level" property about the sum of 2 util signals actually does not hold, and the real property is significantly more tricky to check).

So essentially, all new tests must be designed as:
* no margins (or only on last resort, for like 1 or 2% on an absolute scale)
* reimplement the kernel bit you want to test in Python
* expose the inputs and outputs of the kernel bit, and check against the Python implementation

When the test fails, we know exactly where, and it's much more straightforward to understand why. Also it works on all platforms, and when it does not, it's usually easy to understand why and the fix is basically just exposing some kernel state that was hidden.

lisa/tests/staging/utilclamp.py
vdonnefort: I don't think we can easily predict the schedutil selected frequency:
* util_est doesn't use the clamped value of util.
* the usage of the uclamp value for schedutil depends _when_ schedutil updates the frequency. (i.e. If it is updated while the task is enqueued or not)
* other things running on the board that would change the schedutil value, including bits that aren't frozen during the run, like IRQs.

Notice also, that this max value is more a sanity check that the solution to a UtilClamp value isn't just using the highest capacity available. Ultimately, we only test the min value of UClamp.

I can try to compute what would be the outcome of get_next_freq() for a given util, but that would mean doing many assumptions anyway.

douglas-raillard-arm: > util_est doesn't use the clamped value of util.

You can just use that:
https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/trace_analysis.html#lisa.analysis.load_tracking.LoadTrackingAnalysis.df_task_signal

> the usage of the uclamp value for schedutil depends when schedutil updates the frequency. (i.e. If it is updated while the task is enqueued or not)
We have `cpu_frequency` event to tell us when schedutil took a decision to change freq.

> other things running on the board that would change the schedutil value, including bits that aren't frozen during the run, like IRQs.
We have a couple of options:
* simulate schedutil behavior on each kind of util so we know what to expect for a given CFS task util (`total/1024 * max_freq` plus checking that no iowait boost or similar occured. If they did, we can just ignore the "unpredictable" sample) 
* make schedutil a bit more verbose on what it does and why.

The alternative is simply only check the placement with uclamp, and not the frequency selection. It might look artificial and "why don't we do it since we can ?" but in reality, if we cannot implement what I described, we cannot actually check what is happening. Tests relying on margins as a cheap way of "simulating" kernel subsystems have all turned into maintenance liability, and some of them have empirically been shown to not be able to even catch using the energy model description of another board than the one under test.

The only sort-of-exception to that are tests checking one-sided interval. If for example we should *never* end up selecting a frequency below a threshold (taking into account all the edge conditions), we can check for that, and this property should hold across all platforms.

lisa/tests/staging/utilclamp.py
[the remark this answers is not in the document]
vdonnefort: My bad, that seems to work well

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+    @FrequencyAnalysis.df_cpus_frequency.used_events
+    @LoadTrackingAnalysis.df_tasks_signal.used_events
+    def test_freq_selection(self) -> ResultBundle:
+        """
+        For each phase, checks if the task placement and frequency selection
+        is compatible with UtilClamp requirements. This is done by comparing
+        the current CPU capacity on which the task has been placed, with the
+        UtilClamp value.
+
+        The expected capacity is :
```
douglas-raillard-arm, line 275: Extra space before `:`
Also since there is no blank line after `:`, this becomes a definition list, rather than just a bullet list:
https://docutils.sourceforge.io/docs/user/rst/quickref.html#definition-lists

The expected formatting is:
image
But the one we get from the definition list is:
image


Should be reformatted as:
```
        The expected capacity is:

          - min: The first capacity below the UtilClamp value.
          - max: The estimated selected schedutil capacity.
```

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+        cpu_capacities = self._collect_capacities(self.plat_info)
+
+        def schedutil_map_util_cap(cpu, util):
+            """
+            Returns, for a given util on a given CPU, the capacity that
+            schedutil would select.
+            """
+            max_freq = max(cpu_capacities[cpu])
+            max_cap = max(cpu_capacities[cpu].values())
+            schedutil_freq = (max_freq + (max_freq >> 2)) * util / max_cap
```
douglas-raillard-arm, line 291: We can use readable code here (and tbf, the kernel can as well: https://godbolt.org/z/38v8aW ...):
```
			 schedutil_freq = max_freq * (1/self.CAPACITY_MARGIN) * util / max_cap
```

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+
+            # Find the first available freq that meet the schedutil freq
+            # requirement.
+            for freq in sorted(cpu_capacities[cpu]):
```
douglas-raillard-arm, line 295: The call to `sorted()` should be done outside of the closure so we don't sort the iterable again at each row

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+        df = self._get_test_df()
+        capacities = self._collect_capacities_flatten(self.plat_info)
+        cpu_capacities = self._collect_capacities(self.plat_info)
+
+        def schedutil_map_util_cap(cpu, util):
+            """
+            Returns, for a given util on a given CPU, the capacity that
+            schedutil would select.
+            """
+            max_freq = max(cpu_capacities[cpu])
```
douglas-raillard-arm, line 289: let's have `capacities = cpu_capacities[cpu]` and reuse it, a lookup is not the end of the world but that function will be called a very large number of time, so we might as well save some (as well as some typing)

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+            for freq in sorted(cpu_capacities[cpu]):
+                if freq >= schedutil_freq:
+                    break
+
+            return cpu_capacities[cpu][freq]
+
+        def expected_minmax(row):
+            uclamp_val = row['uclamp_val']
+            band_min = row['band'][0]
+            low_caps = capacities[:capacities.index(band_min) + 1]
```
douglas-raillard-arm, line 304: Not sure what was the intent of:
```
            low_caps = capacities[:capacities.index(band_min) + 1]
            low_caps[-1]
```
Since AFAICT this is equivalent to:
```
band_min
```

lisa/tests/staging/utilclamp.py
douglas-raillard-arm, line 289: All the values that do not depend on `cpu` or `util` should be computed ahead of time in the caller, since the function is going to be called a very large amount of time. We can have a dict that gives `(max_freq, max_cap)` for each CPU:
```
cpu_maxes = {
    cpu: (max(freq_capa.keys()), max(freq_capa.values()), sorted(freq_capa.keys()))
    for cpu, freq_capa in cpu_capacities.items()
}
```
So we can then just use it as:
```
            max_freq, max_cap, freqs = cpu_maxes[cpu]
            ...
            for freq in freqs:
                if freq >= schedutil_freq:
                    break
            ...
```

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+        df['capacity_max'] = df.apply(
+                                lambda line:
+                                max(capacities[line.cpu].values()),
+                                axis=1
+                            )
+
+        df.set_index('Time', inplace=True)
+
+        return (
+                series_mean(df['capacity']),
```
douglas-raillard-arm, line 198: too much indentation

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+        df['Time'] = df['Time'].cumsum()
+
+        # Frequency to capacity
+        capacities = self._collect_capacities(self.plat_info)
+        for cpu, freq_to_capa in capacities.items():
+            df[cpu] = df[cpu].map(freq_to_capa)
+
+        df['capacity'] = df.apply(lambda line: line[line.cpu], axis=1)
+        df['capacity_max'] = df.apply(
+                                lambda line:
```
douglas-raillard-arm, line 190: too much indentation. Once the unneeded indentation is removed, the lambda fits on its own line:
```
        df['capacity_max'] = df.apply(
            lambda line: max(capacities[line.cpu].values()),
            axis=1
        )
```

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+        df_freq = df_freq.pivot(index=None,
+                                columns='cpu',
+                                values='frequency').reset_index()
+        df_freq.set_index('Time', inplace=True)
+        df = df.merge(df_freq, how='outer', left_index=True, right_index=True)
+
+        end = end if not np.isnan(end) else df.last_valid_index()
+
+        # Skip the first half to skip any task migration
+        start += (end - start) / 2
```
douglas-raillard-arm, line 165: are we sure it's enough ? Since we got the activation dataframe, we can easily check if there is any migration in the window, and if there is any, set the start to the first activation after it.

```
df = self.trace.analysis.tasks.df_task_activation(task)

# First row after each migration
df = df.copy(deep=False)
df['migration'] = df['cpu'] != df['cpu'].shift()
start_after_migration = df[df['migration']].first_valid_index()

if start_after_migration is not None:
	start = start_after_migration
```
Note the shallow copy of the dataframe since we don't modify existing columns but want to add a new one.

If for some reason we expect multiple migrations, we can take the `last_valid_index()`.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+          (uclamp_val, util, (band))
+        """
+        capacities = cls._collect_capacities_flatten(plat_info)
+        bands = cls._get_bands(capacities)
+
+        def band_mid(band):
+            return int((band[1] + band[0]) / 2)
+
+        return [
+                (
```
douglas-raillard-arm, line 127: unneeded extra indentation

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+        if (start >= end):
+            raise ValueError('Phase ends before it has even started')
+
+        df = df.ffill()
+        df = df_window(df, (start, end), method='exclusive')
+
+        # Fake Time, representing only the task activation period
+        df_add_delta(df, col='Time', inplace=True)
+        df = df.dropna()
+        df = df[df.active == 1]
```
douglas-raillard-arm, line 180: s/df.active/df.['active']/

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+                                max(capacities[line.cpu].values()),
+                                axis=1
+                            )
+
+        df.set_index('Time', inplace=True)
+
+        return (
+                series_mean(df['capacity']),
+                df['capacity_max'].max(),
+                df.cpu.unique()[-1]
```
douglas-raillard-arm, line 200: s/df.cpu/df['cpu']/

douglas-raillard-arm (changes requested):
Small bits here and there, other than that LGTM and very close to merge

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+    @LoadTrackingAnalysis.df_tasks_signal.used_events
+    def test_placement(self) -> ResultBundle:
+        """
+        For each phase, checks if the task placement is compatible with
+        UtilClamp requirements. This is done by comparing the maximum capacity
+        of the CPU on which the task has been placed, with the UtilClamp
+        value.
+        """
+
+        df = self._get_test_df()
```
valschneider, line 248: I'm not fond of what `_get_task_activation_mean_capacities()` does under the hood. We don't need to compute any of the capacity means here, do we? Couldn't we simplify this test to something like:

```
task = self.rtapp_task_ids_map['task'][0]
df_activations = self.trace.analysis.tasks.df_task_activation(task)

for start, end, uclamp_val in XXX:
    df = df_activations[start:end]['cpu']
    fitting_cpus = [cpu for cpu, cap in self.plat_info['cpu-capacities']['orig'].items() if cap > uclamp_val * 1.25]
    df_notfit = df[~df['cpu'].isin(fitting_cpus)]
    if df_notfit.any():
        # add metrics, store e.g. first timestamp failure, maybe event plot them

```

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+
+        dict(cpu, dict(freq, capacity))
+
+        where capacity = max_cpu_capacity * freq / max_cpu_frequency.
+        """
+
+        max_capacities = plat_info['cpu-capacities']['rtapp']
+
+        return {
+                cpu: {
```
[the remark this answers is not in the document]
vdonnefort: done

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+
+        df = self._get_test_df()
+        capacity_margin = self.CAPACITY_MARGIN
+        max_capacities = sorted(set(
+            self.plat_info['cpu-capacities']['rtapp'].values()))
+
+        def expected_minmax(row):
+            uclamp_val = row['uclamp_val']
+            expected = 1024
```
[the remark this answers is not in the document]
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: That code is gone

lisa/tests/staging/utilclamp.py
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: that code is gone

lisa/tests/staging/utilclamp.py
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: After many discussions we finally went for the "let's not slice the phase at all" solution.

lisa/tests/staging/utilclamp.py
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: Couldn't reuse all the proposal here, because I need some common function that slice the phases. However I reused the fitting_cpus list idea. With a set, that's even simpler than it used to be, to check if actual CPUs are a subset of the expected and if not, what are the wrong ones.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+        return res
+
+    def _plot_phases(self, test, failures, signal=None):
+        task = self.rtapp_task_ids_map['task'][0]
+        ax = self.trace.analysis.tasks.plot_task_activation(task,
+                                                            which_cpu=True)
+        ax = self.trace.analysis.rta.plot_phases(task, axis=ax)
+        for failure in failures:
+            ax.axvline(failure, alpha=0.5, color='r')
+        if isinstance(signal, pd.DataFrame):
```
douglas-raillard-arm, line 230: The idiom is to check for None:

```
if signal is not None
```

Which is easier to spot and will avoid swallowing the input silently instead of:
* failing loudly with some exception if it gets passed wrongfully e.g. an int (we want exceptions in these cases, since there is no "compile-time" check for that)
* Simply working with duck-typing, e.g. with a Series since Series.plot() is essentially the same as DataFrame.plot()

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            signal.plot(ax=ax.twinx(), drawstyle='steps-post')
+        filepath = os.path.join(self.res_dir, 'utilclamp_{}.png'.format(test))
+        self.trace.analysis.rta.save_plot(ax.figure, filepath=filepath)
```
douglas-raillard-arm, line 233: adding a ``return ax`` is a good idea if someone needs a reference to the axis in a debugging notebook

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            df['activation'] = df['active'].cumsum()
+
+            # Only keep the activations
+            df.ffill(inplace=True)
+            df = df[df['active'] == 1]
+
+            # Actual capacity at which the task is running
+            for cpu, freq_to_capa in cpu_capacities.items():
+                df[cpu] = df[cpu].map(freq_to_capa)
+                df['capacity'] = df.apply(lambda line: line[line.cpu], axis=1)
```
douglas-raillard-arm, line 354: There seems to be something fishy here:
* we keep overwriting `df['capacity']` inside the loop
* we then use it after the loop is finished

Just taking it out of the loop should be enough, since it only relies on the columns labeled by CPU IDs, which are updated independently at each loop iteration.

douglas-raillard-arm (changes requested):
A few nits and some fishy business around TestMetric lifecycle but other than that the approach looks sane :)

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            """
+
+            schedutil_factor, frequencies = cpu_frequencies[cpu]
+            schedutil_freq = schedutil_factor * util
+
+            # Find the first available freq that meet the schedutil freq
+            # requirement.
+            for freq in frequencies:
+                if freq >= schedutil_freq:
+                    break
```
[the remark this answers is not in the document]
vdonnefort: frequencies being sorted in the structure `cpu_frequencies`, `freq` should already be the highest available if we can't find any that fits.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            #   where C = 1.25
+            schedutil_factor,
+
+            # list of frequencies available for a given CPU.
+            frequencies,
+        )
+        """
+        cpu_frequencies = {
+            cpu: (
+                (max(capacities) * (1 / self.CAPACITY_MARGIN)) /
```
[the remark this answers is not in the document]
vdonnefort: done on a single line, breaking pep8 recommendations though.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+        UtilClamp value.
+
+        The expected capacity is the schedutil projected frequency selection
+        for the given uclamp value.
+        """
+
+        metrics = {}
+        test_failures = []
+        capacity_dfs = []
+        """
```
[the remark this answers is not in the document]
vdonnefort: done.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            df.ffill(inplace=True)
+            df = df[df['active'] == 1]
+
+            # Actual capacity at which the task is running
+            for cpu, freq_to_capa in cpu_capacities.items():
+                df[cpu] = df[cpu].map(freq_to_capa)
+                df['capacity'] = df.apply(lambda line: line[line.cpu], axis=1)
+
+            failures = df[df['capacity'] != expected]
+            num_failures = len(failures['activation'].unique())
```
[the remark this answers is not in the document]
vdonnefort: done.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            ].index.tolist()
+            num_failures = len(failures)
+            test_failures.extend(failures)
+
+            phase_str = 'Phase-{}'.format(phase['phase'])
+            metrics[phase_str] = TestMetric({
+                'uclamp-min': uclamp_val,
+                'cpu-placements': cpus,
+                'expected-cpus': fitting_cpus,
+                '%bad-activations': num_failures * 100 / num_activations,
```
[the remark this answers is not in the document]
vdonnefort: done

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            })
+
+            return cpus.issubset(fitting_cpus)
+
+        res = self._parse_phases(parse_phase, metrics)
```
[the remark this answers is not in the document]
vdonnefort: I removed the _parse_phases() function which in the end doesn't factorise much, to simplify the problem.

lisa/tests/staging/utilclamp.py
```diff
@@ -0,0 +1,376 @@
...
+            num_failures = len(failures['activation'].unique())
+
+            test_failures.extend(failures.index.tolist())
+            capacity_dfs.append(df[['capacity']])
+
+            phase_str = 'Phase-{}'.format(phase['phase'])
+            metrics[phase_str] = TestMetric({
+                'uclamp-min': uclamp_val,
+                'expected-capacity': expected,
+                '%bad-activations': num_failures * 100 / num_activations,
```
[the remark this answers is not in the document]
vdonnefort: done.

lisa/tests/staging/utilclamp.py
vdonnefort: done. Can't recall why I did that ...

lisa/tests/staging/utilclamp.py
vdonnefort: done

lisa/tests/staging/utilclamp.py
vdonnefort: done.

lisa/tests/staging/utilclamp.py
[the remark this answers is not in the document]
douglas-raillard-arm: Good point, I forgot that `for` loops don't introduce a new scope, so if `break` never fires we will indeed end up with `frequencies[-1]`

douglas-raillard-arm (approved):
LGTM, thanks for adding that test :)

PR #1495, opened 2020-10-01
lisa/trace: Bail out if trace file does not exist

bea-arm:
Drop initializing Trace object if specified trace file does not exist.
There is not much point in proceeding if the trace itself cannot be located.

### DIFF
--- a/lisa/trace.py
+++ b/lisa/trace.py
@@ -3210,6 +3210,12 @@ class Trace(Loggable, TraceBase):
         max_swap_size=None,
         write_swap=True,
     ):
+        # Bail-out straightaway if the specified trace file does not exist
+        # There is not much point in continuing here if that is the case
+        if not os.path.exists(trace_path):
+            raise FileNotFoundError("Unable to locate specified trace file: {}"
+                                    .format(trace_path))
+
         super().__init__()
 
         sanitization_functions = sanitization_functions or {}


lisa/trace.py
```diff
@@ -3210,6 +3210,12 @@ def __init__(self,
         max_swap_size=None,
         write_swap=True,
     ):
+        # Bail-out straightaway if the specified trace file does not exist
+        # There is not much point in continuing here if that is the case
+        if trace_path is None or not os.path.exists(trace_path):
```
douglas-raillard-arm, line 3215: > trace_path is None

this can be dropped, it's not legal to pass None in any case (and also rather stupid, if we make a Trace object it's because there is a trace to access, or at least we think there is supposed to be something at the provided path. Passing None cannot happen "by mistake", someone could equally pass 42 and expect things to work):
```
    :param trace_path: File containing the trace
    :type trace_path: str
```

douglas-raillard-arm (changes requested):
In the commit message:
"lisa/trace" should be "lisa.trace" since it designates a module

lisa/trace.py
bea-arm: One could argue that this might actually happen if the trace file path is not being provided explicitly but is being retrieved by some scripting around it, but sure - will drop that one. Thanks.

douglas-raillard-arm: > but is being retrieved by some scripting around it

The only thing I can think of that can provide None is argparse if the argument is (mistakenly) not flagged with `required=True`. And in this case, we don't want a FileNotFoundError, we want argparse to sanitize the parameters directly when possible, or manually and then call that in case of error:
https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error

PR #1500, opened 2020-10-15
energy_model: fix single CPU cluster energy estimate

vdonnefort:
Some topologies like the Pixel4 have a single CPU on a cluster. In this
case, the CPU and the cluster are described similarly:

e.g. multiple CPUs cluster:

 Single CPU description: (4, )
 Cluster description: (4, 5, 6)

e.g. single CPU cluster:

 Single CPU description: (7, )
 Cluster description: (7, )

This is problematic in _estimate_from_active_time(), as the energy computed
would be overwritten. Ensure a single result table initialization when
several nodes have the same description. This avoid losing the energy
computed for the first nodes of an identical description.

Signed-off-by: Vincent Donnefort <vincent.donnefort@arm.com>

### DIFF
--- a/lisa/energy_model.py
+++ b/lisa/energy_model.py
@@ -626,12 +626,19 @@ class EnergyModel(Serializable, Loggable):
             _idle_power = max(node.idle_states[idle_states[c]] for c in cpus)
             idle_power = _idle_power * (1 - active_time)
 
+            if cpus not in ret:
+                if combine:
+                    ret[cpus] = 0
+                else:
+                    ret[cpus] = {}
+                    ret[cpus]["active"] = 0
+                    ret[cpus]["idle"] = 0
+
             if combine:
-                ret[cpus] = active_power + idle_power
+                ret[cpus] += active_power + idle_power
             else:
-                ret[cpus] = {}
-                ret[cpus]["active"] = active_power
-                ret[cpus]["idle"] = idle_power
+                ret[cpus]["active"] += active_power
+                ret[cpus]["idle"] += idle_power
 
         return ret
 


lisa/energy_model.py
```diff
@@ -626,12 +626,19 @@ def _estimate_from_active_time(self, cpu_active_time, freqs, idle_states,
...
+                if combine:
+                    ret[cpus] = 0
+                else:
+                    ret[cpus] = {}
+                    ret[cpus]["active"] = 0
+                    ret[cpus]["idle"] = 0
+
             if combine:
-                ret[cpus] = active_power + idle_power
+                ret[cpus] += active_power + idle_power
```
DouglasRaillard, line 638: Is the returned dict supposed to have a different key for each CPU and for each cluster ? If so, this will basically assume cluster energy = 0, since we add all the cost to `(n,)`, which is recognized as a CPU by the rest of the code. Is there is anything CPU-specific somewhere, i.e. that needs to make the difference between `(n,)-the-cluster` and `(n,)-the-cpu` ?

vdonnefort: I don't think anyone in the code path needs the distinction between `(n,)-the-cluster` and `(n,)-the-cpu` estimate_from_cpu_util(). The callers are:

**1. get_optimal_placements():**

```
power = self.estimate_from_cpu_util(util, freqs=freqs)
candidates[util] = sum(power.values())
```

This shouldn't be a problem here.

**2. tests/test_energy_model.py**

The topology used doesn't have a single CPU cluster. The test still passes.

**3. lisa/tests/staging/schedutil.py  estimate_nrg()**

Unused.

PR #1534, opened 2020-12-07
base: Print out info about the total noise

qais-yousef:
When checking for noise in the experiment to verify there was minimal
interference, we do the check against the total noise in the system but
only print out the info about the noisiest task.

In latest integration branch, RampDown:test_task_placement is undecided
in one case although the noisiest task is 0.24% and the threshold is 1%.
This pattern is repeated, hence we need to print out more info about the
total noise to help with the diagnosis.

### DIFF
--- a/lisa/tests/base.py
+++ b/lisa/tests/base.py
@@ -1330,6 +1330,8 @@ class RTATestBundle(FtraceTestBundle, DmesgTestBundle):
         if noise_threshold_ms is not None:
             threshold_s = min(threshold_s, noise_threshold_ms * 1e3)
 
+        threshold_pct = threshold_s * 100 / self.trace.time_range
+
         df = self.trace.analysis.tasks.df_tasks_runtime()
 
         # We don't want to account the test tasks
@@ -1366,7 +1368,9 @@ class RTATestBundle(FtraceTestBundle, DmesgTestBundle):
         if df_noise.empty:
             return ResultBundle.from_bool(True)
 
-        res = ResultBundle.from_bool(df_noise['runtime'].sum() < threshold_s)
+        total_noise_duration_s = df_noise['runtime'].sum()
+        total_noise_duration_pct = df_noise['runtime_pct'].sum()
+        res = ResultBundle.from_bool(total_noise_duration_s < threshold_s)
 
         pid = df_noise.index[0]
         comm = df_noise['comm'].iloc[0]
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
Are we sure it's not an issue with the current calculation (i.e. 0.24% would be a wrong value) ? In this specific case (the trace that gave the 0.24%), what is the noise added by let's say the top 10 noisy tasks ?

qais-yousef:
The task only ran once on the trace. So I don't think the 0.24% is wrong. I thought initially we check the noise against the noisiest task, but then found out we check against the total noise, hence this patch. I think we need to print this info to give a better picture of why we could have had UNDECIDED.

FWIW, the noisiest task beside it only ran once, it ran on irrelevant CPU. So it didn't really interfere, it just ran somewhere else we don't care about what we take it into account.

I'm trying to extract info about the total noise from the trace, so can't answer your question about the top 10 noisy tasks yet.

douglas-raillard-arm:
> I'm trying to extract info about the total noise from the trace, so can't answer your question about the top 10 noisy tasks yet.

You can take the TestBundle this trace comes from and re-run with the modified code: https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/workflows/ipynb/tests/synthetics_example.html#Offline-replaying
(or using `exekall run --load-db VALUE_DB.pickle.xz --replay ....`). This will be required in any case to see what exactly went wrong in that trace. All we will see with this set of metric is that we get some noise (which we already know), and it won't tell us more about where it comes from (either lots of small tasks or a bug in `test_noisy_tasks()`). In other words, metrics will only take you so far, at some point `breakpoint()` (or `import pdb; pdb.set_trace()`) is required.

valschneider:
> FWIW, the noisiest task beside it only ran once, it ran on irrelevant CPU. So it didn't really interfere, it just ran somewhere else we don't care about what we take it into account.

That could be something for us to improve, e.g. as an option to `test_noisy_tasks()` to filter out "noisy" tasks that ran on CPUs not used by the rt-app tasks.

It isn't obvious to me which tests could benefit from it, however. A task running on a different CPU but same cluster can drive frequency up. Even if it doesn't raise frequency, EAS would see things being run on that CPU and that could interfere with energy computations.

>  All we will see with this set of metric is that we get some noise (which we already know), and it won't tell us more about where it comes from (either lots of small tasks or a bug in test_noisy_tasks()).

I tend to prefer having metrics directly linked with the `ResultBundle.result` decision, and it's true that here it is rather indirect. However, as @douglas-raillard-arm says, the threshold is passed as parameter, so a failure means whatever noise was computed exceeded this readily-available value.

Now, we already have some tests producing debug-oriented plots. Could we add say a plot of noisy tasks (maybe not *all* of them, say the top 10), perhaps a stacked histogram like https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/trace_analysis.html#lisa.analysis.tasks.TasksAnalysis.plot_tasks_total_residency ?

PR #1538, opened 2020-12-09
lisa.wa: Avoid error logging for mixed WA workloads

douglas-raillard-arm:
Allow artifact collectors to declare which workload name they expect,
and swallow any exception if the expected name is set and does not match
the given JobOutput.


credp (approved):
LGTM

PR #1543, opened 2020-12-14
Adding df feec

PierreARM:
New version of the pull request.
I didn't create a lisa/analysis/eas.py module for now as the df_delta function is only used for feec. This could still be done.

### DIFF
--- a/lisa/analysis/eas.py
+++ b/lisa/analysis/eas.py
@@ -1,0 +1,72 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2020, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pandas as pd
+import numpy as np
+
+from lisa.analysis.base import TraceAnalysisBase
+from lisa.trace import requires_events
+from lisa.datautils import df_delta
+
+class EASAnalysis(TraceAnalysisBase):
+    """
+	Energy Aware Scheduler (EAS) specific analysis.
+
+    :param trace: input Trace object
+    :type trace: lisa.trace.Trace
+    """
+    name = 'eas'
+
+###############################################################################
+# DataFrame Getter Methods
+###############################################################################
+
+    @TraceAnalysisBase.cache
+    @requires_events('sched_pre_feec', 'sched_post_feec')
+    def df_feec_delta(self, tasks=None):
+        """
+        DataFrame containing ``find_energy_efficient_cpu`` (feec) related
+        information.
+
+        :param tasks: Task names or PIDs or ``(pid, comm)`` to look for.
+        :type tasks: list(int or str or tuple(int, str))
+
+        :returns: a :class:`pandas.DataFrame` indexed by ``Time`` with:
+
+          - A ``pid`` column.
+          - A ``comm`` column.
+          - A ``__cpu`` column (the cpu feec was executed on).
+          - A ``prev_cpu`` column (the cpu the task was running on).
+          - A ``dst_cpu`` column (the cpu selected by feec).
+          - A ``delta`` column (duration of feec function call).
+        """
+        pre_df = self.trace.df_event('sched_pre_feec')
+        post_df = self.trace.df_event('sched_post_feec')
+
+        # Filter the tasks.
+        if tasks:
+            pre_df = df_filter_task_ids(pre_df, tasks)
+            post_df = df_filter_task_ids(post_df, tasks)
+
+        pre_df = pre_df.drop(columns=['__comm', '__pid'])
+        post_df = post_df.drop(columns=['__comm', '__pid'])
+
+        # Also group by '__cpu': the events must be emitted from the same cpu.
+        return df_delta(pre_df, post_df, ['pid', 'comm', '__cpu'])
+
+
+# vim :set tabstop=4 shiftwidth=4 expandtab textwidth=80
--- a/lisa/datautils.py
+++ b/lisa/datautils.py
@@ -2043,6 +2043,8 @@ _SIGNALS = [
     SignalDesc('cpu_frequency', ['cpu_id']),
     SignalDesc('userspace@cpu_frequency_devlib', ['cpu_id']),
     SignalDesc('sched_compute_energy', ['comm', 'pid']),
+    SignalDesc('sched_pre_feec', ['comm', 'pid', 'prev_cpu']),
+    SignalDesc('sched_post_feec', ['comm', 'pid', 'dst_cpu']),
 
     SignalDesc('sched_pelt_se', ['comm', 'pid']),
     SignalDesc('sched_load_se', ['comm', 'pid']),
--- a/lisa/datautils.py
+++ b/lisa/datautils.py
@@ -1632,7 +1632,7 @@ class SignalDesc:
 
 
 @SeriesAccessor.register_accessor
-def series_convert(series, dtype):
+def series_convert(series, dtype, nullable=None):
     """
     Convert a :class:`pandas.Series` with a best effort strategy.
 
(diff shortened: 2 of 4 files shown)


douglas-raillard-arm (changes requested):
Looks pretty good, I was wondering if we could generalize that to any pair of event, and then add a lisa/analysis/eas.py module with an EASAnalysis class for the feec-specific parts (this would just amount to renaming and dropping some columns) ?

lisa/analysis/latency.py
```diff
@@ -172,6 +171,65 @@ def df_runtimes(self, task):
...
+        post_df['curr_state'] = False
+
+        df = post_df.append(pre_df)
+
+        df.sort_index(inplace=True)
+        df.index.name = 'Time'
+        df.reset_index(inplace=True)
+
+        # Assign new columns.
+        grouped = df.groupby('pid', observed=True, sort=False)
```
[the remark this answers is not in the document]
PierreARM: ok will do

lisa/analysis/latency.py
```diff
@@ -172,6 +171,65 @@ def df_runtimes(self, task):
...
+        new_columns = dict(
+            dst_cpu=grouped['cpu'].shift(-1),
+            next_state=grouped['curr_state'].shift(-1),
```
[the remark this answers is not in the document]
PierreARM: ok yes, I forgot to remove it

lisa/analysis/latency.py
```diff
@@ -172,6 +171,65 @@ def df_runtimes(self, task):
...
+            delta=grouped['Time'].transform(lambda time: time.diff().shift(-1)),
+        )
+        df = df.assign(**new_columns)
+        df.rename(columns={'cpu': 'prev_cpu'}, inplace=True)
+
+        # Only select the rows coming from the 'sched_pre_feec' event
+        # since the row has all the information necessary.
+        df = df[df['curr_state'] == True]
```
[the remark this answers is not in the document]
PierreARM: ok yes, will do

lisa/datautils.py
```diff
@@ -416,6 +416,85 @@ def merge(df1, df2):
...
+    This is usefull to concatenate two dataframe containing consecutive events.
+    A typical usecase would be adding pre/post events at the entrance/exit of a
+    function.
+
+    If a 'pre' event has no corresponding 'post' event, the 'pre' event is
+    dropped.
+
+    :param pre_df: Dataframe containing the events that start a record. All
+        the columns of this dataframe are available in the returned dataframe.
+    :type df: pandas.DataFrame
```
douglas-raillard-arm, line 436: s/:type df:/:type pre_df:/

lisa/datautils.py
```diff
@@ -416,6 +416,85 @@ def merge(df1, df2):
...
+    # Merge. Don't allow column renaming.
+    df = pd.merge(pre_df, post_df, left_index=True, right_index=True, on=on_col,
+                  how='outer', suffixes=(False, False))
+    df.index.name = 'Time'
+    df.reset_index(inplace=True)
+
+    # Create 'delta' column + additionnal post_col.
+    grouped = df.groupby(on, observed=True, sort=False)
+    new_columns = dict(
+        delta=grouped['Time'].transform(lambda time: time.diff().shift(-1)),
```
douglas-raillard-arm, line 484: This assumes that:
* Events are perfectly interleaved, which might not be the case. If for some reason we end up with a "pre->post->post->pre" sequence, the delta will become silently meaningless.
* The dataframe starts with a "pre" event, which is not enforced. If it stats with a "post" event, we will end up measuring the delta between a post and the next pre, which is not what we want.

You could add an "is_pre" column that would be True if it's a pre-event and False otherwise:
```
pre_df['is_pre'] = True
post_df['is_pre'] = False
```
And the on each group:
```
# Only keep the rows for which the "is_pre" status is different from the previous one
df = df.loc[df['is_pre'] != df['is_pre'].shift()]
# Ignore the first row if it's a "post" event
if len(df) >= 1 and not df['is_pre'].iat[0]:
     df = df.iloc[1:]
```
You will have to use the iterable interface of "grouped" that yields `(group, subdf)` tuples with `subdf` being a subset of the initial dataframe, so you can apply the processing using multiple columns on that `subdf` (rather than one processing on the `Time` column directly on the `grouped` object).

douglas-raillard-arm (changes requested):
General shape LGTM, what's left is mostly a bunch of small changes and moving the feec-specific function into an new EASAnalysis class, and making df_add_delta() more robust.

lisa/datautils.py
```diff
@@ -416,6 +416,85 @@ def merge(df1, df2):
...
+        * All the columns from the ``pre_df`` dataframe.
+        * A ``delta`` column (duration between the emission of the ``pre_df``
+            and its consecutive ``post_df``).
+        * All the ``post_col`` columns from the ``post_df``.
+    """
+    pre_df = pre_df.copy(deep=False)
+    post_df = post_df.copy(deep=False)
+
+    # Only keep useful columns in post_df.
+    post_df = post_df[on + post_col]
```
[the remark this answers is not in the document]
PierreARM: thx for all the details

lisa/datautils.py
PierreARM: > You will have to use the iterable interface of "grouped" that yields (group, subdf) tuples

If we process each group individually to remove the isolated pre/post events, we will have to reorder the groups since one single dataframe is returned. Would it be worth creating a `is_next_pre` column like this to select the rows to keep ? This would allow to conserve the order.

```
new_columns = dict(
        delta=grouped['Time'].transform(lambda time: time.diff().shift(-1)),
        is_next_pre=grouped["is_pre"].shift(-1),
)
```

douglas-raillard-arm: You will still need to detect if the first event of each group is a "post", which can be done with something like that:
```
def f(df):
    df = df.shift(-1) if not df['is_pre'].iat[0] else df
    df['delta'] = df['Time'].diff()
    return df

grouped.apply(f).reset_index('Time').sort_index()
```
This is pretty much the same thing as I mentioned earlier except that pandas is doing the concatenation of each group dataframe on its own. In my experiment, it does mess up the order so it needs to be restored by re-sorting against the time.

This is the point where pandas API falls a little bit apart: `GroupBy.apply()` returns a dataframe rather than another GroupBy object, which means you cannot easily chain operations on it (you have to put all of it in one function and use it with `apply()`). This means that any "advanced" processing *has* to be done with `GroupBy.apply/map/transform()`, unless it can be done in a single step with a `GroupBy` method like `GroupBy.diff()`.

Wrt to broken sequences of pre/post, it probably makes more sense to detect them and just raise a ValueError rather than attempt to silently cover it, as it is probably a sign of grouping on the wrong columns (or bug in the kernel code that emitted the events).

The only case that would really make sense would be sequences like `pre1 -> pre2 -> post2 -> post1` that could occur e.g. in function entry/exit events for a recursive one.

PierreARM: I don't understand why the case where the first event of a group is a "post"
event is problematic, we are only keeping the "pre" events with
`     df = df.loc[df["is_pre"]]`
so all the deltas between a "post" and a "pre" event should be tossed in the
end.

About the broken sequences, I agree throwing an error when the sequence is
wrong would be rightful.
Replacing:
```
# For each group, select the rows that are correctly interleaving
# (pre-post-pre-post-...). If multiple rows coming from the same dataframe
# (pre_df or post_df), the first row is kept.
grouped = df.groupby(group_on, observed=True, sort=False)
df = df.loc[grouped["is_pre"].transform(lambda x: x != x.shift(-1)).fillna(False)]
```

by this:
```
# In each group, search for a faulty sequence (where pre/post events are not interleaving,
# e.g. pre1->pre2->post1->post2). In such case, raise an error.
df_faulty = df.loc[grouped["is_pre"].transform(lambda x: x == x.shift(-1)).fillna(False)]
if (not df_faulty.empty):
    raise ValueError
```

should allow to detect this.

douglas-raillard-arm: >  df = df.loc[df["is_pre"]]
Good point, I got confused, I think I was under the wrong assumption that we would filter the results by getting one row over two, which would only work if everything is properly aligned (it would probably be a bit faster but is much less safe).

># In each group, search for a faulty sequence (where pre/post events are not interleaving,
# e.g. pre1->pre2->post1->post2). In such case, raise an error.
df_faulty = df.loc[grouped["is_pre"].transform(lambda x: x == x.shift(-1)).fillna(False)]
if (not df_faulty.empty):
    raise ValueError

You can avoid creating a new dataframe with something like:
```
if grouped['is_pre'].transform(lambda x: x == x.shift()).any():
    raise ValueError('Unexpected sequence of pre and post event (more than one "pre" or "post" in a row')
```

lisa/datautils.py
```diff
@@ -416,6 +416,90 @@ def merge(df1, df2):
...
+    # Thus, let the on_col non-nullable and converts the others to nullable.
+    pre_df_cols = list(set(pre_df) - set(on_col))
+    post_df_cols = list(set(post_df) - set(on_col))
+    pre_df[pre_df_cols] = df_convert_to_nullable(pre_df[pre_df_cols])
+    post_df[post_df_cols] = df_convert_to_nullable(post_df[post_df_cols])
+
+    # Merge. Don't allow column renaming.
+    df = pd.merge(pre_df, post_df, left_index=True, right_index=True, on=on_col,
+                  how='outer', suffixes=(False, False))
+    df.index.name = 'Time'
```
douglas-raillard-arm, line 474: nit: we could save the current index name and reapply it afterwards. May be a good idea to ensure there is no conflict with existing columns by using a UUID as temp name: 
```
import uuid
index_col = uuid.uuid4().hex
```

lisa/datautils.py
```diff
@@ -416,6 +416,90 @@ def merge(df1, df2):
     return functools.reduce(merge, df_list)
```
douglas-raillard-arm, line 418: Missing `@DataFrameAccessor.register_accessor` decorator

lisa/analysis/eas.py
```diff
@@ -0,0 +1,72 @@
...
+import pandas as pd
+import numpy as np
+
+from lisa.analysis.base import TraceAnalysisBase
+from lisa.trace import requires_events
+from lisa.datautils import df_delta
+
+class EASAnalysis(TraceAnalysisBase):
+    """
+	Energy Aware Scheduler (EAS) specific analysis.
```
douglas-raillard-arm, line 27: there seems to be a mix of space and tabs for indentation, all tabs should be replaced by 4 spaces

douglas-raillard-arm (changes requested):
A few nits but otherwise LGTM

PierreARM:
Closing as the first patches adding a `df_delta()` function have been merged, and the last patches are relying on `sched_feec_[pre|post]` events which are not going to be integrated.

PR #1550, opened 2020-12-31
tools: batch-rebase: Add required positional argument

deggeman:
get_help() is missing the required positional argument 'style'. Add it.


douglas-raillard-arm:
The underlying issue was a missing default value, fixed by:
https://github.com/ARM-software/lisa/pull/1551

PR #1568, opened 2021-01-15
lisa.tests.base: Add df_noisy_tasks

vdonnefort:
This new interface allows to get the list of noisy tasks for a test.


vdonnefort:
part of ignore_threaded_irqs

PR #1581, opened 2021-01-27
lisa.stats: Add plotly support for plot_stats()

vdonnefort:
Plotly is a data visualization framework which offers a more user friendly
interface, compared to Matplotlib. This patch allows to use this framework
with the Stats plot_stats() function. The later still defaults to
Matplotlib. Plotly can be enabled by passing backend='plotly'.

### DIFF
--- a/lisa/stats.py
+++ b/lisa/stats.py
@@ -22,6 +22,9 @@ from itertools import combinations
 
 import scipy.stats
 import pandas as pd
+import plotly.express as px
+import plotly.graph_objects as go
+from plotly.subplots import make_subplots as plotly_make_subplots
 import numpy as np
 
 from lisa.utils import Loggable, memoized, FrozenDict, deduplicate, fold
@@ -806,8 +809,21 @@ class Stats(Loggable):
         df[val_col] = df[val_col].round(10)
         return df
 
-    def _plot(self, df, title, plot_func, facet_rows, facet_cols, collapse_cols, filename=None, interactive=None):
+    def _plot(self,
+            df,
+            title,
+            plot_func,
+            facet_rows,
+            facet_cols,
+            collapse_cols,
+            tick_group=None,
+            filename=None,
+            interactive=None,
+            backend='matplotlib'):
         unit_col = self._unit_col
+
+        if tick_group and backend == 'matplotlib':
+            raise ValueError(f'tick_group not supported with matplotlib')
 
         group_on = list(facet_rows) + list(facet_cols)
         facet_rows_len = len(facet_rows)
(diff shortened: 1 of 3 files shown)


lisa/stats.py
```diff
@@ -840,6 +848,35 @@ def split_row_col(group):
             collapsed_col = None
             collapse_group = {}
 
+        if backend == 'plotly':
```
douglas-raillard-arm, line 851: The 2nd part of this function needs to be split into 2 subfunctions, one for each backend.

Currently it's structured like that, with an arbitrary imbalance in the nesting:
```
def f():
	common()
	if cond:
		return backend1()

	return backend2()
```

Should be turned into:
```
def f():
	common()
	if cond:
		return backend1()
	else:
		return backend2()
```

with bakckend1() and backend2() split into actual separate functions or methods.

lisa/stats.py
```diff
@@ -840,6 +848,35 @@ def split_row_col(group):
...
+            for group, subdf in grouped:
+                group = fixup_tuple(group)
+                row, col = split_row_col(group)
+                group_dict = dict(zip(group_on, group))
+                x = rows.index(row) + 1
+                y = cols.index(col) + 1
+                subdf = subdf.drop(columns=group_on)
+                subdf = self._collapse_cols(subdf, collapse_group)
+                group_dict = dict(zip(group_on, group))
+                plot_func(subdf, (figure, x, y), collapsed_col,
```
douglas-raillard-arm, line 869: should use named parameters, the risk of confusion becomes to great with signatures that long.

douglas-raillard-arm, line 869: The value of having a `plot_func` callback lies in having a fixed type for the callback, i.e. a fixed signature, otherwise we end up with spaghetti. The "subplot" can become opaque (an axis for matplotlib or a tuple figure,x,y for plotly) but the rest should be the same.

While at it, you can make all parameters of plot_func keyword-only so there is no risk of mixup and we can easily add more in the future:
```
# foo and bar are keyword-only since the are after a lone start
def f(*, foo, bar): ...
```
https://www.python.org/dev/peps/pep-3102/

Once this is done, we may realize that the loop calling plot_func can actually be shared by both backends

douglas-raillard-arm (changes requested):

lisa/stats.py
```diff
@@ -920,6 +959,9 @@ def plot_stats(self, filename=None, remove_ref=None, interactive=None, groups_as
             **kwargs
         )
 
+        if kind not in ['horizontal_bar', 'vertical_bar']:
```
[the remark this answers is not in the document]
vdonnefort: The check was done in the matplotlib version of 'plot'. Plotly needs a similar version, so I moved it into some common code before the callback plot_func is called.

lisa/stats.py
```diff
@@ -1057,15 +1135,27 @@ def format_val(val):
         if groups_as_row:
             facet_rows, collapse_cols = collapse_cols, facet_rows
 
+        # plotly backend supports another grouping layer, before having the
+        # need for collapsing columns.
+        tick_group = None
+        if backend != 'matplotlib':
+            for col in collapse_cols:
```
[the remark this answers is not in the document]
vdonnefort: It is a bit different. It can groups per tick several colors see the example below.

I can try to improve a bit the column selection though, by picking the one with the highest nunique() ?

horizontal_bar-groups_as_row-matplotlib
horizontal_bar-groups_as_row

douglas-raillard-arm:
Closing as interactive plots are now available with holoviews

PR #1536, opened 2020-12-07
Jankbench collector

credp:
Here's my WIP Jankbench collector, plus a couple of notebooks showing how they can be used.

PR #1583, opened 2021-01-28
Jankbench collector demo notebooks

credp:
These are the notebooks split off from #1536 as discussed. They will not have any jankbench data until the collector is merged, but I dropped it from the set to make a clean PR.


douglas-raillard-arm:

PR #1586, opened 2021-02-02
lisa.analysis.tasks: Catch empty task_state df's

valschneider:
Running trace.analysis.plot_tasks_activation() on a recent integration
trace yields:

```
  lisa/lisa/analysis/tasks.py in df_task_states(self, task, stringify)
      459         """
      460         tasks_df = list(self._df_tasks_states(tasks=[task]))
  --> 461         task_id, task_df = tasks_df[0]
      462         task_df = task_df.drop(columns=["pid", "comm"])
      463
  IndexError: list index out of range
```

It turns out that the task causing this exception *is* present in the
trace, and has associated sched_switch / sched_wakeup events, but is
eclipsed by the TraceView slice done on the downstream TestBundle.

Catch empty state df's in df_task_states(). Silently skip unplottable tasks
in plot_tasks_activation().


lisa/analysis/tasks.py
```diff
@@ -458,6 +458,10 @@ def df_task_states(self, task, stringify=False):
             this state)
         """
         tasks_df = list(self._df_tasks_states(tasks=[task]))
+
+        if not tasks_df:
+            raise ValueError(f'Task "{task}" has no/missing state events')
```
douglas-raillard-arm, line 463: Might be good to say what are the events:
```
             raise ValueError(f'Task "{task}" has no associated events among: {self._df_tasks_states.used_events}') 
```

lisa/analysis/tasks.py
```diff
@@ -1125,7 +1129,10 @@ def plot_tasks_activation(self, hide_tasks: TypedList[TaskID]=None, which_cpu: b
             if pid in hidden_pids:
                 continue
 
-            self.plot_task_activation(TaskID(pid=pid, comm=None), which_cpu=which_cpu, **kwargs)
+            try:
+                self.plot_task_activation(TaskID(pid=pid, comm=None), which_cpu=which_cpu, **kwargs)
+            except ValueError:
```
douglas-raillard-arm, line 1134: Could use a comment like:
```
# The task might not be present in that slice, or might only be visible through events that are not used for the task state such as PELT events
```

douglas-raillard-arm (changes requested):

PR #1592, opened 2021-02-05
tools: add '--native-build' option to lisa-build-asset

PierreARM:
The '--native-build' option allows to natively compile an asset
instead of cross-compiling it. It uses alpine-chroot-install \[1\],
which installs Alpine Linux into a chroot and uses qemu to emulate
the desired architecture.

Only a few asset/architecture couples are currently available, so
the option only natively compiles an asset if possible.

As chroot requires root privileges, the '--native-build' option
also requires root privileges.

\[1\]: https://github.com/alpinelinux/alpine-chroot-install

TODO list:

- [ ] perf: wait for some things to be accepted on the upstream project
- [ ] All assets: update the binaries


tools/lisa-build-asset
```diff
@@ -60,22 +70,54 @@ def list_assets(recipe_dir):
...
+    # Install the chroot
+    subprocess.check_call(['sudo', os.path.join(alpine_source_dir, 'alpine-chroot-install'),
+            '-d', arch_chroot_dir, '-i', build_dir, '-a', QEMU_ARCH_NAMES[arch],
+            '-p', ' '.join(ALPINE_PKG),
+            '-k', ' '.join(env['LISA_ASSET_RECIPE_USED_ENV'].split())],
+            cwd=build_dir)
+
+    # Copy the recipe where the sources are and append the function to execute
+    shutil.copyfile(get_asset_recipe(asset, recipe_dir), os.path.join(build_dir, asset + ".recipe"))
+    with open(os.path.join(build_dir, get_recipe_name(asset)), "a") as f:
```
douglas-raillard-arm, line 86: That's neither robust nor necessary (it's not really supposed to happen but nothing prevents anyone from having an `exit(0)` in the middle of the script, and it prevents sourcing the script again to do the install stage for example).

It is probably possible to fuse that function with `make()` with a `chroot=None` parameter that is set to `arch_chroot_dir` when we want to use the Alpine chroot:
```
cmd  = ... # set by existent make() code

if chroot:
    cmd = [
        os.path.join(arch_chroot_dir, 'enter-chroot'),
        '-u', os.environ['USER'],
        '--', # If the script supports it, it's a very good idea to use -- to avoid conflicts between the custom command and the script options
        *cmd,
    ]

return subprocess.check_output(cmd, env=env, cwd=build_dir)
```

tools/lisa-build-asset
```diff
@@ -60,22 +70,54 @@ def list_assets(recipe_dir):
 def get_asset_recipe(asset, recipe_dir):
     return os.path.join(recipe_dir, '{}.recipe'.format(asset))
 
-def make(asset, actions, arch, toolchain, recipe_dir, build_dir):
+def get_recipe_name(asset):
+    return asset + ".recipe"
+
+def make_native_build(asset, arch, recipe_dir, build_dir, alpine_source_dir, arch_chroot_dir, env):
+    # Install the chroot
+    subprocess.check_call(['sudo', os.path.join(alpine_source_dir, 'alpine-chroot-install'),
```
douglas-raillard-arm, line 78: How long does that take (roughly) ? If it's more than let's say 20s we probably want to factor out the setup and only do it once for each arch, rather than each arch x each recipe.

douglas-raillard-arm, line 78: We cannot call `sudo` from a script like that, it can hang depending on the system-level settings of sudo that are outside of our control. If sudo is required, the user must run the whole script as root. (maybe we can try to drop privileges when we can)

douglas-raillard-arm, line 78: We can get rid of the alpine_source_dir parameter and rely on`alpine-chroot-install` being in PATH (we can download the script in tools/ in install_base.sh --install-toolchains, and tools/ is already in PATH)

tools/lisa-build-asset
```diff
@@ -60,22 +70,54 @@ def list_assets(recipe_dir):
...
             recipe=recipe,
             action=action,
         )
         cmd = ['bash', '-c', script]
         subprocess.check_call(cmd, env=env, cwd=build_dir)
 
-ARCHITECTURES = ['arm64', 'armeabi', 'x86_64']
+        if native_build and action == 'install':
+            # Clean-up the chroot
+            subprocess.check_call([os.path.join(arch_chroot_dir, 'destroy'), '--remove'])
```
douglas-raillard-arm, line 118: should be handled with a context manager like the temporary build dir in `make_asset()`.
Using `contextlib.contextmanager`, you can safely encapsulate the chroot creation and teardown, so that it will always be cleaned up properly no matter what exception is raised in the middle: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

It also seems that `make_asset()` is the right level to use that context manager, and `make()` will therefore only need to take the `chroot` parameter when appropriate to run the action inside the chroot:

```
def make_asset(asset, arch_list, build_dir, recipe_dir, toolchains, native):
    ... 
    for arch in arch_list:
        if native:
            arch_chroot_dir  = os.path.join(arch_build_dir, 'chroot')
            chroot_cm = make_chroot(arch_chroot_dir)
        else: 
             arch_chroot_dir = None
             chroot_cm = nullcontext(None)
         ...
         make(...., chroot=arch_chroot_dir)
```

tools/lisa-build-asset
```diff
@@ -84,28 +126,32 @@ def make_asset(asset, arch_list, build_dir, recipe_dir, toolchains):
...
-
-                toolchain = toolchains[arch]
-                make(asset, actions, arch, toolchain, recipe_dir, arch_build_dir)
-
-        # Only one architecture was requested
-        else:
-            arch = arch_list[0]
-            actions = ['download', 'build', 'install']
+        alpine_source_dir = None
+        if native_build:
```
douglas-raillard-arm, line 130: to be done in install_base.sh --install-toolchain switch

tools/lisa-build-asset
```diff
@@ -164,6 +215,9 @@ def main():
     parser.add_argument('--build-dir', help='Build director. Defaults to temp folder')
     args = parser.parse_args()
 
+    if args.build_dir:
```
douglas-raillard-arm, line 218: I usually  avoid modifying the args inplace, so it makes it more obvious what is straight from argparse and what is the result of some processing:
```
build_dir = os.path.abspath(args.build_dir) if args.build_dir else None
```

douglas-raillard-arm (changes requested):

tools/lisa-build-asset
```diff
@@ -60,22 +70,54 @@ def list_assets(recipe_dir):
...
-def make(asset, actions, arch, toolchain, recipe_dir, build_dir):
+def get_recipe_name(asset):
+    return asset + ".recipe"
+
+def make_native_build(asset, arch, recipe_dir, build_dir, alpine_source_dir, arch_chroot_dir, env):
+    # Install the chroot
+    subprocess.check_call(['sudo', os.path.join(alpine_source_dir, 'alpine-chroot-install'),
+            '-d', arch_chroot_dir, '-i', build_dir, '-a', QEMU_ARCH_NAMES[arch],
+            '-p', ' '.join(ALPINE_PKG),
+            '-k', ' '.join(env['LISA_ASSET_RECIPE_USED_ENV'].split())],
```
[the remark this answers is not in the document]
PierreARM: There should at least be `LISA_ARCH_ASSETS` (where the assets finally copied to) in `LISA_ASSET_RECIPE_USED_ENV`, so it should never be an empty list.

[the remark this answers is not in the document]
douglas-raillard-arm: For now, this may change in the future, better safe than sorry (especially since the extra code is minimal and makes it clear what "-k" applies to)

tools/lisa-build-asset
PierreARM: I agree this is not really gracious. I did this as a workaround not to modify the `enter-chroot` script. I will make a PR to the alpine-install-chroot repository once I have the right to.

tools/lisa-build-asset
PierreARM: >  it can hang depending on the system-level settings of sudo that are outside of our control

I am not sure to understand the case. The issue is that the script will hang asking for sudo privilieges ?
On the other hand, running the script as sudo wipes out the lisa environment / functions sourced, preventing the script to run properly.

tools/lisa-build-asset
PierreARM: Since the alpine-chroot-install repository is only used in this script, I am not setting a LISA_... env variable for it

tools/lisa-build-asset
PierreARM: It takes ~15s.
Another thing is that the alpine script sets the env variables and Alpine packages required for an asset. Sharing a container is possible, but would mean re-writing parts of the alpine script instead of just passing the right parameters to the script.

tools/lisa-build-asset
```diff
@@ -10,6 +10,16 @@ import sys
 from lisa.utils import nullcontext
 
 LISA_HOME = os.environ['LISA_HOME']
+ARCHITECTURES = ['arm64', 'armeabi', 'x86_64']
+
+# Packages required to build assets
+ALPINE_PKG = ['autoconf', 'automake', 'libtool', 'linux-headers', 'git',
```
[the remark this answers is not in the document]
douglas-raillard-arm: I suspect we should install bash in that case in the container, `bash` is already a braindead language but ash is even more. We don't really want to start loosing time debugging things in tools' build systems to end up realizing all we needed was bash to run them.

tools/lisa-build-asset
douglas-raillard-arm: as discussed offline, let's keep sudo. If environment needs to be preserved, we can use `sudo -E`.

tools/lisa-build-asset
douglas-raillard-arm: Not sure what the code was but I suppose it was the build dependencies.

install_base.sh
```diff
@@ -314,6 +314,8 @@ for arg in "$@"; do
         # gettext for autopoint
         pacman_packages+=(gettext autoconf libtool bison)
 
+        `git clone https://github.com/PierreARM/alpine-chroot-install -b pg/use_eval -C ${LISA_HOME}/tools`
```
douglas-raillard-arm, line 317: backquotes don't seem to be needed here (and in places where they have any use, they are better replaced by `$()` since 1. it nests correctly and 2. it's a bit less confusing when other quotes are in use

tools/lisa-build-asset
```diff
@@ -58,24 +87,70 @@ def list_assets(recipe_dir):
...
+            cwd=build_dir)
 
-def make(asset, actions, arch, toolchain, recipe_dir, build_dir):
+    try:
+        yield arch_chroot_dir
+    finally:
+        # Clean-up the chroot
+        subprocess.check_call([os.path.join(arch_chroot_dir, 'destroy'), '--remove'])
+
+def make(asset, actions, arch, toolchain, recipe_dir, build_dir, arch_chroot_dir=None, native_build=False):
```
douglas-raillard-arm, line 131: Is `native_build` parameter needed ? It looks like a check on `arch_root_dir` would achieve the same thing.

tools/lisa-build-asset
```diff
@@ -8,8 +8,45 @@ import shutil
...
+    # Locate where 'alpine-chroot-install' was installed.
+    return os.path.join(os.environ['LISA_HOME'], 'tools', 'alpine-chroot-install')
+
+def support_native_build(asset, recipe_dir, arch):
+    # Not all assets/arch couple support native building.
+    # Check that the arch is suported
+    if not QEMU_ARCH_NAMES.get(arch, None):
+        return False
+
+    # Read the recipe, searching for the 'ALPINE_BUILD_DEPENDENCIES' definition.
```
douglas-raillard-arm, line 33: That should be split separately, as building under QEMU is orthogonal to building on Alpine.

tools/lisa-build-asset
```diff
@@ -8,8 +8,45 @@ import shutil
...
+}
+
+def get_alpine_dir():
+    # Locate where 'alpine-chroot-install' was installed.
+    return os.path.join(os.environ['LISA_HOME'], 'tools', 'alpine-chroot-install')
+
+def support_native_build(asset, recipe_dir, arch):
+    # Not all assets/arch couple support native building.
+    # Check that the arch is suported
+    if not QEMU_ARCH_NAMES.get(arch, None):
```
douglas-raillard-arm, line 30: Is the definition of QEMU_ARCH_NAMES exhaustive ? we need to support armv7 (`arm` value for `--arch`).

Also a couple of nits:
* dict.get()'s `default` default is None already
* It's a better idea to just catch the exception, as it will not make any assumption on the values:
```
try:
    foo[bar]
except KeyError:
    return False
```

Or equivalently, exploit the facts that all Mappings are also Iterables that provide their keys:
```
if bar not in foo:
    return False
```

douglas-raillard-arm (changes requested):
Starting to look good :)
Could you provide an example on what things are not working with the upstream alpine scripts ?

install_base.sh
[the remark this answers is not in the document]
PierreARM: There is a missing `eval` when receiving the commands to run in the chroot, preventing to run multiple commands. (This was why I was copying the recipe and appending the 'build' at the end)

tools/lisa-build-asset
PierreARM: So this is actually working for armv7, there is just a slightly annoying
`qemu: Unsupported syscall: 397`
message popping

install_base.sh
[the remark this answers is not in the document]
douglas-raillard-arm: Can't we workaround by passing one command with `bash -c` ?

tools/lisa-build-asset
douglas-raillard-arm: ok, I suppose it's not too bad as long as the resulting binary are working. Hopefully we won't run into more issues with that

install_base.sh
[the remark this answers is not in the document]
PierreARM: and this is a yes ...

PierreARM:
This was actually possible to use the upstream alpine script as such, so no need to modify anythin

tools/recipes/rt-app.recipe
```diff
@@ -1,5 +1,7 @@
 #! /bin/bash
 
+ALPINE_BUILD_DEPENDENCIES=(autoconf automake bash bison gcc gettext git flex libtool linux-headers make musl-dev numactl-dev)
```
douglas-raillard-arm, line 3: Do we actually need `numactl-dev` and `linux-headers` to compile rt-app ?
`numactl` is compiled inside this recipe already, and rt-app ships with some syscall headers, which is the only use I can see for kernel headers in rt-app.

Things like bison and flex also seem weird, we don't build anything that has a parser in it except the JSON one, which is a hand-written recursive parser (json-c)

douglas-raillard-arm (changes requested):
A bunch of small things but otherwise LGTM. The main bit is making property reading fail if the property is not defined in the recipe (with a more fancy command) rather than checking for a string in the script that could appear in a comment or as part of another name.

tools/recipes/rt-app.recipe
PierreARM: There is a dependency over the linux-headers in rt-app/libdl/dl_syscalls.h
But you were right, we don t need all of this.

tools/lisa-build-asset
```diff
@@ -8,8 +8,42 @@ import shutil
...
+def get_alpine_dir():
+    # Locate where 'alpine-chroot-install' was installed.
+    return os.path.join(os.environ['LISA_HOME'], 'tools', 'alpine-chroot-install')
+
+def support_native_build(asset, recipe_dir):
+    # Not all assets support native building.
+    # Read the recipe, searching for the 'ALPINE_BUILD_DEPENDENCIES' definition.
+    # Not defining it means not supporting native building.
+    with open(get_asset_recipe(asset, recipe_dir)) as f:
+        if 'ALPINE_BUILD_DEPENDENCIES' not in f.read():
```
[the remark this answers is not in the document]
PierreARM: yes, this should actually be faster

tools/lisa-build-asset
```diff
@@ -57,25 +83,89 @@ def list_assets(recipe_dir):
...
-            action=action,
-        )
-        cmd = ['bash', '-c', script]
-        subprocess.check_call(cmd, env=env, cwd=build_dir)
+        cmd = ['bash', '-c', f'source {recipe} && {action}']
 
-ARCHITECTURES = ['arm64', 'armeabi', 'x86_64']
+        if arch_chroot_dir and action == 'build':
+            cmd = ' '.join(map(quote, cmd))
+            cmd = [os.path.join(arch_chroot_dir, 'enter-chroot'), '-u', os.environ['USER'], 'eval', cmd]
```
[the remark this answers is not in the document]
PierreARM: yes, this comes from the alpine-chroot-install script at:
https://github.com/alpinelinux/alpine-chroot-install/blob/master/alpine-chroot-install#L191
The script receives a list of commands, the last ones being what we want to execute in the chroot. Without the `eval`, it is evaluated as a string. This was what I wanted to modify, but just adding the `eval` here solves it.

tools/lisa-build-asset
```diff
@@ -57,25 +83,89 @@ def list_assets(recipe_dir):
...
+    if arch_chroot_dir:
+        # Since the build is native, no need for a toolchain.
+        toolchain = None
+
+    if get_output:
+        # make() is sometimes expected to return the output of the 'actions'.
+        # If so, use check_output() to exectute the 'actions'.
+        output = dict()
+        def run_cmd(*args, **kwargs):
+            output[action] = subprocess.check_output(*args, **kwargs).decode('utf-8')
```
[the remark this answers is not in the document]
PierreARM: It took me some time to understand, but yes, I didn't realize how dodgy this was

PierreARM:
Thx for the patient review. I haven't update the assets yet, I can do it tonight (I would like to avoid having a slow computer during the day)

tools/lisa-build-asset
```diff
@@ -57,25 +73,108 @@ def list_assets(recipe_dir):
...
+        '-i', build_dir,
+        '-a', QEMU_ARCH_NAMES.get(arch, arch),
+        *alpine_dependencies_args,
+        *used_env_variables_args],
+        cwd=build_dir
+    )
+
+    # Manually bind additionnal the recipe directory to the chroot.
+    # We are joining two abspaths, so the second needs to be stripped.
+    arch_chroot_recipe_dir = os.path.join(arch_chroot_dir, recipe_dir.strip(os.path.sep))
```
douglas-raillard-arm, line 129: `os.path.realpath` should be used instead of direct manipulation like `.strp(os.path.sep)`, it will provide a much more robust result: the path is normalized and is made absolute

tools/lisa-build-asset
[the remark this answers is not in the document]
douglas-raillard-arm: Somewhat related, there is a Python gotcha related to that:
```
functions = [
    lambda: print(i)
    for i in range(2)
]

for f in functions:
    f()
```
This will print:
```
1
1
```

This is because `i` is referenced by name in the function, and the name will get resolved at runtime by looking it up in enclosing scopes until it's found. When the loop finishes, the `i` name is bound to the value `1`, and any closure defined in this scope will therefore get `1` when trying to refer to `i`.

A workaround is to use a function factory like that:
```def make_f(i):
    return lambda: print(i)

functions = [
    make_f(i)
    for i in range(2)
]

for f in functions:
    f()
```

Or if you are able to use `def`, you can use a parameter with default value, which will introduce a new name `i` in the scope of `f`, bound to the value that the loop's `i` had when the function was defined:
```
functions = []
for i in range(2):
    def f(i=i):
        print(i)
    functions.append(f)

for f in functions:
    f()
```

douglas-raillard-arm (changes requested):
One last small change and it's ready to go :)

I'll go ahead and merge this one:
https://github.com/ARM-software/lisa/pull/1590

Once it's done you can just rebase on master, check that `cmake` is part of the Alpine dependencies for rt-app and we should be good to go.
Also make sure you use the right fork and branch for rt-app (currently in lisa master) to include the uclamp fix we need.

tools/lisa-build-asset
```diff
@@ -57,25 +73,108 @@ def list_assets(recipe_dir):
...
+    return os.path.join(recipe_dir, get_recipe_name(asset))
+
+def get_asset_property(asset, arch, property, **kwargs):
+    # Read asset's recipe to get a property value.
+    # Raise a ValueError if property is not set.
+    action = f"if [ -z ${{{property}+x}} ]; then exit 1; else printf '%s\\n' ${{{property}[@]}}; fi"
+    try:
+        val = make(asset, arch, [action], **kwargs, get_output=True)
+    except subprocess.CalledProcessError:
+        raise ValueError(f'The property "{property}" if not set.')
```
douglas-raillard-arm, line 89: s/if not set/is not set/

tools/lisa-build-asset
```diff
@@ -57,25 +73,108 @@ def list_assets(recipe_dir):
...
+    recipe = get_asset_recipe(asset, recipe_dir)
+    env = get_env(
+        asset=asset,
+        arch=arch,
+        recipe=recipe,
+        build_dir=build_dir
+    )
+
+    # List the required Alpine packages.
+    alpine_dependencies = get_asset_property(
```
douglas-raillard-arm, line 104: `get_asset_property` now raises so it should be handled:
```
try:
    alpine_dependencies
 = get_asset_property(...)
except ValueError:
    alpine_dependencies_args = None
else:
    alpine_dependencies_args = ('-p' ...)
```

install_base.sh
```diff
@@ -314,6 +314,8 @@ for arg in "$@"; do
         # gettext for autopoint
         pacman_packages+=(gettext autoconf libtool bison)
 
+        git clone https://github.com/alpinelinux/alpine-chroot-install.git --depth=1 -C ${LISA_HOME}/tools
```
douglas-raillard-arm, line 317: This command fails, should be replaced with:
```
        git -C ${LISA_HOME}/tools clone https://github.com/alpinelinux/alpine-chroot-install.git --depth=1
```

douglas-raillard-arm, line 317: Looks like we miss the relevant qemu packages for userspace emulation, they should be added to apt_packages and pacman_packages (the latter is for archlinux)

douglas-raillard-arm, line 317: Just realized this needs to be split in a separate function, otherwise the repo could be cloned twice depending on the command line invocations. See how `install_nodejs_snap` is added to the list of functions to call and to the ordered list.

PierreARM: The alpine script should install `qemu-user-static` automatically, did you hit an error ?

douglas-raillard-arm: I did not notice the script was doing it. Would still be a better idea to add it to install_base.sh since:
1. this lists our actual dependencies (important for review and if we want to enable LISA on other distro)
2. the installation procedure is done so that after install_base.sh, all things that are pulled have an impact that is strictly local to the repository, with absolutely no global change. This is important in order to avoid unexpected pollution of users' setup.

douglas-raillard-arm:
Looks like compiling perf has some issues with "diff" and "find", which seem to come from busybox. We should probably install the GNU coreutils and diffutils in Alpine to avoid that.

tools/lisa-build-asset
```diff
@@ -57,25 +73,119 @@ def list_assets(recipe_dir):
...
+    # Raise a ValueError if property is not set.
+    action = f"if [ -z ${{{property}+x}} ]; then exit 1; else printf '%s\\n' ${{{property}[@]}}; fi"
+    try:
+        val = make(asset, arch, [action], **kwargs, get_output=True)
+    except subprocess.CalledProcessError:
+        raise ValueError(f'The property "{property}" is not set.')
+    return val[action].splitlines()
+
+@contextmanager
+def make_chroot(asset, arch, build_dir=None, recipe_dir=None, arch_chroot_dir=None):
```
douglas-raillard-arm, line 93: `arch_chroot_dir=None` does not seem to be valid, it looks like this parameter should not have any default value. Same for all the other *_dir parameters, it seems the function won't work if it's not given all of them

PierreARM: I did this to make it easier to identify the parameters (as requested in a previous comment)

PierreARM:
Due to some issues while building trace-cmd, I disabled the cross-compilation build. Indeed, `pkg-conf` doesn't have the `--with-path` option on ubuntu, and the build system relies on this. I think this is better to rely on 1 working version for now

tools/lisa-build-asset
```diff
@@ -57,25 +73,151 @@ def list_assets(recipe_dir):
...
+        try:
+            get_asset_property(
+                asset, arch, "BROKEN_CROSS_COMPILATION",
+                build_dir=build_dir,
+                recipe_dir=recipe_dir
+            )
+        except ValueError:
+            builds_on_alpine = False
+        else:
+            builds_on_alpine = True
```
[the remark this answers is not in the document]
PierreARM: The logic was effectively not right

tools/lisa-build-asset
```diff
@@ -57,25 +73,147 @@ def list_assets(recipe_dir):
...
+    def bind_mount(path):
+        # join() will ignore everything before an absolute path so make it
+        # relative first
+        rel = os.path.relpath(path, start='/')
+        path_ = os.path.join(arch_chroot_dir, rel)
+        subprocess.check_call(['sudo', 'mkdir', '-p', path_])
+        subprocess.check_call(['sudo', 'mount', '--bind', path, path_])
+
+    # If build_dir is not in LISA_HOME, bind it aswell.
+    if not build_dir.startswith(LISA_HOME):
```
douglas-raillard-arm, line 167: This would incorrectly return `True` for 
```
build_dir='/home/foo/lisa-build'
LISA_HOME='/home/foo/lisa'
```
So we need something like:
```
def is_parent(parent, child):
    parent = os.path.abspath(parent)
    child = os.path.abspath(child)
    rel = os.path.relpath(parent, child)
    # If the relative path only includes ., .. and /, it's a subfolder
    return not rel.replace('.', '').replace('/', '')
```

tools/recipes/utils.sh
```diff
@@ -32,6 +33,14 @@ install_readme() {
 
     cat "$LISA_ASSET_RECIPE" >> "$src_info"
 
+    # If compiled with musl-libc, add the COPYRIGHT
+    if [ -n ${musl_libc:+x} ]
+    then
+        printf "\n\nThe sources were compiled with musl-libc (content of COPYRIGHT):\n\n" >> "$src_info"
+        wget https://git.musl-libc.org/cgit/musl/plain/COPYRIGHT
+        cat COPYRIGHT >> "$src_info"
```
douglas-raillard-arm, line 41: This would do the same without potentially overriding an existing COPYRIGHT file:
```
wget https://git.musl-libc.org/cgit/musl/plain/COPYRIGHT -O - >> "$src_info"
```

tools/recipes/utils.sh
```diff
@@ -10,6 +10,7 @@ install_readme() {
     repo=$2
     licence_file="$repo/$3"
     src_info="$LISA_ARCH_ASSETS/README.$name"
+    musl_libc=$4
```
douglas-raillard-arm, line 13: As discussed, this can be replaced with an environment variable indicating an alpine build

tools/lisa-build-asset
```diff
@@ -57,25 +73,147 @@ def list_assets(recipe_dir):
...
+    )
+
+    for action in actions:
+        print(f'\nMaking {arch} {asset} {action} in: {build_dir}\n')
+
+        cmd = ['bash', '-c', f'source {recipe} && {action}']
+
+        if arch_chroot_dir and action in ('build', 'install'):
+            # Notify the recipe the build is on alpine.
+            cmd[-1] = f'{cmd[-1]} alpine'
```
douglas-raillard-arm, line 208: Can be replaced with an environment variable to indicate we are building on alpine

douglas-raillard-arm (changes requested):

tools/lisa-build-asset
PierreARM: Even though this would be redundant to bind build_dir in this case, this is still working fine if we bind it 2 times. I removed the check

PR #1593, opened 2021-02-08
lisa.target: Add Target(use_scp=False) parameter

marcbonnici:
Expose devlib's `use_scp` parameter to utilise scp
for file transfer on targets that do not support SFTP.

### DIFF
--- a/lisa/target.py
+++ b/lisa/target.py
@@ -158,6 +158,8 @@ class TargetConf(SimpleMultiSrcConf, HideExekallID):
                 KeyDesc('args', 'Keyword arguments to build the Platform object', [Mapping]),
             )),
             KeyDesc('excluded-modules', 'List of devlib modules to *not* load', [TypedList[str]]),
+            KeyDesc('file-xfer', 'File transfer method. Can be "sftp" (default) or "scp". (Only valid for linux targets)', [TypedList[str]]),
+
         ))
     ))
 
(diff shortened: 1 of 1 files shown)


lisa/target.py
```diff
@@ -142,6 +142,7 @@ class TargetConf(SimpleMultiSrcConf, HideExekallID):
         KeyDesc('device', 'ADB device. Takes precedence over "host"', [str, None]),
         KeyDesc('keyfile', 'SSH private key file', [str, None]),
         KeyDesc('strict-host-check', 'Equivalent to StrictHostKeyChecking option of OpenSSH', [bool, None]),
+        KeyDesc('use-scp', 'Use SCP as method of file transfer instead of the default SFTP', [bool, None]),
```
douglas-raillard-arm, line 145: Would be best exposed under the "devlib" key (since it's more a devlib-specific workaround for unusual setup) and as "file-xfer" key, with either "scp" or "sftp" string values, so that we can reuse it if any other connection type have alternate backends.

In the doc string, it's a good idea to mention that it only applies to the "linux" kind (i.e. it's ignored for android).

In `Target.__init__` it will map automatically to an `devlib_file_xfer=None` parameter (just like for `devlib_excluded_modules`).

douglas-raillard-arm (changes requested):

lisa/target.py
douglas-raillard-arm: Actually I don't remember if it's automatic, you may have to "manually" map it in `Target.INIT_KWARGS_KEY_MAP`

marcbonnici: Updated.

douglas-raillard-arm:
LGTM, thanks for your PR

PR #1606, opened 2021-03-03
tools/sched_tp: Update sched_tp module to latest

mishradk:
update sched_tp module to include new events as provided
by Qais.Yousef@arm.com
removed 0001-sched-add-a-module-to-convert-tp-into-events.patch


tools/kmodules/sched_tp/Makefile
```diff
@@ -1,3 +1,31 @@
...
 EXTRA_CFLAGS = -I$(src)
+
+VMLINUX_DEPS_H = vmlinux_deps.h
+VMLINUX_H = vmlinux.h
+
+VMLINUX_DEPS_TXT = vmlinux_deps.txt
+VMLINUX_TXT = vmlinux.txt
+
+ifeq ($(wildcard $(KERNEL_SRC)/vmlinux), )
+	VMLINUX = /sys/kernel/btf/vmlinux
```
douglas-raillard-arm, line 12: We should probably remove this case, LISA is expected to be installed on a host that is different from the target and everything cross compiled, so this Makefile will be called on the host machine, not the development board.
Therefore we can't use the host's kernel to infer anything about the kernel's target.

tools/kmodules/sched_tp/Makefile
```diff
@@ -1,3 +1,31 @@
...
+
+install:
+	make -C $(KERNEL_SRC) M=$(PWD) modules_install
+
+clean:
+	make -C $(KERNEL_SRC) M=$(PWD) clean
+	rm -f $(VMLINUX_H) $(VMLINUX_DEPS_H)
+
+$(VMLINUX_DEPS_H): $(VMLINUX_DEPS_TXT) $(VMLINUX)
+	pahole -C file://vmlinux_deps.txt $(VMLINUX) > $@
```
douglas-raillard-arm, line 28: Looks like `file://vmlinux_deps.txt` should be replaced with `$(VMLINUX_DEPS_TXT)`

tools/kmodules/sched_tp/Makefile
```diff
@@ -1,3 +1,31 @@
...
+
+$(VMLINUX_H): $(VMLINUX_DEPS_H) $(VMLINUX_TXT) $(VMLINUX)
+	pahole -C file://vmlinux.txt $(VMLINUX) > $@
```
douglas-raillard-arm, line 31: ditto for the hardcoded string

douglas-raillard-arm (changes requested):

tools/kmodules/sched_tp/Makefile
mishradk: Updated the changed.

tools/kmodules/sched_tp/Makefile
mishradk: Modified the code.

tools/kmodules/sched_tp/Makefile
mishradk: Modified the code to use $(VMLINUX_TXT)

douglas-raillard-arm:
LGTM, let's see how the pahole version issue is handled before merging

douglas-raillard-arm:
Superseded by https://github.com/ARM-software/lisa/pull/1713

PR #1610, opened 2021-03-05
lisa.tests.scheduler.eas_behaviour: Detect unsuitable workloads

douglas-raillard-arm:
Workloads that will not have any valid placement (satisfying the util of
each task and the capacity of each CPU) are unsuitable to test EAS, so
catch this case and raise CannotCreateError.

### DIFF
--- a/lisa/tests/scheduler/eas_behaviour.py
+++ b/lisa/tests/scheduler/eas_behaviour.py
@@ -170,10 +170,23 @@ class EASBehaviour(RTATestBundle):
             df = self.trace.analysis.rta.df_phases(task, wlgen_profile=rtapp_profile)
             df = df[df['properties'].transform(lambda phase: phase['meta']['from_test'])]
 
-            phases_util = {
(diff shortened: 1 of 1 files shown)


lisa/tests/scheduler/eas_behaviour.py
```diff
@@ -281,12 +296,7 @@ def _get_expected_power_df(self, nrg_model, capacity_margin_pct):
 
         def exp_power(row):
             task_utils = row.to_dict()
-            try:
-                expected_utils = nrg_model.get_optimal_placements(task_utils, capacity_margin_pct)[0]
-            except EnergyModelCapacityError:
-                raise CannotCreateError(
-                    'The workload will result in overutilized status for all possible task placement, making it unsuitable to test EAS on this platform'
-                )
+            expected_utils = self._get_optimal_placements(nr_model, task_utils, capacity_margin_pct)[0]
```
ionela-voinescu, line 299: s/nr_model/nrg_model

PR #1620, opened 2021-03-12
lisa.wlgen.rta: Run calibration at min freq

douglas-raillard-arm:
Avoid thermal capping while calibrating rt-app by running at the lowest
frequency, and then scaling back the calibration value we measured.


douglas-raillard-arm:
The main problem with doing that is that it will amplify any noise in the measurement, as the CPU will be extra slow. On top of that, it relies on frequency invariance working really well, which may or may not be the case. Some CPUs have a really slow lowest frequency, so it might be best to pick a midpoint instead of the lowest. Ideally the highest sustainable freq.

douglas-raillard-arm:

PR #1624, opened 2021-03-18
recipe: add pahole v1.20 recipe

mishradk:
add pahole v1.20 recipe to generate pahole binary file to be used while
building sched_tp kernel module

### DIFF
--- a/tools/recipes/pahole.recipe
+++ b/tools/recipes/pahole.recipe
@@ -1,20 +1,21 @@
 #! /bin/bash
 
 download() {
-    git clone -b 'v1.20' --single-branch --depth 1 git://git.kernel.org/pub/scm/devel/pahole/pahole.git pahole_src
-    echo "Pahole download completed"
+	git clone -b 'master' --single-branch --depth 1 git://git.kernel.org/pub/scm/devel/pahole/pahole.git pahole_src
+	cd pahole_src
+	git checkout d124926baf2366a2ef9db8ac796874cae2b63831
 }
 
 build() {
     cd pahole_src
     mkdir build
     cd build
-    cmake -D__LIB=lib ..
+    cmake -D__LIB=lib -DBUILD_SHARED_LIBS=OFF ..
     make
-    echo "Pahole build completed"
 }
 
 install() {
     cp -v pahole_src/build/pahole "$LISA_ARCH_ASSETS/pahole"
-    echo "Pahole install completed"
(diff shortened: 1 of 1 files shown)


tools/recipes/pahole.recipe
```diff
@@ -0,0 +1,20 @@
+#! /bin/bash
+
+download() {
+    git clone -b 'v1.20' --single-branch --depth 1 git://git.kernel.org/pub/scm/devel/pahole/pahole.git pahole_src
+    echo "Pahole download completed"
```
douglas-raillard-arm, line 5: There is no need for extra echo, "lisa-build-asset" already logs when it runs a given action.

tools/recipes/pahole.recipe
```diff
@@ -0,0 +1,20 @@
...
+}
+
+build() {
+    cd pahole_src
+    mkdir build
+    cd build
+    cmake -D__LIB=lib ..
```
douglas-raillard-arm, line 12: We need to build static binaries, so that will involve:
* most likely build all dependencies in the recipe
* ask cmake to create a static binary

You can have a look at the rt-app recipe to see how we build dependencies (one build function called in a separate subshell for each).

tools/recipes/pahole.recipe
```diff
@@ -0,0 +1,20 @@
...
+    make
+    echo "Pahole build completed"
+}
+
+install() {
+    cp -v pahole_src/build/pahole "$LISA_ARCH_ASSETS/pahole"
```
douglas-raillard-arm, line 18: also needs to ensure the license makes its way to the README like the other recipes, see install_readme function used by all the other recipes.

douglas-raillard-arm (changes requested):

tools/recipes/pahole.recipe
mishradk: Resolved in new version

tools/recipes/pahole.recipe
mishradk: Resolved in new version. As updated version of pahole is not released yet, at present it uses
    d124926baf2366a2ef9db8ac796874cae2b63831 commit from master branch.

tools/recipes/pahole.recipe
mishradk: Resolved in new version

douglas-raillard-arm:
Superseded by https://github.com/ARM-software/lisa/pull/1740

PR #1630, opened 2021-03-23
lisa.tests.hotplug.rollback

vdonnefort:
This new test verifies the hotplug rollback mechanism.


lisa/tests/hotplug/rollback.py
```diff
@@ -0,0 +1,195 @@
...
+        try:
+            target.hotplug.fail(cpu, state)
+            try:
+                cls._online(target, cpu, up)
+                target.hotplug.fail(cpu, -1)
+                cls._online(target, cpu, not up)
+                return False
+            except TargetStableError:
+                return True
+        except TargetStableError:
```
douglas-raillard-arm, line 62: looks like this could be flattened into:
```
try:
    target.hotplug.fail(cpu, state)
except TargetStableError:
    return False

try:
    cls._online(target, cpu, up)
    target.hotplug.fail(cpu, -1)
    cls._online(target, cpu, not up)
except TargetStableError:
    return True
else:
    return False

```

lisa/tests/hotplug/rollback.py
```diff
@@ -0,0 +1,195 @@
...
+        cls._mark_trace(target, collector, start=1, up=up, fail=fail)
+        cls._online(target, cpu, online=up, verify=False)
+        cls._mark_trace(target, collector, start=0)
+
+    @classmethod
+    def _from_target(cls, target, *,
+                     res_dir: ArtifactPath = None) -> 'HotplugRollback':
+        target.hotplug.online_all()
+        cpu = target.hotplug.list_hotpluggable_cpus()[0]
+        collector = FtraceCollector(target, events=['cpuhp*'])
```
douglas-raillard-arm, line 103: The collector configuration will be provided by FtraceTestBundle, see RTATestBundle.run_rtapp() for an example

lisa/tests/hotplug/rollback.py
```diff
@@ -0,0 +1,195 @@
...
+        return df['idx'].dropna().tolist()
+
+    def test_hotplug_rollback(self) -> ResultBundle:
+        """
+        Test that the hotplug can rollback to its previous state after a
+        failure. All possible steps, up/down combinations will be tested. For
+        each combination, also verify that the hotplug is going through all the
+        steps it is supposed to.
+        """
+        trace = Trace(ArtifactPath.join(self.res_dir, self.TRACE_PATH))
```
douglas-raillard-arm, line 143: Will be provided by FtraceTestBundle with the `.trace` property

lisa/tests/hotplug/rollback.py
```diff
@@ -0,0 +1,195 @@
...
+
+        return cls(res_dir, target.plat_info)
+
+    @classmethod
+    def _get_expected_states(cls, df, up):
+        df = df[(df['expected'] == 1.0) & (df['up'] == float(up))]
+
+        return df['idx'].dropna().tolist()
+
+    def test_hotplug_rollback(self) -> ResultBundle:
```
douglas-raillard-arm, line 136: Will need :
```
@requires_events('cpuhp_enter', 'userspace@HotplugRollback')
```

lisa/tests/hotplug/rollback.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Test that the hotplug can rollback to its previous state after a
+        failure. All possible steps, up/down combinations will be tested. For
+        each combination, also verify that the hotplug is going through all the
+        steps it is supposed to.
+        """
+        trace = Trace(ArtifactPath.join(self.res_dir, self.TRACE_PATH))
+        df = pd.concat([
```
douglas-raillard-arm, line 144: `lisa.datautils.df_merge` should work here

lisa/tests/hotplug/rollback.py
```diff
@@ -0,0 +1,195 @@
...
+                'up': up,
+                'result': df['idx'].tolist() == expected
+            })
+
+        df['fail'].ffill(inplace=True)
+        df.dropna(inplace=True)
+        df = df[df['expected'] == 0.0]
+        df = df.groupby(['up', 'fail']).apply(_verify_rollback)
+        df.drop_duplicates(inplace=True)
+        res = ResultBundle.from_bool(df['result'].all)
```
douglas-raillard-arm, line 191: s/all/all()/, otherwise the whole test is a glorified `return True` :)

douglas-raillard-arm (changes requested):
A few surface comments, I'll review the logic itself later.

General question without having taken the time to really understand the code: is it necessary to write to the trace_marker file ?

If we could just remember what is supposed to happen at step N and then match that with the trace, it would be better. Generally, tests that need to "remember" things fall into 2 categories:
1. they don't actually need to remember anything, e.g. the rtapp profile, as it's generated from already-saved data (platform info)
2. they do need to remember some context, so their `__init__` saves it (like lisa.tests.scheduler.load_tracking.InvarianceItem` that needs to remember its CPU and frequency)

vdonnefort:
Due to several online/offline commands to setup the CPU, using the marker is much more efficient and easy to use to filter actual interesting datas from the trace.

Uploaded a new version with all the requested changes above.

lisa/tests/hotplug.py
```diff
@@ -269,4 +271,174 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+
+        def _get_expected_rollback(up, failing_state):
+            if up == 'True':
+                expected = [
+                    s for s in expected_up if s < int(failing_state)]
+                expected.extend([
+                    s for s in expected_down if s < int(failing_state)
+                ])
+            else:
+                expected = [
```
douglas-raillard-arm, line 416: Actually, the whole function can be heavily deduplicated:
```
import operator
from itertools import chain

def _get_expected_rollback(up, failing_state):
    comp = operator.lt if up else operator.gt
    return [
        s
        for s in chain(expected_up, expected_down)
        if comp(s, failing_state)
    ]

```

Or event actually turned into a single expression
```

from functools import partial
def _get_expected_rollback(up, failing_state):
    return list(filter(
        partial(
            operator.gt if up else operator.lt,
            failing_state,
        ),
        chain(expected_up, expected_down)
    ))

```

lisa/tests/hotplug.py
```diff
@@ -269,4 +271,174 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+            return pd.DataFrame(data={
+                'failing_state': df['failing_state'],
+                'up': up,
+                'result': df['idx'].tolist() == expected
+            })
+
+        df['failing_state'].ffill(inplace=True)
+        df.dropna(inplace=True)
+        df = df[df['expected'] == 0.0]
+        df = df.groupby(['up', 'failing_state']).apply(_verify_rollback)
```
douglas-raillard-arm, line 438: should pass `observed=True` to all groupby() calls to be more robust against categorical dtypes

lisa/tests/hotplug.py
```diff
@@ -269,4 +271,174 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+
+            return pd.DataFrame(data={
+                'failing_state': df['failing_state'],
+                'up': up,
+                'result': df['idx'].tolist() == expected
+            })
+
+        df['failing_state'].ffill(inplace=True)
+        df.dropna(inplace=True)
+        df = df[df['expected'] == 0.0]
```
douglas-raillard-arm, line 437: Looks like expected is a boolean ? If so it should be cast to the correct type as soon as possible in the code with DataFrame.astype()

lisa/tests/hotplug.py
```diff
@@ -269,4 +271,174 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
         for cpu in cur_off_cpus:
             yield cpu, 1
 
-# vim :set tabstop=4 shiftwidth=4 textwidth=80 expandtab
+
+class HotplugRollback(FtraceTestBundle):
+
+    @classmethod
+    def _online(cls, target, cpu, online, verify=True):
+        try:
```
douglas-raillard-arm, line 279: As a general rule the try/except should only include what we strictly want to check:

```@classmethod
def _online(cls, target, cpu, online, verify=True):
    hp = target.hotplug
    f = hp.online if online else hp.offline
    try:
        f(cpu)
    except Exception:
        if verify:
            raise
```

`if foo: ...` is equivalent to `if bool(foo): ...`, itself equivalent (at least for all classes defined in Python) to `if foo.__bool__(): ...`: 
```
>>> class C:
...  def __bool__(self):
...   raise ValueError()
... 
>>> c=C()
>>> if c: pass
... 
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in __bool__
ValueError
```
Other example:
```
>>> import pandas as pd
>>> df=pd.DataFrame()
>>> if df: pass
... 
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/site-packages/pandas/core/generic.py", line 1442, in __nonzero__
    raise ValueError(
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

So there is no guarantee that even a check like that won't raise an exception. The problem is made worse by the fact that `Exception` is caught, which is probably way too wide (I suppose we actually expect a TargetStableError).

lisa/tests/hotplug.py
```diff
@@ -269,4 +271,174 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+        target.hotplug.fail(cpu, failing_state)
+        cls._mark_trace(target, ftrace_coll, start=1, up=up,
+            failing_state=failing_state)
+        cls._online(target, cpu, online=up, verify=False)
+        cls._mark_trace(target, ftrace_coll, start=0)
+
+    @classmethod
+    def _from_target(cls, target, *,
+                     res_dir: ArtifactPath = None) -> 'HotplugRollback':
+        target.hotplug.online_all()
```
douglas-raillard-arm, line 348: Given the manipulation of hotplug we do at multiple levels of helpers in that function, it should be encapsulated into something like that:
```
@classmethod
def _from_target(cls, target, *,
                    res_dir: ArtifactPath = None) -> 'HotplugRollback':
    try:
        return cls._do_from_target(target=target, res_dir=res_dir)
    finally:
        target.hotplug.online_all()
        # plus other things potentially to disable hotplug programmatic fail
```

douglas-raillard-arm (changes requested):
Later in the week it will need to be rebased on:
https://github.com/ARM-software/lisa/pull/1632

But we can wait until the rest is finished, it will be a straightforward change.

external/devlib/devlib/module/hotplug.py
```diff
@@ -57,3 +57,25 @@ def hotplug(self, cpu, online):
             return
         value = 1 if online else 0
         self.target.write_value(path, value)
+
+    def fail(self, cpu, state):
+        path = self.base_path + '/cpu{}/hotplug/fail'.format(cpu)
```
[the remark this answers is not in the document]
vdonnefort: Looks like we'll have to stick to the old format ... devlib still supports python3.5

external/devlib/devlib/module/hotplug.py
```diff
@@ -57,3 +57,25 @@ def hotplug(self, cpu, online):
...
+        if not self.target.file_exists(path):
+            return
+
+        return self.target.write_value(path, state)
```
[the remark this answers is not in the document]
vdonnefort: I just removed file_exists() everywhere.

lisa/tests/hotplug.py
```diff
@@ -269,4 +271,174 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+        except Exception as e:
+            if verify:
+                raise e
+
+    @classmethod
+    def _state_can_fail(cls, target, cpu, state, up):
+        """
+        There are no way of probing the kernel for a list of hotplug states
+        that can fail and for which we can test the rollback. We need therefore
+        to try:
```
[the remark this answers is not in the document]
douglas-raillard-arm: This goes into the direction of adding a newline as well:
http://rst.ninjs.org/#VGhlcmUgYXJlIG5vIHdheSBvZiBwcm9iaW5nIHRoZSBrZXJuZWwgZm9yIGEgbGlzdCBvZiBob3RwbHVnIHN0YXRlcwp0aGF0IGNhbiBmYWlsIGFuZCBmb3Igd2hpY2ggd2UgY2FuIHRlc3QgdGhlIHJvbGxiYWNrLiBXZSBuZWVkIHRoZXJlZm9yZQp0byB0cnk6Ci0gSWYgd2UgY2FuIHNldCB0aGUgc3RhdGUgaW4gdGhlIGtlcm5lbCAnZmFpbCcgaW50ZXJmYWNlLgotIElmIHRoZSBob3RwbHVnIGlzIHJlc2V0IGFjdHVhbGx5IGZhaWxpbmcgKHNvbWUgc3RhdGVzIGNhbiBmYWlsIG9ubHkKICB3aGVuIGdvaW5nIHVwIG9yIGRvd24K

external/devlib/devlib/module/hotplug.py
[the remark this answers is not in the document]
douglas-raillard-arm: LISA does not, we converted to f-string the entire lisa package

lisa/tests/hotplug.py
```diff
@@ -264,4 +269,193 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+    @classmethod
+    def _prepare_hotplug(cls, target, cpu, up):
+        cls._reset_fail(target, cpu)
+        cls._online(target, cpu, not up)
+
+    @classmethod
+    def _get_states(cls, target, cpu, up):
+        states = target.hotplug.get_states()
+        cls._prepare_hotplug(target, cpu, not up)
+        return [
```
douglas-raillard-arm, line 326: list comprehensions lend themselves to a clean readable implementation that highlights the mapped function, the thing iterated over and the filter:
```
x = [
    f(x)
    for x in ...
    if ...
]
```
The 59 spaces indentation before `state` is not really helping anything, harming readability, and will have to be undone the second a single char is added or removed on the previous line.

lisa/tests/hotplug.py
```diff
@@ -264,4 +269,193 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+        try:
+            target.hotplug.fail(cpu, state)
+        except TargetStableError:
+            return False
+
+        try:
+            cls._online(target, cpu, up)
+            cls._reset_fail(target, cpu)
+            cls._online(target, cpu, not up)
+            """
```
douglas-raillard-arm, line 309: Should be a comment, not a string literal. String literals are only "replacing comments" in docstrings, as it has a special support from the Python parser (and Sphinx autodoc extends that to all class attribute and top-level definitions in modules)

lisa/tests/hotplug.py
```diff
@@ -264,4 +269,193 @@ def cpuhp_seq(cls, nr_operations, hotpluggable_cpus, max_cpus_off, random_gen):
...
+            for failing_state in states_up:
+                cls._test_rollback(target, collector, cpu=cpu,
+                                   failing_state=failing_state, up=True)
+
+        return cls(res_dir, target.plat_info)
+
+    @classmethod
+    def _from_target(cls, target, *,
+                     res_dir: ArtifactPath = None, collector=None) -> 'HotplugRollback':
+        cpu = random.choice(target.hotplug.list_hotpluggable_cpus())
```
douglas-raillard-arm, line 384: We want reproducible tests, so we need to pick a specific one, e.g.:
```
# Pick any CPU
cpu = min(target.hotplug.list_hotpluggable_cpus())
```

douglas-raillard-arm (changes requested):

PR #1632, opened 2021-03-24
Improve tests' collector handling

douglas-raillard-arm:
Create a composed collector from generic infrastructure to pass to `_from_target`, so that:
* Tests don't have to fiddle with the details of the collector they use, which can then be fully encapsulated in a mixin class.
* Mixin classes like FtraceTestBundle and DmesgTestBundle can register their own collector automatically
* The user can pass a custom collector that will be ran as well

### DIFF
--- a/lisa/test_example.py
+++ b/lisa/test_example.py
@@ -59,7 +59,7 @@ class ExampleTestBundle(RTATestBundle):
         self.shell_output = shell_output
 
     @classmethod
-    def _from_target(cls, target: Target, *, res_dir: ArtifactPath, ftrace_coll: FtraceCollector = None) -> 'ExampleTestBundle':
+    def _from_target(cls, target: Target, *, res_dir: ArtifactPath, collector=None) -> 'ExampleTestBundle':
         """
         :meta public:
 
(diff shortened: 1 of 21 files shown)


lisa/trace.py
```diff
@@ -5180,10 +5181,17 @@ def __getattr__(self, attr):
...
 
     def __exit__(self, *args, **kwargs):
-        return self._collector.__exit__(*args, **kwargs)
+        swallow = self._collector.__exit__(*args, **kwargs)
+
+        path = self._output_path
+        if path is not None:
+            self.get_data(path)
+
+        return swallow
```
valschneider, line 5203: We should specify whether this is an African or European swallow :coconut:.

lisa/test_example.py
```diff
@@ -38,7 +38,7 @@
 ################################################################################
 
 
-class ExampleTestBundle(RTATestBundle):
+class ExampleTestBundle(RTATestBundle, TestBundle):
```
valschneider, line 41:
```
class RTATestBundle(FtraceTestBundle, DmesgTestBundle)
class TestBundle(FtraceTestBundle, OptionalDmesgTestBundle, TestBundleBase)
```

So this is only adding the OptionalDmesgTestBundle; how does that work vs DmesgTestBundle? If I get [1] right, then DmesgTestBundle "wins", but in that case what do we gain from the added inheritance?

[1]: https://docs.python.org/3/tutorial/classes.html#multiple-inheritance

douglas-raillard-arm: The goal is to provide a TestBundle class you can safely inherit from and that comes with some convenience features. If a given class inherits explicitly from DmesgTestBundle before TestBundle, it will ensure the non-optional API will "win", so the test can be relied on.

The OptionalDmesgTestBundle will bail out with a warning if it cannot collect the output, and `test_dmesg` will return UNDECIDED with a warning instead of raising an exception if the log file does not exist.

valschneider (approved):

PR #1643, opened 2021-04-06
wlgen: Uclamp default values vs kernel uclamp support

valschneider:


lisa/wlgen/rta.py
```diff
@@ -1318,9 +1323,53 @@ class ConcretePropertyBase(PropertyBase):
...
+            # Dependency could be built as a module, in which case we'd
+            # need to check whether it is loaded. Assert against Y for
+            # now.
+            if val is not KernelConfigTristate.YES:
+                msg = f'Kernel config {key}={val}, expected Y'
+                return raise_err(msg, TargetStableError(msg))
+            return True
+
+    def to_json(self, plat_info, **kwargs):
+        self.check_kconfig(plat_info, True)
```
douglas-raillard-arm, line 1364: Might want to add a comment stating that raising is *mandatory*, otherwise the value of the property in following phases is basically an undefined behavior, which violates a very important invariant.

lisa/wlgen/rta.py
```diff
@@ -1318,9 +1323,53 @@ class ConcretePropertyBase(PropertyBase):
...
+        return self._to_json(plat_info, **kwargs)
```
douglas-raillard-arm, line 1365: would be best to pass plat_info as a keyword argument, so we don't need to change the signatures of the implementations

lisa/wlgen/rta.py
```diff
@@ -1318,9 +1323,53 @@ class ConcretePropertyBase(PropertyBase):
...
+            except KeyError as e:
+                return raise_err(f'Kernel config does not have key: {key}', e)
+
+            # Dependency could be built as a module, in which case we'd
+            # need to check whether it is loaded. Assert against Y for
+            # now.
+            if val is not KernelConfigTristate.YES:
+                msg = f'Kernel config {key}={val}, expected Y'
+                return raise_err(msg, TargetStableError(msg))
+            return True
```
douglas-raillard-arm, line 1361: `return True` seems to have one too many indentation level. It will currently be executed at the first iteration of the loop, bypassing the check for the subsequent keys

douglas-raillard-arm (changes requested):

lisa/wlgen/rta.py
```diff
@@ -2374,11 +2426,12 @@ def __mul__(self, n):
     def to_events(self, **kwargs):
         pass
 
-    def to_json(self, **kwargs):
+    def _to_json(self, plat_info, **kwargs):
         """
         Deduplicate the event names and turn the whole workload into a loop if
         possible.
         """
+        kwargs['plat_info'] = plat_info
```
valschneider, line 2434: So I faced a similar issue with `to_events()` as with `to_json()`: there isn't a unified signature (not that there should be one). This gets around aligning them.

douglas-raillard-arm: Not sure what you mean exactly, this function does not use plat_info (beyond adding it to its positional argument list), so there should not be any need to mention it in the signature. If the function is simply called with keyword arguments, most `to_json` can just be renamed to `_to_json` without affecting their signature (unless they actually need to use the `plat_info` of course)

douglas-raillard-arm (approved):
LGTM, we can merge as soon as the tests pass

PR #1666, opened 2021-05-07
Migr test cases

deggeman:


lisa/tests/scheduler/load_tracking.py
```diff
@@ -310,7 +315,8 @@ def get_simulated_pelt(self, task, signal_name):
             # executing
             preempted_value=0,
         )
-        df = trace.analysis.load_tracking.df_task_signal(task, signal_name)
+        cpu = self.wlgen_task.children['test']['cpus']
```
douglas-raillard-arm, line 318: We could add the following property (taken from the deleted CPUMigrationBase and slightly modified to not fail if a phase is not pinned) to ensure we don't miss any CPU and so that we don't need to repeat that bit if it's needed anywhere else:
```
    @property
    def cpus(self):
        """
        All CPUs used by RTapp workload.
        """
        return set(itertools.chain.from_iterable(
            phase.get('cpus', [])
            for task in self.rtapp_profile.values()
            for phase in task.phases
        ))
```

lisa/tests/scheduler/load_tracking.py
```diff
@@ -808,432 +825,13 @@ def make_group_bundle(cpu, item_group):
...
-
-
-class NTasksCPUMigration(NTasksCPUMigrationBase):
-    """
-    N tasks on N CPUs, and try all permutations of tasks and CPUs.
-    """
-
-    @classmethod
-    def get_nr_required_cpu(cls, plat_info):
+class RqInvariance(InvarianceBase):
```
douglas-raillard-arm, line 828: nit: 2 blank lines between module-level declarations

lisa/tests/scheduler/load_tracking.py
```diff
@@ -694,51 +636,126 @@ def item_test(test_item):
...
+
+        :attr:`~lisa.tests.base.Result.UNDECIDED` is ignored.
+        """
+        item_res_bundles = [
+            item_test(item)
+            for item in self.invariance_items
+        ]
+        return AggregatedResultBundle(item_res_bundles, 'cpu')
+
+class TaskInvariance(InvarianceBase):
```
douglas-raillard-arm, line 653: nit: 2 blank lines between module-level declarations

lisa/tests/scheduler/load_tracking.py
douglas-raillard-arm, line 318: s/cpu/cpus/ as this is a set, not a single CPU ID

douglas-raillard-arm (changes requested):
A couple of nits but otherwise LGTM, it's nice to see that basic rq signal testing was within reach with just a bit of refactoring :)

douglas-raillard-arm:
One thing: do we plan on having a different workload for the rq tests in the future ? 
If we plan on using the exact same workload for both cases, we might want to think of a way to avoid running it twice, as we run on all kinds of CPUs in the system, and up to 8 freqs, so it's a fair amount of runtime and artifacts.

douglas-raillard-arm:
Also it looks like LoadTrackingBase could be incorporated into InvarianceItemBase, and LoadTrackingBase.is_almost_equal removed as it's not used anywhere anymore

lisa/tests/scheduler/load_tracking.py
deggeman: Will change it.

lisa/tests/scheduler/load_tracking.py
deggeman: Will change it.

deggeman:
> Also it looks like LoadTrackingBase could be incorporated into InvarianceItemBase, and LoadTrackingBase.is_almost_equal removed as it's not used anywhere anymore

WIll add a patch on top.

deggeman:
> One thing: do we plan on having a different workload for the rq tests in the future ?
> If we plan on using the exact same workload for both cases, we might want to think of a way to avoid running it twice, as we run on all kinds of CPUs in the system, and up to 8 freqs, so it's a fair amount of runtime and artifacts.

Not sure yet. I would like to get the current patch-set in first before starting to think about additional tests.

lisa/tests/scheduler/load_tracking.py
douglas-raillard-arm: The property should be used unless there is a good reason, along with an assert if we can only deal with len(cpus) == 1 (which is the case, as the test currently can only check the signals of one rq at a time). This will make it obvious in the future if a subclass tries to make a workload with more than one CPU. Alternatively to the assert, you can use:
```
cpu, = ...
```
which will unpack a 1-item iterable into `cpu`, or raise a ValueError otherwise.

deggeman:
```
index e3d5891b138e..966f7d5456d5 100644
@@ -193,6 +193,17 @@ Time in seconds for util_avg to converge (i.e. ignored time)
         assert len(tasks) == 1
         return tasks[0]
 
+    @property
+    def cpus(self):
+        """
+        All CPUs used by RTapp workload.
+        """
+        return set(itertools.chain.from_iterable(
+            phase['cpus']
+            for task in self.rtapp_profile.values()
+            for phase in task.phases
+        ))
+
     def get_tags(self):
         return {'cpu': f'{self.cpu}@{self.freq}'}
 
@@ -308,7 +319,6 @@ Time in seconds for util_avg to converge (i.e. ignored time)
         logger = self.get_logger()
         trace = self.trace
         task = trace.get_task_id(task)
-        cpus = trace.analysis.tasks.cpus_of_tasks([task])
 
         df_activation = trace.analysis.tasks.df_task_activation(
             task,
@@ -316,8 +326,11 @@ Time in seconds for util_avg to converge (i.e. ignored time)
             # executing
             preempted_value=0,
         )
-        cpu = self.wlgen_task.children['test']['cpus']
-        df = self._get_trace_signal(task, cpu, signal_name)
+
+        cpus = sorted(self.cpus)
+        assert len(cpus) == 1
+        df = self._get_trace_signal(task, cpus, signal_name)
+
         df = df.copy(deep=False)
 
         # Ignore the first activation, as its signals are incorrect
@@ -337,8 +350,7 @@ Time in seconds for util_avg to converge (i.e. ignored time)
         except KeyError:
             if any(
                 self.plat_info['cpu-capacities']['rtapp'][cpu] != UTIL_SCALE
-                for phase in self.wlgen_task.phases
-                for cpu in phase['cpus']
+                for cpu in self.cpus
             ):
                 ResultBundle.raise_skip('PELT time scaling can only be simulated when the PELT clock is available from the trace')
 
@@ -346,6 +358,7 @@ Time in seconds for util_avg to converge (i.e. ignored time)
             clock = None
 
         try:
+            cpus = trace.analysis.tasks.cpus_of_tasks([task])
             capacity = trace.analysis.load_tracking.df_cpus_signal('capacity', cpus)
         except MissingTraceEventError:
             capacity = None
```

deggeman: OK, I'll use:

```
+        cpus = sorted(self.cpus)
+        assert len(cpus) == 1
+        df = self._get_trace_signal(task, cpus, signal_name)
```

and:

```
         try:
+            cpus = trace.analysis.tasks.cpus_of_tasks([task])
             capacity = trace.analysis.load_tracking.df_cpus_signal('capacity', cpus)
```
in def get_simulated_pelt(self, task, signal_name).

lisa/tests/scheduler/load_tracking.py
```diff
@@ -345,6 +359,7 @@ def get_simulated_pelt(self, task, signal_name):
             clock = None
 
         try:
+            cpus = trace.analysis.tasks.cpus_of_tasks([task])
```
douglas-raillard-arm, line 362: Would be best to use a different variable name than `cpus` to avoid confusion with the previous definition. Maybe the first one could be renamed "pinned_cpus"

deggeman: OK, let me change this.

deggeman:
Latest push bases on home_pr306) (exekall: Fix annotation resolution in methods) and has the 'pinned_cpus' fix.

CI Tests on sd845c, tc2, juno-r0, hikey960, juno-r2 all passing.

The missing bit is the refactoring that the Task and Rq tests use shared testruns.

lisa/tests/scheduler/load_tracking.py
```diff
@@ -118,59 +119,36 @@ def correct_expected_pelt(cls, plat_info, cpu, signal_value):
...
-        Verify that ``value``` is reasonably close to ``target```
-
-        :returns: A tuple (bool, delta_pct)
-        """
-        delta = value - target
-        delta_pct = delta / target * 100
-        equal = abs(delta_pct) <= allowed_delta_pct
-
-        return (equal, delta_pct)
+#class LoadTrackingBase(RTATestBundle, LoadTrackingHelpers, TestBundle):
```
douglas-raillard-arm, line 122: Should be removed rather than commented

douglas-raillard-arm:
After thinking about it for a while, maybe we want to keep the workloads for the task and rq separate. It's probably not necessary right now, but we are likely to extend the test in the future (e.g. for the RT rq, or even IRQ pressure), at which point we might need to have more flexibility.

tools/exekall/exekall/engine.py
```diff
@@ -2401,6 +2401,10 @@ class UnboundMethod:
     .. note:: It is generally assumed that if a given method is wrapped in an
         :class:`UnboundMethod`, all subclasses will also have that method
         wrapped the same way.
+
```
douglas-raillard-arm, line 2404: This commit is now in master, so it should be removed

lisa/tests/scheduler/load_tracking.py
deggeman: OK, will fix it.

tools/exekall/exekall/engine.py
deggeman: OK.

PR #1668, opened 2021-05-11
Remove CannotCreateError

douglas-raillard-arm:


lisa/exekall_customize.py
```diff
@@ -30,7 +31,7 @@
 from lisa.target import Target, TargetConf
 from lisa.utils import HideExekallID, ArtifactPath, Serializable, get_nested_key, ExekallTaggable
 from lisa.conf import MultiSrcConf
-from lisa.tests.base import TestBundle, ResultBundleBase, Result
+from lisa.tests.base import TestBundle, ResultBundleBase, Result, CannotCreateError
```
valschneider: Doesn't this conflict with the next commit which removes CannotCreateError?

PR #1685, opened 2021-06-09
exekall: Fix annotation resolution in methods

douglas-raillard-arm:
Allow a method to refer to a TypeVar class attribute by name, such as:

    class X:
	T = TypeVar('T')
	def f(self) -> 'T':
	    pass


deggeman:
LGTM. With this patch underneath https://github.com/ARM-software/lisa/pull/1666 I can change the function annotation of InvarianceBase.iter_invariance_items() from 'InvarianceBase.ITEM_CLS' to 'ITEM_CLS' and `lisa-test --list` still lists all the TaskInvariance and RqInvariance tests.

PR #1730, opened 2021-08-04
Compile kernel modules on the fly

douglas-raillard-arm:
Adds dynamic module compilation to LISA.
The module is compiled on the fly and loaded/unloaded automatically by the FtraceCollector, when ftrace events provided by the module are requested. Compilation stages are cached as much as possible to minimize the overhead, but it might take a couple minutes on the first use depending on the setup.

When using this, please ensure that:
* You are not using any other patch that adds the trace events (e.g. EAS integration branch). If you do, the module from lisa will either not be used, or be partially used and create duplicates events for the overlapping events.
* Use the following kernel config. `CONFIG_DEBUG_INFO_BTF=y` will require a recent pahole, which is shipped in LISA, so just `source init_env` before building your kernel:
```
CONFIG_IKHEADERS=y
CONFIG_DEBUG_INFO=y
CONFIG_DEBUG_INFO_BTF=y
CONFIG_DEBUG_INFO_REDUCED=n
```
* You will probably need to point at the kernel sources that go with your kernel in the `target_conf.yml` file:
```
target-conf:
    kernel:
        src: /home/foobar/linux/
```

* If you want a more reproducible setup, add that too. This will use an Alpine chroot and use clang by default:
```
target-conf:
    kernel:
        modules:
            build-env: alpine
```


douglas-raillard-arm:
TODO: add better doc for options required for the dynamic module and examples (attached
[kmod_example.py.txt](https://github.com/ARM-software/lisa/files/7528414/kmod_example.py.txt)
)

PR #1733, opened 2021-08-16
lisa.wlgen.rta: Fix recursive calibration

douglas-raillard-arm:
FIX

Avoid accidentally re-entering the calibration function while running
it.


deggeman:
LGTM. Running github_dourai01/_home_pr90018 lets this issue disappear when executing an initial

    calib = RTA.get_cpu_calibrations(target, res_dir=path)

PR #1737, opened 2021-09-01
tools/recipes/perf.recipe: Avoid compilation failure because of warning

douglas-raillard-arm:


douglas-raillard-arm:
Does not seem to be actually fixing the issue, EXTRA_CFLAGS does not seem to have any impact anymore

PR #1746, opened 2021-09-22
Fix LISA android target with latest

Leo-Yan:
These two patches are to address the issue found with executing command:

  lisa-test 'OneSmallTask*'

The first patch is to address the issue introduced by the latest AOSP with emulated daemons for storage; the second patch is to fix the root permission issue which is caused by the change in devlib.

After applying these two patches, the lisa-test command can execute successfully on my Hikey960 board.

### DIFF
--- a/lisa/target.py
+++ b/lisa/target.py
@@ -633,9 +633,9 @@ class Target(Loggable, HideExekallID, ExekallTaggable, Configurable):
                 device = 'DEFAULT'
 
             conn_settings['device'] = device
-            # If the username was explicitly set to "root", root the target as
-            # early as possible
-            conn_settings['adb_as_root'] = (username == 'root')
+            # If the username was explicitly set to "root", or username is None,
+            # root the target as early as possible
+            conn_settings['adb_as_root'] = (username == 'root' or username is None)
 
         elif kind == 'linux':
             devlib_target_cls = devlib.LinuxTarget
@@ -691,14 +691,6 @@ class Target(Loggable, HideExekallID, ExekallTaggable, Configurable):
         )
 
         target.connect(check_boot_completed=wait_boot, timeout=wait_boot_timeout)
-
-        # None as username means adb root will be attempted, but failure will
-        # not prevent from connecting to the target.
-        if kind == 'android' and username is None:
-            try:
-                target.adb_root(enable=True)
-            except Exception as e: # pylint: disable=broad-except
-                logger.warning(f'"adb root" failed: {e}')
 
         logger.debug(f'Target info: {dict(abi=target.abi, cpuinfo=target.cpuinfo, workdir=target.working_directory)}')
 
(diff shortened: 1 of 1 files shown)


lisa/target.py
```diff
@@ -629,9 +633,9 @@ def _init_target(self, kind, name, workdir, device, host,
                 device = 'DEFAULT'
 
             conn_settings['device'] = device
-            # If the username was explicitly set to "root", root the target as
-            # early as possible
-            conn_settings['adb_as_root'] = (username == 'root')
+            # If the username was explicitly set to "root", or username is None,
+            # root the target as early as possible
+            conn_settings['adb_as_root'] = (username == 'root' or username is None)
```
douglas-raillard-arm, line 638: only `username=='root'` should trigger the "mandatory" rooting of the device, otherwise non-rooted device will not be usable anymore.

lisa/target.py
```diff
@@ -688,14 +692,6 @@ def _init_target(self, kind, name, workdir, device, host,
 
         target.connect(check_boot_completed=wait_boot, timeout=wait_boot_timeout)
 
-        # None as username means adb root will be attempted, but failure will
-        # not prevent from connecting to the target.
-        if kind == 'android' and username is None:
-            try:
-                target.adb_root(enable=True)
```
douglas-raillard-arm, line 695: looks like this should be turned into something like that, following the original devlib implementation (with the extra necessary import for AdbConnection):
```
conn = target.conn
if isinstance(conn, AdbConnection) and username is None:
	try:
		conn.adb_root(enable=enable)
	except Exception as e: # pylint: disable=broad-except
		logger.warning(f'"adb root" failed: {e}')
```

douglas-raillard-arm (changes requested):

lisa/target.py
douglas-raillard-arm: Hopefully that will be enough. If devlib does not appreciate late rooting anymore, we could also try to connect to the target as root and bail out if it does not work when username is None (but I'd rather try this solution first, catching the "non rooted" error will be problematic and we will end up catching network error etc as well)

douglas-raillard-arm: Actually had a thought about that: intervening directly on `target.conn` is not really going to cut it. This attribute is thread local and new connections will be instantiated and destroyed as threads are used. Setting it for the main thread would provide a partial workaround but it's not really a fix unfortunately.

douglas-raillard-arm:
Would you mind opening a separate PR for the rs.media.module so we can merge it independently please ?

For the rooting matter, could you confirm that passing `username='root'` solves the issue ? The information whether to root or not has unfortunately to be passed to the connection settings and therefore needs to be provided by the user in advance according to how the board/phone is configured. The only way we can make it work otherwise is to attempt to root no matter what, and fallback to non-root if `target.connect()` raises an exception, but I'd rather get this information first hand from the user configuration.

douglas-raillard-arm:
Closing with the remaining non fixed issue fixed in https://github.com/ARM-software/lisa/pull/1875

PR #1754, opened 2021-10-05
install_base: Do not install VirtualBox on aarch64 machines

vdonnefort:
VirtualBox currently does not support aarch64 machines.

### DIFF
--- a/install_base.sh
+++ b/install_base.sh
@@ -189,8 +189,10 @@ pacman_packages=(
     kernelshark
 )
 
+HOST_ARCH="$(uname -m)"
+
 # ABI-specific packages
-case $(uname -m) in
+case $HOST_ARCH in
     aarch64)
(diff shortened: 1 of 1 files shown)


install_base.sh
```diff
@@ -310,6 +312,8 @@ for arg in "${args[@]}"; do
         vm=$(systemd-detect-virt 2>/dev/null)
         if [[ $vm == 'oracle' ]] ; then
             echo "VirtualBox detected, not installing virtualbox apt packages" >&2
+        elif [[ $HOST_ARCH == 'aarch64' ]]; then
+            echo "VirtualBox not supported on $HOST_ARCH"
```
douglas-raillard-arm, line 316: missing `>&2` but otherwise LGTM

douglas-raillard-arm (changes requested):

PR #1766, opened 2021-11-03
lisa.trace: Avoid forking multiprocessing.Pool

douglas-raillard-arm:
FIX

Since they can cause deadlocks in multithreaded applications, use the
"spawn" method even though it is slower.


douglas-raillard-arm:

PR #1773, opened 2021-11-12
Support for Artifactory in bisector

ionela-voinescu:
This adds support for ArtifactoryService and modifies or adds to the documentation about uploading and downloading artifacts.

### DIFF
--- a/tools/bisector/bisector/bisector.py
+++ b/tools/bisector/bisector/bisector.py
@@ -1819,7 +1819,7 @@ class LISATestStep(ShellStep):
             export_db=BoolOrStrParam('export a merged exekall ValueDB, merging it with existing ValueDB if the file exists', allow_empty=False),
             export_logs=BoolOrStrParam('export the logs and artifact directory symlink to the given directory'),
             download=BoolParam('Download the exekall artifact archives if necessary'),
-            upload_artifact=BoolParam('upload the artifact directory to Artifactorial and update the in-memory report. Following env var are needed: ARTIFACTORIAL_FOLDER set to the folder URL and ARTIFACTORIAL_TOKEN. Note: --export should be used to save the report with updated paths'),
+            upload_artifact=BoolParam('upload the artifact directory to an artifacts service and update the in-memory report. Following env var are needed: ARTIFACTORY_FOLDER or ARTIFACTORIAL_FOLDER set to the folder URL, and ARTIFACTORY_TOKEN or ARTIFACTORIAL_TOKEN. Note: --export should be used to save the report with updated paths'),
         )
 
     )
@@ -5139,11 +5139,12 @@ command line""")
         option.""")
 
     run_parser.add_argument('--upload-report', action='store_true',
-        help="""Continuously upload the report to Artifactorial after every
-        iteration. This relies on the following environment variables:
-        ARTIFACTORIAL_FOLDER set to the folder's URL and ARTIFACTORIAL_TOKEN
-        set to the token. Remember to use the right step option to upload the
-        results as they are computed if desired.
+        help="""Continuously upload the report to an artifacts service after
+        every iteration. This relies on the following environment variables:
+        ARTIFACTORY_FOLDER or ARTIFACTORIAL_FOLDER set to the folder's URL
+        and ARTIFACTORY_TOKEN or ARTIFACTORIAL_TOKEN set to the token.
+        Remember to use the right step option to upload the results as they
+        are computed if desired.
         """)
 
     dbus_group = run_parser.add_mutually_exclusive_group()
--- a/tools/bisector/bisector/bisector.py
+++ b/tools/bisector/bisector/bisector.py
@@ -2574,6 +2574,96 @@ class ArtifactorialService(ArtifactsService):
         urlretrieve(url, path)
 
 
+class ArtifactoryService(ArtifactsService):
+    """Upload/download files to/from Artifactory. """
+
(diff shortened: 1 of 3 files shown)


tools/bisector/bisector/bisector.py
```diff
@@ -2558,6 +2563,105 @@ def upload(self, path):
...
+
+        :param path: path to the file to download
+        """
+
+        headers = {
+            "X-JFrog-Art-Api": self.token
+        }
+
+        response = requests.get(url, headers=headers)
+        # Raise an exception is the request failed
```
douglas-raillard-arm, line 2655: Maybe we should just emit a warn() instead ? Raising an exception will prevent downloading any other link if one is dead for some reason. If that is the most convenient function from `requests`, we can simply:
```
try:
    ....
except Exception as e:
    warn(f'Fail to retrieve URL {url}: {e}')
```

tools/bisector/bisector/bisector.py
```diff
@@ -5138,10 +5249,13 @@ def _main(argv):
 
     SHOW_TRACEBACK = args.debug
     service_hub = ServiceHub()
+
     try:
-        service_hub.register_service('upload', ArtifactorialService())
+        artifactorial = ArtifactorialService()
+        service_hub.register_service('upload', artifactorial)
+        service_hub.register_service('download', artifactorial)
```
douglas-raillard-arm, line 5256: I guess we want to add ArtifactoryService here too. We could try ArtifactoryService first and if it raises an exception then we register the other one. This just means we have to check for the expected env var in `ArtifactoryService.__init__` and raise e.g. `ValueError` if they are not present.

douglas-raillard-arm (commented):
Some nits but otherwise LGTM

tools/bisector/bisector/bisector.py
```diff
@@ -2493,37 +2490,45 @@ def urlretrieve(url, path):
...
+        :param path: path to the file to upload
+        :param url: URL of the Artifactorial folder to upload to,
+                    or env var ARTIFACTORIAL_FOLDER
         """
 
-        url = self.url
-        token = self.token
+        token = self.token or os.getenv('ARTIFACTORIAL_TOKEN')
+        if not token:
+            raise ValueError("An Artifactorial token must be provided through ARTIFACTORIAL_TOKEN env var")
```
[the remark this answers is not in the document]
ionela-voinescu: I'm tempted to move this to __init__ but keep the format:

`self.token = token or os.getenv('ARTIFACTORIAL_TOKEN')`

as I think it's better for token to be used if given as an argument, and if not, we'll use ARTIFACTORIAL_TOKEN.

tools/bisector/bisector/bisector.py
ionela-voinescu: We do catch these exceptions when we call the download function, as in:

```
                          try:
                              download_service.download(url=archive_path,
                                                        path=archive_dst)
                          except Exception as e:
                              error('Could not retrieve {archive_path}: {e}'.format(
                                  archive_path=archive_path,
                                  e=e
                              ))
```

I'm tempted to either leave it as is, or modify the download functions for both the ArtifactorialService and the ArtifactoryService to catch and warn inside the download function. 

Any preferences?

tools/bisector/bisector/bisector.py
```diff
@@ -2558,6 +2563,105 @@ def upload(self, path):
...
+
+        headers = {
+            "X-Checksum-Md5": md5sum,
+            "X-JFrog-Art-Api": token
+        }
+
+        response = requests.put(dest, headers=headers, data=content)
+
+        if response.ok:
+            url = response.json().get("downloadUri")
```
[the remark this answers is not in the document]
ionela-voinescu: No, it cannot. It always returns a response object.

tools/bisector/bisector/bisector.py
```diff
@@ -5138,10 +5251,21 @@ def _main(argv):
...
     SHOW_TRACEBACK = args.debug
     service_hub = ServiceHub()
+
+    artifacts_service = None
     try:
-        service_hub.register_service('upload', ArtifactorialService())
-    except Exception as e:
-        info('Artifactorial upload will not be available: ' + str(e))
+        artifacts_service = ArtifactoryService()
+        info('Artifactory Service was initialized.')
```
douglas-raillard-arm, line 5258: Should be moved out in an `else` clause of the try/except/else statement as we only want to catch exceptions raised by `ArtifactoryService()`

tools/bisector/bisector/bisector.py
```diff
@@ -5138,10 +5251,21 @@ def _main(argv):
...
+    except:
```
douglas-raillard-arm, line 5259: bare `except:` are a very bad idea. They will catch any subclass of `BaseException`, which include things like `KeyboardInterrupt`. In this sort of context, the most general exception that should be caught is `Exception`

https://docs.python.org/3/library/exceptions.html#exception-hierarchy

tools/bisector/bisector/bisector.py
```diff
@@ -5138,10 +5251,21 @@ def _main(argv):
...
+        try:
+            artifacts_service = ArtifactorialService()
+            info('Artifactorial Service was initialized.')
+        except:
```
douglas-raillard-arm, line 5263: ditto

douglas-raillard-arm (changes requested):

PR #1775, opened 2021-11-15
sched_tp: Handle a couple of compilation errors

qais-yousef:
sched_cpu_capacity depends on a tracepoint that was added on 5.10, so
protect it with the right kernel version.

Also the tracepoint makes a call to arch_scale_freq_capacity() which
doesn't compile on x86 as it points to non-exported function. Just wrap
it with a warning until someone who cares finds a better alternative for
those archs.

### DIFF
--- a/lisa/_assets/kmodules/sched_tp/sched_events.h
+++ b/lisa/_assets/kmodules/sched_tp/sched_events.h
@@ -301,6 +301,7 @@ TRACE_EVENT_CONDITION(uclamp_util_cfs,
 #define trace_uclamp_util_cfs_enabled() false
 #endif /* CONFIG_UCLAMP_TASK */
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,10,0) && defined(CONFIG_ARM64)
 TRACE_EVENT(sched_cpu_capacity,
 
 	TP_PROTO(struct rq *rq),
@@ -327,6 +328,7 @@ TRACE_EVENT(sched_cpu_capacity,
 	TP_printk("cpu=%d capacity=%lu capacity_orig=%lu capacity_curr=%lu",
 		  __entry->cpu, __entry->capacity, __entry->capacity_orig, __entry->capacity_curr)
 );
+#endif
 
 #endif /* _SCHED_EVENTS_H */
 
--- a/lisa/_assets/kmodules/sched_tp/sched_tp.c
+++ b/lisa/_assets/kmodules/sched_tp/sched_tp.c
@@ -160,10 +160,12 @@ static void sched_util_est_se(void *data, struct sched_entity *se)
 		_trace_se(se, trace_sched_util_est_se);
 }
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,10,0) && defined(CONFIG_ARM64)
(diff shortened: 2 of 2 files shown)


tools/kmodules/sched_tp/sched_events.h
```diff
@@ -315,7 +316,13 @@ TRACE_EVENT(sched_cpu_capacity,
 	),
 
 	unsigned long scale_cpu = rq->cpu_capacity_orig;
+#ifdef CONFIG_ARM64
 	unsigned long scale_freq = arch_scale_freq_capacity(rq->cpu);
+#else
+#warning "arch_scale_freq_capacity() support on non ARM64 platforms needs attention"
+	unsigned long scale_freq = 0;
```
douglas-raillard-arm, line 323: I'm not sure this default is sane. At the end of the day, if we cannot scale for the frequency the `sched_cpu_capacity` will be a lie so we might as well just not define the event, or at least not the `capacity_curr` field.

qais-yousef: This not supposed to be a default nor sane :-)

What will happen if we remove capacity_curr? Wouldn't LISA fail because it can't find it? Similarly if we remove the event altogether?

Which brings an interesting question if we should just emit an error if we can't compile sched_cpu_capcity in older kernels?

The question really is how much of a hard requirement we want to enforce certain conditions. In this particular case; it made sense to emit a warning and carry on so that one can still use it; albeit with a caveat. You can't miss the warnings when you build it; and you'd read always 0 in case someone really missed the warnings. 2 should give obvious indications for the users of these platforms; which are not a lot. Hopefully this can promote them to send proper fixes for what works for them. It might require kernel side changes or use a different function for these archs.

In my case I was just testing things on x86 and I needed the module to produce the events; wasn't running anything with LISA.

douglas-raillard-arm: > This not supposed to be a default nor sane :-)

```
#ifdef anything
the_main_bit();
#else 
	unsigned long scale_freq = 0;
#endif
```
the `scale_freq = 0` is very much a default for anything that is not `x86_64`.

> What will happen if we remove capacity_curr? Wouldn't LISA fail because it can't find it?

If by "fail" you mean unusable, no. The pieces of lisa that actually need `capacity_curr` would rightfully complain. Feeding a value of 0 to shut it up is not going to bring anything good to any code that actually needs to know the capacity_curr. All it will do is "hide" the problem and make some computation barf in a non obvious way. We don't want that at all.
Now the question is whether to remove the event as a whole or just that specific field. Since the rest of the event is still usable and might be enough for some users we could simply remove the field.

> Which brings an interesting question if we should just emit an error if we can't compile sched_cpu_capcity in older kernels?

The comment was specifically targeted at the alternative path of `#ifdef CONFIG_ARM64`. What happens to old kernels is sane as we just don't add the event. If something does require the event, `FtraceCollector` will complain about it. We might want to add descriptive `#warning` if we want to have a better debug log but it should be obvious to anyone coming to this source code what happened anyway.

douglas-raillard-arm (changes requested):
The PR will also need to be rebased on current master. In case git rebase is not smart enough, the module sources have been moved to `lisa/_assets/kmodules/` with a symlink left at `tools/kmodules`.

PR #1787, opened 2021-11-30
Fix compilation errors on android v5.4

qais-yousef:
This is a tricky one. sched_nr_running I believe is optional, but @ionela-voinescu reported that not including sched_util_est_{se,cfs} leads to these errors.

```
[2021-11-29 21:59:22,712][EXEKALL] ERROR  ValueError: Events are missing in the kernel. Enable kmod_auto_load=True to attempt setting them up: (sched_cpu_capacity and sched_cpu_capacity and sched_util_est_cfs and sched_util_est_cfs and sched_util_est_se and sched_util_est_se)
```

It's either we fail to build the module for these required events. Or make LISA less fussy about them being missing and just skip the tests that require them.

### DIFF
--- a/lisa/_assets/kmodules/sched_tp/sched_events.h
+++ b/lisa/_assets/kmodules/sched_tp/sched_events.h
@@ -148,6 +148,7 @@ TRACE_EVENT(sched_overutilized,
 		  __entry->overutilized, __entry->span)
 );
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
 TRACE_EVENT(sched_update_nr_running,
 
 	    TP_PROTO(int cpu, int change, unsigned int nr_running),
@@ -227,6 +228,7 @@ TRACE_EVENT(sched_util_est_cfs,
 		  __entry->cpu, __entry->path, __entry->enqueued,
 		 __entry->ewma, __entry->util)
 );
+#endif
 
 #ifdef CONFIG_UCLAMP_TASK
 
--- a/lisa/_assets/kmodules/sched_tp/sched_tp.c
+++ b/lisa/_assets/kmodules/sched_tp/sched_tp.c
@@ -138,6 +138,7 @@ static void sched_overutilized(void *data, struct root_domain *rd, bool overutil
 	}
 }
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
 static void sched_update_nr_running(void *data, struct rq *rq, int change)
 {
 	if (trace_sched_update_nr_running_enabled()) {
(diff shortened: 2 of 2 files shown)


douglas-raillard-arm:
We clearly shouldn't fail to build the module just because of that since it contains many other unrelated events. Tests that require the events are util est tests, so they are expected to raise on systems that don't have util est (or no util est in the same form as current mainline does). Since there is no real way of checking what kind of util est a kernel has (if any) we cannot really do any better than that.

qais-yousef:
> We clearly shouldn't fail to build the module just because of that since it contains many other unrelated events. Tests that require the events are util est tests, so they are expected to raise on systems that don't have util est (or no util est in the same form as current mainline does). Since there is no real way of checking what kind of util est a kernel has (if any) we cannot really do any better than that.

Hmm. Should we add a build flag to force building these events if someone took the trouble to back port these required deps for these trace events to compile?

qais-yousef:
> > We clearly shouldn't fail to build the module just because of that since it contains many other unrelated events. Tests that require the events are util est tests, so they are expected to raise on systems that don't have util est (or no util est in the same form as current mainline does). Since there is no real way of checking what kind of util est a kernel has (if any) we cannot really do any better than that.
> 
> Hmm. Should we add a build flag to force building these events if someone took the trouble to back port these required deps for these trace events to compile?

It seems @ionela-voinescu can carry out of tree kernel patches to have these events provided by the kernel instead from the sched_tp. So we're good.

douglas-raillard-arm:
> Hmm. Should we add a build flag to force building these events if someone took the trouble to back port these required deps for these trace events to compile?

Ideally the kernel would have a builtin way of declaring versioned features so that this can be detected at runtime. There is no unified framework but if e.g. a macro (in a public header) or a struct is only defined when the kernel has util est, then it could be used as a way of detecting that.

douglas-raillard-arm:
Also it should be noted that from the test perspective, we only support kernels with the behavior of the current mainline kernel, so even if we have the trace event, it does not mean the tests will do anything useful. The problem still stands for other things like e.g. cpu capacity though. Kernel version checks are sadly only an approximation of feature set, feature backports and Android land blur the lines

qais-yousef:
> Also it should be noted that from the test perspective, we only support kernels with the behavior of the current mainline kernel, so even if we have the trace event, it does not mean the tests will do anything useful. The problem still stands for other things like e.g. cpu capacity though. Kernel version checks are sadly only an approximation of feature set, feature backports and Android land blur the lines

Fair enough. I think we asked the right questions and I don't have a preference to what needs to be done to be honest. As long as we explored them. It sounds good to me. It'll be hard to get it right for every thing sadly, as you say.

PR #1799, opened 2022-02-17
Async wlgen API

douglas-raillard-arm:
Add an async wlgen API as an alternative to the phase tree


douglas-raillard-arm:
Superseded by https://github.com/ARM-software/lisa/pull/1808

PR #1807, opened 2022-03-23
lisa.wa: Add collector for sysfs-extractor

vdonnefort:
WA has a sysfs-extractor which copies the content of a folder before, after
and generates a diff of it. The collector 'sysfs-extractor' leverages those
artifacts.

Here's an example for the Pixel6 energy meters:

def pixel6_energy_meter(df):
    # Keep only CPU's meters
    df = df[df.value.str.contains('S4M_VDD_CPUCL0|S3M_VDD_CPUCL1|S2M_VDD_CPUCL2')]
    df[['variable', 'value']] = df.value.str.split(', ', expand=True)
    df['unit'] = "bogo-joules"

    return df

df = WAOutput('pixel6/wa_output').get_collector(
	'sysfs-extractor',
        path='/sys/bus/iio/devices/iio:device0/energy_value',
        df_postprocess=pixel6_energy_meter
     ).df

### DIFF
--- a/lisa/wa.py
+++ b/lisa/wa.py
@@ -645,3 +645,89 @@ class WAJankbenchCollector(WAArtifactCollectorBase):
             agg_cols=['iteration', 'frame_id'],
             **kwargs,
         )
+
+
+class WASysfsExtractorCollector(WAArtifactCollectorBase):
+    """
+    WA collector for the syfs-extractor augmentation.
+
+    **Example**::
+
+        def pixel6_energy_meter(df):
+            # Keep only CPU's meters
+            df = df[df.value.str.contains('S4M_VDD_CPUCL0|S3M_VDD_CPUCL1|S2M_VDD_CPUCL2')]
+            df[['variable', 'value']] = df.value.str.split(', ', expand=True)
+
+            def _clean_variable(variable):
+                if 'S4M_VDD_CPUCL0' in variable:
+                    return 'little-energy'
+                if 'S3M_VDD_CPUCL1' in variable:
+                    return 'mid-energy'
+                if 'S2M_VDD_CPUCL2' in variable:
+                    return 'big-energy'
+                return ''
+
+            df['variable'] = df['variable'].apply(_clean_variable)
+            df['value'] = df['value'].astype(int)
+            df['unit'] = "bogo-ujoules"
+
+            # Add a total energy variable
+            df = pd.concat([
+                df,
+                pd.DataFrame(data={
+                    'variable': 'total-energy',
+                    'value': [df['value'].sum()]
+                })
+            ])
+            df.ffill(inplace=True)
+
+            return df
+
+        df = WAOutput('.').get_collector(
+                'sysfs-extractor',
+                path='/sys/bus/iio/devices/iio:device0/energy_value',
+                df_postprocess=pixel6_energy_meter
+        ).df
+    """
+    NAME = 'sysfs-extractor'
+
+    def __init__(self, wa_output, path, type='diff', **kwargs):
+        allowed_types = ['diff', 'before', 'after']
+
+        if type not in allowed_types:
+            ValueError(f'{self.__class__.__qualname__} type must be one of {allowed_types}')
+
+        self._ARTIFACT_NAME = f'{path} [{type}]'
+
+        # TODO: WA's sysfs-extractor augmentation can actually use a path
+        # to collect several files not only one...
+        self._filename = os.path.basename(path)
+
+        super().__init__(wa_output, **kwargs)
+
+    def _get_artifact_df(self, path):
+        # WA given path is actually dirname
+        path = os.path.join(path, self._filename)
+
+        # Expects a sysfs/procfs file e.g
+        # $ cat /sys/bus/iio/devices/iio:device0/energy_value
+        # t=199784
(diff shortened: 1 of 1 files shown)


lisa/wa.py
```diff
@@ -645,3 +645,38 @@ def get_stats(self, **kwargs):
...
         )
+
+class WASysfsExtractorCollector(WAArtifactCollectorBase):
+    NAME = 'sysfs-extractor'
+
+    def __init__(self, wa_output, path, type='diff', df_postprocess=None):
+        allowed_types = ['diff', 'before', 'after']
+
+        if type not in allowed_types:
+            self.logger.warn(
```
douglas-raillard-arm, line 656: Should raise `ValueError(f'{self.__class__.__qualname__} type must be one of {allowed_types}')`

lisa/wa.py
```diff
@@ -645,3 +645,38 @@ def get_stats(self, **kwargs):
...
+                f'SysfsExtractor type must be one of {allowed_types}')
+            type = 'diff'
+
+        self._ARTIFACT_NAME = path + ' [' + type + ']'
```
douglas-raillard-arm, line 660: nit: you can use f-strings, i.e. `f'{path}[{type}]'`

lisa/wa.py
```diff
@@ -645,3 +645,38 @@ def get_stats(self, **kwargs):
             agg_cols=['iteration', 'frame_id'],
             **kwargs,
         )
+
+class WASysfsExtractorCollector(WAArtifactCollectorBase):
+    NAME = 'sysfs-extractor'
+
+    def __init__(self, wa_output, path, type='diff', df_postprocess=None):
```
douglas-raillard-arm, line 652: Unless `df_postprocess` is directly used here, it's probably best to just use `**kwargs`

lisa/wa.py
```diff
@@ -645,3 +645,38 @@ def get_stats(self, **kwargs):
...
+        super().__init__(wa_output, df_postprocess=df_postprocess)
+
+    def _get_artifact_df(self, path):
+        # WA given path is actually dirname
+        path = os.path.join(path, self._filename)
+
+        with open(path, 'r') as f:
+            raw_file = f.readlines()
+
+        raw_file = [line.strip() for line in raw_file]
```
douglas-raillard-arm, line 675: If the format is simply one numeric value per line, `pd.read_csv` would be faster and probably lower the peak memory usage

lisa/wa.py
```diff
@@ -645,3 +645,38 @@ def get_stats(self, **kwargs):
             agg_cols=['iteration', 'frame_id'],
             **kwargs,
         )
+
+class WASysfsExtractorCollector(WAArtifactCollectorBase):
```
douglas-raillard-arm, line 649: missing docstring

douglas-raillard-arm (changes requested):
A few small things but overall looks good

lisa/wa.py
vdonnefort: It's supposed to read any type of sysfs/procfs file so the format can be anything.

lisa/wa.py
```diff
@@ -645,3 +645,75 @@ def get_stats(self, **kwargs):
...
+            df['value'] = df['value'].astype(int)
+            df['unit'] = "bogo-ujoules"
+
+            # Add a total energy variable
+            df = pd.concat([
+                df, pd.DataFrame(data={
+                    'variable': 'total-energy',
+                    'value': [df['value'].sum()]
+                })]
+            )
```
douglas-raillard-arm, line 680: nit: the indentation/parenthesing/line return is a bit of a mess:
```
df = pd.concat((
    df,
    pd.DataFrame(data={
        'variable': 'total-energy',
        'value': [df['value'].sum()]
    })
))
```

Also seems to be missing a `unit` column

lisa/wa.py
```diff
@@ -645,3 +645,75 @@ def get_stats(self, **kwargs):
...
+            df.ffill(inplace=True)
+
+            return df
+
+        df = WAOutput('.').get_collector('sysfs-extractor',
```
douglas-raillard-arm, line 685:
```
df = WAOutput('.').get_collector(
    'sysfs-extractor',
    path='/sys/bus/iio/devices/iio:device0/energy_value',
    df_postprocess=pixel6_energy_meter
).df
```

lisa/wa.py
```diff
@@ -645,3 +645,75 @@ def get_stats(self, **kwargs):
...
+
+        with open(path, 'r') as f:
+            raw_file = f.readlines()
+
+        raw_file = [line.strip() for line in raw_file]
+
+        df = pd.DataFrame(data={
+                'variable': self._filename,
+                'value': raw_file,
+                'unit': 'unknown'
```
douglas-raillard-arm, line 717: AFAICT the other collectors use `''` for unknown units. There is also a chance it can work without the column at all (but maybe not, I can't remember)

lisa/wa.py
douglas-raillard-arm: Can you add in comment in the file a snippet of "raw_file" please ? It's useful to know what is to be expected, and if it's something like `path/to/file: the content` we probably want to parse it and separate the path and content part in 2 columns.

PR #1816, opened 2022-04-01
Fixes for peripheral clock

Leo-Yan:
This patch set is to fixing peripheral clock analysis and support plotting the frequency.

The first two patches are to fix the clock event analysis, so that we can get the correct data frame for peripheral clocks.  The third patch adds method for plotting the peripheral clock frequency.

### DIFF
--- a/lisa/analysis/frequency.py
+++ b/lisa/analysis/frequency.py
@@ -338,15 +338,19 @@ class FrequencyAnalysis(TraceAnalysisBase):
         return series_mean(freq)
 
     @TraceAnalysisBase.cache
-    @requires_events('clock_set_rate', 'clock_enable', 'clock_disable')
+    @requires_events('clk_set_rate', 'clk_enable', 'clk_disable')
     def df_peripheral_clock_effective_rate(self, clk_name):
-        rate_df = self.trace.df_event('clock_set_rate')
-        enable_df = self.trace.df_event('clock_enable')
-        disable_df = self.trace.df_event('clock_disable')
+        rate_df = self.trace.df_event('clk_set_rate')
+        enable_df = self.trace.df_event('clk_enable')
+        disable_df = self.trace.df_event('clk_disable')
 
-        freq = rate_df[rate_df.clk_name == clk_name]
-        enables = enable_df[enable_df.clk_name == clk_name]
-        disables = disable_df[disable_df.clk_name == clk_name]
+        # Add 'state' for enable and disable events
+        enable_df['state'] = 1
+        disable_df['state'] = 0
+
+        freq = rate_df[rate_df['name'] == clk_name]
+        enables = enable_df[enable_df['name'] == clk_name]
+        disables = disable_df[disable_df['name'] == clk_name]
 
         freq = pd.concat([freq, enables, disables], sort=False).sort_index()
         freq['start'] = freq.index
@@ -357,7 +361,7 @@ class FrequencyAnalysis(TraceAnalysisBase):
         freq.ffill(inplace=True)
         freq['effective_rate'] = np.where(
             freq['state'] == 0, 0,
-            np.where(freq['state'] == 1, freq['state'], float('nan'))
+            np.where(freq['state'] == 1, freq['rate'], float('nan'))
         )
         return freq
 
--- a/lisa/analysis/frequency.py
+++ b/lisa/analysis/frequency.py
@@ -573,4 +573,31 @@ class FrequencyAnalysis(TraceAnalysisBase):
             )
         ).cols(1)
 
+    @TraceAnalysisBase.plot_method
+    @df_peripheral_clock_effective_rate.used_events
+    def plot_peripheral_frequency(self, clk_name, average: bool=True):
+        """
+        Plot frequency for the specified peripheral clock frequency
+
+        :param clk_name: The clock name for which to plot frequency
+        :type clk_name: str
+
+        :param average: If ``True``, add a horizontal line which is the
+            frequency average.
+        :type average: bool
+
+        """
+        df = self.df_peripheral_clock_effective_rate(clk_name)
+
+        freq = series_refit_index(df['effective_rate'], window=self.trace.window)
+        avg = series_mean(freq)
+
+        df = df_refit_index(df, window=self.trace.window)
+        fig = plot_signal(df['effective_rate'], name=f'Frequency of {clk_name} (Hz)')
+
+        if avg > 0:
+            fig *= hv.HLine(avg, group='average').opts(color='red')
+
+        return fig
+
 # vim :set tabstop=4 shiftwidth=4 expandtab textwidth=80
--- a/lisa/trace.py
+++ b/lisa/trace.py
@@ -3920,11 +3920,9 @@ class Trace(Loggable, TraceBase):
                 available_events=self.available_events,
             )
 
-        # Ensure the event name is set, as this sort of metadata might not be
-        # saved in the swap
-        df.name = event
-        # This is how it should be done. Keep the previous line in case someone
-        # depends on it.
+        # We used to set a ".name" attribute, but:
+        # 1. There is no central way of saving these metadata when serializing
+        # 2. It conflicts with a "name" column in the dataframe.
         df.attrs['name'] = event
         return df
 
--- a/doc/plot_conf.yml
+++ b/doc/plot_conf.yml
@@ -19,6 +19,9 @@ doc-plot-conf:
             plat_info: &plat_info0 !call:lisa.platforms.platinfo.PlatformInfo.from_yaml_map
                 path: !env:interpolate $LISA_HOME/doc/traces/plat_info.yml
 
+            plat_info1: &plat_info1 !call:lisa.platforms.platinfo.PlatformInfo.from_yaml_map
+                path: !env:interpolate $LISA_HOME/doc/traces/juno_r2_plat_info.yml
+
             trace: &trace0 !call:lisa.trace.Trace
                 trace_path: !env:interpolate $LISA_HOME/doc/traces/trace.dat
                 normalize_time: true
@@ -101,6 +104,15 @@ doc-plot-conf:
         FrequencyAnalysis.plot_cpu_frequency_residency:
             kwargs:
                 cpu: *cpu0
+
+        FrequencyAnalysis.plot_peripheral_frequency:
+            trace: !call:lisa.trace.Trace
+                trace_path: !env:interpolate $LISA_HOME/doc/traces/trace_peripheral_clk.dat
+                normalize_time: true
+                # Either refer to the existing anchor, or make a new plat_info for the board that was used
+                plat_info: *plat_info1
+            kwargs:
+                clk_name: juno:uartclk
 
         CpusAnalysis.plot_orig_capacity:
             kwargs:


lisa/analysis/rta.py
```diff
@@ -269,7 +269,7 @@ def __missing__(self, key):
                    raise ValueError(f'Unexpected phase number "{key}"')
 
             def f(df):
-                pid, comm = df.name
```
douglas-raillard-arm, line 272: This case is "interestingly" not affected. Pandas manages to set this attribute from `DataFrame.apply()` in a "special" way that will not set a column named "name". This means that in that specific case, `df.name != df['name']`.

lisa/trace.py
```diff
@@ -3921,10 +3921,10 @@ def df_event(self, event, raw=None, window=None, signals=None, signals_init=True
 
         # Ensure the event name is set, as this sort of metadata might not be
         # saved in the swap
-        df.name = event
```
douglas-raillard-arm, line 3924: And once more pandas gets in the way of reliable code. As the next line suggests, this was already left for backward compat only. The "new" way of adding custom metadata is by using `attrs` as done below:
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.attrs.html

I'll make a fix that solves that problem in LISA, following which I expect we will be able to drop this patch in the PR.

lisa/trace.py
```diff
@@ -3921,10 +3921,10 @@ def df_event(self, event, raw=None, window=None, signals=None, signals_init=True
...
+        df.event_name = event
         # This is how it should be done. Keep the previous line in case someone
         # depends on it.
-        df.attrs['name'] = event
```
douglas-raillard-arm, line 3927: that's actually not an issue, attrs is a dict that is not conflicting in any way with the actual dataframe columns (that's probably one of the reason it was introduced):
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.attrs.html

lisa/analysis/frequency.py
```diff
@@ -338,15 +338,19 @@ def get_average_cpu_frequency(self, cpu):
...
+        disable_df = self.trace.df_event('clk_disable')
 
-        freq = rate_df[rate_df.clk_name == clk_name]
-        enables = enable_df[enable_df.clk_name == clk_name]
-        disables = disable_df[disable_df.clk_name == clk_name]
+        # Add 'state' for enable and disable events
+        enable_df['state'] = 1
+        disable_df['state'] = 0
+
+        freq = rate_df[rate_df.name == clk_name]
```
douglas-raillard-arm, line 351: column access must be done with `[]`, e.g. `rate_df]'name']`. Pandas has a quite flaky API so doing that prevents a bunch of issues and make some others more clear.

lisa/analysis/frequency.py
```diff
@@ -569,4 +573,34 @@ def plot_domain_frequency_transitions(self, pct: bool=False):
...
+            frequency average.
+        :type average: bool
+
+        """
+        logger = self.logger
+        df = self.df_peripheral_clock_effective_rate(clk_name)
+
+        freq = series_refit_index(df['effective_rate'], window=self.trace.window)
+        avg = series_mean(freq)
+        logger.info(
```
douglas-raillard-arm, line 595: Plot methods should be pure, aka only return a plot. If there is useful information that is non trivial to extract from the dataframes, it can be done in separate methods (or maybe even properties ?)

In this specific case I don't think it's really needed as:
1. the info is trivial to compute from the df
2. the tooltip for the HLine on the plot should provide the value (if someone is still using matplotlib backend, they really, really should move to bokeh backend to get interactive stuff like that)

lisa/analysis/frequency.py
```diff
@@ -569,4 +573,34 @@ def plot_domain_frequency_transitions(self, pct: bool=False):
...
+            "Average frequency for {} : {:.3f} GHz".format(clk_name, avg / 1e6))
+
+        df = df_refit_index(df, window=self.trace.window)
+        fig = plot_signal(df['effective_rate'], name=f'Frequency of {clk_name} (Hz)')
+
+        if average and avg > 0:
```
douglas-raillard-arm, line 601: `avg > 0` can be simplified into `avg`, in case it becomes negative for some reason (would probably be a bug but at least it would make it visible)

douglas-raillard-arm (changes requested):
A few nits but otherwise LGTM. I'll make a PR to solve the `df.name` issue today, I'll report back here when it's merged so you can rebase on latest master.

Thanks for your contribution !

douglas-raillard-arm:
Turns out the `df.name` issue was already solved, the only reason we still set the attribute is for backward compat with user code, so I simply dropped it:
https://github.com/ARM-software/lisa/pull/1817 

By now plotting libraries etc should be able to understand `df.attrs` I hope.

lisa/analysis/frequency.py
Leo-Yan: Okay, will drop the info.

douglas-raillard-arm:
> Seems like the checking failure is caused by the trace data misses clock related ftrace events. Please let me know if anything need me to follow up.

The documentation builds example plots as well so it fails because it does not have the events in the trace used. In order to enable that, we would need:
* A trace containing the event in `doc/traces/`
* Ideally collected on a juno-r0. That would allow sharing the same `doc/traces/plat_info.yml` file. If that is not possible, you will need to run `lisa-platinfo-extract` to get a plat_info.yml and clean the output [1].
* Update `doc/plot_conf` to specify the parameters to pass to the plot method when building the doc, e.g.:
```diff
diff --git a/doc/plot_conf.yml b/doc/plot_conf.yml
index 2fef4d415..bd7391b2b 100644
--- a/doc/plot_conf.yml
+++ b/doc/plot_conf.yml
@@ -102,6 +102,14 @@ doc-plot-conf:
             kwargs:
                 cpu: *cpu0
 
+        FrequencyAnalysis.plot_peripheral_frequency:
+            kwargs:
+                trace: !call:lisa.trace.Trace
+                    trace_path: !env:interpolate $LISA_HOME/doc/traces/trace.dat
+                    normalize_time: true
+                    # Either refer to the existing anchor, or make a new plat_info for the board that was used
+                    plat_info: *plat_info0
+
         CpusAnalysis.plot_orig_capacity:
             kwargs:
                 cpu: *cpu0
```


[1] The plat_info.yml will need to be cleaned up so that all the `!python/object` tags are removed and replaced by a public API call as seen in the current `doc/traces/plat_info.yml`. 
For example, the `version` key is not a !python/object` dump of the internal state of `KernelVersion` but is a call to the constructor, which takes a string as parameter.
```yaml
            version: !call:devlib.target.KernelVersion
                version_string: '5.12.0-rc8-g34ced2a7bbf1 #1 SMP PREEMPT Fri Jun 11 15:15:11 BST 2021'
``` 
The reason we do that is to avoid relying on some unstable internal object representation that could break at any time. Instead, we use the Python public APIs to build instances of the objects using a stable interface.

Leo-Yan:
Have rebased patches based on the PR #1817.  I can confirm that PR#1817 fixes the issue.

Regard your suggestion for the trace.dat, it contians many other events.  So what's the suggested way to record ftrace data?  I am a bit concern that my captured trace data will break other testing cases.  So one way is you could share the suggested command for recording the ftrace data and then I get the trace data from my Juno-r2 board?

Alternatively, could you help to get the trace data at your side?

douglas-raillard-arm:
>  it contians many other events
not an issue per say, maybe we want to trim it down a bit to avoid a gigantic file but that's all. ``trace-cmd`` has the ability to split a trace.dat into smaller parts in case the trace is needlessly long for the plot example

>  I am a bit concern that my captured trace data will break other testing cases
With the snippet of plot_conf I shared, this trace would only be used for this specific plot (obviously it would need to have another name, like ``doc/traces/trace_peripheral_clocks.dat``. All the other plots would keep using the default trace that is currently there. 

> Alternatively, could you help to get the trace data at your side?
If you have e.g. an rt-app experiment that generates an "interesting" example trace I could do that relatively easily (bonus point if Juno R0 works for that to avoid cleaning up a new plat_info.yml). For that, I would use the `lisa.trace.FtraceCollector` API like here: https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/workflows/ipynb/examples/typical_experiment.html#Running-the-workload (ipynb/examples/typical_experiment.ipynb in the git repo)

The actual way you generate your trace.dat does not really matters at the end of the day, as long as the resulting file is not gigantic.

Leo-Yan:
Thanks a lot for the suggestion @douglas-raillard-arm.  Seems finally the testing is passed :)  I followed your suggestion to create separate juno_r2_plat_info.yml and trace data file, so these two files are used for clock event testing.

I am curious what's the command I should use to test for 'doc/plot_conf.yml', I tried two below commands but neither is used for testing doc/plot_conf.yml:

~~~
[LISAShell lisa-us] \> exekall run lisa.tests --conf /home/leoy/Work2/tools/lisa-us/target_conf.yml --symlink-artifact-dir-to home/leoy/Work2/tools/lisa-us/results_latest --share \*.Target --select-multiple --list
[LISAShell lisa-us] \> bash ./tools/tests.s
~~~

douglas-raillard-arm:
> I am curious what's the command I should use to test for 'doc/plot_conf.yml', I tried two below commands but neither is used for testing doc/plot_conf.yml:

Building the doc (e.g. `source init_env && lisa-doc-build` or manually calling sphinx) makes use of it. Sphinx allows custom hooks [1] to update docstrings, so we take advantage of that and include some raw HTML (holoviews plot rendered with bokeh backend). The outcome is in the analysis page of the doc, e.g. https://lisa-linux-integrated-system-analysis.readthedocs.io/en/master/trace_analysis.html#lisa.analysis.frequency.FrequencyAnalysis.plot_cpu_frequencies

`tools/tests.sh` simply builds the doc to ensure it builds without warning and also checks pre-built manpages are up-to-date wrt their reStructuredText sources.

[1] https://github.com/ARM-software/lisa/blob/62268e46bea6adc79f29f3cef70044b8f50657b6/doc/conf.py#L468

douglas-raillard-arm:
LGTM, thanks for this new analysis method :)

douglas-raillard-arm:
@Leo-Yan I just realized that on my kernel [1] the events are named `clock_*` and not `clk_*`. Is it a recent upstream change ?

[1] Ubuntu 20.04, kernel Linux e126130 5.13.0-35-generic #40~20.04.1-Ubuntu SMP Mon Mar 7 09:18:32 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

douglas-raillard-arm:
I checked the latest mainline kernel and found the following events in `./include/trace/events/power.h`:

```
DEFINE_EVENT(clock, clock_enable,

	TP_PROTO(const char *name, unsigned int state, unsigned int cpu_id),

	TP_ARGS(name, state, cpu_id)
);

DEFINE_EVENT(clock, clock_disable,

	TP_PROTO(const char *name, unsigned int state, unsigned int cpu_id),

	TP_ARGS(name, state, cpu_id)
);

DEFINE_EVENT(clock, clock_set_rate,

	TP_PROTO(const char *name, unsigned int state, unsigned int cpu_id),

	TP_ARGS(name, state, cpu_id)
);
```

So I'll change the event names used by LISA since we want to match mainline. If your kernel cannot be modified to match mainline we can always add a compat layer like we did for PELT events.

Leo-Yan:
> So I'll change the event names used by LISA since we want to match mainline. If your kernel cannot be modified to match mainline we can always add a compat layer like we did for PELT events.

No, when you search the kernel for these three trace events, they are not used anymore:

~~~
leoy@leo-build-box:~/Work2/linux$ git grep trace_clock_enable
leoy@leo-build-box:~/Work2/linux$ git grep trace_clock_disable
leoy@leo-build-box:~/Work2/linux$ git grep trace_clock_set_rate
~~~

douglas-raillard-arm:
Indeed, I did not notice your previous message while posting mine. It indeed looks like the `clock_*` variant is somewhat deprecated. Thanks for pointing this out

Leo-Yan:
> Indeed, I did not notice your previous message while posting mine. It indeed looks like the `clock_*` variant is somewhat deprecated. Thanks for pointing this out

Welcome!  :)

PR #1818, opened 2022-04-04
lisa._kmod: Use HOSTCC == CC

douglas-raillard-arm:
FEATURE

When possible, use HOSTCC == CC to avoid having to install two different
compilers.


douglas-raillard-arm:
Closing as we already specify LLVM=1, which is a "superset" of HOSTCC=clang

PR #1819, opened 2022-04-05
lisa.datautils: Add SignalDesc for clk_* events

douglas-raillard-arm:
FIX

Add SignalDesc for clk_set_rate, clk_enable, clk_disable so that
requesting the dataframe for one of these will always provide the
occurence before the beginning of the trace window for each clock name.


douglas-raillard-arm:
@Leo-Yan this might affect the plot (in a good way) as it will ensure that e.g. `trace.df_event('clk_set_rate`) will provide an "initial condition" event for each clock name when windowing the trace. With the `df_refit_index` used in the plot method, the timestamp of that initial event will be moved to match the trace window start so the plot should be as expected.

Otherwise, the plot would only start on the first occurrence of the event inside the window.

Leo-Yan:
Thanks for the info, @douglas-raillard-arm.  Based  on your description, the change for "the initial event will be moved to match the trace window start" looks good to me.

PR #1821, opened 2022-04-06
Fix plot_peripheral_frequency

douglas-raillard-arm:
@Leo-Yan I made a few changes to plot_peripheral_frequency/df_peripheral_clock_effective_rate, let me know if everything still works for you.

Wrt to the trace for the documentation plot, it seems that `juno:uartclk` does not have any event for `clk_set_rate`, so I switched to `aplclk` to get a more interesting plot.

The plot can be quickly viewed (with that PR) using:
```
lisa-plot doc/traces/trace_peripheral_clk.dat --plot frequency:peripheral-frequency interactive -X clk_name aplclk
```

### DIFF
--- a/doc/plot_conf.yml
+++ b/doc/plot_conf.yml
@@ -112,7 +112,7 @@ doc-plot-conf:
                 # Either refer to the existing anchor, or make a new plat_info for the board that was used
                 plat_info: *plat_info1
             kwargs:
-                clk_name: juno:uartclk
+                clk_name: aplclk
 
         CpusAnalysis.plot_orig_capacity:
             kwargs:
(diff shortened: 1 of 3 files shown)


douglas-raillard-arm:
I'll go ahead and merge it today as this is delaying v3.0.0 release, please open an issue or PR if this change turns out to be problematic in your use case.

Leo-Yan:
> the reason I dropped the nested `np.where` is because it did not play well when `state` is NaN, leading to an `effective_rate` full of NaN even though the rate is actually known. This happened on the documentation trace for `juno:uartclk` since it does not contain any enable/disable event for it.

Yeah, this makes sense to me.  I verified the latest LISA master branch with a local trace data, both ipynb file and below command (and I really like the command) can work well at my side:

`LISAShell ipynb] \> lisa-plot trace.dat --plot frequency:peripheral-frequency interactive -X clk_name bimc_clk`

Thanks a lot for the fixing.  And sorry that I missed the github message for this thread.

PR #1850, opened 2022-06-21
tools/batch-rebase: Fix tag action

douglas-raillard-arm:
FIX

Fix tag action that were resulting in an exception: KeyError: 'tip'


mishradk:
This works for me.

PR #1854, opened 2022-06-21
WIP

douglas-raillard-arm:
Add a small C parser combinator API in order to write things like sysfs
files parsers. It's a bit clunky to manually use since C does not allow
closures but otherwise work and should be pretty efficient.


douglas-raillard-arm:
Closing as it's now part of: https://github.com/ARM-software/lisa/pull/1849

PR #1856, opened 2022-06-24
doc/setup.rst: Remove references to now defunct clean_module script

douglas-raillard-arm:
FIX

Remove reference to a script that does not exist anymore.


douglas-raillard-arm:
The tool actually still exists

PR #1861, opened 2022-07-13
Refresh rt-app

douglas-raillard-arm:
Fix tracefs hard coded path: https://github.com/scheduler-tools/rt-app/pull/126


douglas-raillard-arm:
These binaries have been tested on Juno R0 with integration kernel and everything seems to build and work fine.

Waiting on @credp Pixel 6 test to validate that everything is fine there as well before merging.

PR #1866, opened 2022-07-26
WIP lisa.energy_model: Use LISA kernel module fallback

douglas-raillard-arm:
TODO:

- [ ] Get code in the LISA kmod to create a folder tree matching the kernel behavior

FEATURE

If the energy model is not exposed on the target in sysfs, load the lisa
kernel module with the right feature enabled and retry. This feature
will expose the EM as a folder hierarchy using in-kernel APIs


douglas-raillard-arm:

PR #1870, opened 2022-07-29
lisa._assets.kmodules.sched_tp: Add EM sysfs tree for latest phones

lukaszluba-arm:
Latest Android removed debugfs dir and config. There is a need to expose
Energy Model in some other way. Use the kobject and sysfs infrastructure
to do that.

Signed-off-by: Lukasz Luba <lukasz.luba@arm.com>

### DIFF
--- a/lisa/_assets/kmodules/sched_tp/Makefile
+++ b/lisa/_assets/kmodules/sched_tp/Makefile
@@ -33,7 +33,7 @@ ifneq ($(KERNELRELEASE),)
 
 LISA_KMOD_NAME ?= sched_tp
 obj-m := $(LISA_KMOD_NAME).o
-$(LISA_KMOD_NAME)-y := main.o tp.o wq.o features.o pixel6.o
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
"cpu files" need to be added, as shown in @mrkajetanp's commit: https://github.com/ARM-software/lisa/pull/1871/commits/f6452b80bda964253b310adad4da67f480ba01d1

douglas-raillard-arm:
Also need to add symlinks in tools/kmodules for each source file this PR adds.

douglas-raillard-arm:
Looking at this doc, it seems like it's not necessary anymore:
https://source.android.com/docs/core/architecture/kernel/using-debugfs-12

So at least for Android 12 we should be able to mount it manually, which could be done upon target init using that PR: https://github.com/ARM-software/lisa/pull/2025

douglas-raillard-arm:

PR #1871, opened 2022-08-04
[v2] lisa._assets.kmodules.sched_tp: Add EM sysfs tree for latest phones

mrkajetanp:
From Lukasz's original PR:

Latest Android removed debugfs dir and config. There is a need to expose
Energy Model in some other way. Use the kobject and sysfs infrastructure
to do that.

This PR adds another commit on top of the previous one to include the missing 'cpus' files that are expected by Lisa's LinuxEnergyModel. This avoids Lisa errors and allows us to load the EM using `LinuxEnergyModel.from_target(target, directory='/sys/kernel/energy_model')` which would previously result in errors caused by the missing 'cpus' files.

Expands upon #1870.

### DIFF
--- a/lisa/_assets/kmodules/sched_tp/Makefile
+++ b/lisa/_assets/kmodules/sched_tp/Makefile
@@ -33,7 +33,7 @@ ifneq ($(KERNELRELEASE),)
 
 LISA_KMOD_NAME ?= sched_tp
 obj-m := $(LISA_KMOD_NAME).o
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
Closing as the PR this is based on is still under review, so I forwarded the comment: https://github.com/ARM-software/lisa/pull/1870#issuecomment-1208287806

PR #1875, opened 2022-08-09
lisa.target: Allow mount to work on Android

douglas-raillard-arm:
FIX

It has been reported that rs.media.module process is necessary for the
mount command to work on Android, so add it to the list of critical
tasks to not freeze.


douglas-raillard-arm:
This is the last non-merged part of https://github.com/ARM-software/lisa/pull/1746

PR #1907, opened 2022-11-04
lisa._assets.kmodules.sched_tp: Avoid using pahole --expanded_prefix

douglas-raillard-arm:
Remove the use of --expanded_prefix so that we can use a pahole version closer to mainstream (we still require --expand_types_once, but it should be easier to finish upstreaming).

Replace the effect of that option by a sed regex to rename all types, then rename back the types we did not want to rename in the first place (the ones listed in private_types.txt)

### DIFF
--- a/lisa/_assets/kmodules/sched_tp/Makefile
+++ b/lisa/_assets/kmodules/sched_tp/Makefile
@@ -54,6 +54,8 @@ else
     VMLINUX := $(_DWARF_VMLINUX)
 endif
 
+VMLINUX_H_TYPE_PREFIX=KERNEL_PRIVATE_
+
 $(VMLINUX_H): $(VMLINUX_TXT) $(VMLINUX)
 # Some options are not upstream (yet) but they have all be published on the
 # dwarves mailing list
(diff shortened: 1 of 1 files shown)


KJTsanaktsidis:
Hi @douglas-raillard-arm - I stumbled across this because I found your patch for `--expanded_prefix`  on the dwarves mailing list - https://lore.kernel.org/dwarves/20211207173151.2283946-1-douglas.raillard@arm.com/ 

This feature is precisely what I've spent all day looking for! I can't see from the mailing list thread why this didn't make its way upstream - is there some reason why this is fundamentally not a good idea? Or is it just a case of having the time to work on the string-manipulation stuff?

My use-case is that I'm working on (as in, only a few days of hacking!) profiling tools for the Ruby programming language. I want to extract user-space structure definitions about the Ruby VM from dwarf debuginfo into header files that can be included in eBPF prorams compiled with the libbpf toolchain, so that my eBPF program can introspect the state of the ruby VM in various ways.

I have the actual header files, but they're not at all suitable for including in an eBPF program; they contain a ton of type definitions (most coming from the C standard library headers) that conflict majorly with vmlinux.h generated by `bpftool`. My thinking was to use `pahole` to extract only the structure definitions I actually need out of the ruby binary. 

Having the functionality to do this in pahole would be really neat! Is there any way I could help to get some of these options upstreamed?

KJTsanaktsidis:
Oh. Hum. I thought about this a bit more and even with `--expand_prefix` I can't do what I want - plenty of the structures in question depend on structures from third party libraries, and we're not guaranteed to have the DWARF debuginfo for those libs.

Still, your patch is definitely cool!

PR #1913, opened 2022-11-16
lisa._assets.kmodules.sched_tp: Add sched_pelt_thermal trace event

deggeman:
The corresponding trace point is already supported in Linux kernel mainline:

Commit 77cf151b7bbdf ("sched/core: Export pelt_thermal_tp") Commit 765047932f153 ("sched/pelt: Add support to track thermal pressure")

### DIFF
--- a/lisa/_assets/kmodules/sched_tp/tp.c
+++ b/lisa/_assets/kmodules/sched_tp/tp.c
@@ -95,6 +95,18 @@ static void sched_pelt_irq_probe(struct feature *feature, struct rq *rq)
 }
 DEFINE_TP_EVENT_FEATURE(sched_pelt_irq, pelt_irq_tp, sched_pelt_irq_probe);
 
+static void sched_pelt_thermal_probe(struct feature *feature, struct rq *rq)
+{
+	const struct sched_avg *avg = sched_tp_rq_avg_thermal(rq);
+	int cpu = sched_tp_rq_cpu(rq);
+
+	if (!avg)
+		return;
+
+	trace_sched_pelt_thermal(cpu, avg);
+}
+DEFINE_TP_EVENT_FEATURE(sched_pelt_thermal, pelt_thermal_tp, sched_pelt_thermal_probe);
+
 static void sched_pelt_se_probe(struct feature *feature, struct sched_entity *se)
 {
 	_trace_se(se, trace_sched_pelt_se);
--- a/lisa/trace.py
+++ b/lisa/trace.py
@@ -1486,6 +1486,14 @@ class TxtTraceParser(TxtTraceParserBase):
                 'util': _KERNEL_DTYPE['util'],
             },
         ),
+        'sched_pelt_thermal': dict(
+            fields={
+                'cpu': _KERNEL_DTYPE['cpu'],
+                'load': _KERNEL_DTYPE['util'],
+                'rbl_load': _KERNEL_DTYPE['util'],
+                'util': _KERNEL_DTYPE['util'],
+            },
+        ),
         'sched_process_wait': dict(
             fields={
                 'comm': _KERNEL_DTYPE['comm'],
--- a/lisa/_assets/kmodules/sched_tp/ftrace_events.h
+++ b/lisa/_assets/kmodules/sched_tp/ftrace_events.h
@@ -95,6 +95,10 @@ DEFINE_EVENT(sched_pelt_rq_template, sched_pelt_irq,
 	TP_PROTO(int cpu, const struct sched_avg *avg),
 	TP_ARGS(cpu, avg));
 
+DEFINE_EVENT(sched_pelt_rq_template, sched_pelt_thermal,
+	TP_PROTO(int cpu, const struct sched_avg *avg),
+	TP_ARGS(cpu, avg));
+
 TRACE_EVENT(sched_pelt_se,
 
 	TP_PROTO(int cpu, char *path, char *comm, int pid, const struct sched_avg *avg),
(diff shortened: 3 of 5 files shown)


lisa/_assets/kmodules/sched_tp/tp.c
```diff
@@ -95,6 +95,18 @@ static void sched_pelt_irq_probe(struct feature *feature, struct rq *rq)
...
+{
+	const struct sched_avg *avg = sched_tp_rq_avg_thermal(rq);
+	int cpu = sched_tp_rq_cpu(rq);
+
+	if (!avg)
+		return;
+
+	trace_sched_pelt_thermal(cpu, avg);
+}
+DEFINE_TP_EVENT_FEATURE(sched_pelt_thermal, pelt_irq_tp, sched_pelt_thermal_probe);
```
douglas-raillard-arm, line 108: Could use a comment as to why the probe is attached to that irq related tp. Is it fundamentally looking at an irq-only signal or is it just a convenient tracepoint triggered "often enough" or something like that ?

douglas-raillard-arm (changes requested):
Would need an entry there as well
https://github.com/ARM-software/lisa/blob/06d12762d91a553799c863fe9f519c64dab44b66/lisa/datautils.py#L206

And there as well if we want parsing to be not extra slow:
https://github.com/ARM-software/lisa/blob/06d12762d91a553799c863fe9f519c64dab44b66/lisa/trace.py#L1473

lisa/_assets/kmodules/sched_tp/tp.c
deggeman: This is a copy&paste error. It didn't show up as an issue while testing it since the `tp_name` parameter of DEFINE_TP_ENABLE_DISABLE() is only there for testing whether a trace_point with this name exists. And `pelt_irq_tp` existed so everything went fine enabling the pelt_thermal_tp related trace_event. Has to be corrected to use `pelt_thermal_tp` instead of `pelt_irq_tp`. thermal and irq are independent EPLT trace signals, like dl and rt as well.

douglas-raillard-arm: The tracepoint is not just checked it it exists, it's also the one that the probe is attached as well but if they have the same signature that code must have "worked" in your test

douglas-raillard-arm:
Yes, the datatutils modification is necessary to have correct windowing behavior (i.e. have an event at the beginning of any window if the information exists). If we don't do that, any analysis like taking an average value or plotting a signal will be somewhat broken when using trace windows

douglas-raillard-arm:
@deggeman 
Looks like my link to datautils was pointing at the wrong line, the correct one is:
https://github.com/ARM-software/lisa/blob/06d12762d91a553799c863fe9f519c64dab44b66/lisa/datautils.py#L2034

lisa/trace.py
```diff
@@ -1486,6 +1486,14 @@ class TxtTraceParser(TxtTraceParserBase):
                 'util': _KERNEL_DTYPE['util'],
             },
         ),
+        'sched_pelt_thermal': dict(
+            fields={
+                'cpu': _KERNEL_DTYPE['cpu'],
+                'load': _KERNEL_DTYPE['util'],
+                'rbl_load': _KERNEL_DTYPE['util'],
+                'util': _KERNEL_DTYPE['util'],
+            },
```
douglas-raillard-arm, line 1495: The event has a `update_time` not listed here. That means the dataframe will not have this column. It's fine if it's somewhat useless for thermal signals (it's used to dump the PELT clock). 

Once you confirm that, it's good to merge

deggeman: 'update_time' () could be a valuable information for sched_pelt_thermal but it's use case is very limited.
We would have the same issue for sched_pelt_[rt\|dl\|irq]. They also don't list the 'update_time' field but it appears in the trace event.

We have another issue with these trace events. Since we use the sched_pelt_rq_template for all of them, we plot `load`, `rbl_load` and `util` although only one has meaning to each of them and the other 2 are undefined.

rt, dl, irq: util_avg (load_avg & runnable_avg are meaningless)
thermal: load_avg (util_avg and runnable_avg are meaningless)

I would say we use sched_pelt_thermal w/o 'update_time' for now.

And then later we have to make a decision to whether:
(1) do we want to add 'update_time' to all of them?
(2) do we want to show only the valid load/rbl_load/util value for those 
    4 trace_events?
(I'm not sure why load_avg was chosen for thermal? Should be util_avg IMHO.

douglas-raillard-arm: > and the other 2 are undefined.
Undefined as in: the member of `struct sched_avg` is never initialized so we get arbitrary junk ? Or is it just that the field is set to some value but never updated by the kernel to something sensible ?

Either way, it sounds a bit strange that we end up attaching sched_avg structs to runqueues if only part of the data is used, and it can actually be harmful if some junk data become accessbile. AFAIU people usually try to keep structs as small as possible, but maybe it matters less for rq ? 

> I would say we use sched_pelt_thermal w/o 'update_time' for now.

Sounds good, we can always add it later if someone needs it.

If it turns out some values really don't make any sense, would you mind making a PR to split the events of each (i.e. not use a common template that forces the same format) and remove the corresponding senseless event fields in trace.py ?

deggeman: Yes, I can do this. I'll add patches to take care of (2). And I could also add 'update_time' to all, solving (1).

douglas-raillard-arm:
Drive-by comment: you will need to update any tracepoint probe you used to ensure their first param is a `void *`, as was done in that PR: https://github.com/ARM-software/lisa/pull/2044

douglas-raillard-arm:

PR #1916, opened 2022-12-16
lisa._assets.kmodules.sched_tp: Hide UTIL_AVG_UNCHANGED in sched_util…

deggeman:
…_est_se

UTIL_AVG_UNCHANGED is an internal util_est state bit and should not be exposed via util_est.enqueued in the sched_util_est_se trace event.


douglas-raillard-arm:
This will need the corresponding header to be included in the C source to avoid relying on implicit dependencies. Can you confirm the macro is defined in a file available in /sys/kernel/kheaders.tar.xz ? (CONFIG_IKHEADERS).

deggeman:
Yes, I can. Checked on a Juno-r0 running eas/next/integration-20221121 :
`root@juno:/# tar -xf /sys/kernel/kheaders.tar.xz -C /root/tmp/
root@juno:/# find /root/tmp/ -type f | xargs grep UTIL_AVG_UNCHANGED
/root/tmp/include/linux/sched.h:#define UTIL_AVG_UNCHANGED              0x80000000`

PR #1924, opened 2023-01-10
lisa.analysis.latency: Include CPUs in latency DFs

mrkajetanp:
Latency-related functions such as df_latency_wakeup and df_latency_preemption at the moment do not return any information about which CPUs the latency ocurred on.

Add a 'cpu' column to df_latency_preemption along with 'cpu' & 'target_cpu' columns to df_latency_wakeup.


douglas-raillard-arm:
Ah one last thing, can you add "FEATURE" on an otherwise empty line after the title line in the commit message please ? This will make it appear in the changelog (see git history for examples)

PR #1934, opened 2023-01-18
install_base.sh: Do not install selenium geckodriver on recent Ubuntu

douglas-raillard-arm:
FIX

firefox-geckodriver does not exist as of Ubuntu 22.04 as it now is part of the firefox snap package.

### DIFF
--- a/install_base.sh
+++ b/install_base.sh
@@ -28,6 +28,12 @@ test_os_release(){
     fi
 }
 
+lower_or_equal() {
+    local x=$(printf "$1" | sed 's/\.//2g')
+    local y=$(printf "$2" | sed 's/\.//2g')
+    [[ "$(printf "$x\n$y\n" | sort -g | head -1)" == "$x" ]]
+}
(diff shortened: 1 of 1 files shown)


install_base.sh
```diff
@@ -28,6 +28,12 @@ test_os_release(){
     fi
 }
 
+lower_or_equal() {
+    local x = $(printf "$1" | sed 's/\.//2g')
```
credp, line 32: this just needs to be local x=$ (i.e. lose the spacing around the equals)

install_base.sh
```diff
@@ -28,6 +28,12 @@ test_os_release(){
...
+    local y = $(printf "$2" | sed 's/\.//2g')
```
credp, line 33: Here too.

credp (changes requested):
Works fine with the syntax fixed up - correctly skips geckodriver on my ubuntu 22.04.

PR #1938, opened 2023-02-10
lisa.analysis: Add Pixel 6 analysis

mrkajetanp:
Add a file for Pixel 6-specific analysis.
Include methods df_power_meter() for getting power meter readings in mW and plot_power_meter() for plotting out the meter readings from different channels.

### DIFF
--- a/lisa/analysis/pixel6.py
+++ b/lisa/analysis/pixel6.py
@@ -1,0 +1,92 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023, ARM Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pandas as pd
+import holoviews as hv
+from holoviews import opts
+
+from lisa.datautils import df_add_delta
+from lisa.analysis.base import TraceAnalysisBase
+from lisa.trace import requires_events
+from lisa.notebook import plot_signal
+
+
+class Pixel6Analysis(TraceAnalysisBase):
+    """
+    Support for Pixel 6-specific data analysis
+
+    :param trace: input Trace object
+    :type trace: lisa.trace.Trace
+    """
+
+    name = 'pixel6'
+
+    EMETER_CHAN_NAMES = {
+        'S4M_VDD_CPUCL0': 'little',
+        'S3M_VDD_CPUCL1': 'mid',
+        'S2M_VDD_CPUCL2': 'big',
+    }
+
+###############################################################################
+# DataFrame Getter Methods
+###############################################################################
+    @TraceAnalysisBase.cache
+    @requires_events('pixel6_emeter')
+    def df_power_meter(self):
+        """
+        Get the power meter readings across the trace.
+
+        :retuns: A :class:`pandas.DataFrame` with:
+
+            * A ``channel`` column (description of the power usage channel)
+            * A ``value`` column (average power usage in mW since the last measurement)
+        """
+        df = self.trace.df_event('pixel6_emeter')
+        df = df[df['chan_name'].isin(Pixel6Analysis.EMETER_CHAN_NAMES)]
+        df = df.groupby(['chan_name'], observed=True).apply(
+            lambda x: df_add_delta(x, col='value_diff', src_col='value', window=self.trace.window)['value_diff'] / df_add_delta(x, col='ts_diff', src_col='ts', window=self.trace.window)['ts_diff']
+        ).reset_index().rename(columns={0:'value', 'chan_name':'channel'}).dropna().set_index('Time')
+        df['channel'] = df['channel'].astype('category').cat.rename_categories(Pixel6Analysis.EMETER_CHAN_NAMES)
+
+        return df
+
+###############################################################################
+# Plotting methods
+###############################################################################
+
+    @TraceAnalysisBase.plot_method
+    @df_power_meter.used_events
+    def plot_power_meter(self, channels=None):
+        """
+        Plot the power meter readings from various channels.
+
+        :param channels: List of channels to plot
+        :type channels: list(str)
+
+        The channels needs to correspond to values in the ``channel`` column of df_power_meter().
+        """
+        df = self.df_power_meter()
+
+        channels = channels if channels is not None else list(df['channel'].cat.categories)
+        if any(channel not in df['channel'].cat.categories for channel in channels):
+            raise ValueError('Specified channel not found')
+
+        channel_data = dict(iter(df[df['channel'].isin(channels)].groupby(['channel'], observed=True)))
+        return hv.Overlay([
+            plot_signal(channel_data[channel]['value'], name=channel, vdim=hv.Dimension('value', label='Power', unit='mW'))
+            for channel in channels
+        ]).opts(title='Power usage per channel over time')
--- a/lisa/notebook.py
+++ b/lisa/notebook.py
@@ -376,7 +376,7 @@ def make_figure(width, height, nrows, ncols, interactive=None, **kwargs):
     return (figure, axes)
 
 
-def plot_signal(series, name=None, interpolation=None, add_markers=True):
+def plot_signal(series, name=None, interpolation=None, add_markers=True, vdim=None):
     """
     Plot a signal using ``holoviews`` library.
 
@@ -393,6 +393,9 @@ def plot_signal(series, name=None, interpolation=None, add_markers=True):
 
     :param add_markers: Add markers to the plot.
     :type add_markers: bool
+
+    :param vdim: Value axis dimension.
+    :type vdim: holoviews.Dimension
     """
     if isinstance(series, pd.DataFrame):
         try:
@@ -409,12 +412,16 @@ def plot_signal(series, name=None, interpolation=None, add_markers=True):
         # We don't set the unit as this will prevent shared_axes to work if
         # the other plots do not set the unit, which is what usually
         # happens, since only name/label is taken from pandas index names.
-        hv.Dimension('Time'),
+        hv.Dimension('Time', unit='s'),
     ]
+    vdims = [
+        vdim
+    ] if vdim is not None else None
     fig = hv.Curve(
         series,
         label=label,
         kdims=kdims,
+        vdims=vdims,
     ).opts(
         interpolation=interpolation,
         title=label,
@@ -429,6 +436,7 @@ def plot_signal(series, name=None, interpolation=None, add_markers=True):
             label=label,
             group='marker',
             kdims=kdims,
+            vdims=vdims,
         )
     return fig
 
--- a/doc/plot_conf.yml
+++ b/doc/plot_conf.yml
@@ -118,6 +118,9 @@ doc-plot-conf:
             kwargs:
                 cpu: *cpu0
 
+        Pixel6Analysis.plot_power_meter:
+            trace: !call:lisa.trace.Trace
+                trace_path: !env:interpolate $LISA_HOME/doc/traces/trace_pixel6.dat
 
         NotebookAnalysis.plot_event_field:
             kwargs:
--- a/doc/trace_analysis.rst
+++ b/doc/trace_analysis.rst
@@ -150,6 +150,12 @@ Thermal
 .. automodule:: lisa.analysis.thermal
    :members:
 
+Pixel 6
++++++++
+
+.. automodule:: lisa.analysis.pixel6
+   :members:
+
 Function profiling
 ++++++++++++++++++
 


lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+import pandas as pd
+import holoviews as hv
+from holoviews import opts
+
+from lisa.datautils import df_refit_index
+from lisa.analysis.base import TraceAnalysisBase
+from lisa.trace import requires_events
+from lisa.notebook import plot_signal
+
+P6_EMETER_CHAN_NAMES = {
```
douglas-raillard-arm, line 27: would make sense to turn it in a class attribute and remove the "P6" prefix

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+    'S3M_VDD_CPUCL1': 'mid',
+    'S2M_VDD_CPUCL2': 'big',
+}
+
+class Pixel6Analysis(TraceAnalysisBase):
+    """
+    Support for Pixel 6-specific analysis data
+
+    :param trace: input Trace object
+    :type trace: :class:`trace.Trace`
```
douglas-raillard-arm, line 38: s/:class:`trace.Trace`/lisa.trace.Trace/

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+
+        :retuns: A :class:`pandas.DataFrame` with:
+            * A ``little`` column (readings from the little cluster)
+            * A ``mid`` column (readings from the mid cluster)
+            * A ``big`` column (readings from the big cluster)
+        """
+        df = self.trace.df_event('pixel6_emeter')
+        df = df.pivot_table(values='value', index='ts', columns='chan_name').reset_index().rename(columns=P6_EMETER_CHAN_NAMES)
+        channels = list(P6_EMETER_CHAN_NAMES.values())
+        df['ts_diff'] = df['ts'].diff()
```
douglas-raillard-arm, line 60: use df_add_delta() or series_add_delta() from lisa.datautils. This will take care of windowing properly

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+        # convert channels to power usage delta in mW
+        for channel in channels:
+            df[channel] = df[channel].diff() / df["ts_diff"]
+        # convert ts to seconds
+        df['Time'] = df['ts'] / 1000
```
douglas-raillard-arm, line 65: that should already be the case, do you have a snippet of the input dataframe ? `Time` is normally the index and timestamps come from ftrace itself.

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+
+        :param channels: List of channels to plot
+        :type channels: list(str)
+
+        The channels needs to correspond to columns in df_power_meter().
+        """
+        df = self.df_power_meter()
+
+        channels = channels if channels is not None else list(df.columns)
+        if any([channel not in df.columns for channel in channels]):
```
douglas-raillard-arm, line 87: can drop the `[` and `]`

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+            raise ValueError('Specified channel not found')
+        idx = df.index.name
```
douglas-raillard-arm, line 89: dataframes returned by any lisa function are read-only, as they can stay in the memory cache for an arbitrary amount of time. Modifications like that need a `.copy()` first (deep=True or False depending on the sort of modification done on it)

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+        ds = hv.Dataset(df, idx, channels)
+        layout = ds.to(hv.Curve, idx, channels[0], label=channels[0])
+        for channel in channels[1:]:
```
douglas-raillard-arm, line 94: you can create a non-nested Overlay in one go with `hv.Overlay(a_list_of_plots)`

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        df = self.df_power_meter()
+
+        channels = channels if channels is not None else list(df.columns)
+        if any([channel not in df.columns for channel in channels]):
+            raise ValueError('Specified channel not found')
+        idx = df.index.name
+
+
+        ds = hv.Dataset(df, idx, channels)
```
douglas-raillard-arm, line 92: you should be able to create the Curve directly without making an explicit Dataset. You might want to try the `plot_signal` function for something like that as it adds some stuff and makes it easier to maintain consistent plots in all analysis

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+        layout = ds.to(hv.Curve, idx, channels[0], label=channels[0])
+        for channel in channels[1:]:
+            layout *= ds.to(hv.Curve, idx, channel, label=channel)
+
+        layout.opts(
+            opts.Curve(xlabel='Time [s]', ylabel='Power [mW]'),
```
douglas-raillard-arm, line 98: Holoviews supports setting the unit separately from the label:
https://holoviews.org/user_guide/Annotating_Data.html#dimension-parameters and merges them when it makes sense

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,100 @@
...
+        df = self.trace.df_event('pixel6_emeter')
+        df = df.pivot_table(values='value', index='ts', columns='chan_name').reset_index().rename(columns=P6_EMETER_CHAN_NAMES)
+        channels = list(P6_EMETER_CHAN_NAMES.values())
+        df['ts_diff'] = df['ts'].diff()
+        # convert channels to power usage delta in mW
+        for channel in channels:
+            df[channel] = df[channel].diff() / df["ts_diff"]
+        # convert ts to seconds
+        df['Time'] = df['ts'] / 1000
+        df = df[['Time'] + channels].set_index('Time').iloc[1:]
```
douglas-raillard-arm, line 66: Why do we need `.iloc[1:]` ?

douglas-raillard-arm (changes requested):

douglas-raillard-arm (changes requested):
Commit message should include `FEATURE` to be part of the changelog:

```
lisa.analysis: Add Pixel 6 analysis

FEATURE

Add a file for Pixel 6-specific analysis.
Include methods df_power_meter() for getting power meter readings in mW
and plot_power_meter() for plotting out the meter readings from
different channels.
```

lisa/analysis/pixel6.py
mrkajetanp: There's something a bit strange happening there. the 'ts' field is included with the event so I was using this for consistency as the power usage is computed based on the deltas in that column.

```
                 __comm  __pid  __cpu       ts  chan      value       chan_name
Time
297.103991  kworker/6:2     85      6  1917075     0    4464332    S10M_VDD_TPU
297.103993  kworker/6:2     85      6  1917075     1   25916055  VSYS_PWR_MODEM
297.103993  kworker/6:2     85      6  1917075     2       9916   VSYS_PWR_RFFE
297.103994  kworker/6:2     85      6  1917075     3   35633092  S2M_VDD_CPUCL2
297.103995  kworker/6:2     85      6  1917075     4   25287151  S3M_VDD_CPUCL1
...                 ...    ...    ...      ...   ...        ...             ...
472.252146  kworker/6:2     85      6  2092223     3  381290070  S2M_VDD_CPUCL2
472.252146  kworker/6:2     85      6  2092223     4  101833620  S3M_VDD_CPUCL1
472.252146  kworker/6:2     85      6  2092223     5  150807532  S4M_VDD_CPUCL0
472.252147  kworker/6:2     85      6  2092223     6   15784792     S5M_VDD_INT
472.252147  kworker/6:2     85      6  2092223     7   76808208     S1M_VDD_MIF
```

What should we do with this?

lisa/analysis/pixel6.py
mrkajetanp: Every analysis file apart from `load_tracking.py` from what I can see says `trace.Trace` so I just did that for consistency. Will update, sure.

douglas-raillard-arm: Thanks for mentioning it, it's broken in 2 ways:
* old versions of Sphinx used to not create a link if provided the `:class:...` part in that specific location. Even today, I'm not sure if e.g. ``list(:class:`trace.Trace`)`` would work.
* It creates a link to that: https://docs.python.org/3/library/trace.html#trace.Trace
which is not at all the right thing :)

I'll fix existing broken locations

douglas-raillard-arm: I dug a bit and it turned out this issue was introduced in 2017 in what is now lisa legacy, before it was sanely packaged into a Python package (rather than a folder to add to PYTHON_PATH). I guess it worked at that time given because of the wonky layout, even though Python 2 already had `trace.Trace`. It has been copy/pasted since then. Sometimes the past still haunts us :ghost:

lisa/analysis/pixel6.py
mrkajetanp: Well what we're interested in is the delta between measurements which makes one of the rows just contain 'NaN's in the end so I was slicing it off. df_add_delta also shifts the deltas up so now the empty row is the last one and I replaced this with `.iloc[:-1]`.

douglas-raillard-arm: Using `df_add_delta(..., window=self.trace.window)` should fix that without having to discard anything (and it will work properly when slicing the trace instead of just resulting in a possibly empty dataframe for a slice where the signal is perfectly defined)

mrkajetanp: Even with window=self.trace.window there were still NaN rows left so I'll just add a `.dropna()` at the end instead of the `iloc`.

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,89 @@
...
+        :retuns: A :class:`pandas.DataFrame` with:
+            * A ``channel`` column (description of the power usage channel)
+            * A ``value`` column (average power usage in mW since the last measurement)
+        """
+        df = self.trace.df_event('pixel6_emeter')
+        channels = list(Pixel6Analysis.EMETER_CHAN_NAMES.keys())
+        df = df.groupby(['chan_name'], observed=True).apply(
+            lambda x: df_add_delta(x, col='value_diff', src_col='value', window=self.trace.window)['value_diff'] / df_add_delta(x, col='ts_diff', src_col='ts', window=self.trace.window)['ts_diff']
+        ).reset_index().query("chan_name in @channels").rename(columns={0: 'value', 'chan_name': 'channel'}).dropna().set_index('Time')
+        df['channel'] = df['channel'].apply(lambda s: Pixel6Analysis.EMETER_CHAN_NAMES[s])
```
douglas-raillard-arm, line 62: this:
```
.apply(lambda s: Pixel6Analysis.EMETER_CHAN_NAMES[s])
```

can be replaced by:
```
.map(Pixel6Analysis.EMETER_CHAN_NAMES)
```

or even (much) better yet, change the column dtype to `category` (df['foo'] = df['foo'].astype('category')) and modify the category with the new names. This will avoid having to lookup the value of each row in the dict

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,89 @@
...
+
+        :retuns: A :class:`pandas.DataFrame` with:
+            * A ``channel`` column (description of the power usage channel)
+            * A ``value`` column (average power usage in mW since the last measurement)
+        """
+        df = self.trace.df_event('pixel6_emeter')
+        channels = list(Pixel6Analysis.EMETER_CHAN_NAMES.keys())
+        df = df.groupby(['chan_name'], observed=True).apply(
+            lambda x: df_add_delta(x, col='value_diff', src_col='value', window=self.trace.window)['value_diff'] / df_add_delta(x, col='ts_diff', src_col='ts', window=self.trace.window)['ts_diff']
+        ).reset_index().query("chan_name in @channels").rename(columns={0: 'value', 'chan_name': 'channel'}).dropna().set_index('Time')
```
douglas-raillard-arm, line 61: is it normal that we end up with a column named `0` ?

douglas-raillard-arm, line 61: I'm not against query() per say but most of the code uses the typical API, i.e. `df[df['chan_name'].isin(channels)]` which is explicit about the temporary series.

lisa/analysis/pixel6.py
douglas-raillard-arm: There likely is still an issue if you end up with NaN at the end. Maybe it's only because some deltas are `0` so we end up with a division by 0 but otherwise it's fishy. Can you provide the end of the dataframe right after groupby ?

lisa/notebook.py
```diff
@@ -376,7 +376,7 @@ def make_figure(width, height, nrows, ncols, interactive=None, **kwargs):
     return (figure, axes)
 
 
-def plot_signal(series, name=None, interpolation=None, add_markers=True):
+def plot_signal(series, name=None, interpolation=None, add_markers=True, vdims=None):
```
douglas-raillard-arm, line 379: the docstring needs to be updated. Does it make sense to allow for multiple vdims even though only a single Series can be plotted by this function ?

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,89 @@
...
+        Get the power meter readings across the trace.
+
+        :retuns: A :class:`pandas.DataFrame` with:
+            * A ``channel`` column (description of the power usage channel)
+            * A ``value`` column (average power usage in mW since the last measurement)
+        """
+        df = self.trace.df_event('pixel6_emeter')
+        channels = list(Pixel6Analysis.EMETER_CHAN_NAMES.keys())
+        df = df.groupby(['chan_name'], observed=True).apply(
+            lambda x: df_add_delta(x, col='value_diff', src_col='value', window=self.trace.window)['value_diff'] / df_add_delta(x, col='ts_diff', src_col='ts', window=self.trace.window)['ts_diff']
```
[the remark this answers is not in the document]
mrkajetanp: Well no that'll give us garbage results. The delta of `ts` is the time between the measurements, otherwise we'd be dividing the measurement by the time between the events being emitted which would just give us nonsense data.

lisa/analysis/pixel6.py
mrkajetanp: Well we're apply a result of a division as a column so yes, from what I've seen this resulting column just ends up being named 0 both here and in other similar cases.

mrkajetanp: Are you sure these two are equivalent? If I save the df before the `.query` calls and try to apply the snippet you wrote it results in an empty dataframe. I think the query is probably fine - clear to read, used as intended and can be chained with other functions easily.

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,89 @@
...
+
+        :param channels: List of channels to plot
+        :type channels: list(str)
+
+        The channels needs to correspond to values in the ``channel`` column of df_power_meter().
+        """
+        df = self.df_power_meter()
+
+        channels = channels if channels is not None else list(df['channel'].unique())
+        if any(channel not in df['channel'].unique() for channel in channels):
```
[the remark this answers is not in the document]
mrkajetanp: I switched the column to a category type so I can just grab the list of categories directly without computing anything, should be better

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError('Specified channel not found')
+
+        channel_data = {channel: data for channel, data in df.groupby(['channel'], observed=True)}
```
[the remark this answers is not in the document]
mrkajetanp: That doesn't work. The return of groupby is a list of tuples `(channel_name, df)`. It can be iterated over like above or I can turn it into a list but trying dict just errors out with `TypeError: attribute of type 'list' is not callable`

lisa/analysis/pixel6.py
mrkajetanp: Ah okay I see, it's because channels contains 'big', 'mid' 'little' instead of the raw channel names, that can be easily fixed

lisa/analysis/pixel6.py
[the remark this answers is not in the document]
douglas-raillard-arm: I guess it depends what triggers the sampling. If it's triggered by reading from sysfs, the event timestamp shouldn't be so bad. If it's something that happens no matter what then yes `Time` is meaningless.

lisa/analysis/pixel6.py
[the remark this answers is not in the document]
douglas-raillard-arm: Of course it does not ...
https://github.com/pandas-dev/pandas/issues/30621
So turns out that:
1. `dict()` treats as a mapping anything that has `.keys()` and `__getitem__` and will use those methods.
2. treats everything else as a key/value iterator.

Pandas' groupby object falls in category 1. except that `.keys()` just raises random errors depending on circumstances (your TypeError). 

So we can dodge it by hiding the `.keys()` behind a layer of `iter()`: `dict(iter(df.groupby(['channel'], observed=True)))`

lisa/analysis/pixel6.py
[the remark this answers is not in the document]
douglas-raillard-arm: Good idea that's even better

lisa/analysis/pixel6.py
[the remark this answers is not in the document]
mrkajetanp: It comes from the driver, if you read the sysfs multiple times within the 50ms window between updates you'll just get the same values in the `ts` column. In any case it's there for a reason, we might as well use it :)

lisa/analysis/pixel6.py
```diff
@@ -0,0 +1,91 @@
...
+from lisa.trace import requires_events
+from lisa.notebook import plot_signal
+
+
+class Pixel6Analysis(TraceAnalysisBase):
+    """
+    Support for Pixel 6-specific analysis data
+
+    :param trace: input Trace object
+    :type trace: :class:`lisa.trace.Trace`
```
douglas-raillard-arm, line 33: s/:class:`lisa.trace.Trace`/lisa.trace.Trace/

douglas-raillard-arm (changes requested):
LGTM, but there seems to be some issue with the vdim added to plot_signals, see github action log:
https://github.com/ARM-software/lisa/actions/runs/4186408402/jobs/7254866757

```
  FAILED tests/test_trace.py::TestTraceNoPlatform::test_sched_load_avg_signals - ValueError: NoneType type could not be interpreted as Dimension.  Dimensions must be declared as a string, tuple, dictionary or Dimension type.
```

PR #1945, opened 2023-02-16
Expose jobs in WAOutput & traces in WATraceCollector

mrkajetanp:
This PR allows acessing a list of jobs in the WAOutput through
wa_output.jobs along with a LazyMapping of raw Trace objects in WATraceCollector
through wa_output['trace'].traces.

Additionally, add deprecation warnings for creating WAOutputs containing multiple different WA runs. This is meant to determine whether any of the existing workflows or use cases rely on the feature with the ultimate goal of removing it and achieving the same functionality in a cleaner, less error-prone way.

### DIFF
--- a/lisa/utils.py
+++ b/lisa/utils.py
@@ -3845,4 +3845,34 @@ class SerializeViaConstructor(metaclass=_SerializeViaConstructorMeta):
         return (self._make_instance, (self._ctor, dct))
 
 
+class LazyMapping(Mapping):
+    """
+    Lazy Mapping dict-like class for elements evaluated on the fly.
+
+    It takes the same set of arguments as a dict with keys as the mapping keys
+    and values as closures that take a key and return the value. The class does
+    no automatic memoization but memoization can easily be achieved using
+    :func:`functools.lru_cache`, as shown in the example below.
+
+    **Example**::
+
+        LazyMapping({
+            x: lru_cache()(lambda k: k + 42)
+            for x in [1, 2, 3, 4]
+        })
+
+    """
+    def __init__(self, *args, **kwargs):
+        self._closures = dict(*args, **kwargs)
+
+    def __getitem__(self, key):
+        return self._closures[key](key)
+
+    def __iter__(self):
+        return iter(self._closures)
+
+    def __len__(self):
+        return len(self._closures)
+
+
 # vim :set tabstop=4 shiftwidth=4 textwidth=80 expandtab
--- a/lisa/wa.py
+++ b/lisa/wa.py
@@ -22,6 +22,7 @@ import abc
 import contextlib
 import sqlite3
 import pathlib
+from functools import lru_cache
 
 import pandas as pd
 
@@ -29,7 +30,7 @@ from wa import discover_wa_outputs, Status
 
 from lisa.version import VERSION_TOKEN
 from lisa.stats import Stats
-from lisa.utils import Loggable, memoized, get_subclasses
+from lisa.utils import Loggable, memoized, get_subclasses, LazyMapping
 from lisa._git import find_shortest_symref, get_commit_message
 from lisa.trace import Trace
 
@@ -250,11 +251,69 @@ class WAOutput(StatsProp, Mapping, Loggable):
         """
         parameters = list(inspect.signature(col).parameters.items())
         return any(
-            param.default == param.empty
+            param.default is param.empty and param.kind not in (param.VAR_POSITIONAL, param.VAR_KEYWORD)
             # The first parameter is provided so we can skip it
             for name, param in parameters[1:]
         )
 
+    @property
+    @memoized
+    def jobs(self):
+        """
+        List containing all the jobs present in the output of 'wa run'.
+        """
+        if len(self._jobs) > 1:
+            raise ValueError("jobs is only meant to be used with one output of 'wa run'. Please, use _jobs instead.")
+        return list(self._jobs.values())[0][1]
+
+    @property
+    @memoized
+    def _jobs(self):
+        return {
+            name: (output, [
+                job
+                for job in output.jobs
+                if job.status == Status.OK
+            ])
+            for name, output in self.outputs.items()
+        }
+
+    @property
+    @memoized
+    def outputs(self):
+        """
+        Dict containing a mapping of 'wa run' names to
+        :class:`wa.framework.RunOutput` objects.
+        """
+        wa_outputs = list(discover_wa_outputs(self.path))
+
+        wa_outputs = {
+            pathlib.Path(
+                wa_output.basepath
+            ).resolve(): wa_output
+            for wa_output in wa_outputs
+        }
+
+        if len(wa_outputs) > 1:
+            common_prefix = pathlib.Path(
+                os.path.commonpath(wa_outputs.keys())
+            )
+        elif wa_outputs:
+            path, = wa_outputs.keys()
+            common_prefix = pathlib.Path(path).parent
+        else:
+            common_prefix = None
+
+        wa_outputs = {
+            str(
+                name.relative_to(common_prefix)
+                if common_prefix else
+                name
+            ): wa_output
+            for name, wa_output in wa_outputs.items()
+        }
+
+        return wa_outputs
 
 class WACollectorBase(StatsProp, Loggable, abc.ABC):
     """
@@ -323,8 +382,6 @@ class WACollectorBase(StatsProp, Loggable, abc.ABC):
     def _get_df(self):
         self.logger.debug(f"Collecting dataframe for {self.NAME}")
 
-        wa_outputs = list(discover_wa_outputs(self.wa_output.path))
-
         def load_df(job):
             def loader(job):
                 cache_path = os.path.join(
@@ -360,39 +417,12 @@ class WACollectorBase(StatsProp, Loggable, abc.ABC):
             else:
                 return self._df_postprocess(df)
 
-        wa_outputs = {
-            pathlib.Path(
-                wa_output.basepath
-            ).resolve(): wa_output
-            for wa_output in wa_outputs
-        }
-
-        if len(wa_outputs) > 1:
-            common_prefix = pathlib.Path(
-                os.path.commonpath(wa_outputs.keys())
-            )
-        elif wa_outputs:
-            path, = wa_outputs.keys()
-            common_prefix = pathlib.Path(path).parent
-        else:
-            common_prefix = None
-
-        wa_outputs = {
-            str(
-                name.relative_to(common_prefix)
-                if common_prefix else
-                name
-            ): wa_output
-            for name, wa_output in wa_outputs.items()
-        }
-
         dfs = [
             self._add_output_info(wa_output, name, df)
-            for name, wa_output in wa_outputs.items()
+            for name, (wa_output, jobs) in self._jobs.items()
             for df in [
                 load_df(job)
-                for job in wa_output.jobs
-                if job.status == Status.OK
+                for job in jobs
             ]
             if df is not None
         ]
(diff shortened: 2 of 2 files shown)


lisa/wa.py
```diff
@@ -255,6 +255,13 @@ def _needs_params(col):
             for name, param in parameters[1:]
         )
 
+    @property
+    @memoized
+    def jobs(self):
+        outputs = list(discover_wa_outputs(self.path))
+        if len(outputs) > 1:
+            raise DeprecationWarning('Passing a directory containing multiple outputs of "wa run" to WAOutput is deprecated. Please, create one object per run.')
```
douglas-raillard-arm, line 263: use this: https://docs.python.org/3/library/warnings.html#warnings.warn instead of raising an exception

lisa/wa.py
```diff
@@ -255,6 +255,13 @@ def _needs_params(col):
...
+        return [job for job in outputs[0].jobs if job.status == Status.OK]
```
douglas-raillard-arm, line 264: We should still use all of `outputs`, not just `[0]`. Maybe in time we can actually get rid of the others but for now it's only going to be deprecated, not removed

lisa/wa.py
```diff
@@ -598,16 +609,28 @@ def trace_idle_analysis(trace):
     NAME = 'trace'
     _ARTIFACT_NAME = 'trace-cmd-bin'
     _PURE_GET_JOB_DF = False
+    _FORCE_AUTO = True
 
-    def __init__(self, wa_output, trace_to_df, **kwargs):
+    def __init__(self, wa_output, trace_to_df=None, **kwargs):
```
douglas-raillard-arm, line 614: We definitely don't want to do that. I've honestly never seen any code where something is either X or None and does not eventually trip itself, it's quite amazing how consistently we fail at this. The only salvation can come from mypy or simply never doing this (current approach). 

Here it's easy, you can simply make `trace_to_df` default to a function that raises an exception and that's even less code.

lisa/wa.py
```diff
@@ -598,16 +609,28 @@ def trace_idle_analysis(trace):
...
+        if self._trace_to_df is None:
+            raise ValueError('trace_to_df was not specified for the TraceCollector')
         trace = Trace(path, **self._trace_kwargs)
         return self._trace_to_df(trace)
 
+    @property
+    @memoized
+    def traces(self):
+        return {
+            f"{job.label}-{job.iteration}":Trace(job.get_artifact_path('trace-cmd-bin'))
```
douglas-raillard-arm, line 629: s/":Trace/": Trace/

lisa/wa.py
```diff
@@ -598,16 +609,28 @@ def trace_idle_analysis(trace):
...
     def _get_artifact_df(self, path):
+        if self._trace_to_df is None:
+            raise ValueError('trace_to_df was not specified for the TraceCollector')
         trace = Trace(path, **self._trace_kwargs)
         return self._trace_to_df(trace)
 
+    @property
+    @memoized
+    def traces(self):
+        return {
```
douglas-raillard-arm, line 628: The `Trace` must be built with the `**self._trace_kwargs` as well

lisa/wa.py
```diff
@@ -598,16 +609,28 @@ def trace_idle_analysis(trace):
...
         super().__init__(wa_output, df_postprocess=None)
 
     def _get_artifact_df(self, path):
+        if self._trace_to_df is None:
+            raise ValueError('trace_to_df was not specified for the TraceCollector')
         trace = Trace(path, **self._trace_kwargs)
         return self._trace_to_df(trace)
 
+    @property
+    @memoized
```
douglas-raillard-arm, line 626: `memoized` is a bit dangerous as Traces will by default take arbitrary amounts of memory. That's a great way to end up with some sort of leak.

lisa/wa.py
douglas-raillard-arm, line 628: In the name of saving some computations, I'd use a lazy mapping object. You can easily make one by inheriting from `collections.abc.Mapping` and implementing the relevant methods. This should go in `lisa.utils`. I can look like:
```
lazy = LazyMapping({'key1': (1, 2), 'key2': (3, 4)}, lambda x1, x2: x1 + x2)
dict(lazy) == {'key1': 3, 'key2': 7}
```

lisa/wa.py
```diff
@@ -168,7 +168,7 @@ def __init__(self, path, kernel_path=None):
         auto_collectors = {
             name: cls
             for name, cls in collector_classes.items()
-            if not self._needs_params(cls)
+            if not self._needs_params(cls) or cls._FORCE_AUTO
```
douglas-raillard-arm, line 171: That's broken, if a class needs some parameters you simply cannot create the collector with this `__getitem__` API as it would force calling a function with too few parameters.

Instead simply fix `_needs_params()` so it ignores parameters with `.kind in (VAR_POSITIONAL, VAR_KEYWORD)` : https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind

douglas-raillard-arm (changes requested):

lisa/wa.py
mrkajetanp: Hmm should I just drop it then? Without it the traces take a good moment to load every time so it definitely makes the experience worse :/

mrkajetanp:
I rebased this on the latest main as well

lisa/utils.py
```diff
@@ -3845,4 +3845,21 @@ def __reduce__(self):
         return (self._make_instance, (self._ctor, dct))
 
 
+class LazyMapping(Mapping):
+    """
+    Lazy Mapping dict-like class for elements evaluated on the fly.
```
douglas-raillard-arm, line 3850: Parameters passed to the constructor need to be documented here

lisa/wa.py
```diff
@@ -250,11 +252,18 @@ def _needs_params(col):
...
             for name, param in parameters[1:]
         )
 
+    @property
+    @memoized
+    def jobs(self):
+        outputs = list(discover_wa_outputs(self.path))
+        if len(outputs) > 1:
+            warnings.warn('Passing a directory containing multiple outputs of "wa run" to WAOutput is deprecated. Please, create one object per run.', DeprecationWarning)
+        return [job for output in outputs for job in output.jobs if job.status == Status.OK]
```
douglas-raillard-arm, line 266: needs to be split, one clause per line

lisa/wa.py
```diff
@@ -250,11 +252,18 @@ def _needs_params(col):
         """
         parameters = list(inspect.signature(col).parameters.items())
         return any(
-            param.default == param.empty
+            param.default is param.empty and param.kind not in (param.VAR_POSITIONAL, param.VAR_KEYWORD)
             # The first parameter is provided so we can skip it
             for name, param in parameters[1:]
         )
 
+    @property
```
douglas-raillard-arm, line 260: This job discovery is now duplicated with `_get_df`, they need to be unified. It's probably possible to rewrite `_get_df` in terms of that `jobs` property so we don't accidentally start having diverging definition of the set of valid jobs for both uses.

lisa/wa.py
```diff
@@ -599,15 +613,24 @@ def trace_idle_analysis(trace):
     _ARTIFACT_NAME = 'trace-cmd-bin'
     _PURE_GET_JOB_DF = False
 
-    def __init__(self, wa_output, trace_to_df, **kwargs):
+    def __init__(self, wa_output, trace_to_df=_stub_trace_to_df, **kwargs):
         self._trace_to_df = trace_to_df
         self._trace_kwargs = kwargs
         super().__init__(wa_output, df_postprocess=None)
 
+
```
douglas-raillard-arm, line 621: empty line can be removed

douglas-raillard-arm (changes requested):
Some adjustments needed but otherwise looks good

lisa/utils.py
douglas-raillard-arm: We need to state that it takes the arguments in the same shape as a dict with the keys being the mapping keys and then document the value format (closures with key as param and returning the value, no automatic memoization).

Maybe stating that the parameters are the same as dict and a commented example would do it

lisa/utils.py
```diff
@@ -3845,4 +3845,34 @@ def __reduce__(self):
...
+
+    It takes the same set of arguments as a dict with keys as the mapping keys
+    and values as  closures that take a key and return the value. The class does
```
douglas-raillard-arm, line 3853: s/as  closures/as closures/

lisa/utils.py
```diff
@@ -3845,4 +3845,34 @@ def __reduce__(self):
...
+    no automatic memoization but memoization can easily be achieved using something
+    like lru_cache from functools, as shown in the example below.
```
douglas-raillard-arm, line 3855: can be shortened into:
```
but memoization can be achieved using :func:`functools.lru_cache`.
```

lisa/utils.py
```diff
@@ -3845,4 +3845,34 @@ def __reduce__(self):
...
+
+    **Example**::
+
+        LazyMapping({
+            f"{job.label}-{job.iteration}": lru_cache()(lambda k: Trace(job.get_artifact_path('trace-cmd-bin'), **self._trace_kwargs))
```
douglas-raillard-arm, line 3860: `**self._trace_kwargs` must be removed, this is an private attribute so we can definitely not have examples of a public API showing it.

The example can simply show how to compute `key + 42` without requiring the user to know anything about WA

lisa/wa.py
```diff
@@ -250,11 +252,65 @@ def _needs_params(col):
...
         return any(
-            param.default == param.empty
+            param.default is param.empty and param.kind not in (param.VAR_POSITIONAL, param.VAR_KEYWORD)
             # The first parameter is provided so we can skip it
             for name, param in parameters[1:]
         )
 
+    @property
+    @memoized
+    def jobs(self):
```
douglas-raillard-arm, line 262: public property so needs docstring

lisa/wa.py
```diff
@@ -250,11 +252,65 @@ def _needs_params(col):
...
+                job
+                for job in output.jobs
+                if job.status == Status.OK
+            ])
+            for name, output in self.outputs.items()
+        }
+
+    @property
+    @memoized
+    def outputs(self):
```
douglas-raillard-arm, line 281: public property so needs docstring

lisa/wa.py
```diff
@@ -608,6 +637,14 @@ def _get_artifact_df(self, path):
         trace = Trace(path, **self._trace_kwargs)
         return self._trace_to_df(trace)
 
+    @property
+    def traces(self):
```
douglas-raillard-arm, line 641: public property so needs docstring

lisa/wa.py
```diff
@@ -250,11 +252,71 @@ def _needs_params(col):
...
+                if job.status == Status.OK
+            ])
+            for name, output in self.outputs.items()
+        }
+
+    @property
+    @memoized
+    def outputs(self):
+        """
+        Dict containing a mapping of 'wa run' names to RunOutput objects.
```
douglas-raillard-arm, line 286: s/RunOutput/:class:`wa.framework.RunOutput`/

lisa/wa.py
```diff
@@ -608,6 +643,18 @@ def _get_artifact_df(self, path):
         trace = Trace(path, **self._trace_kwargs)
         return self._trace_to_df(trace)
 
+    @property
+    def traces(self):
+        """
+        LazyMapping that maps job names & iteration numbers to their raw
```
douglas-raillard-arm, line 649: s/LazyMapping/:class:`lisa.utils.LazyMapping`/

douglas-raillard-arm, line 649: s/to their raw trace file/to their corresponding :class:`lisa.trace.Trace`/

douglas-raillard-arm (changes requested):

PR #1946, opened 2023-02-20
Support for PMU event counters

bea-arm:
Add support  for reading Arm PMU event counters through perf  framework and expose the readings  as a dedicated ftrace event. 

This is being achieved by introducing new feature to the sched_tp kernel module for tracing the PMU event counters, plus changes to the lisa/tests/base.py, that handle the post-processing for CPU_CYCLES  counter events, which allows drawing conclusions on the estimated CPU frequency.

Additionally both util_tracking and load_tracking tests get amended to benefit from the above mentioned changes.

### DIFF
--- a/lisa/_assets/kmodules/lisa/Makefile
+++ b/lisa/_assets/kmodules/lisa/Makefile
@@ -34,7 +34,7 @@ ifneq ($(KERNELRELEASE),)
 
 LISA_KMOD_NAME ?= lisa
 obj-m := $(LISA_KMOD_NAME).o
-$(LISA_KMOD_NAME)-y := main.o tp.o wq.o features.o pixel6.o
+$(LISA_KMOD_NAME)-y := main.o tp.o wq.o features.o pixel6.o perf_counters.o
 ldflags-y += -T "$(M)/features.lds"
 
 clean-files := vmlinux.h
--- a/lisa/_assets/kmodules/lisa/ftrace_events.h
+++ b/lisa/_assets/kmodules/lisa/ftrace_events.h
@@ -359,6 +359,25 @@ TRACE_EVENT(lisa__pixel6_emeter,
 		  __entry->ts, __entry->device, __entry->chan, __entry->chan_name, __entry->value)
 );
 
+TRACE_EVENT(lisa__perf_counter,
+	TP_PROTO(unsigned int cpu, unsigned int counter_id, u64 value),
+	TP_ARGS(cpu, counter_id, value),
+
+	TP_STRUCT__entry(
+		__field(	unsigned int,	cpu		)
+		__field(	unsigned int,	counter_id	)
+		__field(	u64,		value		)
+	),
+
+	TP_fast_assign(
+		__entry->cpu		= cpu;
+		__entry->counter_id	= counter_id;
+		__entry->value		= value;
+	),
+
+	TP_printk("cpu=%u counter_id=%u value=%llu",
+		  __entry->cpu, __entry->counter_id, __entry->value)
+);
 #endif /* _FTRACE_EVENTS_H */
 
 /* This part must be outside protection */
--- a/lisa/_assets/kmodules/lisa/perf_counters.c
+++ b/lisa/_assets/kmodules/lisa/perf_counters.c
@@ -1,0 +1,492 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (C) 2023 ARM Ltd.
+#include <linux/perf_event.h>
+#if defined(CONFIG_HW_PERF_EVENTS) && defined(CONFIG_ARM_PMU)
+#include <linux/perf/arm_pmu.h>
+#endif
+#include "main.h"
+#include "ftrace_events.h"
+#include "tp.h"
+
+#define MAX_PERF_COUNTERS	6
+
+#define __PERFCTR_PARAM(name, param_name, type, param_type, desc)	\
+	static type param_name[MAX_PERF_COUNTERS];			\
+	static unsigned int param_name##_count;				\
+	module_param_array_named(name, param_name, param_type,		\
+				 &param_name##_count, 0644);		\
+	MODULE_PARM_DESC(name, desc);
+
+#define PERFCTR_PARAM(name, type, param_type, desc)	\
+	__PERFCTR_PARAM(perf_counter_##name, name##_param, type, param_type, desc)
+
+/* Set of perf counters to enable - comma-separated names of events */
+PERFCTR_PARAM(generic_perf_events, char *, charp,
+	      "Comma-separated list of symbolic names for generic perf events");
+/* Set of perf counters to enable - comma-separated PMU raw counter ids */
+PERFCTR_PARAM(pmu_raw_counters, unsigned int , uint,
+	      "Comma-separated list of raw PMU event counter ids");
+
+/* Initial set of supported counters to be enabled through module params */
+struct perfctr_desc {
+	/* unique name to identify the counter */
+	const char 		*name;
+	/* counter id (may be generic or raw) */
+	u64			id;
+	enum perf_type_id	type;
+	/* enable by default if no counters requested */
+	bool			default_on;
+};
+
+#define PERFCTR_DESC(__name, __id, __type, __en)				\
+	((struct perfctr_desc) {						\
+		.name = __name, .id = __id, .type = __type, .default_on = __en,	\
+	})
+
+#define PERFCTR_DESC_COUNT_HW(__name, __id, __en)	\
+	PERFCTR_DESC(__name, __id, PERF_TYPE_HARDWARE, __en)
+
+/* Initial set of supported counters to be enabled based on provided event names */
+static const struct perfctr_desc perfctr_generic_lt [] = {
+	PERFCTR_DESC_COUNT_HW("cpu_cycles", PERF_COUNT_HW_CPU_CYCLES, 1),
+	PERFCTR_DESC_COUNT_HW("inst_retired", PERF_COUNT_HW_INSTRUCTIONS, 0),
+	PERFCTR_DESC_COUNT_HW("cache_references", PERF_COUNT_HW_CACHE_REFERENCES, 0),
+	PERFCTR_DESC_COUNT_HW("cache_misses", PERF_COUNT_HW_CACHE_MISSES, 0),
+	PERFCTR_DESC_COUNT_HW("branch_retired", PERF_COUNT_HW_BRANCH_INSTRUCTIONS, 0),
+	PERFCTR_DESC_COUNT_HW("branch_mispred", PERF_COUNT_HW_BRANCH_MISSES, 0),
+	PERFCTR_DESC_COUNT_HW("bus_cycles", PERF_COUNT_HW_BUS_CYCLES, 0),
+	PERFCTR_DESC_COUNT_HW("stall_frontend", PERF_COUNT_HW_STALLED_CYCLES_FRONTEND, 0),
+	PERFCTR_DESC_COUNT_HW("stall_backend", PERF_COUNT_HW_STALLED_CYCLES_BACKEND, 0),
+};
+
+struct perfctr_event_entry {
+	struct hlist_node		node;
+	struct hlist_node		group_link;
+	struct perf_event		*event;
+	struct perfctr_event_group	*group;
+	struct rcu_head			rcu_head;
+};
+
+struct perfctr_event_group {
+	struct list_head	node;
+	struct hlist_head	entries;
+	u64			raw_id;
+};
+
+struct perfctr_pcpu_data {
+	struct hlist_head	events;
+};
+
+struct perfctr_core {
+	struct list_head			events;
+	struct perfctr_pcpu_data __percpu	*pcpu_data;
+	unsigned int				nr_events;
+	unsigned int				max_nr_events;
+};
+
+static inline void perfctr_show_supported_generic_events(void)
+{
+	int i;
+
+	pr_info("Possible (subject to actual support) generic perf events: ");
+	for (i = 0; i < ARRAY_SIZE(perfctr_generic_lt); ++i)
+		printk(KERN_CONT "%s, ", perfctr_generic_lt[i].name);
+}
+
+static void perfctr_event_release_entry(struct perfctr_event_entry *entry);
+
+static int perfctr_event_activate_single(struct perfctr_core *perf_data,
+					 struct perf_event_attr *attr)
+{
+	struct perfctr_event_entry *entry= NULL;
+	struct perfctr_event_group *group;
+	struct hlist_node *next;
+	cpumask_var_t active_mask;
+	int cpu;
+
+	group = kzalloc(sizeof(*group), GFP_KERNEL);
+	if (!group)
+		return -ENOMEM;
+
+	if (!zalloc_cpumask_var(&active_mask, GFP_KERNEL)) {
+		kfree(group);
+		return -ENOMEM;
+	}
+
+	group->raw_id = PERF_COUNT_HW_MAX;
+
+	for_each_online_cpu(cpu) {
+		entry = kzalloc(sizeof(*entry), GFP_KERNEL);
+		if (!entry)
+			goto activate_failed;
+
+		entry->event =
+			/* No overflow handler, at least not at this point */
+			perf_event_create_kernel_counter(attr, cpu, NULL,
+							 NULL, NULL);
+		if (IS_ERR(entry->event)) {
+			pr_err("Failed to create counter id=%llu on cpu%d\n",
+			       attr->config, cpu);
+			kfree(entry);
+			continue;
+		}
+
+		perf_event_enable(entry->event);
+		/*
+		 * the PMU driver might still fail to assign a slot for a given
+		 * counter (@see armpmu_add) which leaves the event ineffective
+		 */
+		if (entry->event->state != PERF_EVENT_STATE_ACTIVE) {
+			pr_err("Failed to enable counter id=%llu on cpu%d\n",
+			       attr->config, cpu);
+			perf_event_disable(entry->event);
+			perf_event_release_kernel(entry->event);
+			kfree(entry);
+			continue;
+		}
+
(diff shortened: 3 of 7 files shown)


lisa/_assets/kmodules/sched_tp/wq.c
```diff
@@ -31,17 +31,33 @@ struct work_item *start_work(worker_t f, int delay, void *data) {
...
 		item->f = f;
+		item->cpu = cpu;
 		item->data = data;
 
 		item->__delay = delay;
 		item->__wq = wq;
 		INIT_DELAYED_WORK(&item->__dwork, worker);
 
-		queue_delayed_work(wq, &item->__dwork, delay);
+		if ((unsigned)cpu >= nr_cpu_ids)
```
douglas-raillard-arm, line 41: this logic is duplicated with start_work_on, can we make a helper to handle that ?

lisa/_assets/kmodules/sched_tp/wq.c
```diff
@@ -31,17 +31,33 @@ struct work_item *start_work(worker_t f, int delay, void *data) {
...
+			queue_delayed_work(wq, &item->__dwork, delay);
+		else
+			queue_delayed_work_on(cpu, wq, &item->__dwork, delay);
 	}
 	return item;
 }
 
+void restart_work(struct work_item *item, int delay)
```
douglas-raillard-arm, line 49: work_item->__delay needs to be updated with the new delay

lisa/_assets/kmodules/sched_tp/wq.c
```diff
@@ -31,17 +31,33 @@ struct work_item *start_work(worker_t f, int delay, void *data) {
...
+{
+	struct workqueue_struct *wq = FEATURE(__worqueue)->data;
+
+	if (!wq || !item)
+		return;
+	if ((unsigned)item->cpu >= nr_cpu_ids)
```
douglas-raillard-arm, line 55: can't we simply test for `item->cpu < 0` ?

lisa/_assets/kmodules/sched_tp/ftrace_events.h
```diff
@@ -357,6 +357,26 @@ TRACE_EVENT(pixel6_emeter,
...
+		__field(	u64,		value		)
+	),
+
+	TP_fast_assign(
+		__entry->cpu		= cpu;
+		__entry->counter_id	= counter_id;
+		__entry->value		= value;
+	),
+
+	TP_printk("cpu=%u counter_id=%u value=%lu",
```
douglas-raillard-arm, line 376: `%lu` is not appropriate for u64 value

douglas-raillard-arm (changes requested):
Had a look at wq changes, I'll look at perf_counters.c next and once the kernel module side is done I'll do the Python part

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+#define MAX_PERF_COUNTERS	6
+#define HW_PERF_UNSUPPORTED	0xFFFFFFFF
+
+
+/* Initial set of supported counters to be enabled through module params */
+struct perfctr_desc {
+	const char 	*name;		/* unique name to identify the counter */
+	u64		id;		/* counter id (may be generic or raw) */
+	unsigned int	type;
+	unsigned int	default_on:1;	/* enable by default if no counters requested */
```
douglas-raillard-arm, line 21: I'd rather not have bitfields given the can of worms they are in general. You can simply use `bool` which is the appropriate type here.

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+
+#define MAX_PERF_COUNTERS	6
+#define HW_PERF_UNSUPPORTED	0xFFFFFFFF
+
+
+/* Initial set of supported counters to be enabled through module params */
+struct perfctr_desc {
+	const char 	*name;		/* unique name to identify the counter */
+	u64		id;		/* counter id (may be generic or raw) */
+	unsigned int	type;
```
douglas-raillard-arm, line 20: s/unsigned int/enum perf_type_id/

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+{
+	struct perfctr_core *perf_data = data;
+	struct perfctr_event_entry *entry;
+	struct hlist_head *entry_list;
+	int cpu = smp_processor_id();
+	u64 value, enabled, running;
+	int id;
+
+	if (!trace_perf_counter_enabled())
+		goto leave;
```
douglas-raillard-arm, line 86: can avoid goto with just appropriate scoping for the `if` block

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+	struct kobject				kobj;
+	struct completion			kobj_unregister;
+	struct srcu_struct			srcu;
+};
+
+static int perfctr_worker_fn(void *data)
+{
+	struct perfctr_core *perf_data = data;
+	struct perfctr_event_entry *entry;
+	struct hlist_head *entry_list;
```
douglas-raillard-arm, line 80: variable can be declared at the spot where it is initialized

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+static __maybe_unused
+void perfctr_reactivate_workers(struct perfctr_core *perf_data)
+{
+	int cpu;
+
+	if (!perf_data->nr_events)
+		return;
+
+	for_each_online_cpu(cpu) {
+		 struct work_item *work;
```
douglas-raillard-arm, line 116: can merge declaration and initialization of `work`

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+
+static int perfctr_event_activate_single(struct perfctr_core *perf_data,
+					 struct perf_event_attr *attr)
+{
+	struct perfctr_event_group *group;
+	struct perfctr_event_entry *entry;
+	struct hlist_node *next;
+	int cpu;
+
+	WARN_ON(!mutex_is_locked(&perf_data->lock));
```
douglas-raillard-arm, line 135: is WARN_ON appropriate ? If we are supposed to take that lock, we really should never mess with anything it covers if the lock is not taken, not just merely warn about it.

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+
+struct perfctr_match {
+	void *data;
+	enum perfctr_match_type type;
+};
+
+static int perfctr_event_activate(struct perfctr_core *perf_data,
+				  const struct perfctr_match *match)
+{
+	int result = -EINVAL, i;
```
douglas-raillard-arm, line 210: I'd prefer splitting definitions each on their own line, especially if there is a mix of init and non-init

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+}
+
+/* Lookup match type */
+enum perfctr_match_type {
+	PERFCTR_MATCH_NAME,
+	PERFCTR_MATCH_STATUS
+};
+
+struct perfctr_match {
+	void *data;
```
douglas-raillard-arm, line 203: should be a union to reflect the fact the type is chosen at runtime

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+	mutex_lock(&perf_data->lock);
+	list_for_each_entry(group, &perf_data->events, node) {
+		length += sprintf(buf + length, "%d ", group->raw_id);
+		++count;
+	}
+	mutex_unlock(&perf_data->lock);
+
+	for(; count < perf_data->max_nr_events; ++count)
+		length += sprintf(buf + length, "%d ", HW_PERF_UNSUPPORTED);
+	length += sprintf(buf + length, "\n");
```
douglas-raillard-arm, line 305: I'm not sure about what is being printed here but it seems that we never do any bound checking on `buf` to ensure it's big enough

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+	LIST_HEAD(drop_zone);
+	int i;
+
+	perf_data = container_of(kobj, struct perfctr_core, kobj);
+
+	/*
+	 * The below relies on the max number of counters per-core,
+	 * if one changes, the other should follow
+	 */
+	BUILD_BUG_ON(MAX_PERF_COUNTERS != 6);
```
douglas-raillard-arm, line 327: Should be done at runtime, we don't want to make the whole build break because of a single feature. It might be appropriate to do that check in the feature init and bail with an explanatory pr_err()

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+	}
+	mutex_unlock(&perf_data->lock);
+
+	for(; count < perf_data->max_nr_events; ++count)
+		length += sprintf(buf + length, "%d ", HW_PERF_UNSUPPORTED);
+	length += sprintf(buf + length, "\n");
+	return length;
+}
+
+static ssize_t pmu_raw_counters_store(struct kobject *kobj,
```
douglas-raillard-arm, line 309: dead code

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+			}
+		}
+	}
+	list_for_each_entry_safe(group, next, &drop_zone, node)
+		list_move(&group->node, &perf_data->events);
+	mutex_unlock(&perf_data->lock);
+	return -EINVAL;
+}
+
+static ssize_t sampling_period_ms_show(struct kobject *kobj,
```
douglas-raillard-arm, line 418: dead code

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+	perf_data = container_of(kobj, struct perfctr_core, kobj);
+	mutex_lock(&perf_data->lock);
+	length = sprintf(buf, "%lu\n",
+			 jiffies_to_msecs(perf_data->sampling_period));
+	mutex_unlock(&perf_data->lock);
+
+	return length;
+}
+
+static ssize_t sampling_period_ms_store(struct kobject *kobj,
```
douglas-raillard-arm, line 433: dead code

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+		return result;
+
+	mutex_lock(&perf_data->lock);
+	perf_data->sampling_period = __msecs_to_jiffies(period_ms);
+	mutex_unlock(&perf_data->lock);
+
+	return count;
+}
+
+static ssize_t max_events_show(struct kobject *kobj,
```
douglas-raillard-arm, line 453: dead code

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,694 @@
...
+
+	perf_data = kzalloc(sizeof(*perf_data), GFP_KERNEL);
+	if (!perf_data) {
+		DISABLE_FEATURE(__worqueue);
+		return 1;
+	}
+
+	if (init_srcu_struct(&perf_data->srcu)) {
+		kfree(perf_data);
+		DISABLE_FEATURE(__worqueue);
```
douglas-raillard-arm, line 599: In the context of the previous comment, this will break refcounting: `perfctr_disable()` will always disable run `DISABLE_FEATURE(__worqueue)` provided `perf_data != NULL`, but that block also runs `DISABLE_FEATURE(__worqueue);` with `perf_data != NULL`.

douglas-raillard-arm (changes requested):

lisa/tests/scheduler/util_tracking.py
```diff
@@ -227,6 +228,10 @@ def test_means(self) -> ResultBundle:
             apply_phase_window = functools.partial(df_refit_index, window=(phase.start, phase.end))
 
             ue_phase_df = apply_phase_window(ue_df)
+            if ue_phase_df.shape[0] <= 1:
+                self.logger.warning(f"Not enough events recorded for phase {phase.id} to proceed")
```
douglas-raillard-arm, line 232: I'm not sure this belongs to this PR but the only reason I can see for having an empty dataframe here is the usual cpufreq Juno R0 issue that leads to util=1023 during the whole test and therefore no util_est values at all.

douglas-raillard-arm (commented):
I realized I did not ask yet: what is the use case for PMU counters ?
If it is just to catch the cases where Juno R0 fails to switch frequency, we can catch that without the PMU in case of rt-app:
* Given the profile, we can compute the expected task duty cycle on any CPU at any freq
* df_task_activations() tells us the current cpu of the task, and df_cpu_frequency() can provide the expected freq
* df_task_activations() provides the instantaneous duty cycle of the task as well, so we can compared that to the expected one.

If the mismatch is big enough, then cpufreq is busted. 

However, we might still have a use for having perf counters in the trace in general so I don't think it's a waste.

douglas-raillard-arm:
@bea-arm FYI I exposed feature parameters to `lisa-load-kmod`:
https://github.com/ARM-software/lisa/pull/1986

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,490 @@
...
+	})
+
+#define PERFCTR_DESC_COUNT_HW(__name, __id, __en)	\
+	PERFCTR_DESC(__name, __id, PERF_TYPE_HARDWARE, __en)
+
+/* Initial set of supported counters to be enabled based on provided event names */
+static const struct perfctr_desc perfctr_generic_lt [] = {
+	PERFCTR_DESC_COUNT_HW("cpu_cycles", PERF_COUNT_HW_CPU_CYCLES, 1),
+	PERFCTR_DESC_COUNT_HW("inst_retired", PERF_COUNT_HW_INSTRUCTIONS, 0),
+	PERFCTR_DESC_COUNT_HW("l1d_cache", PERF_COUNT_HW_CACHE_REFERENCES, 0),
```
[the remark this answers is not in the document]
bea-arm: This is how it's mapped on Arm, I think perf uses 'cache-references' for a generic name .
I could potentially switch to that one to keep it consistent with perf on other platforms

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,490 @@
...
+
+#define __PERFCTR_PARAM(name, param_name, type, param_type, desc)	\
+	static type param_name[MAX_PERF_COUNTERS];			\
+	static unsigned int param_name##_count;				\
+	module_param_array_named(name, param_name, param_type,		\
+				 &param_name##_count, 0644);		\
+	MODULE_PARM_DESC(name, desc);
+
+#define PERFCTR_PARAM(name, type, param_type, desc)	\
+	__PERFCTR_PARAM(name, name##_param, type, param_type, desc)
```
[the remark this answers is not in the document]
bea-arm: Done

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,490 @@
...
+			goto activate_failed;
+
+		entry->event =
+			/* No overflow handler, at least not at this point */
+			perf_event_create_kernel_counter(attr, cpu, NULL,
+							 NULL, NULL);
+		if (IS_ERR(entry->event)) {
+			pr_err("Failed to create counter id=%llu on cpu%d\n",
+			       attr->config, cpu);
+			kfree(entry);
```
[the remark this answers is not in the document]
bea-arm: Done

lisa/_assets/kmodules/sched_tp/perf_counters.c
```diff
@@ -0,0 +1,490 @@
...
+
+static int perfctr_disable(struct feature *feature)
+{
+	struct perfctr_core *perf_data = feature->data;
+
+	if (!perf_data)
+		return 0;
+
+	if (perf_data->pcpu_data) {
+		perfctr_events_release(perf_data);
```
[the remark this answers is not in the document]
bea-arm: Yes, but if there are no events registered it will bail out

lisa/tests/base.py
```diff
@@ -1860,6 +1863,135 @@ def test_noisy_tasks(self, *, noise_threshold_pct=None, noise_threshold_ms=None)
...
+            * A ``runtime`` column with total runtime reported on given CPU
+            * A ``counter`` column with CPU_CYCLES event counter
+            * A ``freq``column with the estimated frequency
+        """
+        try:
+            df_perf = self.trace.df_event('perf_counter')
+        except:
+            return pd.DataFrame()
+
+        # CPU_CYCLES counter:
```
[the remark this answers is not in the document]
bea-arm: As agreed offline I will drop the changes from lisa tests that pull in the PMU counters support so I will leave this and relevant comments .

douglas-raillard-arm:
Drive-by comment: you might need to update any tracepoint probe you used to ensure their first param is a `void *`, as was done in that PR: https://github.com/ARM-software/lisa/pull/2044

bea-arm:
Dropped changes validating expected frequency from Lisa tests.
Ongoing wok on supporting runtime configuration through vfs interface.

douglas-raillard-arm:

PR #1950, opened 2023-03-09
lisa._assets.kmodules: Support hot-reloading in kernel module

douglas-raillard-arm:
FEATURE

Add a lisa/reload debugfs file that can hot-reload the module when "all\n" is written to it.


douglas-raillard-arm:
Before merging, we need to make sure the module will work correctly without debugfs, as in some Android kernels config

douglas-raillard-arm:

PR #1957, opened 2023-03-13
Fix trace parser detection

douglas-raillard-arm:


derkling:
This fixes the bug reported in:
https://pastebin.com/1t0PSVap

PR #1986, opened 2023-05-19
[WIP] Add lisa-load-kmod --feature-param

douglas-raillard-arm:
WIP DO NOT MERGE

This is to be considered a technological preview until questions around managing multiple feature instances and different option sets are resolved.

### DIFF
--- a/lisa/_cli_tools/lisa_load_kmod.py
+++ b/lisa/_cli_tools/lisa_load_kmod.py
@@ -20,6 +20,12 @@ def main():
             action='append',
             help='Enable a specific module feature. Can be repeated. By default, the module will try to enable all features and will log in dmesg the ones that failed to enable'
         ),
+        'feature-param': dict(
+            action='append',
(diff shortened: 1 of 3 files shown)


douglas-raillard-arm:
I'll have to think a bit more on how multiple simultaneous uses of the module will coexist e.g. the user used lisa-load-kmod to enable a given feature, and then run a test that itself requires the module with another feature set. 

So far it was not too bad as the only features around were ftrace-related, and FtraceCollector therefore was simply in charge of loading the module. Also, it would have been quite easy to take the union of features somehow.

But with feature parameters, that opens the door to the same feature being enabled multiple times with different configurations. Either the same loaded module will need to support that, or we will need to support loading the same module multiple times. This is an issue as the module name is simply `lisa`, and that is somewhat necessary to be able to interact with it in 3rd party scripts (if only for lsmod/insmod/rmmod)

douglas-raillard-arm:

PR #2025, opened 2023-06-02
[RFC] lisa.target: Add hooks

douglas-raillard-arm:
FEATURE

Add lisa.target.Target(hooks={'post-connect': ['echo hello', 'echo world']}, ...) parameter. This allows running custom commands on the target after connection, which can be useful to do custom setup on some devices such as disabling SE Linux.


douglas-raillard-arm:

PR #2038, opened 2023-06-26
lisa/_assets/kmodules/lisa: Support in-tree builds

mrkajetanp:
FEATURE

Modify the Makefile for the lisa module to distinguish between in-tree and out-of-tree building. Adjust the imports and definitions in the source files accordingly.

With this change, the module can now be copy-pasted into a kernel tree and built with kbuild like any other kernel module. The default, out-of-tree build method with Lisa should work unchanged as before.

### DIFF
--- a/doc/setup.rst
+++ b/doc/setup.rst
@@ -198,6 +198,33 @@ To help expose these tracepoints (and any additional one we might require in
 the future) as trace events, an external module is required and is provided
 under the name of "lisa" in $LISA_HOME/tools/kmodules/lisa
 
+Pre-requisites
+..............
+
+CFI
+~~~
+
+Using the out-of-tree build method for kernels with CONFIG_CFI_CLANG=y as all
+Android kernels come by default requires the module to be built with at least
+clang-16. This can either be achieved by using the ``alpine`` build environment,
+by having it installed on host and using ``LLVM=1`` or forcing the version with
+``LLVM=-16`` in ``target-conf/kernel/modules/make-variables``.
+
+Kernel symbols needed for reading files on Android product kernels
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to use some Lisa module features (e.g. the lisa__pixel6_emeter ftrace event)
+on a product kernel, some symbols forbidden by Google need to be re-enabled.
+
+In order to do that, the kernel will need to be built with:
+
+.. code-block:: sh
+
+    ./update_symbol_list.sh
+
+The script should be included in the product kernel tree. It will ensure that the required
+symbols are not stripped from the final kernel image and the module does not get rejected.
+
 Enabling a module
 .................
 
@@ -337,6 +364,53 @@ Integrating the module in your kernel tree
 If you're rebuilding your kernel tree anyway, it might be easier to integrate
 the module into your kernel tree as a built-in module so that it's always
 present.
+
+.. warning::
+   This method is less supported than the out-of-tree method above.
+   Proceed only if this method has specific benefits for your use case or
+   for some reason you cannot build the module out-of-tree. The most common
+   reason would be accessing the kernel sources remotely thus not having a
+   kernel checkout on the machine that Lisa runs on. Otherwise, using the
+   default out-of-tree build method will likely be less error-prone.
+
+In order to do that, follow the steps below:
+
+* Disable Google's ABI symbols checks by applying the patch found under
+  ``tools/kmodules/lisa-in-tree/android/abi`` to the tree in ``build/abi``.
+
+* Apply the patches in ``tools/kmodules/lisa-in-tree/linux``
+  to include a stub Kbuild Makefile structure for the module.
+  For Android product kernels it should be applied under ``private/gs-google``,
+  for Android mainline kernels under ``common``.
+
+.. note:: Older Android product kernels might be missing some internal header
+   import guards present in newer mainline versions. For this method to work
+   make sure your kernel tree includes mainline commits 95458477f5b2dc436e3aa6aa25c0f84bb83e6195
+   and d90a2f160a1cd9a1745896c381afdf8d2812fd6b.
+
+* Additionally, on Android kernels it can be useful to apply the patches in
+  ``tools/kmodules/lisa-in-tree/android`` as well. It will include the module
+  in the vendor modules list for Android so that it is automatically loaded
+  at boot-time. The patch is specific to the Pixel 6 source tree
+  and very likely should be adjusted accordingly for any other platform.
+
+* Then, put the script found under ``tools/kmodules/lisa-in-tree/fetch_lisa_module.py``
+  and follow the instructions in ``--help`` to link or fetch the Lisa module sources into
+  the source tree.
+
+With all these steps complete, rebuild the kernel:
+
+.. code-block:: sh
+
+    ./update_symbol_list.sh
+
+The module should be built in-tree and then loaded at boot-time.
+
+.. note:: The order at which the module is loaded at boot time is not guaranteed and
+   Android will not perform any of the Lisa module setup steps. Usually e.g. ``pixel6_emeter``
+   will fail to load on boot and the module will have to be reloaded with ``rmmod lisa && modprobe (..)``.
+   As loading the module in ways different than through Lisa is not officially supported, any such
+   setup is the user's responsibility.
 
 Updating
 ========
--- a/lisa/_assets/kmodules/lisa/Makefile
+++ b/lisa/_assets/kmodules/lisa/Makefile
@@ -25,9 +25,6 @@ else
 	KERNEL_SRC ?= /lib/modules/`uname -r`/build
 endif
 
-VMLINUX_H = $(MODULE_OBJ)/vmlinux.h
-SYMBOL_NAMESPACES_H = $(MODULE_OBJ)/symbol_namespaces.h
-
 # kbuild part of makefile. Only Kbuild-related targets should be used here to
 # avoid any sort of clash.
 ifneq ($(KERNELRELEASE),)
@@ -35,12 +32,28 @@ ifneq ($(KERNELRELEASE),)
 LISA_KMOD_NAME ?= lisa
 obj-m := $(LISA_KMOD_NAME).o
 $(LISA_KMOD_NAME)-y := main.o tp.o wq.o features.o pixel6.o
-ldflags-y += -T "$(M)/features.lds"
-
-clean-files := vmlinux.h
 
 # -fno-stack-protector is needed to possibly undefined __stack_chk_guard symbol
 ccflags-y = "-I$(MODULE_SRC)" -std=gnu11 -fno-stack-protector -Wno-declaration-after-statement
+
+FEATURES_LDS := features.lds
+
+SYMBOL_NAMESPACES_H = $(MODULE_OBJ)/symbol_namespaces.h
+
+# in-tree build
+ifneq ($(srctree),.)
+
+ccflags-y += -I$(srctree) -D_IN_TREE_BUILD
+ldflags-y += -T $(srctree)/$(obj)/$(FEATURES_LDS)
+
+# out-of-tree build
+else
+
+VMLINUX_H = $(MODULE_OBJ)/vmlinux.h
+
+ldflags-y += -T $(M)/$(FEATURES_LDS)
+
+clean-files := vmlinux.h
 
 VMLINUX_TXT = $(MODULE_SRC)/private_types.txt
 
@@ -91,6 +104,9 @@ $(VMLINUX_H): $(VMLINUX_TXT) $(VMLINUX)
 	sed -r -n 's/.*(struct|union|enum) ([0-9a-zA-Z_]*) .*/#define TYPE_DEFINED_\1_\2/p' _header_no_comment | sort -u >> _fwd_decl
 	cat _fwd_decl _header > $@
 # cat $@
+
+# out-of-tree build
+endif
 
 # Some kernels require the use of MODULE_IMPORT_NS() before using symbols that are part of the given namespace:
 # https://docs.kernel.org/core-api/symbol-namespaces.html
--- a/lisa/_assets/kmodules/lisa/Makefile
+++ b/lisa/_assets/kmodules/lisa/Makefile
@@ -39,6 +39,7 @@ ccflags-y = "-I$(MODULE_SRC)" -std=gnu11 -fno-stack-protector -Wno-declaration-a
 FEATURES_LDS := features.lds
 
 SYMBOL_NAMESPACES_H = $(MODULE_OBJ)/symbol_namespaces.h
+MODULE_VERSION_H = $(MODULE_OBJ)/module_version.h
 
 # in-tree build
 ifneq ($(srctree),.)
@@ -121,8 +122,13 @@ $(SYMBOL_NAMESPACES_H):
 	find "$(KERNEL_SRC)" '(' -name '*.c' -o -name '*.h' ')' -print0 | xargs -0 sed -n 's/MODULE_IMPORT_NS([^;]*;/\0/p' | sort -u > $@
 	echo "MODULE_IMPORT_NS(VFS_internal_I_am_really_a_filesystem_and_am_NOT_a_driver);" >> $@
 
+$(MODULE_VERSION_H):
+	echo -n '#ifndef _MODULE_VERSION_H\n#define _MODULE_VERSION_H\n#define LISA_MODULE_VERSION "' > $@
+	cat $$(find $(KERNEL_SRC)/$(MODULE_SRC)/ -name '*.*' | sort) | sha1sum | awk '{printf "%s\"\n", $$1}' >> $@
+	echo '#endif' >> $@
+
 # Make all object files depend on the generated sources
-$(addprefix $(MODULE_OBJ)/,$($(LISA_KMOD_NAME)-y)): $(VMLINUX_H) $(SYMBOL_NAMESPACES_H)
+$(addprefix $(MODULE_OBJ)/,$($(LISA_KMOD_NAME)-y)): $(VMLINUX_H) $(SYMBOL_NAMESPACES_H) $(MODULE_VERSION_H)
 
 # Non-Kbuild part
 else
--- a/lisa/_kmod.py
+++ b/lisa/_kmod.py
@@ -127,6 +127,7 @@ import glob
 import collections
 import collections.abc
 import hashlib
+import errno
 from operator import itemgetter
 from shlex import quote
 from io import BytesIO
@@ -137,7 +138,7 @@ from elftools.elf.elffile import ELFFile
 
 from devlib.target import KernelVersion, TypedKernelConfig, KernelConfigTristate
 from devlib.host import LocalConnection
-from devlib.exception import TargetStableError
+from devlib.exception import TargetStableError, TargetStableCalledProcessError
 
 from lisa.utils import nullcontext, Loggable, LISA_CACHE_HOME, checksum, DirCache, chain_cm, memoized, LISA_HOST_ABI, subprocess_log, SerializeViaConstructor, destroyablecontextmanager, ContextManagerExit, ignore_exceps, get_nested_key
 from lisa._assets import ASSETS_PATH, HOST_PATH, ABI_BINARIES_FOLDER
@@ -2284,5 +2285,42 @@ class LISAFtraceDynamicKmod(FtraceDynamicKmod):
     @classmethod
     def _event_features(cls, events):
         return set(f'event__{event}' for event in events)
+
+    def install(self, kmod_params=None):
+        try:
+            self.uninstall()
+        except Exception:
+            pass
+
+        busybox = quote(self.target.busybox)
+        modules_path_base = '/lib/modules'
+        modules_version = self.target.kernel_version.release
+
+        if self.target.os == 'android':
+            modules_path_base = f'/vendor_dlkm{modules_path_base}'
+            try:
+                modules_version = Path(self.target.execute(
+                    f"{busybox} find {modules_path_base} -maxdepth 1 -mindepth 1 | {busybox} head -1"
+                ).strip()).name
+            except TargetStableCalledProcessError:
+                pass
+
+        modules_path = f"{modules_path_base}/{modules_version}"
+        lisa_module_filename = f"{self.src.mod_name}.ko"
+
+        try:
+            lisa_module_path = self.target.execute(
+                f"{busybox} find {modules_path} -name {quote(lisa_module_filename)}"
+            ).strip()
+
+            self.target.execute(
+                f"{busybox} insmod {lisa_module_path} version={self.src.checksum_sources}"
+            )
+        except TargetStableCalledProcessError as e:
+            if e.returncode == errno.EPROTO:
+                raise ValueError('In-tree module version does not match what Lisa expects.')
+            super().install(kmod_params=kmod_params)
+        else:
+            self.logger.debug(f'Loaded module {self.src.mod_name} from kernel modules directory')
 
 # vim :set tabstop=4 shiftwidth=4 expandtab textwidth=80
--- a/lisa/_kmod.py
+++ b/lisa/_kmod.py
@@ -1708,17 +1708,35 @@ class KmodSrc(Loggable):
             if name.endswith('.c')
         }
 
+    def _checksum(self, sources_only=False):
+        m = hashlib.sha1()
+        sources = {
+            name: content
+            for name, content
+            in self.src.items()
+            if '.' in name
+        } if sources_only else self.src
+        for name, content in sorted(sources.items()):
+            if not sources_only:
+                m.update(name.encode('utf-8'))
+            m.update(content)
+        return m.hexdigest()
+
     @property
     @memoized
     def checksum(self):
         """
+        Checksum of the module's sources & Makefile.
+        """
+        return self._checksum(sources_only=False)
+
+    @property
+    @memoized
+    def checksum_sources(self):
+        """
         Checksum of the module's sources.
         """
-        m = hashlib.sha256()
-        for name, content in sorted(self.src.items()):
-            m.update(name.encode('utf-8'))
-            m.update(content)
-        return m.hexdigest()
+        return self._checksum(sources_only=True)
 
     @property
     @memoized
--- a/tools/kmodules/lisa-in-tree/fetch_lisa_module.py
+++ b/tools/kmodules/lisa-in-tree/fetch_lisa_module.py
@@ -1,0 +1,74 @@
+#! /usr/bin/env python3
+
+import shutil
+import argparse
+import subprocess
+
+from pathlib import Path
+
+
+def main():
+    parser = argparse.ArgumentParser(description="""
+    Fetch the Lisa module from GitHub or a local copy and integrate it
+    inside the kernel tree.
+
+    Method 1 - using LISA_HOME.
+
+    If the script finds the LISA_HOME environment variable (present after sourcing the Lisa shell)
+    it will symlink the Lisa module sources from the local Lisa git checkout into the kernel tree.
+    That way updating the local Lisa tree will automatically update the module sources in the kernel
+    tree and the changes will be present after the next kernel rebuild.
+    In this case, the script only needs to be run once.
+
+    Method 2 - Sparse checkout.
+
+    If the script doesn't find LISA_HOME (e.g. building on an external server without a Lisa checkout)
+    the alternative mode is to directly clone the module sources into the tree.
+                                     """)
+    parser.add_argument('-g', '--git-ref', help='Lisa git reference to checkout')
+    parser.add_argument('-f', '--force', action='store_true', help='Override the module checkout if it already exists')
+    parser.add_argument('--module-kernel-path',
+                        help='Path relative to the kernel tree root where the module will be stored')
+    parser.add_argument('--git-remote', help='Git remote to pull the module from',
+                        default='git@github.com:ARM-software/lisa.git')
+    args = parser.parse_args()
+
+    module_kernel_path = Path(args.module_kernel_path).resolve()
+
+    # --git-ref passed, clone the sparse checkout
+    if args.git_ref:
+        module_git_path = module_kernel_path.parent / f"{module_kernel_path.name}-git"
+
+        # module has already been cloned
+        if module_git_path.exists():
+            if not args.force:
+                parser.error('Module checkout already exists, pass --force to override')
+            shutil.rmtree(module_git_path)
+            module_kernel_path.unlink()
+
+        clone_cmd = [
+            'git', 'clone', '--verbose', '--no-checkout', '--no-tags', '--filter=tree:0',
+            args.git_remote, module_git_path
+        ]
+        subprocess.check_call(clone_cmd)
+        subprocess.check_call(['git', '-C', module_git_path,
+                               'sparse-checkout', 'set', 'lisa/_assets/kmodules/'])
+        subprocess.check_call(['git', '-C', module_git_path, 'checkout', args.git_ref])
+
+        module_src_kernel_git_path = module_git_path / 'lisa/_assets/kmodules/lisa'
+        module_kernel_path.symlink_to(
+            module_src_kernel_git_path.relative_to(module_kernel_path.parent)
+        )
+        return
+
+    # --git-ref not passed, try linking from LISA_HOME
+    try:
+        import lisa._assets.kmodules.lisa as kmod_mod
+    except ImportError:
+        parser.error('Either lisa Python package needs to be importable or --git-ref needs to be used')
+    else:
+        module_src_lisa_path = Path(kmod_mod.__path__[0]).resolve()
+        module_kernel_path.symlink_to(module_src_lisa_path)
+
+
+main()
--- a/lisa/_assets/kmodules/lisa/main.c
+++ b/lisa/_assets/kmodules/lisa/main.c
@@ -3,6 +3,7 @@
 
 #include "main.h"
 #include "features.h"
+
 /* Import all the symbol namespaces that appear to be defined in the kernel
  * sources so that we won't trigger any warning
  */
--- a/lisa/_assets/kmodules/lisa/main.c
+++ b/lisa/_assets/kmodules/lisa/main.c
@@ -29,7 +29,7 @@ static int __init modinit(void) {
 	pr_info("Loading Lisa module version %s\n", LISA_MODULE_VERSION);
 	if (strcmp(version, LISA_MODULE_VERSION)) {
 		pr_err("Lisa module version check failed. Got %s, expected %s\n", version, LISA_MODULE_VERSION);
-		return -EINVAL;
+		return -EPROTO;
 	}
 
 	ret = init_features(features, features_len);
--- a/lisa/_assets/kmodules/lisa/main.c
+++ b/lisa/_assets/kmodules/lisa/main.c
@@ -3,15 +3,18 @@
 
 #include "main.h"
 #include "features.h"
-
+#include "module_version.h"
 /* Import all the symbol namespaces that appear to be defined in the kernel
  * sources so that we won't trigger any warning
  */
 #include "symbol_namespaces.h"
 
+static char* version = LISA_MODULE_VERSION;
+module_param(version, charp, 0);
+MODULE_PARM_DESC(version, "Module version defined as sha1sum of the module sources");
+
 static char *features[MAX_FEATURES];
 unsigned int features_len = 0;
-
 module_param_array(features, charp, &features_len, 0);
 MODULE_PARM_DESC(features, "Comma-separated list of features to enable. Available features are printed when loading the module");
 
@@ -21,7 +24,15 @@ static void modexit(void) {
 }
 
 static int __init modinit(void) {
-	int ret = init_features(features, features_len);
+	int ret;
+
+	pr_info("Loading Lisa module version %s\n", LISA_MODULE_VERSION);
+	if (strcmp(version, LISA_MODULE_VERSION)) {
+		pr_err("Lisa module version check failed. Got %s, expected %s\n", version, LISA_MODULE_VERSION);
+		return -EINVAL;
+	}
+
+	ret = init_features(features, features_len);
 
 	if (ret) {
 		pr_err("Some errors happened while loading LISA kernel module\n");
--- a/lisa/_assets/kmodules/lisa/pixel6.c
+++ b/lisa/_assets/kmodules/lisa/pixel6.c
@@ -4,6 +4,10 @@
 #include <linux/fs.h>
 #include <linux/types.h>
 
+#ifdef _IN_TREE_BUILD
+#include <kernel/sched/sched.h>
+#endif
+
 #include "main.h"
 #include "features.h"
 #include "wq.h"
--- a/lisa/_assets/kmodules/lisa/pixel6.c
+++ b/lisa/_assets/kmodules/lisa/pixel6.c
@@ -5,6 +5,7 @@
 #include <linux/types.h>
 
 #ifdef _IN_TREE_BUILD
+#include <linux/sched/cputime.h>
 #include <kernel/sched/sched.h>
 #endif
 
--- a/lisa/_assets/kmodules/lisa/sched_helpers.h
+++ b/lisa/_assets/kmodules/lisa/sched_helpers.h
@@ -7,8 +7,11 @@
 
 #include <linux/cgroup.h>
 
-/* private kernel headers are missing, redefine some functions */
-#ifndef cap_scale
+#ifdef _IN_TREE_BUILD
+
+#include <kernel/sched/autogroup.h>
+
+#else
 
 #include "vmlinux.h"
 
@@ -79,7 +82,7 @@ unsigned long uclamp_rq_util_with(struct rq *rq, unsigned long util,
 #endif
 }
 
-#endif /* cap_scale */
+#endif /* _IN_TREE_BUILD */
 
 static inline void cfs_rq_tg_path(struct cfs_rq *cfs_rq, char *path, int len)
 {
--- a/lisa/_assets/kmodules/lisa/sched_helpers.h
+++ b/lisa/_assets/kmodules/lisa/sched_helpers.h
@@ -7,8 +7,10 @@
 
 #include <linux/cgroup.h>
 
-#include "vmlinux.h"
+/* private kernel headers are missing, redefine some functions */
+#ifndef cap_scale
 
+#include "vmlinux.h"
 
 #ifdef CONFIG_FAIR_GROUP_SCHED
 static inline struct rq *rq_of(struct cfs_rq *cfs_rq)
@@ -56,6 +58,29 @@ static int autogroup_path(struct task_group *tg, char *buf, int buflen)
 #endif
 }
 
+/* A cut down version of the original. @p MUST be NULL */
+static __always_inline
+unsigned long uclamp_rq_util_with(struct rq *rq, unsigned long util,
+				  struct task_struct *p)
+{
+#ifdef CONFIG_UCLAMP_TASK
+	unsigned long min_util;
+	unsigned long max_util;
+
+	min_util = READ_ONCE(rq->uclamp[UCLAMP_MIN].value);
+	max_util = READ_ONCE(rq->uclamp[UCLAMP_MAX].value);
+
+	if (unlikely(min_util >= max_util))
+		return min_util;
+
+	return clamp(util, min_util, max_util);
+#else
+	return util;
+#endif
+}
+
+#endif /* cap_scale */
+
 static inline void cfs_rq_tg_path(struct cfs_rq *cfs_rq, char *path, int len)
 {
 	if (!path)
@@ -86,28 +111,6 @@ static inline struct cfs_rq *get_se_cfs_rq(struct sched_entity *se)
 	return se->cfs_rq;
 #else
 	return NULL;
-#endif
-}
-
-
-/* A cut down version of the original. @p MUST be NULL */
-static __always_inline
-unsigned long uclamp_rq_util_with(struct rq *rq, unsigned long util,
-				  struct task_struct *p)
-{
-#ifdef CONFIG_UCLAMP_TASK
-	unsigned long min_util;
-	unsigned long max_util;
-
-	min_util = READ_ONCE(rq->uclamp[UCLAMP_MIN].value);
-	max_util = READ_ONCE(rq->uclamp[UCLAMP_MAX].value);
-
-	if (unlikely(min_util >= max_util))
-		return min_util;
-
-	return clamp(util, min_util, max_util);
-#else
-	return util;
 #endif
 }
 
--- a/lisa/_assets/kmodules/lisa/tp.c
+++ b/lisa/_assets/kmodules/lisa/tp.c
@@ -1,6 +1,10 @@
 /* SPDX-License-Identifier: GPL-2.0 */
 #include <linux/slab.h>
+#ifndef _IN_TREE_BUILD
 #include <linux/sched.h>
+#else
+#include <kernel/sched/sched.h>
+#endif
 #include <trace/events/sched.h>
 
 #define CREATE_TRACE_POINTS
--- a/lisa/_assets/kmodules/lisa/tp.c
+++ b/lisa/_assets/kmodules/lisa/tp.c
@@ -1,8 +1,10 @@
 /* SPDX-License-Identifier: GPL-2.0 */
 #include <linux/slab.h>
+
 #ifndef _IN_TREE_BUILD
 #include <linux/sched.h>
 #else
+#include <linux/sched/cputime.h>
 #include <kernel/sched/sched.h>
 #endif
 #include <trace/events/sched.h>
--- a/tools/kmodules/lisa-in-tree/android/0002-arm-vh-Include-Lisa-module-in-the-vendor-modules.patch
+++ b/tools/kmodules/lisa-in-tree/android/0002-arm-vh-Include-Lisa-module-in-the-vendor-modules.patch
@@ -1,0 +1,38 @@
+From ef77f0e015f41cf5b53e9d6089044b8a260e4831 Mon Sep 17 00:00:00 2001
+From: Kajetan Puchalski <kajetan.puchalski@arm.com>
+Date: Tue, 27 Jun 2023 14:07:42 +0000
+Subject: [PATCH] arm: vh: Include Lisa module in the vendor modules
+
+Include lisa.ko (the Lisa tracing module) in the list of
+vendor modules included in the kernel so that it is copied
+to dist and loaded at boot time automatically alongside all
+the other vendor modules on Pixel 6.
+---
+ BUILD.bazel               | 1 +
+ vendor_boot_modules.gs101 | 1 +
+ 2 files changed, 2 insertions(+)
+
+diff --git a/BUILD.bazel b/BUILD.bazel
+index 2c6b1564d390..5a3070bb5653 100644
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -210,6 +210,7 @@ kernel_build(
+         "zcomp_eh.ko",
+         "zram.ko",
+         "zsmalloc.ko",
++        "lisa.ko",
+     ],
+     deps = [
+         "//prebuilts/misc/linux-x86/libufdt:mkdtimg",
+diff --git a/vendor_boot_modules.gs101 b/vendor_boot_modules.gs101
+index 4d913d9d4788..79447f2d5de0 100644
+--- a/vendor_boot_modules.gs101
++++ b/vendor_boot_modules.gs101
+@@ -209,3 +209,4 @@ zcomp_cpu.ko
+ zcomp_eh.ko
+ zram.ko
+ zsmalloc.ko
++lisa.ko
+-- 
+2.34.1
+
--- a/tools/kmodules/lisa-in-tree/android/abi/0001-abi-Allow-GKI-restricted-symbols.patch
+++ b/tools/kmodules/lisa-in-tree/android/abi/0001-abi-Allow-GKI-restricted-symbols.patch
@@ -1,0 +1,29 @@
+From 7b6f4db61675012148ce872b1e86bbe0806e57ba Mon Sep 17 00:00:00 2001
+From: Kajetan Puchalski <kajetan.puchalski@arm.com>
+Date: Mon, 31 Jul 2023 11:18:50 +0000
+Subject: [PATCH] abi: Allow GKI restricted symbols
+
+---
+ abi/symbols.deny | 6 +++---
+ 1 file changed, 3 insertions(+), 3 deletions(-)
+
+diff --git a/abi/symbols.deny b/abi/symbols.deny
+index fd643b7..b15771c 100644
+--- a/abi/symbols.deny
++++ b/abi/symbols.deny
+@@ -17,9 +17,9 @@
+ # File access symbols that are forbidden because drivers should never
+ # try to access files directly.  Instead proper user/kernel apis should
+ # always be used.
+-filp_open	Drivers should not open files directly
+-kernel_read	Drivers should never read from a file
+-kernel_write	Drivers should never write to a file
++# filp_open	Drivers should not open files directly
++# kernel_read	Drivers should never read from a file
++# kernel_write	Drivers should never write to a file
+ 
+ # Kprobe needs to be left alone
+ disable_kprobe		kprobe should not be touched by any driver
+-- 
+2.34.1
+
--- a/tools/kmodules/lisa-in-tree/linux/0001-arm-vh-Include-Lisa-module-stub.patch
+++ b/tools/kmodules/lisa-in-tree/linux/0001-arm-vh-Include-Lisa-module-stub.patch
@@ -1,0 +1,56 @@
+From 4251bcc1e0641992e3a99f19c654245d2b6aff87 Mon Sep 17 00:00:00 2001
+From: Kajetan Puchalski <kajetan.puchalski@arm.com>
+Date: Tue, 27 Jun 2023 14:05:40 +0000
+Subject: [PATCH] arm: vh: Include Lisa module stub
+
+Include the directory structure and necessary makefile changes
+for a stub Lisa module to be built in-tree as part of the kernel.
+For the module to work this commit requires fetching the actual
+sources from GitHub either manually or using a script.
+
+Signed-off-by: Kajetan Puchalski <kajetan.puchalski@arm.com>
+---
+ drivers/soc/Makefile                 | 1 +
+ drivers/soc/arm/Makefile             | 1 +
+ drivers/soc/arm/vh/kernel/.gitignore | 2 ++
+ drivers/soc/arm/vh/kernel/Makefile   | 2 ++
+ 4 files changed, 6 insertions(+)
+ create mode 100644 drivers/soc/arm/Makefile
+ create mode 100644 drivers/soc/arm/vh/kernel/.gitignore
+ create mode 100644 drivers/soc/arm/vh/kernel/Makefile
+
+diff --git a/drivers/soc/Makefile b/drivers/soc/Makefile
+index 71ef89c79ea7..af7be45a2776 100644
+--- a/drivers/soc/Makefile
++++ b/drivers/soc/Makefile
+@@ -30,3 +30,4 @@ obj-$(CONFIG_PLAT_VERSATILE)	+= versatile/
+ obj-y				+= xilinx/
+ obj-$(CONFIG_ARCH_ZX)		+= zte/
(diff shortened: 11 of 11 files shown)


douglas-raillard-arm:
That looks reasonable to me

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
 
+Pre-requisites
+.................
+
+CFI
+~~~~~
+
+First of all, building the module outside the kernel will not work if the kernel
+has CFI (Control-Flow Integrity) enabled as it will not allow loading of outside
+modules. Android kernels come with CFI by default so it will have to be disabled
```
douglas-raillard-arm, line 209: s/allow loading of outside module/dynamic loading of modules/

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
+in the config (``CONFIG_CFI_CLANG``).
+If you want to use the module with a CFI-enabled kernel, follow the steps under
+"Integrating the module in your kernel tree" below.
```
douglas-raillard-arm, line 212: use target and reference:
https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#explicit-targets

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
 the future) as trace events, an external module is required and is provided
 under the name of "lisa" in $LISA_HOME/tools/kmodules/lisa
 
+Pre-requisites
+.................
```
douglas-raillard-arm, line 202: match the number of dots with the length of the text.

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
+
+CFI
+~~~~~
```
douglas-raillard-arm, line 205: ditto for number of ~

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
+
+First of all, building the module outside the kernel will not work if the kernel
+has CFI (Control-Flow Integrity) enabled as it will not allow loading of outside
+modules. Android kernels come with CFI by default so it will have to be disabled
+in the config (``CONFIG_CFI_CLANG``).
+If you want to use the module with a CFI-enabled kernel, follow the steps under
+"Integrating the module in your kernel tree" below.
+
+Kernel symbols needed for pixel6_emeter on Android product kernels
+~~~~~
```
douglas-raillard-arm, line 215: ditto

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
+
+First of all, building the module outside the kernel will not work if the kernel
+has CFI (Control-Flow Integrity) enabled as it will not allow loading of outside
+modules. Android kernels come with CFI by default so it will have to be disabled
+in the config (``CONFIG_CFI_CLANG``).
+If you want to use the module with a CFI-enabled kernel, follow the steps under
+"Integrating the module in your kernel tree" below.
+
+Kernel symbols needed for pixel6_emeter on Android product kernels
```
douglas-raillard-arm, line 214: s/pixel6_emeter/``lisa__pixel6_emeter`` ftrace event/

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
+~~~~~
+
+In order to use the pixel6_emeter on a product kernel, some symbols forbidden
```
douglas-raillard-arm, line 217: s/pixel6_emeter/``lisa__pixel6_emeter`` ftrace event/

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
 the future) as trace events, an external module is required and is provided
 under the name of "lisa" in $LISA_HOME/tools/kmodules/lisa
 
+Pre-requisites
```
douglas-raillard-arm, line 201: This section should go under "Manual route". We really don't want people to start getting into that unless they really have to, so the first thing the user must read is:
> LISA Python package will compile and load the module automatically when required for tracing so there is usually no reason to do so manually

This should be properly "under" that manual section as this is a child of it, it is not a sibling.

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
+.. code-block:: sh
+    BUILD_AOSP_KERNEL=1 ./build_slider.sh
+
+If any of the symbols are changed, the kernel needs to be rebuilt with:
+
+.. code-block:: sh
+    ./update_symbol_list.sh
+
+* Firstly, apply the patch found in the Lisa source tree under
+  ``tools/kmodules/lisa-in-tree/0001-fs-Move-filp_open-kernel_read-write-outside-the-name.patch``
```
douglas-raillard-arm, line 235: Don't include the file name in that path, as this will likely change over time. Just say to apply all the patches found in ``tools/kmodules/lisa-in-tree``

douglas-raillard-arm (changes requested):
Looks good generally speaking, it mainly needs:
* A bit of "details removal" to ease maintenance (e.g. less references to pixel6_emeter as this could change, and other features might start needing things as well)
* Clear separation in patches needed for android and what works on mainline

doc/setup.rst
mrkajetanp: The pre-requisites are not related to the build method, they apply to the automatic build method just as much as they do to the manual.

Putting the pre-requisites for the automatic build under the manual build section makes no sense, since it applies to both it should be a sibling section that comes before the two imo.

Without following the steps in pre-requisites the automatic method will fail on every Android build - because of the symbols and CFI and product kernels and because of CFI on mainline ones.

The automatic method without pre-requisites pretty much only works on Juno and some servers that can run pure mainline Arm kernels, very few devices in the grand scheme of things.

douglas-raillard-arm: Ok that makes sense.

>pretty much only works on Juno and some servers that can run pure mainline Arm kernels, very few devices in the grand scheme of things.

There are hundreds if not thousands of devices that fall into that category. Any amd64 desktop, laptop, server in the past 17 years basically. We just happen to be more interested in the few locked down devices android phones has brought in the recent years, but that's hardly the majority of devices, and that's definitely not the focus of the entire LISA userbase so we very much want to keep android-isms as a layer on top of the rest, not as a base assumption.

lisa/_assets/kmodules/lisa/Makefile
```diff
@@ -25,23 +25,36 @@ else
...
 
 # -fno-stack-protector is needed to possibly undefined __stack_chk_guard symbol
 ccflags-y = -I$(MODULE_SRC) -std=gnu11 -fno-stack-protector -Wno-declaration-after-statement
 
+FEATURES_LDS := features.lds
+
+# in-tree build
+ifneq ($(srctree),.)
+
+EXTRA_CFLAGS += -I$(srctree) -D_IN_TREE_BUILD
```
douglas-raillard-arm, line 44: `EXTRA_CFLAGS` is deprecated, use ccflags-y instead (make sure to append to and not discard the existing ccflags-y):

```
    ccflags-y, asflags-y and ldflags-y
	These three flags apply only to the kbuild makefile in which they
	are assigned. They are used for all the normal cc, as and ld
	invocations happening during a recursive build.
	Note: Flags with the same behaviour were previously named:
	EXTRA_CFLAGS, EXTRA_AFLAGS and EXTRA_LDFLAGS.
	They are still supported but their usage is deprecated.
```
https://www.kernel.org/doc/Documentation/kbuild/makefiles.txt

douglas-raillard-arm (changes requested):

doc/setup.rst
```diff
@@ -198,6 +198,55 @@ To help expose these tracepoints (and any additional one we might require in
...
+  to the ``aosp/`` tree.
+
+* Secondly, apply the patch found under
```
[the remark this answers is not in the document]
mrkajetanp: I'll just put them in separate folders, the two patches need to be applied to different git trees inside the kernel anyway. It'll avoid the explicit naming parts.

lisa/_assets/kmodules/lisa/Makefile
mrkajetanp: Ah right, I copied that line from Google's tracing module, of course it's deprecated :))

mrkajetanp:
I pushed the doc & Makefile changes, I'll rewrite the script in python in a moment.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+LISA_MODULE_GIT_DIR = "lisa-module-git"
+
+parser = argparse.ArgumentParser(prog='Lisa module in-tree helper')
+init_group = parser.add_mutually_exclusive_group()
+init_group.add_argument('hash', nargs='?', help='Lisa repository hash to checkout')
+parser.add_argument('-u', '--update', action='store_true', help='Run git fetch before checking out the hash')
+args = parser.parse_args()
+
+# LISA_HOME is present, use the bundled module
+module_kernel_path = os.path.join('.', LISA_MODULE_DIR_KERNEL_PATH, LISA_MODULE_DIR)
```
douglas-raillard-arm, line 49: Might be worth adding a CLI option to override that, defaulting to the value you currently use (`parser.add_argument(..., default=...)`)

douglas-raillard-arm, line 49: This should be either be moved inside the `if` branch or re-used in the `else` branch.
The `else` branch seems to be using a non-relative version of that path, but I'm not sure why.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # module has already been cloned
+    if os.path.exists(module_git_path):
+        if not args.hash:
+            parser.print_help()
+            sys.exit(1)
+
+        if args.update:
+            subprocess.check_call(['git', '-C', module_git_path, 'fetch', 'origin', LISA_MODULE_REMOTE_BRANCH])
+        subprocess.check_call(['git', '-C', module_git_path, 'checkout', args.hash])
```
douglas-raillard-arm, line 70: I'd skip that and just always make a new clone, using the recipe for a partial clone and checkout I shared with you offline. If for some reason the folder belongs to another git tree, we might end up checking out somewhere else, or the checkout could fail because of modified files or whatever.

All we want is the script to create the folder with the sources of the module with a clean set of files, and possibly raise an error if the folder already exists with an option to force overriding the path.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+    else:
+        clone_cmd = [
+            'git', 'clone', '--verbose', '--no-checkout', '--no-tags', '--filter=tree:0',
+            LISA_MODULE_REMOTE, module_git_path
+        ]
+        subprocess.check_call(clone_cmd)
+        subprocess.check_call(['git', '-C', module_git_path, 'sparse-checkout', 'set', 'lisa/_assets/kmodules/'])
+        subprocess.check_call(['git', '-C', module_git_path, 'checkout', LISA_MODULE_REMOTE_BRANCH])
```
douglas-raillard-arm, line 78: The checkout should be done on the commit passed by the user

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+import argparse
+import subprocess
+
+# module directory setup
+LISA_MODULE_DIR = "lisa"
+LISA_MODULE_DIR_KERNEL_PATH = "private/gs-google/drivers/soc/arm/vh/kernel"
+
+# sparse clone remote setup
+LISA_MODULE_REMOTE = "git@github.com:mrkajetanp/lisa.git"
+LISA_MODULE_REMOTE_BRANCH = "kmod-in-tree"
```
douglas-raillard-arm, line 39: Looks like this global should be removed

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+import sys
+import argparse
+import subprocess
+
+# module directory setup
+LISA_MODULE_DIR = "lisa"
+LISA_MODULE_DIR_KERNEL_PATH = "private/gs-google/drivers/soc/arm/vh/kernel"
+
+# sparse clone remote setup
+LISA_MODULE_REMOTE = "git@github.com:mrkajetanp/lisa.git"
```
douglas-raillard-arm, line 38: This should be turned into an argparse option, defaulting to the upstream LISA repo

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+        ]
+        subprocess.check_call(clone_cmd)
+        subprocess.check_call(['git', '-C', module_git_path, 'sparse-checkout', 'set', 'lisa/_assets/kmodules/'])
+        subprocess.check_call(['git', '-C', module_git_path, 'checkout', LISA_MODULE_REMOTE_BRANCH])
+
+        module_src_kernel_git_path = os.path.join(module_git_path, 'lisa/_assets/kmodules/lisa')
+        link_cmd = [
+            'ln', '-sr', module_src_kernel_git_path, module_kernel_path
+        ]
+        subprocess.check_call(link_cmd)
```
douglas-raillard-arm, line 84: ditto about the symlink command

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+# LISA_HOME is present, use the bundled module
+module_kernel_path = os.path.join('.', LISA_MODULE_DIR_KERNEL_PATH, LISA_MODULE_DIR)
+
+if 'LISA_HOME' in os.environ:
+    LISA_HOME = os.environ['LISA_HOME']
+    lisa_home_module_src = os.path.join(LISA_HOME, 'lisa/_assets/kmodules/lisa')
+    link_cmd = [
+        'ln', '-sr', lisa_home_module_src, module_kernel_path
+    ]
+    subprocess.check_call(link_cmd)
```
douglas-raillard-arm, line 57: Can use [os.symlink](https://docs.python.org/3/library/os.html#os.symlink) or [Path.symlink_to](https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to) instead of relying on a shell command

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
...
+parser = argparse.ArgumentParser(prog='Lisa module in-tree helper')
+init_group = parser.add_mutually_exclusive_group()
+init_group.add_argument('hash', nargs='?', help='Lisa repository hash to checkout')
+parser.add_argument('-u', '--update', action='store_true', help='Run git fetch before checking out the hash')
+args = parser.parse_args()
+
+# LISA_HOME is present, use the bundled module
+module_kernel_path = os.path.join('.', LISA_MODULE_DIR_KERNEL_PATH, LISA_MODULE_DIR)
+
+if 'LISA_HOME' in os.environ:
```
douglas-raillard-arm, line 51: this branch should raise an exception if --commit is used, as the commit is whatever is already checked out (and we don't want to go and possibly mess things up by checkout something in LISA_HOME, if the user wants a different commit they checkout first themselves)

tools/kmodules/lisa-prerequisites/android/aosp/0001-fs-Move-filp_open-kernel_read-write-outside-the-name.patch
```diff
@@ -0,0 +1,53 @@
...
+
+diff --git a/fs/open.c b/fs/open.c
+index b9d55ca3763b..d29008121b8a 100644
+--- a/fs/open.c
++++ b/fs/open.c
+@@ -1158,7 +1158,7 @@ struct file *filp_open(const char *filename, int flags, umode_t mode)
+ 	}
+ 	return file;
+ }
+-EXPORT_SYMBOL_NS(filp_open, ANDROID_GKI_VFS_EXPORT_ONLY);
```
douglas-raillard-arm, line 24: I think we can avoid touching exports just to remove ANDROID_GKI_VFS_EXPORT_ONLY namespace, we already import the namespace (VFS_internal_I_am_really_a_filesystem_and_am_NOT_a_driver) in the module source. Reducing this patch will reduce the chances of conflict when applying it.

EDIT: if that works that actually allows removing the entire patch, so even better.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,84 @@
+#! /usr/bin/env python3
+
+#
+# Usage:
```
douglas-raillard-arm, line 4: This doc comment should be cleaned up to remove CLI option doc (that's autogenerated by argparse --help) and the rest should go in  `ArgumentParser(description=...)` so it is shown in `--help`

douglas-raillard-arm (changes requested):

mrkajetanp:
All right, updated

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,77 @@
...
+        LISA_HOME = os.environ['LISA_HOME']
+
+        if args.commit:
+            parser.error("Can't use --commit when LISA_HOME is set")
+
+        # create relative symlink
+        pwd = os.getcwd()
+        os.chdir(os.path.dirname(args.module_kernel_path))
+        os.symlink(os.path.join(LISA_HOME, 'lisa/_assets/kmodules/lisa'), os.path.basename(args.module_kernel_path))
+        os.chdir(pwd)
```
douglas-raillard-arm, line 46: Please no unnecessary stateful code. Use `pathlib.Path.resolve()` to turn all the paths into absolute paths at the boundaries of the script. I'd generally recommend using `pathlib` now that we mandate high enough versions of Python, it's generally easier to use and some functions actually work better than their `os` counterpart.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,77 @@
...
+        # module has already been cloned
+        if os.path.exists(module_git_path):
+            if not args.force:
+                parser.error('Module checkout already exists, pass --force to override')
+            shutil.rmtree(module_git_path)
+            os.remove(args.module_kernel_path)
+
+        clone_cmd = [
+            'git', 'clone', '--verbose', '--no-checkout', '--no-tags', '--filter=tree:0',
+            args.remote, module_git_path
```
douglas-raillard-arm, line 63: Can't we simply clone the repo in a subfolder (e.g. `repo/`) of --module-kernel-path and simplify the CLI ?

douglas-raillard-arm (changes requested):
A few comments on the Python script but no fundamental issue. 
Before we merge that, we will also need to add a version check for the module:

* The module version will be defined as the sha1 sum of all the sources of the module. That should be easy to add to LISAFtraceDynamicKmod using `self.src.src` dict (file name to content map).
* The same sha1 will be computed by the Makefile and baked in a const string when the module is compiled (e.g. by generating a header from the Makefile. That header won't be part of the sha1-ed sources obviously).
* the module will gain a `version` CLI` parameter that will check the value passed against the const string it knows about. If they differ, module load will fail.

So this way we can have a `modprobe lisa version=XYZ` command and that will only succeed if the module loaded would have been the same as if we built it from source.


We still need to figure out how we deal with switching from one method and the other:

* ideally, we would always try to modprobe first. If that fails (missing lisa.ko, missing modprobe, or wrong lisa.ko version), we try to build from source.
* Problem with that is that failing the module load is done by returning non-zero from module init function, and that makes the module impossible to unload (the kernel assumes the state of the module is unknown and prevents from calling module exit in case that crashes the kernel). So if the version check fails, we can't unload it and try to build from source.

That issue is solvable with a "loading protocol" based on a filesystem where the now CLI parameters are files, and there is a "startup sequence of writes" to start the module and a status file to check for errors. Beata's work on lisafs should bring that, but until then we might have to cobble something up with a status debugfs file and make the module load not fail per-say, but set the error in debugfs and not call `init_features()`. Problem: debugfs is disabled in some android config. But I guess we can just ask for having it enabled ...

Another option is to pull the module file from the target and try to read the sha1 variable from the .ko using:
```
elf = elftools.elf.elffile.ELFFile(...)
section = elf.get_section_by_name('.lisa_version')
if section is not None:
    data = section.data()
```
We already use it in `lisa._kmod` to list the symbols defined in the module to know the list of trace events defined in it. We can probably use it to extract the content of a specific `.lisa_version` section and store the sha1 const string in it.

mrkajetanp:
> Before we merge that, we will also need to add a version check for the module:
> * The module version will be defined as the sha1 sum of all the sources of the module. That should be easy to add to LISAFtraceDynamicKmod using self.src.src dict (file name to content map).
> * The same sha1 will be computed by the Makefile and baked in a const string when the module is compiled (e.g. by generating a header from the Makefile. That header won't be part of the sha1-ed sources obviously).
> * the module will gain a version CLI` parameter that will check the value passed against the const string it knows about. If they differ, module load will fail.

Okay I'm a bit confused here, what exactly are we trying to achieve? We're purposefully adding an obstacle to loading the module unless someone manually confirms the hash the module was built with? What do we gain from this?

The obstacles I see are:
* with the CFI issue now fixed, the Android kernels do not need to be reconfigured for the Lisa module to work - because of the debugfs thing you mentioned this would likely reintroduce this obstacle just making life harder for ourselves
* the main benefit of the in-tree method is that the module comes bundled with the kernel and the Android startup system will load it automatically so that it's always there. It would be loaded without the version parameter so requiring that version parameter would break the loading and effectively result in the module crashing every single time on boot thus taking away a significant chunk of the reason why this PR exists in the first place

Besides, I don't think this is specific to this PR anyway? If the version check would go into the DynamicKmod as well and be performed regardless of the build method then those are two entirely separate module changes. In the already existing build methods it's equally possible to just build the module, keep a lisa.ko around and load it manually - that's what I used to do. If we want to guard against that for some reason the place for that is probably an entirely separate PR and discussion.

douglas-raillard-arm:
> Okay I'm a bit confused here, what exactly are we trying to achieve? 

The interface between the module and the rest of LISA is fully unstable and can change at any time. We are not about to start supporting loading module from LISA version X for Python code along version Y for the module. Therefore, the only thing that will be supported for the foreseeable future is when the builtin module matches exactly what you would have gotten with dynamic module compilation, as I stated right from the beginning of implementing this idea.

> We're purposefully adding an obstacle to loading the module unless someone manually confirms the hash the module was built with?

When the Python API will load the module, it will do the check, there is nothing manual in there. We can make the CLI parameter optional so that a user who only wants the module can load it without troubles, but the Python code needs a way of asserting that the code running in-kernel is matching what it expects. It's easy to do for a string param as a missing CLI value will just make the string a null pointer (or something equivalent).

> What do we gain from this?

Since the interface of the module is fully unstable, the only bug report I ever want to see on that front is "It complained about a version check" to which I'll answer "run the python script with the right commit or use LISA_HOME to share the module code". 
For example when the filesystem interface from Beata becomes a reality, the way of enabling features will become completely different from the current insmod CLI parameter, and the Python code will be changed to reflect that. 

On top of the expected user's accidental misconfiguration, people start sharing prebuilt kernel images+modules (which they will, we know that for a fact since I helped yesterday someone who was doing exactly that). This means you get a kernel compiled with 

* a random compiler (user unaware of which one exactly), that was creating undefined asan symbols issues from mismatching clang versions
* a random version of LISA, and then the user will try to use it with their own random version of LISA.

The version check I proposed is still flexible enough to allow multiple versions of LISA to use the same module, since it's the sha1 of the module source, not the entire git repo. It could technically be made more accurate with some sort of API/ABI check but that's impossible to implement manually, and I don't want to rely on remembering to manually increment a version number.

> because of the debugfs thing you mentioned this would likely reintroduce this obstacle just making life harder for ourselves

Yes, that's why I haven't merged anything based on debugfs. Hopefully if we create our own fs we won't have this problem. We will just need to allow some more symbols, but we already need that anyway.

> the main benefit of the in-tree method is that the module comes bundled with the kernel and the Android startup system will load it automatically so that it's always there. 

I did not know it was a requirement, I explicitly stated in earlier conversations that in the present form, we require modprobe/rmmod lifecycle. That is because enabling the features is done at module load time. This can be changed, such as what this PR does https://github.com/ARM-software/lisa/pull/1950 but it relies on debugfs. We could move to our own custom fs (Beata's work) but that will take time. The normal lifecycle of a module on a linux system is to get loaded only if something does it, which can either be a manual modprobe (or a modprobe config file), or something like udev based on peripheral identification (e.g. for USB).

Otherwise you will end up with things like the emeter workqueue being permanently activated and doing background work every few ms, which is I doubt is something we want to have when we don't care about that measurement. We may get more of those invasive stuff in the future, or even mutually exclusive features. It's like the ftrace event name clash with android events, it was just a matter of time, I knew the day would come. So I'm glad if it works for your use case right now, but it's not the solution of the future unless you are ready to reboot your system every time you want to change the enabled features.

> If the version check would go into the DynamicKmod as well and be performed regardless of the build method then those are two entirely separate module changes

There is some confusion: all the Python code do is compute the expected sha1 to compare it with what is currently loaded. In other words, the in-build tree is just an ahead-of-time version of what happens on the normal flow, which then gets skipped and replaced by a direct `modprobe lisa`. Other than that, there is basically no change to the flow.

>  If we want to guard against that for some reason the place for that is probably an entirely separate PR and discussion.

The purpose of this PR is to allow in-tree build of the module but otherwise should not introduce holes and opportunities for things to explode (beyond what it does to the android tree). It needs to uphold all the guarantees of consistency we already have and rely on, which includes strict alignment between what the Python code expects and what is actually running on the target.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,77 @@
...
+    Method 2 - Sparse checkout.
+
+    If the script doesn't find LISA_HOME (e.g. building on an external server without a Lisa checkout)
+    the alternative mode is to directly clone the module sources into the tree.
+                                     """)
+    parser.add_argument('-c', '--commit', help='Lisa git reference to checkout')
+    parser.add_argument('-f', '--force', action='store_true', help='Override the module checkout if it already exists')
+    parser.add_argument('--module-kernel-path', help='Path to where the module will be stored in-kernel',
+                        default='./private/gs-google/drivers/soc/arm/vh/kernel/lisa')
+    parser.add_argument('--module-kernel-git-dir', help='Directory name to store the module git source in-kernel',
```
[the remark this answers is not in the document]
mrkajetanp: This one isn't a path though, just the directory name for the sparse checkout. Supporting placing it in arbitrary paths makes creating the symlinks way more annoying and serves no purpose that I can think of. The sparse checkout will just be put in
```
module_git_path = os.path.join(os.path.dirname(args.module_kernel_path), args.module_kernel_git_dir)
```

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
mrkajetanp: Currently `--module-kernel-path` is `./private/gs-google/drivers/soc/arm/vh/kernel/lisa` which means that that's where the sources are, the directory structure comes out like this:
```
❯ pwd
/home/kajpuc01/aosp/kernel/device/private/gs-google/drivers/soc/arm/vh/kernel
❯ l
drwxrwxr-x  - kajpuc01 12 Jul 13:52 .
drwxrwxr-x  - kajpuc01 12 Jul 14:09 ..
.rw-rw-r-- 22 kajpuc01 12 Jul 13:49 .gitignore
lrwxrwxrwx 42 kajpuc01  6 Jul 20:54 lisa -> lisa-module-git/lisa/_assets/kmodules/lisa
drwxrwxr-x  - kajpuc01  6 Jul 20:54 lisa-module-git
```
The cli just allows renaming `lisa-module-git` to something else but not changing the path to not overcomplicate. We can't put `lisa-module-git` inside `lisa` because `lisa` is a symlink from a subfolder of `lisa-module-git`, that's why I just put them next to each other.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
mrkajetanp: That's why I was directly calling into `ln` in the first place, to use the `-r` argument to create a relative symlink. If I turn all the paths absolute then the symlink relies on the entire directory structure so if the kernel directory is moved or renamed the symlink will break whereas in the relative variant it would still work fine. We can do it with absolute paths as well, it'll just be less robust that's all.

mrkajetanp:
I pushed the python script changes including the move to using Pathlib, thanks for the suggestion, it's definitely more readable now. The only downside is the potentially breakable non-relative symlink but we can probably live with that.

mrkajetanp:
> The interface between the module and the rest of LISA is fully unstable and can change at any time.

We're probably thinking about this differently but from a user-perspective how does the module interface with the rest of LISA? It pretty much does not, the module provides events to generate traces then the traces are the thing interfacing with the rest of Lisa. The potential mismatch you're describing is between the trace processing code and the trace, is it not? These sorts of errors already happen whenever someone tries to process an older tracer with a newer version of Lisa, I don't see how we can avoid them. If we did, the way to do it would be on the trace processing side not on the module inserting side imo. Even if we implement this, if I collect a trace today with the newest module, upload it somewhere and then someone tries to analyse it after a few months we'll bump into the same potential mismatch problem.

> When the Python API will load the module, it will do the check, there is nothing manual in there

For the Dynamic module yes, but if you build the module in the kernel the Python API never even sees it. I don't even have Lisa cloned past the sparse checkout with module sources on the machine where I build the kernel and use this PR.

> the Python code needs a way of asserting that the code running in-kernel is matching what it expects.

Again, which Python code are we talking about interacting with the module directly? Lisa Python code only sees the trace, not the device while it's running. The module itself is only relevant to WA as far the available trace events go, by the time the Lisa trace analysis code runs my test device is already disconnected from the laptop and cooling down for the next set of runs.

> the only bug report I ever want to see on that front is "It complained about a version check"

As I said, if we want this type of error checking it should be on the trace processing side. Even if the module was super outdated like the stuff Google bundles in their sched vendor hook it'd still work perfectly fine for collecting traces and analysing them manually, the issue is that it wouldn't play well with Lisa analysis functions. The conclusion here is that such complaints should be given out on the analysis side, not the module loading side.

> (which they will, we know that for a fact since I helped yesterday someone who was doing exactly that)

If anything this just further reinforces my point about this not being related to this PR but a separate issue.

> I did not know it was a requirement, I explicitly stated in earlier conversations that in the present form, we require modprobe/rmmod lifecycle. That is because enabling the features is done at module load time. 

Well yes as I said, this is pretty much the main appeal of this. For example my test platform Pixel 6 gets rebooted 10-20 times a day, it quickly gets incredibly annoying to have to constantly think whether the Lisa module is loaded or not, which features it has enabled and such. That's what this PR is for, if someone only needs to load the module once every few days with a specific set of events the dynamic kmod approach is more than adequate but if you need to do the same thing over and over it's just a massive waste of time. I ended up just manually building the module, keeping it in /data/local and running every test from a script doing `insmod` before `wa run` so the same "potential version mismatch" issue you were talking about above.
Now with this PR it's incomparably better to know that no matter what the module will be there and I won't get a "missing event" error if I forget to load it properly on the 8th reboot of the day causing me to have to cool down the phone all over again and waste another hour.

> Otherwise you will end up with things like the emeter workqueue being permanently activated and doing background work every few ms, which is I doubt is something we want to have when we don't care about that measurement.

Depends on the use case, I've not had a single test where I didn't want the power meter to be active so for me it's much more annoying if the default involves the power meter being off. When we get the lisa-fs stuff working to set features it'll be incomparably better to interact with but still I use Lisa & the module pretty much every day and I've not even once wanted to disable any of the features.

> which then gets skipped and replaced by a direct modprobe lisa. Other than that, there is basically no change to the flow

If the module comes built in the tree then the Android framework will load it automatically, it's already active by default after each reboot. Somehow trying to make the automatic loading fail every time to force users to go through Python code just to load it and fail on version mismatches would be very unpleasant to work with, if I'm being honest I'd just do a git revert on that commit for my local branch and continue as if it was never there.

TLDR:
I think it's worthwhile to guard against the version mismatch issue but in the same way we guard against the "ftrace buffer has overflown issue". If the analysis code that works with events provided by the module fails for some reason, we should include an error about a "potential version mismatch" between the module version the trace was collected with and the Lisa analysis version. That way we'll actually cover the entire surface of the problem including the "old trace" variant rather than just make it more difficult for people to work with the module.

douglas-raillard-arm:
> We're probably thinking about this differently but from a user-perspective how does the module interface with the rest of LISA?

Setting CLI parameters, the format and the name of the events, the name of the features, the mechanism to enable/disable a feature, in the future the mechanism to configure a feature (Beata's perf counter PR requires having config options for the corresponding feature). That plus anything else that might come one day (like in-kernel analysis etc).

When it comes to trace data, we currently can't identify what version of the module was used to generate the trace, and we don't really need to because if we make a breaking change to an event, we will simply make a new one. It's a bit of a grey area but fortunately very little changes like that have been needed so far. Some are planned (e.g. the cpumasks fields that store it as string, which was a workaround for trace-cmd report limitation) though, but we will deal with that in time.

> For the Dynamic module yes, but if you build the module in the kernel the Python API never even sees it. I don't even have Lisa cloned past the sparse checkout with module sources on the machine where I build the kernel and use this PR.

It obviously sees it the second lisa's FtraceCollector will `modprobe lisa`, as I mentionned:

> We can make the CLI parameter optional so that a user who only wants the module can load it without troubles

So if all you care about is the module there is no need to check anything because you are only using one half. If you use the other half as well (the Python API), the check will be required. That does include using `lisa-load-kmod`, as that script needs to know how to enable features, so it definitely needs to assert that the interface of the module is what it expects it to be.

> Again, which Python code are we talking about interacting with the module directly? Lisa Python code only sees the trace, 

No. The Python API obviously knows how to load the module, since it is done:
    * By lisa-load-kmod
    * Automatically by FtraceCollector if you ask for an event defined inside the module
https://github.com/ARM-software/lisa/blob/5b7119fa44fef74898487ff8a4e917f1aef132d9/lisa/_kmod.py#L2010
https://github.com/ARM-software/lisa/blob/5b7119fa44fef74898487ff8a4e917f1aef132d9/lisa/trace.py#L6097

Both these locations of the code expect a specific lifecycle for the module and features. And again: _this WILL change_. We will likely end up with:
* a module that can hot-reload a feature to reconfigure it
* a configuration system based on filesystem rather than module param

These changes are planned and _WILL NOT_ be considered breaking changes. They are 100% implementation details, and as such will stay hidden deep in LISA code. If you take today's `lisa` python package, it realistically will not be able to load what will be the module in 6 months time. When this happens, I want the (old) Python code to complain that the module is not a version it knows about and raise an exception, so that users either downgrade the module or upgrade the Python part.

> keeping it in /data/local and running every test from a script doing insmod before wa run
You could use `lisa-load-kmod` and pretty much get the same benefit. As long as the lisa sources stay the same between invocations and that the kernel stays the same (same /sys/kernel/kheaders.tar.xz, same /proc/config.gz) you will hit the cache. Since it needs to download the said files to make the checksum it takes longer than a simple insmod. On my machine with a qemu target, it takes 17s. It seems that an `rmmod` takes 6 seconds which is a bit fishy, I'll investigate that. In any case, if the performance of `lisa-load-kmod` is a concern, it can certainly be addressed without introducing something like this PR, and would likely benefit to all parts of LISA (and WA if we make devlib faster).

If you are concerned about forgetting to run the command when starting WA, then the best way is probably to make an augmentation that allows running arbitrary commands upon setup, similar to this: https://github.com/ARM-software/lisa/pull/2025. You will also need something like that to disable SELinux. I'm sure other people would be happy to have a way to run stuff from the agenda.

At the very least you can trivially check that events are present before running your workload by grepping `/sys/kernel/tracing/available_events`. Devlib FtraceCollector can already do a strict check for you (https://github.com/ARM-software/devlib/blob/be988bb42b766574bb9bd9191fa3532a65091716/devlib/collector/ftrace.py#L69), so all you need is probably a single `strict: true` line in your WA agenda (and if it's not exposed, I'm sure it's easy to contribute that).

All in all this is hardly a justification for an invasive change such as this PR (mostly invasive because of the requirement to work with different headers, which adds a 2nd orthogonal setup to test for every single kmod change, or else it will fall behind and stop compiling).

> But if you need to do the same thing over and over it's just a massive waste of time

Given the caching, I fail to see where the massive waste of time is. The insmod itself is instantaneous, you are not saving anything by doing it only once at boot time. 

>  I use Lisa & the module pretty much every day and I've not even once wanted to disable any of the features.

Ok, but you do see my point though don't you ? The second I add a feature  _you_ don't want that argument goes away instantly. And there might already be issues with the emeter workqueue when using some synthetic tests, I don't know since I never tried to benchmark the amount of time this workqueue runs. Maybe it's negligible, maybe the read to sysfs synchronously triggers an I2C requests and takes several hundreds of usec.

That said, the default behavior of activating everything possible by default won't go away since the list of feature is private to LISA, so it's the only reasonable default. If one day features become public and part of the stable API, then maybe the default will be to enable nothing and let people opt-in what they need, we will see. That would especially make sense if we start getting things that are not-at-all scheduler related, so different group of users start wanting different presets. We are currently having [this exact issue in devlib](https://github.com/ARM-software/devlib/issues/611), so I'd like to avoid that

> Somehow trying to make the automatic loading fail every time to force users to go through Python code just to load it and fail on version mismatches would be very unpleasant to work with, if I'm being honest I'd just do a git revert on that commit for my local branch and continue as if it was never there.

That has already been addressed every single time we discussed this topic: 
* In the current thread on github:
> . We can make the CLI parameter optional so that a user who only wants the module can load it without troubles,
* on Slack:
> Python APIs will apply a strict version check on that, so you will be forced to rebuild the module/kernel when you change LISA version. If you just want the module for events in other systems, pick any version you like.

If you are just going to use the module on its own, you can pick whatever version you want. If you use it along any of the Python API, there will be a strict version check, period. And yes, the official way of loading the module when used with 3rd party software (your use case) will stay being `lisa-load-kmod`, with our without this PR. And `lisa-load-kmod` falls into the Python API category, so you will need a version that matches the module. This implies that every time `lisa-load-kmod` fits the bill, that is what will be recommended, both internally and externally. If I see someone recommending manual build (of any kind) without an excellent reason, I'll change the material to use `lisa-load-kmod`. In the lifetime of this module, about 3/4 of the support requests I got (before we tried to make it work on recent android kernels) were about issues that already had been 100% solved under the Python build path, such as people getting error -22 on BTF when loading the module.

> When we get the lisa-fs stuff working to set features it'll be incomparably better to interact with

Don't get your hopes too high, it will still be something like:
```
echo foo > /sys/lisa/file1
echo bar > /sys/lisa/file2
echo reload > /sys/lisa/control
```
and then you will wish to have back the CLI way `rmmod lisa; insmod lisa features=foo,bar`.
Since there is currently no externally useful state inside the kmod (like configuration), the difference is really thin. Even in your case, you still end up with a separate .ko anyway so you can insmod and rmmod it as usual.

> we should include an error about a "potential version mismatch" between the module version the trace was collected with and the Lisa analysis version.

Since this was written under the assumption the only interface is the event format, but to answer that specific point the way we have done it so far is by either:
* introducing new events if the format changes too much
* adapting analysis code to deal with various data schema. When it becomes hairy, we create an analysis function that abstracts over the various sources (like for PELT task signals, we have a bazillion legacy events supported from prehistoric versions of android).

douglas-raillard-arm:
>  It seems that an rmmod takes 6 seconds which is a bit fishy,

After investigation, what takes ~5/6s is basically `import lisa.target` that is done again in a subprocess (for user namespaces, so that we can become root and use overlayfs as a regular user). The import time is because it indirectly imports pandas, numpy and bokeh. I tried quite hard fixing that but ultimately the fix will be very brittle, and also would impact devlib. I'll try to see if I can hook into the import machinery to make use of (LazyLoader)[https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader] and hope it does not break anything we care about.

mrkajetanp:
> It obviously sees it the second lisa's FtraceCollector will modprobe lisa

Ah right we're just taking about different workflows here, I think that's where the misunderstanding is coming from. My usual workflow is `load the module -> run and collect traces with wa -> process traces with Lisa` hence none of the Lisa-FtraceCollector stuff is ever used and hence the module only interacts with Lisa at the data processing stage. For the `load the module` stage Android just does the `modprobe` for me so same story but yes you're right that it's different in the other workflow.

> So if all you care about is the module there is no need to check anything because you are only using one half

Yes I now see what you mean, as long as `modprobe lisa` & `insmod lisa.ko` will still work with no arguments provided and enable everything by default that's perfectly fine by me, my concern here was just breaking this workflow that's all.

> mostly invasive because of the requirement to work with different headers, which adds a 2nd orthogonal setup to test for every single kmod change

Arguably that's probably a benefit to be honest, the headers included with the module need to mimic what's inside the kernel anyway. In the module we're re-defining internal kernel functions reliant on non-stable interfaces so if those change we'll have to update our headers regardless of whether we only support one build method or both to match the change. We already have this type of thing here:
```
#if LINUX_VERSION_CODE <= KERNEL_VERSION(5,6,0)
#define RBL_LOAD_ENTRY      rbl_load
#define RBL_LOAD_MEMBER     runnable_load_avg
#define RBL_LOAD_STR        "rbl_load"
#else
#define RBL_LOAD_ENTRY      runnable
#define RBL_LOAD_MEMBER     runnable_avg
#define RBL_LOAD_STR        "runnable"
#endif
```

> echo foo > /sys/lisa/file1

That's how pretty much every other module out there is configured so fine by me, at least there'll be no need to worry whether the module is loaded or not because it'll just be there :)

But yeah fair enough, I think I just misunderstood your idea based on thinking about the other workflow and that's where all the confusion came from. As long as the behaviour is as described below that's perfectly functional imo.

```
modprobe lisa version=<CORRECT> # works & loads the module
modprobe lisa version=<WRONG> # fails and raises an error
modprobe lisa # works & loads the module
```

mrkajetanp:
Are we looking for something like this?
```
$(MODULE_VERSION_H):
    echo -n '#ifndef _MODULE_VERSION_H\n#define _MODULE_VERSION_H\n#define LISA_MODULE_VERSION ' > $@
    cat $(KERNEL_SRC)/$(MODULE_SRC)/*.* | sha1sum | awk '{print $$1}' >> $@
    echo '#endif' >> $@
```

This produces a `module_version.h` file in the output build directory with the following contents:

```
#ifndef _MODULE_VERSION_H
#define _MODULE_VERSION_H
#define LISA_MODULE_VERSION d69254e51206d759467e877c616300f8d7f3b7bb
#endif
```
Then we can match the `version` CLI parameter against this constant and it should get us what we want?

douglas-raillard-arm:
> Are we looking for something like this?
```
$(MODULE_VERSION_H):
    echo -n '#ifndef _MODULE_VERSION_H\n#define _MODULE_VERSION_H\n#define LISA_MODULE_VERSION ' > $@
    cat $(KERNEL_SRC)/$(MODULE_SRC)/*.* | sha1sum | awk '{print $$1}' >> $@
    echo '#endif' >> $@
```

Yes, although the file list should be sorted before being fed to sha1sum. POSIX mandates sorting, but it's locale-dependent so probably a can of worm (unless you set LC_ALL=C). Then we can easily replicate that computation on Python side.

douglas-raillard-arm:
I've updated this comment with the result of my investigations to make `lisa-load-kmod` faster:
https://github.com/ARM-software/lisa/pull/2038#issuecomment-1643926345
Bottom line: lazily loading Python module does not work, and it's not fixable without ecosystem support. If 100% of the code is under your control, you can adopt a style and import hooks that could work, but it's a bit like MISRA C for Python...

mrkajetanp:
> At that point, activating features will require writing to some files etc

As long as the actual loading still works that's fine imo, as you said above it's always possible to just add a hook to some tooling to write to those files etc.

> keep our helpers working AND keep up with the changes in kernel helpers APIs

Realistically I think that's probably a two way street though - for this to actually be a problem we'd need to bump into a situation where a kernel change is at the same time invasive enough to make the module not compile at all with kernel helpers but also not invasive enough to cause problems when using our now-outdated helpers. Most of the time I think either both workflows will break or neither will break so if anything the in-tree build works as an early warning system because we get the warning at build time rather than only when the codepath is hit in the module.

Regardless, not a particularly common occurrence seeing as the only kernel version check we have in there is for 5.6, a version released over 3 years ago, and the mainline is at 6.4 these days. 

I pushed a new commit adding the version check & tested it, works very well. Passing a version string that doesn't match to insmod shows this error:
```
oriole:/data/local # insmod lisa.ko version=boo
insmod: failed to load lisa.ko: Invalid argument
```

douglas-raillard-arm:
> Most of the time I think either both workflows will break or neither will break

Maybe, I can't really judge since so far I don't think we ever had such a breakage (which is a good sign). The worse would be a silent change of meaning of an existing struct member. As long as new things get new name, we should become aware of changes pretty quickly.

mrkajetanp:
I pushed another commit with a small fix to make the in-tree build method imports match what the newer mainline kernels look like rather than older product kernels. The added documentation note mentions two commits that need to be in the tree for this to work on older kernels. The issue is that older kernels don't have import guards around some internal headers and so the build system complains about double imports.

lisa/_kmod.py
```diff
@@ -2267,4 +2285,23 @@ def from_target(cls, target, **kwargs):
...
+            pass
+
+        probe_cmd = 'modprobe {} {} version={}'.format(
+            f"-d /vendor_dlkm/lib/modules/{self.target.kernel_version.release}" if self.target.os == 'android' else "",
+            quote(self.src.mod_name), self.src.checksum_sources
+        )
+
+        try:
+            self.target.execute(probe_cmd)
+            self.logger.info(f'Loaded module {self.src.mod_name} from kernel modules directory')
```
douglas-raillard-arm, line 2301: nit: We are not trying to catch issues in `self.logger.info()`, it should therefore not be in the `try` clause. Instead it should be in the `else` clause.

douglas-raillard-arm (changes requested):

lisa/_kmod.py
```diff
@@ -2267,4 +2285,23 @@ def from_target(cls, target, **kwargs):
     def _event_features(cls, events):
         return set(f'event__{event}' for event in events)
 
+    def install(self, kmod_params=None):
+        try:
+            self.uninstall()
+        except Exception:
+            pass
+
+        probe_cmd = 'modprobe {} {} version={}'.format(
```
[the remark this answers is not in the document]
mrkajetanp: Seeing as the busybox variant doesn't have the option we use, how likely is it that a target device someone tries to run the Lisa module one doesn't have access to modprobe?
If we rely on insmod here we also need to hardcode the path where the module is in the kernel which makes this more error prone. If someone renames the directory the modprobe variant will still work but the insmod variant will break so it's a tradeoff.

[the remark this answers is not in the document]
douglas-raillard-arm: > Seeing as the busybox variant doesn't have the option we use, how likely is it that a target device someone tries to run the Lisa module on doesn't have access to modprobe?

Any busybox-based system will not have `modprobe -d` available (so any Linux board in our CI), and they may not have modprobe at all. The main problem is not so much the availability of modprobe more than the exact flavor of modprobe. With that PR, we already hit the issue that the system modprobe had `-d` in your setup, but won't in CI mainline linux setups.

We have been bitten by that in devlib multiple times (e.g. the `ps -C` issue). We also had cases of slightly different behaviors (can't remember which command). So since we do have a set version of modprobe + insmod available, we should try to use them.

> If we rely on insmod here we also need to hardcode the path where the module is in the kernel which makes this more error prone.

Isn't that what you did already ? My suggestion is try insmod with your hardcoded android path, then fallback on modprobe if insmod failed. This way we don't have to teach modprobe to look anywhere else. We can speedup the process by running a snippet of shell script:

```
path = quote(path)
busybox = quote(target.busybox)
target.execute(f'([ -e {path} ] && {busybox} insmod {path}/lisa.ko {options}) || {busybox} modprobe lisa {options}', as_root=True) 
```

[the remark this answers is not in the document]
mrkajetanp: > Any busybox-based system will not have modprobe -d available (so any Linux board in our CI),

Ah yeah that's fair, I was just thinking about phone and normal linux targets, I guess the CI boards do swing the tradeoff the other way.

> Isn't that what you did already ? 

For half the path, this is the full one:
```
/vendor_dlkm/lib/modules/5.10.107-android13-4-00025-g72ae232e302a-dirty/kernel/drivers/soc/arm/vh/kernel/lisa/lisa.ko
```
I meant that modprobe can detect the part from `kernel` onwards, insmod cannot. But I can replicate that with `busybox find` just as well so sure, insmod it is.

mrkajetanp:
There you go, updated.

mrkajetanp:
Another fix because Android works in mysterious ways..
Essentially what happens is that if you rebuild your Android kernel without wiping the directory every time you'll end up with a mismatch between `uname -r` and the directory in `/vendor_dlkm/lib/modules/`. It happens a lot in my case which is why just inserting `uname -r` might fail. However, there's always only one directory there so I just made it use `ls` to find out what it is in the case of Android and use that instead of the actual kernel version, this fixes the problem. If the `ls` for some reason fails it will fall back on `self.target.kernel_version.release`, then if that fails it'll just start building the module out-of-tree. Seems like a reasonable solution.

lisa/_kmod.py
```diff
@@ -2267,4 +2286,40 @@ def from_target(cls, target, **kwargs):
...
+            lisa_module_path = self.target.execute(
+                f"{busybox} find {modules_path} -name '{self.src.mod_name}.ko'"
+            ).strip()
+
+            self.target.execute(
+                f"{busybox} insmod {lisa_module_path} version={self.src.checksum_sources}"
+            )
+        except TargetStableCalledProcessError as e:
+            if e.returncode == errno.EPROTO:
+                raise ValueError('In-tree module version does not match the Lisa checkout.')
```
douglas-raillard-arm, line 2320: nit: s/Lisa checkout/what lisa expects/

LISA does not necessarily come from a git checkout, it can be installed from PyPI

lisa/_kmod.py
```diff
@@ -2267,4 +2286,40 @@ def from_target(cls, target, **kwargs):
...
+                    f"{busybox} ls {modules_path_base} | {busybox} head -1"
+                ).strip()
+            except TargetStableCalledProcessError:
+                pass
+
+        modules_path = f"{modules_path_base}/{modules_version}"
+
+        try:
+            lisa_module_path = self.target.execute(
+                f"{busybox} find {modules_path} -name '{self.src.mod_name}.ko'"
```
douglas-raillard-arm, line 2312: quote() needs to be used, not actual `'` quotes.

lisa/_kmod.py
```diff
@@ -2267,4 +2286,40 @@ def from_target(cls, target, **kwargs):
...
+
+        busybox = quote(self.target.busybox)
+        modules_path_base = '/lib/modules'
+        modules_version = self.target.kernel_version.release
+
+        if self.target.os == 'android':
+            modules_path_base = f'/vendor_dlkm{modules_path_base}'
+            try:
+                modules_version = self.target.execute(
+                    f"{busybox} ls {modules_path_base} | {busybox} head -1"
```
douglas-raillard-arm, line 2303: `ls` is notoriously variable in its output, which makes it unsuitable for machine processing, even more so without options to restrain its behavior. `printf` and shell globbing is the most reliable way I know to list a folder with.

In the current case, is there any need for that though ? We use the `find` command anyway, so there shouldn't be any problem pointing it at one level above the one that is buggy.

douglas-raillard-arm (changes requested):

lisa/_kmod.py
mrkajetanp: It's because if we point the already existing call to `find` one level higher it'll detect the multiple modules directories present on Linux targets. I just replaced the `ls` call with a different call to `find` which will have the same result. The options used are present in `busybox find`.

mrkajetanp:
Updated again

lisa/_kmod.py
douglas-raillard-arm: > it'll detect the multiple modules directories present on Linux targets

I see, that's pretty hacky since we basically allow a broken setup as long as there is one folder only (but only on android)

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
douglas-raillard-arm: You can still achieve the same result without being stateful. You can use Path.relative_to() to make a symlink target relative to whatever base you want:
```
from pathlib import Path

path1 = Path('./a').resolve()
path2 = Path('./a/b').resolve()
path3 = Path('./a/c').resolve()

# At this point, all the paths are absolute, and we will make a symlink `./a/c -> ./a/b`

path1.mkdir()
path2.mkdir()

target = path2.relative_to(path3.parent)
path3.symlink_to(target)
```

lisa/_kmod.py
mrkajetanp: How do you mean? It's just because Linux targets don't usually clean their `/lib/modules` on kernel upgrades, my thinkpad has 39 different ones there. But since Linux targets keep their modules and their base kernel in the same repo the modules for the running kernel will always match `uname -r`.

On Android the two are in separate repos which if you do incremental builds can result in a mismatch but the modules partition is always swapped out so there's always only one directory there, no leftovers.

Ie on Android we don't necessarily know the name but we know there's only one directory. On Linux we know the name but there might be multiple. This implementation caters for both, nothing particularly broken about that as far as I can see.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
douglas-raillard-arm: I think I was thinking of a structure like:
```
./private/gs-google/drivers/soc/arm/vh/kernel/lisa
     lisa/ # git tree
     src -> lisa/lisa/_assets/kmodules/lisa
```

But I guess the build system expects to find a Makfile in the top-level folder of the module ?
If so I guess there is no better way to do it, but I don't think it's useful to expose that as a CLI parameter. You can just build a name based on the provided path:
```
module_kernel_path = Path(module_kernel_path)
module_git_path = module_kernel_path.parent / f'.{module_kernel_path.parts[-1]}-git'
```

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,74 @@
...
+    Method 2 - Sparse checkout.
+
+    If the script doesn't find LISA_HOME (e.g. building on an external server without a Lisa checkout)
+    the alternative mode is to directly clone the module sources into the tree.
+                                     """)
+    parser.add_argument('-g', '--git-ref', help='Lisa git reference to checkout')
+    parser.add_argument('-f', '--force', action='store_true', help='Override the module checkout if it already exists')
+    parser.add_argument('--module-kernel-path',
+                        help='Path relative to the kernel tree root where the module  will be stored',
+                        default='./private/gs-google/drivers/soc/arm/vh/kernel/lisa')
```
douglas-raillard-arm, line 33: I don't think we should provide a default value for something like that. 6 months from now that path will change to whatever Google decides is the new flavor and we will have a broken default forever in that script, with people relying on that and others scratching their head why they don't get the module build because they used the now-wrong default value.

douglas-raillard-arm (commented):
Answered to some old-ish comment I didn't realized were not solved

lisa/_kmod.py
douglas-raillard-arm: > Ie on Android we don't necessarily know the name but we know there's only one directory.

That's my point. There are only 2 sane behaviors:
* like what most distro do, i.e. a folder per kernel (`/lib/modules/$(uname -r)`)
* Support only a single kernel, and therefore do not bother with the `uname -r` level

Having a single level named as the `uname -r` output of another kernel is not sane in any way. AFAICU from what you said that is the situation on android.

This situation is indistinguishable from a Linux setup that would simply have no module at all installed for the current kernel, and a leftover modules folder from another kernel. In that situation, it would be completely pointless to try to load anything found there, since we know it does not match the current kernel. We only try to do that because it's android, and we know that the build system is basically broken.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
mrkajetanp: yeah this works too, thanks!

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
mrkajetanp: Yeah the build system will complain/ignore it without the Makefile there. I removed the argument and replaced it with just appending `-git` to the basename of the given path, good suggestion.

mrkajetanp:
Updated with the suggested script changes, I also slightly refactored the code to not redo the same path operations multiple times.

lisa/_kmod.py
mrkajetanp: Yeah that's why we only do this 'trick' after checking if we're working with an Android target.
It's not necessarily broken, because the KMI is frozen the Android modules will work fine even against different base kernel versions for the most part, that's intentional. If someone is in the position to try to load the Lisa module that means that the phone booted which means that the setup works - it wouldn't boot without the modules being compatible. The main reason for the mismatch for us is that if you're rebuilding the kernels you probably want them tagged with the git hash which obviously changes every commit even if none of the changes affect the modules.
This is just what the environment we work with is, at least it's a problem that's pretty simple to solve.

The GKI itself comes prebuilt by default so pretty much every even stock Android setup will have a mismatch between their modules and the base kernel to begin with.

douglas-raillard-arm: > It's not necessarily broken, because the KMI is frozen

It is, if the modules are tied to a KMI more than a specific kernel name they could use a hash of the ABI. Any way you slice it, that is a dirty side of that build system. Yes we can work around, yes we can explain it, does not make it right. It's just like this: https://github.com/ARM-software/lisa/blob/150f9ed4e3adc3f62847126f880813491af6f412/lisa/_kmod.py#L968

I'm wondering how it works inside android itself, since that obviously breaks modprobe, but maybe they have a patched version that ignores the kernel name (or they don't use modprobe at all).

> at least it's a problem that's pretty simple to solve.

Yes somewhat. That opens the door to trying to load something knowingly broken (in case the module is actually not compatible) since we can't distinguish that anymore, but we have to be able to deal with load failure anyway.  It turns out that busybox rmmod does support -f but only if compiled with !ENABLE_MODPROBE_SMALL so I'll check that's what we have in devlib

mrkajetanp: > That opens the door to trying to load something knowingly broken

This is still a non-issue though, this scenario pretty much cannot happen. There's only one set of modules on Android and they are necessary for the system to work - this includes the scheduler vendor hooks and even the old Lisa module they have included as `sched_tp`. If someone is in a position to load the Lisa module this means that the system has booted so the modules have all been loaded and did not crash. Pondering the hypothetical incompatibilities is all well and good but this isn't actually a real problem.

The worst case scenario is that the module fails to load and the out-of-tree build starts regardless. Even then for someone to get there all of the following would need to be true:
* the modules are similar enough to the old version for the incremental build to not trigger a rebuild
* the modules are similar enough for every single system module to load on boot and not crash the boot
* the modules are different enough for specifically the Lisa module to fail to load in some way while the `vh_sched` and `sched_tp` modules using **the same internal sched headers** work completely fine

All in all, very unlikely if not close to impossible.

douglas-raillard-arm: Or be working with a pre-GKI or post-this-build-system-nonsense kernel :) Here we don't have a choice and it's the most sensible way forward, but that's definitely the sort of cruft that end up creating issues (if only by creating an unmaintained path the day this problem is fixed in android).

> while the vh_sched and sched_tp modules using the same internal sched headers work completely fine

Out of curiosity, are these modules systematically loaded or are they just available for someone to modprobe dynamically ? I can't imagine they ship production system with those enabled.

tools/kmodules/lisa-in-tree/fetch_lisa_module.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+        return
+
+    # --git-ref not passed, try linking from LISA_HOME
+    try:
+        LISA_HOME = os.environ['LISA_HOME']
+    except KeyError:
+        parser.error('Either LISA_HOME needs to be present or --git-ref needs to be used')
+    else:
+        module_src_lisa_path = Path(LISA_HOME).resolve() / 'lisa/_assets/kmodules/lisa'
```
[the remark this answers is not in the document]
mrkajetanp: Oh nice, even better! Updated :)

lisa/_kmod.py
douglas-raillard-arm: > That's very optimistic I'd say haha..

I consider "fixed" a problem where the original problematic software got replaced wholesale by something else :) 

> The sched_tp module is so tightly integrated into their scheduler hooks that Google's Android scheduler doesn't work without it.

I wish I didn't asked ...

PR #2073, opened 2023-08-14
[docs/setup] Fix order of commands when installing LISA from github

luis-machado-arm:
The current order runs a git fetch prior to moving into the repository directory. Fix this by changing the order of commands.


douglas-raillard-arm:
Indeed, thanks

PR #2128, opened 2023-10-11
Trace task uclamp instead of rq uclamp in uclamp_util_se

honxia02:
FIX

If we really want to trace the rq uclamp values, they are currently available in other events anyway.


douglas-raillard-arm:
As discussed offline, since that value is not a property of the tasks, they should be logged in another rq-level event

honxia02:
Okay. Closing.

PR #2145, opened 2023-11-28
lisa._assets.kmodules.lisa: Don't rely on rq->cpu_capacity_orig anymore

deggeman:
The kernel commit 7bc263840bc3 ("sched/topology: Consolidate and clean up access to a CPU's max compute") removed rq->cpu_capacity_orig so we can't use it anymore to retrieve the original CPU capacity nor for setting the FREQ_INVARIANCE kernel feature in LISA.

Replace rq->cpu_capacity_orig with arch_scale_cpu_capacity(cpu) in Lisa trace module's rq_cpu_orig_capacity() and remove the HAS_MEMBER(struct, rq, cpu_capacity_orig) check from the FREQ_INVARIANCE test.

Since arch_scale_cpu_capacity(cpu) exists in the kernel for a long time we don't have to take care of activating this only for newer Linux kernel versions.

### DIFF
--- a/lisa/_assets/kmodules/lisa/introspection.json
+++ b/lisa/_assets/kmodules/lisa/introspection.json
@@ -29,6 +29,7 @@
     "SCHED_AVG_RBL": "HAS_MEMBER(struct, sched_avg, runnable_load_avg) || HAS_MEMBER(struct, sched_avg, runnable_avg)",
 
     "FILE_IO": "HAS_SYMBOL(kernel_read) && HAS_SYMBOL(kernel_write) && HAS_SYMBOL(filp_open)",
-    "FREQ_INVARIANCE": "HAS_SYMBOL(arch_freq_scale) && HAS_MEMBER(struct, rq, cpu_capacity_orig)"
+    "FREQ_INVARIANCE": "HAS_SYMBOL(arch_freq_scale)",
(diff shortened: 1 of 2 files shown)


deggeman:
Related to PWRSOFT-5115

deggeman:
The kernel commit 7bc263840bc3 ("sched/topology: Consolidate and clean
up access to a CPU's max compute") removed rq->cpu_capacity_orig so we
can't use it anymore to retrieve the original CPU capacity nor for
setting the FREQ_INVARIANCE kernel feature in LISA.
    
Replace rq->cpu_capacity_orig with arch_scale_cpu_capacity(cpu) in Lisa
trace module's rq_cpu_orig_capacity() and remove the HAS_MEMBER(struct,
rq, cpu_capacity_orig) check from the FREQ_INVARIANCE test.

https://github.com/ARM-software/lisa/pull/2145

credp:
We probably need to figure out how to make this work for pre and post-patch kernels in order to complete this.

deggeman:
Fortunately not in this case:

I use arch_scale_cpu_capacity(cpu) now instead as accessing rq->cpu_capacity_orig directly.

Since arch_scale_cpu_capacity(cpu) exists in the kernel for a long time we don't have to take care of activating this only for newer Linux kernel versions.

deggeman:
I updated the fix by implementing choice 1.

PR #2146, opened 2023-12-01
lisa/utils: Fix wrong function signature in dispatch_kwargs()

msrasmussen:
Fixes bug in dispatch_kwargs where the object 'self' reference is passed in kwargs causing 'self' to be dispatched twice: Positional argument and kw argument.

The bug can be reproduced by creating an default RTA object, RTA(). Normally RTA objects are created through other functions, so the problem only occurs when trying to create an RTA object from a provided rtapp json file.

The bug is caused by inspect.signature() misbehaving for unbound methods. This problem is already fixed in kwargs_dispatched(), so it needs to be fixed in dispatch_kwargs() too.

### DIFF
--- a/lisa/utils.py
+++ b/lisa/utils.py
@@ -2273,10 +2273,18 @@ def dispatch_kwargs(funcs, kwargs, call=True, allow_overlap=False):
     """
     funcs = list(funcs)
 
+    def get_sig(f):
+        # If this is a method, we need to bind it to something to get rid
+        # of the "self" parameter.
+        if isinstance(f, UnboundMethodType):
+            f = f.__get__(0)
+
+        return inspect.signature(f)
+
     params = {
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
I don't think it's a good idea to do it at this level, as this can be fixed by simply being consistent between what we tell the decorator (`_early_init` function) and what we actually call (`_early_init` function rather than a bound method made out of it):
https://github.com/ARM-software/lisa/pull/2147

douglas-raillard-arm:
Closing as I merged this one: https://github.com/ARM-software/lisa/pull/2147
We can keep the decorator as it is since it currently has the most straightforward behavior and this behavior covers everything that we would want to do  (the callable give to the decorator should match the one that is actually called by the user)

PR #2148, opened 2023-12-11
lisa.wlgen.fio: Introduce Fio as Workload

cloehle:
Fio is a highly-configurable I/O tester used by the block layer for testing and as a general benchmark tool for storage. Even for scheduler investigations Fio is interesting as in_iowait-setting tasks are treated differently. To be able to create tests for them we introduce a simplistic lisa frontend.

Fio Workloads can now be created like:

ftest = Fio(target, filename="/dev/nvme0n1", name="test", rw="randread", runtime=50, bs='4k') and then ran like any other:

with ftest:
    result = fiotestmmc.run()
    print(result.read_iops)

### DIFF
--- a/lisa/wlgen/fio.py
+++ b/lisa/wlgen/fio.py
@@ -1,0 +1,140 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023, Arm Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


cloehle:
Just a couple of notes:
Fio has the capability to specify jobs using a file, I didn't implement this and just used command line.
If this is preferred I can do that, I was also thinking about the possibility of combining a RTA workload and a Fio one, which I wouldn't say is too far fetched.

The work is really for investigating schedutil / intel_pstate `iowait_boost` behavior but of course the entire I/O workload w.r.t. scheduling events and their analysis could prove useful for other use cases like io_uring workloads more generally.

douglas-raillard-arm:

PR #2154, opened 2023-12-14
lisa.trace: Allow Trace(normalize_time=<float>, ...)

douglas-raillard-arm:
FEATURE

Allow normalizing time with an arbitrary offset. Use of True/False still have the same meaning as before.


douglas-raillard-arm:
As discussed offline with @msrasmussen , instead of modifying the base Trace class, we should make TraceView more powerful and pass a post-processing function that allows modifying the dataframe returned by `df_event()` in arbitrary ways. Then we can build up on that with a `Trace.with_time_offset()` method simply creating a `TraceView` with a function applying an offset on the index.

PR #2157, opened 2023-12-20
Pg/mod config v1

pierregondois:


chrisswinchatt-arm:
I haven't fully reviewed this but just noticed a couple of things.

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	if (!lisa_fs_create_single(parent, "available_features",
+			      NULL,
+			      &lisa_available_features_fops,
+			      S_IFREG | S_IRUGO, &lisa_features_param))
+		return -ENOMEM;
+
+	/* activate: activate the selected (and configured) features - RW. */
+	if (!lisa_fs_create_single(parent, "activate",
+			      NULL,
+			      &lisa_activate_fops,
```
bea-arm, line 562: minor really but to keep the naming consistent here maybe lisa_active_features_fops ?

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+			      S_IFREG | S_IRUGO | S_IWUGO, cfg))
+		return -ENOMEM;
+
+	/* configs: Dir containing configurations, only setup at the top level. */
+	if (is_top_level) {
```
bea-arm, line 567: I guess this could simply check  if parent is  sb_root

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+		return -ENOMEM;
+
+	init_feature_param_entry(entry, cfg, &lisa_features_param);
+
+	/* set_features: enable a feature - RW. */
+	if (!lisa_fs_create_single(parent, "set_features",
+			      NULL,
+			      &lisa_param_feature_fops,
+			      S_IFREG | S_IRUGO | S_IWUGO, entry)) {
+		free_feature_param_entry(entry);
```
bea-arm, line 548: It escaped me why this is the only case when the earlier allocated and initialized entry gets properly cleaned upon failure. Why this is not being done for other cases ?  Will that be covered by free_lisa_cfg somehow ? Either way, unless I missed smth it would be best to either clean thing up on each failure, or document why it is not needed.

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+}
+
+/* Called with interface_lock */
+static int
+lisa_fs_create_files(struct dentry *parent, bool is_top_level, struct lisa_cfg *cfg)
+{
+	struct feature_param_entry *entry;
+	struct feature *feature;
+
+	entry = allocate_feature_param_entry();
```
bea-arm, line 537: Does that really need a wrapper ? It gives an impression that there is more to it than just allocating memory .
If you still want to keep it maybe make an __always_inline out of it .

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	d_add(dentry, inode);
+	if (mode & S_IFDIR) {
+		inc_nlink(d_inode(parent));
+		inc_nlink(inode);
+	}
+
+	return dentry;
+}
+
+/* Called with interface_lock */
```
bea-arm, line 530: Right, though why ? At this point vfs is still not valid, so what is this locking protecting from ?

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	if (!cfg)
+		goto error2;
+
+	mutex_lock(&interface_lock);
+
+	ret = lisa_fs_create_files(sb->s_root, true, cfg);
+	if (ret)
+		goto error3;
+
+	ret = init_lisa_cfg(cfg, &cfg_list, NULL);
```
bea-arm, line 653: is this supposed to be 'global' config ? If so why not passing the s_root dentry ?

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+		goto error4;
+
+	mutex_unlock(&interface_lock);
+
+	return 0;
+
+error4:
+	free_lisa_cfg(cfg);
+error3:
+	mutex_lock(&interface_lock);
```
bea-arm, line 664: Shouldn't this be unlock ?

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	for_each_feature(feature)
+		for_each_feature_param(param, pparam, feature)
+			INIT_LIST_HEAD(&param->global_value);
+}
+
+int init_lisa_fs(void)
+{
+	int ret;
+
+	init_feature_param();
```
bea-arm, line 730: Why not making it part of module init instead ? Might be better to distinguish vfs and features.

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	mutex_init(&interface_lock);
+	INIT_HLIST_HEAD(&cfg_list);
```
bea-arm, line 732: I'm not entirely convinced this is the right place to init the hash list , config.c seems to be more suitable place to do that and as the list is global (for conig.c) you could use HLIST_HEAD there and drop the list from the parameters on 'external' functions that do require it and manage list maintenance from within single compilation unit -> config.c

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+			init_feature_param_entry(entry, cfg, param);
+
+			if (!lisa_fs_create_single(dentry, param->name, NULL,
+						   &lisa_param_feature_fops,
+						   S_IFREG | S_IRUGO, entry)) {
+				free_feature_param_entry(entry);
+				return -ENOMEM;
+			}
+		}
+	}
```
bea-arm, line 607: Will all the dentry created on the way before an error is triggered be clean-up ? I do  not recall how fs framework handles such cases (as the fs_fill_super will fail in this case)
Also, same question about allocated entries - will those be covered as well and memory release ?
(aside : did you try to run the module with kmemleak being enabled - it should show if smth falls between the crack )

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	inode_lock_nested(inode, I_MUTEX_PARENT);
+	inode_lock(d_inode(dentry));
+
+	return 0;
+}
+
+const struct inode_operations lisa_fs_dir_inode_operations = {
+	.lookup = simple_lookup,
+	.mkdir = lisa_fs_syscall_mkdir,
+	.rmdir = lisa_fs_syscall_rmdir,
```
bea-arm, line 489: It's just a preference but I would drop the 'syscall' from those names, as mkdir/rmdir as inode dir ops are rather self explanatory

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	inode_unlock(inode);
+	inode_unlock(d_inode(dentry));
+
+	mutex_lock(&interface_lock);
+
+	cfg = find_lisa_cfg(dentry->d_name.name);
+	if (!cfg)
+		pr_err("Failed to find config: %s\n", dentry->d_name.name);
+
+	free_lisa_cfg(cfg);
```
bea-arm, line 476: It does not seem that _free_lisa_cfg_ is prepared to handle null pointers so I guess this should be placed with else for the if statement above

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+
+	mutex_unlock(&interface_lock);
+
+	inode_lock_nested(inode, I_MUTEX_PARENT);
+	inode_lock(d_inode(dentry));
+
+	return 0;
```
bea-arm, line 483: Is this actually removing the directory ?
Looking at simple_rmdir it seems that it might be missing smth , like drop_nlink calls ?

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+static struct dentry *
+lisa_fs_create_single(struct dentry *parent, const char *name,
+		      const struct inode_operations *i_ops,
+		      const struct file_operations *f_ops, umode_t mode,
+		      void *data);
+
+#define LISA_FS_SUPER_MAGIC 0xcdb11bc9
+
+/* Protect the interface. */
+static struct mutex interface_lock;
```
bea-arm, line 23: Any reason for not using fs specific data (s_fs_info) for that ?
Is that mutex there to serialize access to fs nodes or to protect the config list ? or both

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+
+int feature_param_lisa_validate(struct feature_param_entry_value *val)
```
bea-arm, line 25: Why not placing it in feature_param.[c/h] ? Seems bit misplaced 
Same re: lisa_features_param

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+{
+	struct feature *feature;
+
+	for_each_feature(feature) {
+		if (!strcmp(feature->name, val->data))
```
bea-arm, line 30: This does not look safe when the param value is not a pointer

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+
+static int lisa_activate_show(struct seq_file *s, void *data)
+{
+	struct lisa_cfg *cfg = s->private;
+
+	seq_printf(s, "%d\n", cfg->activated);
+	return 0;
+}
+
+static int update_global_value(struct lisa_cfg *cfg, int new_value)
```
bea-arm, line 101: Why  not placing it in config,c where it feels like a more obvious choice ?

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+				pr_err("Could not rollback config values\n");
+				return ret;
+			}
+		}
+	}
+
+	return ret;
+}
+
+static bool is_feature_set(char *name)
```
bea-arm, line 144: Ditto

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	struct feature_param_entry_value *val;
+
+	/* Check whether the feature is in the global set_features list. */
+	list_for_each_entry(val, &lisa_features_param.global_value, node)
+		if (lisa_features_param.ops->is_equal(&name, val))
+			return true;
+	return false;
+}
+
+int lisa_activate_config(bool value, struct lisa_cfg *cfg)
```
bea-arm, line 155: Ditto

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+	struct feature_param_entry_value *val;
+	struct feature_param_entry *entry;
+	struct feature_param *param;
+
+	entry = *(struct feature_param_entry **)s->private;
+	param = entry->param;
+
+	val = hlist_entry(v, struct feature_param_entry_value, node);
+
+	if (param->ops->stringify) {
```
bea-arm, line 258: What if the value is already a string ? Seems this is no being shown then ?

bea-arm (commented):
This is a partial review, mainly of vfs interface. Will try to get back to the rest soon.

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+		return -ENOMEM;
+
+	mutex_lock(&interface_lock);
+
+	my_dentry = lisa_fs_create_single(dentry->d_parent, dentry->d_name.name,
+					  &simple_dir_inode_operations,
+					  &simple_dir_operations,
+					  S_IFDIR | mode, cfg);
+	if (!my_dentry)
+		goto error;
```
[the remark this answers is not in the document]
pierregondois: Yes right, thx

lisa/_assets/kmodules/lisa/fs.c
```diff
@@ -0,0 +1,755 @@
...
+
+	ret = init_lisa_cfg(cfg, &cfg_list, my_dentry);
+	if (ret)
+		return -ENOMEM;
```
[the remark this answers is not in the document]
pierregondois: Yes right

lisa/_assets/kmodules/lisa/fs.c
pierregondois: No specific reason, I will move it

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Done

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Done

lisa/_assets/kmodules/lisa/fs.c
pierregondois: I'm not sure I understand the case, is the concern about having `val->data` not being a string, but a data of another type (like an integer for instance) ?

The input `val` should have been parsed as a string, cf.:
```
struct feature_param lisa_features_param = {
	.name = "lisa_features_param",
       [...]
	.ops = &feature_param_ops_string,
	.validate = feature_param_lisa_validate,
       [...]
};
```

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes right, done

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes correct

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes, this would make more sense

lisa/_assets/kmodules/lisa/fs.c
pierregondois: But it's convenient to just have just one common case. I'll let you decide if a more optimal solution is required to handle strings.

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes right.
I'm not sure the global list is passed as a parameter in any function. However some functions use the same name `cfg_list` name as their input parameter (e.g. below). If you want to change the naming as this is too confusing, let me know.

```
int init_lisa_cfg(struct lisa_cfg *cfg, struct hlist_head *cfg_list,
		  struct dentry *dentry)
{
	cfg->dentry = dentry;
	hlist_add_head(&cfg->node, cfg_list);
	return 0;
}
```

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Ah I did the naming as `lisa_[file name]_fops`, so it is making sense for `lisa_activate_fops` and `lisa_available_features_fops`, but not with `lisa_param_feature_fops` where the fops are used for multiple files

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes right

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes right

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Ok sure

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes ok

lisa/_assets/kmodules/lisa/fs.c
pierregondois: I think it was necessary when the mutex was taken in `lisa_fs_fill_super()`.

It's not the case anymore, so I'm moving the mutex in `s_fs_info`, along with the global `cfg_list`

lisa/_assets/kmodules/lisa/fs.c
pierregondois: The call to `free_feature_param_entry()` should not be necessary actually. If something goes wrong, the caller should call `free_lisa_cfg()`, which will go over the list of entries and free them.
I'll add a comment in that sense.

lisa/_assets/kmodules/lisa/fs.c
pierregondois: The dir removal is done via:
```
lisa_fs_rmdir()
\-free_lisa_cfg()
  \-lisa_fs_remove()
    \-simple_recursive_removal()
```
I.e. the `lisa_cfg` is freed, and while freeing it, its dentry is freed.
This allows to remove the directory associated to a config whenever we free a `lisa_cfg` structure.

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Yes right

lisa/_assets/kmodules/lisa/fs.c
bea-arm: OK, got it - thanks!

lisa/_assets/kmodules/lisa/fs.c
bea-arm: Right, I do see the difference now , thanks for pointing that out.
Still inlining (at least some of) those would be a god idea, though I guess the compiler will pick it up anyway (worth checking).
Aside: very minor but you have:
allocate_xxx -> init_xxx
init is an abrev so why not 'alloc' instead of 'allocate' ?
(Feel free to ignore it)

lisa/_assets/kmodules/lisa/fs.c
bea-arm: Lets rename the global list to distinguish that as yes, I got caught up in that....

lisa/_assets/kmodules/lisa/utils.h
```diff
@@ -12,4 +12,15 @@
 
 #include "linux/kernel.h"
 
+static inline size_t list_count_elements(struct list_head *head)
```
bea-arm, line 15: Minor size_t is usually used for variables that do represent a size of an object, for counting unsigned int seems more appropriate.
Aside : Is it being used anywhere ? Could not find any reference to it...

douglas-raillard-arm: Drive by comment: size_t is definitely the correct type here. The max size of most collection types is proportional to the amount of addressable memory. sizeof() applied on an array type gives a size_t. Beyond that, the restrictions on `unsigned int` are very loose. A strictly conforming implementation in C is perfectly allowed to have 16 bits `int` type with several MB of RAM. These platforms exist in practice even if the kernel or POSIX is not portable to them. The only thing I can think of that is more of a pain with `size_t` is printing it, since printf has no native support for it, you either use the extra macros or just cast to the largest type you expect and print as that. 

Beyond C, Rust systematically uses `usize` for this case (equivalent to size_t), and sloppiness around that would be paid by extra boilerplate (there is no infallible conversion from `usize` to any other integer type beyond the `as` cast operator that is considered to a large extent as a historical accident).

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Ok, will rename and inline

lisa/_assets/kmodules/lisa/fs.c
pierregondois: Should be done

lisa/_assets/kmodules/lisa/utils.h
pierregondois: Thanks for the clarification Douglas.
Unfortunately the function doesn't seem to be used anywhere anymore. So I ll remove it.

douglas-raillard-arm:

PR #2161, opened 2024-01-02
lisa._assets.kmodules.lisa: follow latest upstream

honxia02:
Upstream has completely removed struct util_est. Now util_est is just an unsigned int field.

### DIFF
--- a/lisa/_assets/kmodules/lisa/ftrace_events.h
+++ b/lisa/_assets/kmodules/lisa/ftrace_events.h
@@ -234,8 +234,7 @@ TRACE_EVENT(lisa__sched_util_est_se,
 
 	TP_STRUCT__entry(
 		__field(	unsigned long,	util			)
-		__field( 	unsigned int,	enqueued		)
-		__field( 	unsigned int,	ewma			)
+		__field( 	unsigned int,	util_est		)
 		__field(	int,		cpu			)
 		__field(	int,		pid			)
 		__array(	char,		path,	PATH_SIZE	)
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
I see that this change also renames event fields. Do we really want to do that ? If so this needs to come with the the fixes everywhere it's required in LISA, and ensure it goes through an analysis method that will deal with backward compat of event format.

lisa/_assets/kmodules/lisa/ftrace_events.h
```diff
@@ -267,23 +265,20 @@ TRACE_EVENT(lisa__sched_util_est_cfs,
...
 		__array(	char,		path,	PATH_SIZE	)
 	),
 
 	TP_fast_assign(
 		__entry->cpu		= cpu;
 		strlcpy(__entry->path, path, PATH_SIZE);
-		__entry->enqueued	= avg->util_est.enqueued;
-		__entry->ewma		= avg->util_est.ewma;
 		__entry->util		= avg->util_avg;
+		__entry->util_est	= avg->util_est;
```
douglas-raillard-arm, line 277: This won't fly on less recent kernel, this needs to be done based on the type

douglas-raillard-arm (changes requested):

douglas-raillard-arm:
The commit message will need to be updated to add the FIX tag (see history).

honxia02:
Closed, because so far the conclusion is that the behavior of the trace event has changed and it's not a good idea to modify the old ones. Instead we should treat this as a new signal.

PR #2170, opened 2024-01-15
Trace thermal zone temperatures in the LISA kmodule

chrisswinchatt-arm:
* Temperures of up to 12 thermal zones are traced to `ftrace` by polling. The poll interval can be set by defining `CONFIG_POLL_INTERVAL_MS` during build, which defaults to 100 ms. It will be clamped to be between 1 and 5000 ms inclusive.
* Thermal zones are selected by defining `CONFIG_THERMAL_ZONE_LIST` with a comma-separated list of thermal zone types (the value of `/sys/class/thermal/thermal_zoneX/type`) during build, otherwise, `LITTLE,MID,BIG` is used.


douglas-raillard-arm:

PR #2175, opened 2024-01-17
tools/buildroot_config: Sync rootfs to devlib busybox & add '-o nice' to ps

deggeman:
Set CONFIG_MODPROBE_SMALL=n and CONFIG_STATIC=y to allign the busybox binary of the rootfs with that of devlib (refer to $LISA/tools/recipes/busybox.recipe).

Set CONFIG_MODPROBE_SMALL=n to avoid an issue with insmod related to its return value.

Moreover, set CONFIG_FEATURE_PS_ADDITIONAL_COLUMNS=y so that /bin/ps supports the '-o nice' specifier. This is needed to evaluate issues related to lisa-test during load related PELT Invariance tests on the target they happen. Load depens on task priority, i.e. task nice value.


tools/buildroot_config
```diff
@@ -23,3 +23,4 @@ BR2_PACKAGE_UTIL_LINUX=y
 BR2_PACKAGE_UTIL_LINUX_BINARIES=y
 BR2_PACKAGE_PYTHON3=y
 # BR2_TARGET_ROOTFS_TAR is not set
+BR2_PACKAGE_BUSYBOX_CONFIG_FRAGMENT_FILES="../../tools/buildroot_busybox_fragments"
```
douglas-raillard-arm, line 26: Where does the relative path come from ? Is it relative to where BusyBox gets built ?

deggeman: Yes, it's relative to $LISA/src/buildroot . I figured it's best to put buildroot_busybox_fragments under $LISA/tools next to buildroot_config .

PR #2182, opened 2024-01-23
lisa._kmod: Add util_est_unified trace events

honxia02:
FEATURE

Upstream has now removed struct util_est and effectively merged util_est.enqueued and util_est.ewma in a single int field. Provide a new trace event for it so the new kernel can trace it.

This also fixes LISA not being able to compile and load the module on the new kernel. We detect whether struct util_est exists to decide if the new signal should be used.

### DIFF
--- a/lisa/_assets/kmodules/lisa/ftrace_events.h
+++ b/lisa/_assets/kmodules/lisa/ftrace_events.h
@@ -258,6 +258,38 @@ TRACE_EVENT(lisa__sched_util_est_se,
 );
 #endif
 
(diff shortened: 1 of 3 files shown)


douglas-raillard-arm:
LGTM, can you update the commit message to replace "FEATURE and FIX" with just "FEATURE" ? Otherwise as it stands it will appear twice in the changelog (or not at all, can't remember what the regex looks like).

We could add further support in the load_tracking analysis but let's hold off for now until we had a bit more use of that event in case we want to modify something

honxia02:
Thanks. Will do.

honxia02:
Both the PR and the commit updated.

PR #2183, opened 2024-01-24
Fix a couple of GroupBy.apply() deprecations

msrasmussen:
FIX

Pandas 2.2.0 GroupBy.apply() no longer includes the group column by default. It must be explicitly selected.

### DIFF
--- a/lisa/analysis/pixel6.py
+++ b/lisa/analysis/pixel6.py
@@ -67,7 +67,7 @@ class Pixel6Analysis(TraceAnalysisBase):
             df = pd.DataFrame(dict(power=power, channel=df['chan_name']))
             return df.dropna()
 
-        df = grouped.apply(make_chan_df)
+        df = grouped[df.columns].apply(make_chan_df)
         df['channel'] = df['channel'].astype('category').cat.rename_categories(Pixel6Analysis.EMETER_CHAN_NAMES)
 
         return df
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
It's a bit odd, I can't find anything mentioning that in the change log of pandas. The fix looks like a no-op so it seems a bit fishy. Might be worth having a look at pandas bug tracker in case it's something that will just get fixed soon

msrasmussen:
AFAICT it is deliberate:

> Deprecated including the groups in computations when using [DataFrameGroupBy.apply()](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.apply.html#pandas.core.groupby.DataFrameGroupBy.apply) and [DataFrameGroupBy.resample()](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.resample.html#pandas.core.groupby.DataFrameGroupBy.resample); pass include_groups=False to exclude the groups ([GH 7155](https://github.com/pandas-dev/pandas/issues/7155))

If I understand the PR correctly, some users are fed up having to explicitly exclude the group column, so now it is excluded by default. The proposed fix is to either explicitly specify the columns (like df.columns), or use the new include_groups=True argument in GroupBy.apply(include_groups=True).

